Issue #82, opened 2021-01-21
Add support for wifi in the network configuration


Torxed:
Based off #65, it would be nice with a wifi option.
Since laptop users and other devices more or less defaults to using wifi rather than ethernet, this would be a welcome addition I think.

grazzolini:
I do like the solution that archiso uses, mixing systemd-networkd + iwd. I think that something in that direction is what we should aim for. Because we already have code for systemd-networkd.

Torxed:
![2021-01-25-121031_1096x1029_scrot](https://user-images.githubusercontent.com/861439/105698625-68586c80-5f06-11eb-8bf3-cc903afb8991.png)


Based on a discussion over at IRC, `iwctl` doesn't have a nice machine readable interface for its output.
They recommend [python-dbus](https://gitlab.freedesktop.org/dbus/dbus-python) and to adopt their [python examples](https://git.kernel.org/pub/scm/network/wireless/iwd.git/tree/test/list-known-networks) using dbus.

Torxed:
Appending some nice test code provided by mcon@irc://freenode/#iwd
It's Python bindings for iwd to community and monitor events.

[ZM3UI.txt](https://github.com/Torxed/archinstall/files/5866183/ZM3UI.txt)

Torxed:
As discussed with grazzolini, we'll copy network configuration from the ISO for now.
And let the users use `iwctl` to manually configure wifi before starting archinstall *(since, archinstall needs to be installed it requires functional internet at first anyway)*

grazzolini:
> As discussed with grazzolini, we'll copy network configuration from the ISO for now.
> And let the users use `iwctl` to manually configure wifi before starting archinstall _(since, archinstall needs to be installed it requires functional internet at first anyway)_

This situation will change if archinstall is ever added to the iso. In that case it could be responsible for the network configuration. As things are now, it's fine to have things simple now (copying wifi configuration from the iso) and then move to a more complex solution in the future.

Also, I really don't think archinstall needs to do all the heavy lifting. After all, the idea is to have an installer that incorporates Arch KISS principle. Having the users doing some of the configuration manually is within that spirit.

Torxed:
![2021-01-25-153513_553x150_scrot](https://user-images.githubusercontent.com/861439/105719796-01957c00-5f23-11eb-86a4-31976af186e6.png)

Added the following, the zero-option copies the current network configuration from the ISO (live media).
[...]

PR #96, opened 2021-01-25
Hardware n wifi

Torxed:
As per discussed in #82: 

2021-01-25-153513_553x150_scrot

Added the following, the zero-option copies the current network configuration from the ISO (live media).
It first inserts a dummy-option into the list of interfaces:
* https://github.com/Torxed/archinstall/blob/1c80a893acfb97859dd6c0f61a2cf74c03958595/examples/guided.py#L199
And if zero (injected dummy interface) is selected, it stores it as is:
* https://github.com/Torxed/archinstall/blob/1c80a893acfb97859dd6c0f61a2cf74c03958595/examples/guided.py#L229
And is then interpreted in guided during installation:
* https://github.com/Torxed/archinstall/blob/1c80a893acfb97859dd6c0f61a2cf74c03958595/examples/guided.py#L54-L55

Which in turn copies two locations from ISO to (by default) `/mnt`:
1. Copies PSK files for iwd: https://github.com/Torxed/archinstall/blob/405794d53f1f47e2a19a906dbdfa6ad39ea79c17/archinstall/lib/installer.py#L208
2. Copies .netdev/.network for systemd: https://github.com/Torxed/archinstall/blob/405794d53f1f47e2a19a906dbdfa6ad39ea79c17/archinstall/lib/installer.py#L216

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ SAFETY_LOCK
 **/**.network
 **/**.target
 **/**.qcow2
+**/test.py
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -12,3 +12,4 @@ from .lib.services import *
 from .lib.packages import *
 from .lib.output import *
 from .lib.storage import *
+from .lib.hardware import *
--- a/archinstall/lib/exceptions.py
+++ b/archinstall/lib/exceptions.py
@@ -8,3 +8,5 @@ class SysCallError(BaseException):
 	pass
 class ProfileNotFound(BaseException):
 	pass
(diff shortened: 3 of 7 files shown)


grazzolini:
I'm testing this and if you pick 0 to copy the iso configuration, you end up with a installation that doesn't have iwd installed nor enabled, and also not systemd-networkd nor systemd-resolved enabled either. I'm looking to see if there are other issues as well.

Torxed:
Verified that `iwd` is installed and that any `.psk` files are carried over.
All `systemd-networkd` files are also copied over and both `-networkd` and `-resolved` now properly starts on boot, as well as `iwd.service`.

PR #103, opened 2021-01-29
Fix README.md minimal example link

nikibobi:


Torxed:
We'll spotted, thanks for this one :)

PR #104, opened 2021-01-30
Add correct locale to /etc/locale.gen

phisch:
The method `set_locale` builds a locale string and puts it into `/etc/locale.gen`. Currently it appends `en_US.UTF-8.UTF-8 UTF-8`. This change doesn't contain the encoding in the locale argument, so that the locale string is built correctly.


Torxed:
Great work! I've been meaning to fix this in the next release, as it's caused some issues with translations and especially go-applications, they seam to really like the `LANG` variable :) Thank you!

PR #143, opened 2021-04-02
Join cryptsetup args using shlex

pyfisch:
Ensure that all arguments are properly quoted.
Improve readability by using long-form arguments only.
Use identical arguments for both cryptsetup calls,
to prevent them from accidentally diverging in future commits.

Cryptsetup arguments reference: https://man.archlinux.org/man/cryptsetup.8

Tested using the README method, with everything on default but with disk encryption enabled.

## Checklist:

### DIFF
--- a/archinstall/lib/luks.py
+++ b/archinstall/lib/luks.py
@@ -1,4 +1,5 @@
 import os
+import shlex
 from .exceptions import *
 from .general import *
 from .disk import Partition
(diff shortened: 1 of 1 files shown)


pyfisch:
In PEP8 the insight "code is read much more often than it is written" by Guido van Rossum is quoted. In this spirit I would appreciate it if command arguments are written in a structured form in archinstall (like I did with cryptsetup):

* Always use the long version of parameters. (Easier to read if you don't know the commands by heart)
* Write commands as a list of arguments, not a single string (Eliminates quoting errors.)

Torxed:
Good feedback and a welcome change, definitely a better looking way of doing it. I'll go over most of the command arguments when I also go over the general structure of `sys_command`, it needs some love as well.

PR #170, opened 2021-04-04
Add validation check for username

SecondThundeR:
## Description

Fixes an error during the creation of a new user when the username is invalid

## Bugs and Issues

This issue was described in #168 

## How Has This Been Tested?

**Test Configuration**:
* Hardware: VirtualBox 6.1
* Specific steps: Ran installer and entered the username with incorrect case, characters and length

## Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/archinstall/lib/user_interaction.py
+++ b/archinstall/lib/user_interaction.py
@@ -1,4 +1,4 @@
-import getpass, pathlib, os, shutil
+import getpass, pathlib, os, shutil, re
 from .exceptions import *
 from .profiles import Profile
 from .locale_helpers import search_keyboard_layout
@@ -49,8 +49,15 @@ def print_large_list(options, padding=5, margin_bottom=0, separator=': '):
 
 def ask_for_superuser_account(prompt='Create a required super-user with sudo privileges: ', forced=False):
 	while 1:
-		new_user = input(prompt).strip(' ').lower()
-		
+		new_user = input(prompt).strip(' ')
+
+		if not re.match('[a-z_][a-z0-9_-]*[$]?', new_user) or len(new_user) > 32:
+			log(
+				"The username you entered is invalid. Try again",
+				level=LOG_LEVELS.Warning,
+				fg='red'
+			)
+			continue
 		if not new_user and forced:
 			# TODO: make this text more generic?
 			#       It's only used to create the first sudo user when root is disabled in guided.py
@@ -67,9 +74,16 @@ def ask_for_additional_users(prompt='Any additional users to install (leave blan
 	super_users = {}
 
 	while 1:
-		new_user = input(prompt).strip(' ').lower()
+		new_user = input(prompt).strip(' ')
 		if not new_user:
 			break
+		if not re.match('[a-z_][a-z0-9_-]*[$]?', new_user) or len(new_user) > 32:
+			log(
+				"The username you entered is invalid. Try again",
+				level=LOG_LEVELS.Warning,
+				fg='red'
+			)
+			continue
 		password = get_password(prompt=f'Password for user {new_user}: ')
 		
 		if input("Should this user be a sudo (super) user (y/N): ").strip(' ').lower() in ('y', 'yes'):
--- a/archinstall/lib/user_interaction.py
+++ b/archinstall/lib/user_interaction.py
@@ -18,6 +18,16 @@ def get_terminal_width():
 def get_longest_option(options):
 	return max([len(x) for x in options])
 
+def check_for_correct_username(username):
+    if re.match(r'^[a-z_][a-z0-9_-]*\$?$', username) and len(username) <= 32:
+        return True
+    log(
+		"The username you entered is invalid. Try again",
+		level=LOG_LEVELS.Warning,
+		fg='red'
+	)
+    return False
+
 def get_password(prompt="Enter a password: "):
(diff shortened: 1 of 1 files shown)


pyfisch:
On a related note the complete regex from `useradd` should be checked: https://man.archlinux.org/man/useradd.8#CAVEATS

Personally I prefer to receive an error message if I enter a name containing upper-case letters over lowercasing the name.

Torxed:
> On a related note the complete regex from `useradd` should be checked: https://man.archlinux.org/man/useradd.8#CAVEATS
> 
> Personally I prefer to receive an error message if I enter a name containing upper-case letters over lowercasing the name.

That's handy, so `[a-z_][a-z0-9_-]*[$]?` needs to be matched, otherwise we log an error.
Originally that was part of the discussion, so perhaps that's the route we should go.. It was me who said "just do `.lower()`" which is my bad. The regex would be nicer! And log an error and continue the loop.

SecondThundeR:
Should we implement such a solution where we check if the username is invalid or its length is incorrect?
```python
if not re.match('[a-z_][a-z0-9_-]*[$]?', new_user) or len(new_user) > 32:
  log(
    "Invalid username. ..." # Add text to clarify the cause of the error
    level=LOG_LEVELS.Warning,
    fg='red'
  )
else:
  # Continuation of loop...
```

Torxed:
> Should we implement such a solution where we check if the username is invalid or its length is incorrect?
> 
> ```python
> if not re.match('[a-z_][a-z0-9_-]*[$]?', new_user) or len(new_user) > 32:
>   log(
>     "Invalid username. ..." # Add text to clarify the cause of the error
>     level=LOG_LEVELS.Warning,
>     fg='red'
>   )
> else:
>   # Continuation of loop...
> ```

That looks very ideal!

SecondThundeR:
Updated the code. Waiting for additional verification from your side

archinstall/lib/user_interaction.py
```diff
@@ -49,8 +49,15 @@ def print_large_list(options, padding=5, margin_bottom=0, separator=': '):
 
 def ask_for_superuser_account(prompt='Create a required super-user with sudo privileges: ', forced=False):
 	while 1:
-		new_user = input(prompt).strip(' ').lower()
-		
+		new_user = input(prompt).strip(' ')
+
+		if not re.match('[a-z_][a-z0-9_-]*[$]?', new_user) or len(new_user) > 32:
```
kpcyrd, line 54: I think there's a `^` needed at the start.

archinstall/lib/user_interaction.py
```diff
@@ -67,9 +74,16 @@ def ask_for_additional_users(prompt='Any additional users to install (leave blan
 	super_users = {}
 
 	while 1:
-		new_user = input(prompt).strip(' ').lower()
+		new_user = input(prompt).strip(' ')
 		if not new_user:
 			break
+		if not re.match('[a-z_][a-z0-9_-]*[$]?', new_user) or len(new_user) > 32:
```
kpcyrd, line 80: This code shouldn't be duplicated and live in a function instead.

archinstall/lib/user_interaction.py
SecondThundeR: You're right. Updated this and fixed an issue where it accepts a username with a last character that is not a dollar sign

Torxed:
This looks good to me, I'm ready to merge and will do so if no other objections :)

kpcyrd:
btw, what does `[$]?` do? It seems this means "the end of the string or nothing" which means the regex doesn't actually force that it matches the whole string:
```
>>> re.match('^[a-z_][a-z0-9_-]*[$]?', 'asdf/_\n\x00')
<re.Match object; span=(0, 4), match='asdf'>
```
This would be fine if we're using the match, but we only check if the regex matched _somewhere_ and then use the whole string.

Torxed:
> k if the regex matched _somewhere_ and then use the whole string.

I'm not a regexp ninja, but yea that looks wrong.
A quick search also shows that this is a Arch specific addon to upstream man page for useradd? In which case, perhaps https://archlinux.org/packages/core/x86_64/man-db/ should get a bug report as well.

SecondThundeR:
> This would be fine if we're using the match, but we only check if the regex matched _somewhere_ and then use the whole string.

I tried to think about such a case in the future and added `$` at the end, which means meta-escape in the regular expression. Therefore, it's strictly checked whether there was a dollar sign at the very end or not. If the last character is different from it and isn't part of the previous rule `[a-z0-9_-]`, it returns `None`, but not the part of the string before that character. For other scenarios without this symbol, it is handled correctly. Since the dollar sign is optional, this should work fine.

```
>>> re.match('^[a-z_][a-z0-9_-]*\$?$', 'asdf/_\n\x00')
None

>>> re.match('^[a-z_][a-z0-9_-]*\$?$', 'asdf$')
<re.Match object; span=(0, 5), match='asdf$'>

>>> re.match('^[a-z_][a-z0-9_-]*\$?$', 'asdf123ds@')
None

>>> re.match('^[a-z_][a-z0-9_-]*\$?$', 'asdf123ds')
<re.Match object; span=(0, 9), match='asdf123ds'>
...
```
Also note that here I have used `\$` instead of `[$]`, which is equivalent. Thought it would be better not to use square brackets for single character

PR #178, opened 2021-04-05
Simplify packaging

pyfisch:
My goal with this PR is to simplify packaging archinstall as a Python package (wheel) and as an Arch Linux package. I tested using archinstall both when installed as a Python package and when installed as an Arch package.

Major changes:

* Switch from `setup.py` to `setup.cfg` configuration. Setuptools suggests that project transition to the declarative `setup.cfg`. A minimal `setup.py` file is retained for compatibility and to aid Arch packaging. The current version is now accessible from Python through the standard `__version__` name.
* [Add `pyproject.toml` `[build-system]` section.](https://setuptools.readthedocs.io/en/latest/userguide/quickstart.html#python-packaging-at-a-glance)
* Add the `archinstall` script as a [Python entry point](https://setuptools.readthedocs.io/en/latest/userguide/entry_point.html). This avoids the need for a shell wrapper script. To make this possible I moved `run_as_a_module` to `__init__.py`.
* Remove all `PKGBUILD` files. Since archinstall is now part of the extra repository the canonical PKGBUILD file should be located there. I don't think the distinction between `archinstall` and `python-archinstall` is meaningful anymore, so I wrote a single [PKGBUILD](https://gist.github.com/pyfisch/05a1bcd6e4a7c3c90cfab8f7b089f99b) to package archinstall.


__init__.py
```diff
@@ -1,3 +0,0 @@
-# This __init__ file is just here to support the
```
pyfisch, line 1: I removed this file because I wasn't sure if this "installation" method will still work. Generally Python packages don't support use without installation.

Torxed: One usage of this library is as a git submodule, see for instance [archoffline](https://github.com/Torxed/archoffline) which has this as a submodule. Perhaps this is not needed in the long run now that it's a pypi project as well. I'll have to circle back, but I do like the ability to use it as a git submodule :)

Torxed:
This looks very clean and I like the approach, but I'll ping @grazzolini who may or may not be available for a couple of more days, and he can review these changes since he's the package maintainer for the official package. But overall looks good.

pyfisch:
While I am at it I would also like to move all files from `archinstall/lib` to just `archinstall` as this the standard structure of Python packages and makes imports simpler. (I couldn't figure out how to add a folder to lib and make the imports work. ;-) )

Torxed:
> While I am at it I would also like to move all files from `archinstall/lib` to just `archinstall` as this the standard structure of Python packages and makes imports simpler. (I couldn't figure out how to add a folder to lib and make the imports work. ;-) )

The only down side of this that I can see is that all functions will be exposed during import? there won't be any way to "hide" certain functions? Unless `__init__` still can take care of this?
This might break a few things, but I'm willing to fix it when the time comes. Perhaps a separate PR tho.

grazzolini:
From me removing the PKGBUILD is a no-no. The PKGBUILD is there to help test changes. And the PKGBUILD that's on this repo is different from the one used on the Arch repo.

pyfisch:
> The only down side of this that I can see is that all functions will be exposed during import?

Are there currently functions that aren't exposed? I am not sure because I think that `*` imports everything. Can you give me an example?
`__all__` can be used to modify the names imported with `*`.

>  This might break a few things, but I'm willing to fix it when the time comes. Perhaps a separate PR tho.

Probably. I also would wait until the torxed-v2.2.0 is merged so there aren't any conflicts.

pyfisch:
> From me removing the PKGBUILD is a no-no. The PKGBUILD is there to help test changes. And the PKGBUILD that's on this repo is different from the one used on the Arch repo.

@grazzolini You're welcome. Do you just need the one in the repository root, or also the three in PKGBUILDs/?

grazzolini:
> > From me removing the PKGBUILD is a no-no. The PKGBUILD is there to help test changes. And the PKGBUILD that's on this repo is different from the one used on the Arch repo.
> 
> @grazzolini You're welcome. Do you just need the one in the repository root, or also the three in PKGBUILDs/?

Just the one on the root folder. The others are reminiscent of when the package was on the AUR

Torxed:
> > The only down side of this that I can see is that all functions will be exposed during import?
> 
> Are there currently functions that aren't exposed? I am not sure because I think that `*` imports everything. Can you give me an example?
> `__all__` can be used to modify the names imported with `*`.
> 
> > This might break a few things, but I'm willing to fix it when the time comes. Perhaps a separate PR tho.
> 
> Probably. I also would wait until the torxed-v2.2.0 branch is merged so there aren't any conflicts.

Currently I do lazy imports with `from .x import *`, but the general idea behind it is to hide certain things.
I would also like to avoid circular imports in the future, they are becoming more and more of an issue hehe.

pyfisch:
> Currently I do lazy imports with `from .x import *`

@Torxed  I don't think this is a lazy import. Try this yourself:

1. Add the line `error_on_import = 0 / 0` to `archinstall/lib/disk.py`.
2. Open Python and `import archinstall`.

If names are lazily loaded you will encounter an error when you try to use something from disk. However if names *aren't* lazily loaded the error occurs when you import archinstall. The Python interpreter evaluates  modules on import to know which names they define.

@grazzolini added the PKGBUILD back.

Torxed:
> f names are lazily loaded yo

Ah sorry, I more meant **I was lazy** haha. Didn't consider there's something called lazy imports ^^
I think the PR looks good and I'm happy to merge, can't tell if something will break until I test it properly. And I'm still a bit worried about submodules breaking due to that one shady `__init__.py` file that was more or less empty hehe.

PKGBUILD
```diff
@@ -12,33 +11,14 @@ url="https://github.com/archlinux/archinstall"
 license=('GPL')
 depends=('python')
 makedepends=('python-setuptools')
+conflicts=('archinstall' 'archinstall-python' 'python-archinstall-git')
```
grazzolini, line 14: The conflicts here seem strange. Shouldn't it be:
conflicts=('archinstall' 'python-archinstall' 'python-archinstall-git')

Also, since we're removing the split package, we will need a provides()

pyfisch: That's a typo.

Like that?

```
provides=('python-archinstall')
conflicts=('archinstall' 'python-archinstall' 'python-archinstall-git')
```

grazzolini: Yes, since we are changing the name to archinstall, provides=('python-archinstall') is enough.

PR #179, opened 2021-04-05
i3 and 13-gaps support

advaithm:
# Pull Request Template

## Description

This adds support for i3 and i3-gaps

## How Has This Been Tested?

As an example:

**Test Configuration**:
* Hardware: KVM/QEMU 
* Specific steps: Ran installer with additional packages `zsh` and `wget`

## Checklist:

### DIFF
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -317,6 +317,9 @@ def perform_installation(device, boot_partition, language, mirrors):
 			# Perform a copy of the config
 			if archinstall.arguments.get('nic', None) == 'Copy ISO network configuration to installation':
 				installation.copy_ISO_network_config(enable_services=True) # Sources the ISO network configuration to the install medium.
+			elif archinstall.arguments.get('nic',None) == None:
+				#skip if we don't have a network
+				pass
 			elif archinstall.arguments.get('nic',{}).get('NetworkManager',False):
 				installation.add_additional_packages("networkmanager")
 				installation.enable_service('NetworkManager.service')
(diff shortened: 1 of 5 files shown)


examples/guided.py
```diff
@@ -344,7 +344,11 @@ def perform_installation(device, boot_partition, language, mirrors):
 
 			if (root_pw := archinstall.arguments.get('!root-password', None)) and len(root_pw):
 				installation.user_set_pw('root', root_pw)
-
+			if archinstall.arguments.get('profile', None) == "i3-wm" or archinstall.arguments.get('profile', None) == "i3-gaps":
+				print("the installation of i3/i3-gaps does not conatain any configuerations for the wm. in this shell you should add your configuerations")
+				subprocess.check_call("arch-chroot /mnt")
```
Torxed, line 349: Just a note here, I think we should design a `_post_instruction()` function for profiles, much like we have for `_prep_function()`. As this might grow out of control if other profiles start adding the same code to guided. This way guided can support multiple profiles with the same need - from one block of code rather than having to check `if profile == "someProfile"`.

Perhaps:
```python
if profile.has_post_function():
    profile._post_function()
```
And then we copy and adopt `Profile().has_prep_function()` into a new one called `Profile().has_post_function()`.

Torxed (approved):
See the line-specific comment, I'm willing to merge this as-is, but I'll most likely change this in a future commit to accommodate the mentioned functionality.

examples/guided.py
advaithm: Yeah I'll add that in. Its a bit late for me so I'll do it in the morning

Torxed: I'll merge this for now and you can add that as a new feature in the morning. As it will add new functionality it's nice to have a separate PR for it.

PR #187, opened 2021-04-05
Fix PKGBUILD conflicts and add provides line

pyfisch:
Addition to #178 

@grazzolini You were too fast, I didn't update the PKGBUILD before.


Torxed:
My bad, I didn't check that specific line, I went on the resolved mark hehe.

PR #190, opened 2021-04-06
post install hook

advaithm:
# Pull Request Template

## Description

Please include a summary of the change and which issue is fixed.<br>
It is also helpful to add links to online documentation or to the implementation of the code you are changing.

## Bugs and Issues

this makes a standard hook for profiles to use to manage post_install hooks

## How Has This Been Tested?

As an example:

**Test Configuration**:
* Hardware: VirtualBox 6.1
* Specific steps: Ran installer with additional packages `nano` and `wget`

## Checklist:

### DIFF
--- a/archinstall/lib/profiles.py
+++ b/archinstall/lib/profiles.py
@@ -177,6 +177,21 @@ class Profile(Script):
 					if hasattr(imported, '_prep_function'):
 						return True
 		return False
+	def has_post_install(self):
+		with open(self.path, 'r') as source:
+			source_data = source.read()
+
+			# Some crude safety checks, make sure the imported profile has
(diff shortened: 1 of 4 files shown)


Torxed (approved):
Overall great contribution, really love the post hook!
A few things I tweaked:
 1. Removed the `bg='black'` since we should avoid using it *(it looks like a copy of my old code which I should change too)* as well as rephrase the log to say *"post configuration requirements was not fulfilled"* instead of *"preparation"*:
 2. Instead of `print()` I used `installation.log("...", fg="yellow")` so that the users don't miss the output here.
 3. I added return values from `_post_installation()` so that the log + exit doesn't catch every time.

Issue #166, opened 2021-04-04
Add sound server selection (pipewire or pulseaudio)


dylanmtaylor:
With pipewire being a viable alternative to pulseaudio, it would be useful to have a prompt to install pipewire instead, and pull in packages for compatibility like pipewire-media-session, pipewire-pulse, pipewire-alsa, pipewire-jack and gst-plugin-pipewire.

Torxed:
I think this can be achieved with a application profile under [profiles/application/pipewire.py](https://github.com/archlinux/archinstall/tree/master/profiles/applications). And by matching on packages pull that app profile in automatically if `pipewire` is added as an additional package.

At least this is one concept I've been tinkering with to get multi-profile-installation-support without having to ask the user for which profile**s** they want installed. Would mean they could do:

```
Enter any additional packages to install: pipewire
```

And it would source the profile `profiles/applications/pipewire.py` and do the correct setup steps that's required outside of installing the package itself. Would this be desirable?

dylanmtaylor:
> I think this can be achieved with a application profile under [profiles/application/pipewire.py](https://github.com/archlinux/archinstall/tree/master/profiles/applications). And by matching on packages pull that app profile in automatically if `pipewire` is added as an additional package.
> 
> At least this is one concept I've been tinkering with to get multi-profile-installation-support without having to ask the user for which profile**s** they want installed. Would mean they could do:
> 
> ```
> Enter any additional packages to install: pipewire
> ```
> 
> And it would source the profile `profiles/applications/pipewire.py` and do the correct setup steps that's required outside of installing the package itself. Would this be desirable?

Hmm, treating the sound server like an application seems like an odd choice. Perhaps better:

```
Select the desired sound server:
1. pipewire
2. pulseaudio
3. jack
4. alsa
```

Torxed:
> > I think this can be achieved with a application profile under [profiles/application/pipewire.py](https://github.com/archlinux/archinstall/tree/master/profiles/applications). And by matching on packages pull that app profile in automatically if `pipewire` is added as an additional package.
> > At least this is one concept I've been tinkering with to get multi-profile-installation-support without having to ask the user for which profile**s** they want installed. Would mean they could do:
> > ```
> > Enter any additional packages to install: pipewire
> > ```
> > 
> > 
> > And it would source the profile `profiles/applications/pipewire.py` and do the correct setup steps that's required outside of installing the package itself. Would this be desirable?
> 
[...]

PR #208, opened 2021-04-06
Implement PipeWire sound server option

dylanmtaylor:
VirtualBox_Arch_06_04_2021_19_27_31

Reference: https://wiki.archlinux.org/index.php/PipeWire

This pull request implements an option to install the pipewire audio server (issue https://github.com/archlinux/archinstall/issues/166). 

# Pull Request Template

## Description

Please include a summary of the change and which issue is fixed.<br>
It is also helpful to add links to online documentation or to the implementation of the code you are changing.

## Bugs and Issues

If this pull-request fixes an issue or a bug, please mention the issues with the approriate issue referece *(Example: &#35;8)*.

## How Has This Been Tested?

As an example:

**Test Configuration**:
* Hardware: VirtualBox 6.1
* Specific steps: Ran installer with additional packages `nano` and `wget`

## Checklist:

### DIFF
--- a/archinstall/lib/user_interaction.py
+++ b/archinstall/lib/user_interaction.py
@@ -92,7 +92,7 @@ def ask_for_a_timezone():
 		
 def ask_for_audio_selection():
 	audio = "pulseaudio" # Default for most desktop environments
-	pipewire_choice = input("Would you like to install the pipewire audio server? [Y/n] ").lower()
+	pipewire_choice = input("Would you like to install pipewire instead of pulseaudio as the default audio server? [Y/n] ").lower()
 	if pipewire_choice == "y":
 		audio = "pipewire"
 
--- a/archinstall/lib/user_interaction.py
+++ b/archinstall/lib/user_interaction.py
@@ -89,6 +89,14 @@ def ask_for_a_timezone():
 			level=LOG_LEVELS.Warning,
 			fg='red'
 		)
+		
+def ask_for_audio_selection():
+	audio = "pulseaudio" # Default for most desktop environments
+	pipewire_choice = input("Would you like to install the pipewire audio server? [Y/n] ").lower()
+	if pipewire_choice == "y":
+		audio = "pipewire"
+
+	return audio
 
 def ask_to_configure_network():
 	# Optionally configure one network interface.
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -337,6 +337,9 @@ def perform_installation(device, boot_partition, language, mirrors):
 			if archinstall.arguments.get('audio', None) == 'pipewire':
 				print('Installing pipewire ...')
 				installation.add_additional_packages(["pipewire", "pipewire-alsa", "pipewire-docs", "pipewire-jack", "pipewire-media-session", "pipewire-pulse", "gst-plugin-pipewire", "libpulse"])
+			elif archinstall.arguments.get('audio', None) == 'pulseaudio':
+				print('Installing pulseaudio ...')
+				installation.add_additional_packages("pulseaudio")
 			
 			if archinstall.arguments.get('packages', None) and archinstall.arguments.get('packages', None)[0] != '':
 				installation.add_additional_packages(archinstall.arguments.get('packages', None))
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -182,6 +182,10 @@ def ask_user_questions():
 				)
 				exit(1)
 
+	# Ask about audio server selection (this right now just asks for pipewire and defaults to pulseaudio otherwise)
+	if not archinstall.arguments.get('audio', None):
+		archinstall.arguments['audio'] = archinstall.ask_for_audio_selection()
+
 	# Additional packages (with some light weight error handling for invalid package names)
 	if not archinstall.arguments.get('packages', None):
 		archinstall.arguments['packages'] = [package for package in input('Write additional packages to install (space separated, leave blank to skip): ').split(' ') if len(package)]
@@ -329,7 +333,11 @@ def perform_installation(device, boot_partition, language, mirrors):
 				installation.enable_service('systemd-networkd')
 				installation.enable_service('systemd-resolved')
 
-
+			print('This audio server will be used: ' + archinstall.arguments.get('audio', None))
+			if archinstall.arguments.get('audio', None) == 'pipewire':
+				print('Installing pipewire ...')
+				installation.install_profile('pipewire')
+			
 			if archinstall.arguments.get('packages', None) and archinstall.arguments.get('packages', None)[0] != '':
 				installation.add_additional_packages(archinstall.arguments.get('packages', None))
 
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -333,7 +333,7 @@ def perform_installation(device, boot_partition, language, mirrors):
 				installation.enable_service('systemd-networkd')
 				installation.enable_service('systemd-resolved')
 
-			print('This audio server will be used: ' + archinstall.arguments.get('audio', None))
+			installation.log(f"The {archinstall.arguments.get('audio', None)} audio server will be used.", level=archinstall.LOG_LEVELS.Info)
 			if archinstall.arguments.get('audio', None) == 'pipewire':
 				print('Installing pipewire ...')
 				installation.add_additional_packages(["pipewire", "pipewire-alsa", "pipewire-docs", "pipewire-jack", "pipewire-media-session", "pipewire-pulse", "gst-plugin-pipewire", "libpulse"])
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -336,7 +336,7 @@ def perform_installation(device, boot_partition, language, mirrors):
 			print('This audio server will be used: ' + archinstall.arguments.get('audio', None))
 			if archinstall.arguments.get('audio', None) == 'pipewire':
 				print('Installing pipewire ...')
-				installation.install_profile('pipewire')
+				installation.add_additional_packages(["pipewire", "pipewire-alsa", "pipewire-docs", "pipewire-jack", "pipewire-media-session", "pipewire-pulse", "gst-plugin-pipewire", "libpulse"])
 			
 			if archinstall.arguments.get('packages', None) and archinstall.arguments.get('packages', None)[0] != '':
 				installation.add_additional_packages(archinstall.arguments.get('packages', None))
--- a/profiles/applications/pipewire.py
+++ b/profiles/applications/pipewire.py
@@ -1,5 +1,5 @@
 import archinstall
 
-__packages__ = ["pipewire", "pipewire-alsa", "pipewire-docs", "pipewire-jack", "pipewire-media-session", "pipewire-pulse", "gst-plugin-pipewire", "libpulse"]
+packages = ["pipewire", "pipewire-alsa", "pipewire-docs", "pipewire-jack", "pipewire-media-session", "pipewire-pulse", "gst-plugin-pipewire", "libpulse"]
 
-installation.add_additional_packages(__packages__)
+installation.add_additional_packages(packages)
--- a/profiles/applications/pipewire.py
+++ b/profiles/applications/pipewire.py
@@ -1,0 +1,5 @@
+import archinstall
+
+__packages__ = ["pipewire", "pipewire-alsa", "pipewire-docs", "pipewire-jack", "pipewire-media-session", "pipewire-pulse", "gst-plugin-pipewire", "libpulse"]
+
+installation.add_additional_packages(__packages__)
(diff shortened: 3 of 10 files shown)


dylanmtaylor:
Need to enable/start the following:

pipewire-media-session.service pipewire-pulse.service pipewire.service pipewire.socket pipewire-pulse.socket 

Most guides show this with --user so I'm going to see if I can use the regular mechanism to do this.

Also need to uninstall (if installed) pulseaudio and pulseaudio-bluetooth

dylanmtaylor:
I don't like the install flow - probably need it earlier with other major selections like DE and password, etc. Especially since it's in all DEs. For now, having a choice of pipewire and pulseaudio would probably be good enough as I don't personally care to implement jack.

dylanmtaylor:
@Torxed when you get a chance could you take a look at this work in progress and let me know how you feel about the way it's structured? I think it might work, but it still needs lots of testing.

dylanmtaylor:
Hmm... installation completes and some of the pipewire packages are installed, but it is missing among other things pipewire-pulse.
VirtualBox_Arch_06_04_2021_19_34_40

dylanmtaylor:
With the latest commit, it's at least working. 

VirtualBox_Arch_06_04_2021_20_54_17

dylanmtaylor:
Just did an interactive rebate to clean up my messy commits and I also tested this to work I cinnamon and KDE. 👍 This is ready for review now.

jatin-cbs:
I request you to also implement jack as many people use Pipewire for better jack support (for example doing minimal stuff with cadence), also on latest version of Fedora i.e Fedora 34 both Pipewire pulse and pipewire jack are default because of that reason.

 https://fedoraproject.org/wiki/Changes/DefaultPipeWire

Torxed:
Great work on this @dylanmtaylor, let me know if you have any objections to my last two changes.
I mainly rephrased the question so that users know pulseaudio will be default if they specify no.
Also change to `log` instead of `print`. If you're fine with this I'll merge it later today :)

grazzolini:
What about alsa? jack? pro-audio? No audio at all? This seems to be rushed to get an audio selection at install time, which should be something done post installation.

dylanmtaylor:
> What about alsa? jack? pro-audio? No audio at all? This seems to be rushed to get an audio selection at install time, which should be something done post installation.

I'm open to adding more audio servers if desired, but pipewire/pulseaudio

dylanmtaylor:
> Great work on this @dylanmtaylor, let me know if you have any objections to my last two changes.
> I mainly rephrased the question so that users know pulseaudio will be default if they specify no.
> Also change to `log` instead of `print`. If you're fine with this I'll merge it later today :)

I'm fine with it

dylanmtaylor:
> I request you to also implement jack as many people use Pipewire for better jack support.
> 
> The reason is if jack will still go without pipewire and is using the normal jack daemon it will cause issues ! As Pipewire handles pro audio better. So I recommend you to also enable pipewire jack. @dylanmtaylor thanks for bringing pipewire up ! 
> 
> Update : Sorry I didn't look at the PR changes.. pipewire-jack is there.
> [93750db](https://github.com/archlinux/archinstall/commit/93750dbd223269804c2a9a3a7dbdf30d4f17ac82)

Yes, this is correct, I included the pipewire servers for jack and alsa in there too :) It's great how it tries to be compatible with applications no matter what they use.

dylanmtaylor:
To add other servers in the future, the conditional just needs to be passed a new server name. Feel free to switch it to a menu later on, but pipewire is special in that it is likely going to be the future of audio on Linux for *most* people as it tries to do consumer and professional audio use-cases, has compatibility with most existing audio servers, and supports flatpak and other containerized/sandboxed applications having granular permissions to the audio devices, which no other server supports now.

dylanmtaylor:
> What about alsa? jack? pro-audio? No audio at all? This seems to be rushed to get an audio selection at install time, which should be something done post installation.

Almost nobody uses ALSA directly anymore since it has issues with multiple things using a device at once (which is part of why audio servers are a thing in the first place), so I recommend against that one. No objections to having jack and pro-audio added if someone wanted to do that. Having the audio server you want just work out of the box is great for end-users, and with some audio servers that have multiple packages to find/install it removes considerable confusion for the user having it done for them. I see it as a large quality of life improvement, especially with users who have realized some of the benefits of pipewire, for instance, if they are using flatpaks for audio in a sandbox or want lower-latency audio than pulseaudio can provide.

grazzolini:
My point was: this belongs [post installation](https://wiki.archlinux.org/index.php/General_recommendations#Sound_system). The installer can't accommodate all scenarios. It might be the case someone is installing a server with no audio card whatsoever. Arch is about giving the users choice and freedom to do what they want. You're giving them **some** choices here.

Torxed:
> r with no audio card whatsoever

This is a valid point and as it's post installation, I think the compromise here is that the audio server menu **only** exists in the pre-configured profiles (outside of guided core). And perhaps we can include a "skip audioserver" in the desktop profiles as well.

I do use Xorg + DE on some servers for x11 forward purposes.
And in those cases I do not have a sound card. So an option between pulse and pipewaire that says "Skip audioserver" would be nice. I don't think we need to accommodate all the options - 2 most frequently used + option to skip would be ok.

If this feels like a way forward I'd be happy to see that change in this PR before merge.
Again, good healthy discussion!

dylanmtaylor:
> > My point was: this belongs post installation. The installer can't accommodate all scenarios. It might be the case someone is installing a server with no audio card whatsoever. Arch is about giving the users choice and freedom to do what they want. You're giving them some choices here.
> 
> This is a valid point and as it's post installation, I think the compromise here is that the audio server menu **only** exists in the pre-configured profiles (outside of guided core). And perhaps we can include a "skip audioserver" in the desktop profiles as well _(where this logic should probably be moved to)_.
> 
> I do use Xorg + DE on some servers for x11 forward purposes.
> And in those cases I do not have a sound card. So an option between pulse and pipewaire that says "Skip audioserver" would be nice. I don't think we need to accommodate all the options - 2 most frequently used + option to skip would be ok.
> Or drop to a shell for post-configuration. But the 2+skip option feels simpler from a user-starting-out standpoint.
> 
> If this feels like a way forward I'd be happy to see that change in this PR before merge. _(Either of the two options)_
> Again, good healthy discussion!

Maybe we could do something like this - probably in a future PR. On the first profile prompt, only show 

0: minimal
1: desktop

have all the DEs listed under desktop (these all would require an audio server - even if you don't install pipewire, since their dependencies will pull in something)  - don't expose those profiles directly.

As part of the desktop flow, make the audio server selection and other selections including graphics that only apply to desktop installations.

In an even further in the future PR, rework the prompt into a list of servers for desktop, maybe include pro-audio or jack if someone wants to take that work on.

dylanmtaylor:
Implementing it as suggested in my prior comment would satisfy both https://github.com/archlinux/archinstall/issues/210 and https://github.com/archlinux/archinstall/issues/166

grazzolini:
> > > My point was: this belongs post installation. The installer can't accommodate all scenarios. It might be the case someone is installing a server with no audio card whatsoever. Arch is about giving the users choice and freedom to do what they want. You're giving them some choices here.
> > 

> 
> Maybe we could do something like this - probably in a future PR. On the first profile prompt, only show
> 
> 0: minimal
> 1: desktop
> 
> have all the DEs listed under desktop (these all would require an audio server - even if you don't install pipewire, since their dependencies will pull in something) - don't expose those profiles directly.
> 
> As part of the desktop flow, make the audio server selection and other selections including graphics that only apply to desktop installations.
> 
> In an even further in the future PR, rework the prompt into a list of servers for desktop, maybe include pro-audio or jack if someone wants to take that work on.

I think that, if the user chooses a desktop profile, any of them, then, and only then, it should be asked for audio selection. For example, when installing plasma using plasma-meta, it will pull in pulseaudio as one of the dependencies.

dylanmtaylor:
> > > > My point was: this belongs post installation. The installer can't accommodate all scenarios. It might be the case someone is installing a server with no audio card whatsoever. Arch is about giving the users choice and freedom to do what they want. You're giving them some choices here.
> 
> > Maybe we could do something like this - probably in a future PR. On the first profile prompt, only show
> > 0: minimal
> > 1: desktop
> > have all the DEs listed under desktop (these all would require an audio server - even if you don't install pipewire, since their dependencies will pull in something) - don't expose those profiles directly.
> > As part of the desktop flow, make the audio server selection and other selections including graphics that only apply to desktop installations.
> > In an even further in the future PR, rework the prompt into a list of servers for desktop, maybe include pro-audio or jack if someone wants to take that work on.
> 
> I think that, if the user chooses a desktop profile, any of them, then, and only then, it should be asked for audio selection. For example, when installing plasma using plasma-meta, it will pull in pulseaudio as one of the dependencies.

If I understand your comments right, we are in agreement then. To clarify my thoughts, audio would be a subselection of desktop.

dylanmtaylor:
@Torxed since this pull request is against version 2.2.0, not master, would you be opposed to merging it in and letting me try to implement the solution that I mentioned before? From my understanding this isn't going to ship until June or something anyways. Basically I want to start by separating out desktop and minimal, I think a good way to do this is by adding a 'top level' attribute and filtering on this attribute when displaying the profiles for the first time. This would only show desktop and have all of the other desktop profiles under it. The second phase would be to create a minimal profile. It would basically do nothing special except for install the additional packages selected by the user. From there I went to add conditional logic so that unless the user selects desktop, the graphics driver prompt and audio server prompt will not show, as these are not needed unless you need a graphical interface. This still allows the user to install a graphics driver without a DE by using the Xorg profile. Finally, instead of having a yes no choice if anyone wants to take on the work we can switch it to a menu and add some other options.

Torxed:
> @Torxed since this pull request is against version 2.2.0, not master, would you be opposed to merging it in and letting me try to implement the solution that I mentioned before? From my understanding this isn't going to ship until June or something anyways. Basically I want to start by separating out desktop and minimal, I think a good way to do this is by adding a 'top level' attribute and filtering on this attribute when displaying the profiles for the first time. This would only show desktop and have all of the other desktop profiles under it. The second phase would be to create a minimal profile. It would basically do nothing special except for install the additional packages selected by the user. From there I went to add conditional logic so that unless the user selects desktop, the graphics driver prompt and audio server prompt will not show, as these are not needed unless you need a graphical interface. This still allows the user to install a graphics driver without a DE by using the Xorg profile. Finally, instead of having a yes no choice if anyone wants to take on the work we can switch it to a menu and add some other options. If the changes are not ready in time for 2.2.0 we can always use revert to back them out. Git version control is a thing of beauty. I just want to keep the commit log simple and not have too many different changes in one PR.

Because `v2.2.0` is such a living entity, and nothing will be set in stone until after the patch release, I don't mind merging this into the feature branch. People have been getting a bit nervous tho about *some* merges into this feature branch.
So while I'm ok with merging in this, I also think we should clarify (on the branch) some where in the README that it's a living entity and nothing is set.

I'll also go over the milestone to make it more clear to people.
But unless objections, I can merge for now :)

jatin-cbs:
:+1:  Pipewire   https://fedoraproject.org/wiki/Changes/DefaultPipeWire   You can see the reasons here, this is also default in newer linux distros like fedora 34.

jatin-cbs:
Jack client library and Pulse audio daemon used to handle audio in a different way where integration of applications was really bad. But Pipewire solves this. I personally don't think people object to this, however some people may still prefer old way of handling audio so this should be an option but with a (recommended) flag in installation.

grazzolini:
@jatin-cbs I'm aware fedora is moving to pipewire by default. But, pushing stuff onto the user is not the Arch way. Also, pipewire, although is progressing in a impressive fast pace, is not ideal for most people yet. Don't get me wrong, I use pipewire myself, I just don't think it belongs by default on an Arch installer.

Galcian79:
I tested the last build of archinstall and pipewire-pulse is still missing. Maybe a regression?

maukag:
> I tested the last build of archinstall and pipewire-pulse is still missing. Maybe a regression?

Same for me.

Torxed:
> @Galcian79 & @maukag I tested the last build of archinstall and pipewire-pulse is still missing. Maybe a regression?

The package is installed, but we never run `systemctl start --user pipewire-pulse.service` for each user. Which is [something that has to be done](https://wiki.archlinux.org/title/PipeWire#PulseAudio_clients). I've created a separate branch in order to address this. It probably won't be ready in time for `v2.3.1` release - but if it is it'll be there or the next release : )

You can follow the progress in #821.
But the manual workaround should work :)

maukag:
> The package is installed, but we never run `systemctl start --user pipewire-pulse.service` for each user. Which is [something that has to be done]

When I run `systemctl --user status pipewire-pulse`  I get a unit-not-found error. It seems like this service just wasn't included in the auto-installation?

Torxed:
> > The package is installed, but we never run `systemctl start --user pipewire-pulse.service` for each user. Which is [something that has to be done]
> 
> When I run `systemctl --user status pipewire-pulse` I get a unit-not-found error. It seems like this service just wasn't included in the auto-installation?

I would need a `/var/log/archinstall/install.log` to figure out your specific case, because if pipewire was selected in the guided installer, I'm almost certain it will be included: https://github.com/archlinux/archinstall/blob/69b70caa43d1be75417d17c092916373e70e745f/examples/guided.py#L317

PR #223, opened 2021-04-07
Remove gnome-todo and gnome-sound-recorder from default installation

jatin-cbs:
# Pull Request Template

## Description

gnome todo and gnome sound recorder should not be in default installation, honoring the Arch Linux philosophy. These packages are not even pre installed on other major linux distributions. gnome-tweaks is the only package (that is not part of the gnome group https://archlinux.org/groups/x86_64/gnome/) what 90 percent gnome users still use and that should be in default installation and it is.

## How Has This Been Tested?

Doesn't require any specific tests as these are not required packages for default installation.

## Checklist:

### DIFF
--- a/profiles/applications/gnome.py
+++ b/profiles/applications/gnome.py
@@ -1,4 +1,4 @@
 import archinstall
 
-installation.add_additional_packages("gnome gnome-tweaks gnome-todo gnome-sound-recorder gdm")
+installation.add_additional_packages("gnome gnome-tweaks gdm")
 # Note: gdm should be part of the gnome group, but adding it here for clarity


dylanmtaylor:
This makes sense, I'm fine with it.

Torxed:
> This makes sense, I'm fine with it.

On that note, consider it merged! :)
Remember that `master` is the upcoming patch release, due the ~14:th roughly. So if we want to tweak this further there's still time :)

jatin-cbs:
@Torxed  should also be merged into v2.2.0

Torxed:
> @Torxed should also be merged into v2.2.0

Anything that ends up in master does get synced to 2.2.0 automatically.
I do this manually at the end of every night at the moment tho.
But a second PR never hurts hehe. I'm updating the contribution guidelines now so this is reflected.

jatin-cbs:
Ok ! :+1: I also thought that this will be the case as I was checking commits on v2.2.0 branch but still was in doubt.

PR #219, opened 2021-04-07
Add minimal profile and implement idea of 'top-level' profiles

dylanmtaylor:
This will allow for a much cleaner install (minimal) if desired and can be used to move driver and audio server selection so they're only done on a desktop installation type. This closes issue #210

# Pull Request Template

Make sure you've checked out the [contribution guideline](https://github.com/archlinux/archinstall/blob/master/CONTRIBUTING.md).<br>
Most of the guidelines are not enforced, but is heavily encouraged.

## Description

Please include a summary of the change and which issue is fixed.<br>
It is also helpful to add links to online documentation or to the implementation of the code you are changing.

## Bugs and Issues

If this pull-request fixes an issue or a bug, please mention the issues with the approriate issue referece *(Example: &#35;8)*.

## How Has This Been Tested?

If possible, mention any tests you have made with the current code base included in the pull-requests.<br>
Any core-developer will also run tests, but this helps speed things up. Below is a template that can be used:

As an example:

**Test Configuration**:
* Hardware: VirtualBox 6.1
* Specific steps: Ran installer with additional packages `nano` and `wget`

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code to the best of my abilities
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation where possible/if applicable

PR #230, opened 2021-04-07
Only ask for audio server selection on a desktop profile (depends on #219)

dylanmtaylor:
This has a dependency on #219 already being merged. This makes it so that only desktop profiles are prompted for an audio server selection. We do this by having all the desktop environments under the top-level desktop profile and if (and only if) this is selected, we will ask the user which sound server they prefer (currently pipewire or pulseaudio, but may be expanded later).

🚨 PR Guidelines:

# New features *(v2.2.0)*

# Bug fixes *(v2.1.4)*

Merge against `master` for bug fixes and anything that improves stability and quality of life.<br>
This excludes:
 * New functionality
 * Added complexity
 * Breaking changes

Any changes to `master` automatically gets pulled in to `torxed-v2.2.0` to avoid merge hell.

# Describe your PR

# Testing

### DIFF
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -341,7 +341,7 @@ def perform_installation(device, boot_partition, language, mirrors):
 				installation.enable_service('systemd-networkd')
 				installation.enable_service('systemd-resolved')
 
-			if 	archinstall.arguments['audio'] != 'none':
+			if 	archinstall.arguments.get('audio', None) != None:
 				installation.log(f"The {archinstall.arguments.get('audio', None)} audio server will be used.", level=archinstall.LOG_LEVELS.Info)
 				if archinstall.arguments.get('audio', None) == 'pipewire':
 					print('Installing pipewire ...')
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -182,9 +182,16 @@ def ask_user_questions():
 				)
 				exit(1)
 
-	# Ask about audio server selection (this right now just asks for pipewire and defaults to pulseaudio otherwise)
+	# Ask about audio server selection if one is not already set
 	if not archinstall.arguments.get('audio', None):
-		archinstall.arguments['audio'] = archinstall.ask_for_audio_selection()
+		
+		# only ask for audio server selection on a desktop profile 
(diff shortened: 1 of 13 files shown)


dylanmtaylor:
By the way, this is tested and works. 

Desktop installation selected:

desktop-install-flow


Minimal installation selected:

minimal-install-flow

I opened #231 because the line about the additional packages such as web browsers wrapping like that just doesn't look good.

dylanmtaylor:
The pipewire prompt might make people think pipewire is the default if they press enter, as Y is capitalized.

Wondering if we should use 
```
	if pipewire_choice in ("y",""):
```

instead of
```
	if pipewire_choice == "y":
```

dylanmtaylor:
If you don't like that latest commit  (https://github.com/archlinux/archinstall/pull/230/commits/f4233e73f0b70f236a9fd357ee11bf194224a4fb) or find it too controversial, feel free to revert it, it's not like it took me very long to develop. It just feels like it might confuse people if the Y is capital and they assume that is the default behavior

examples/guided.py
```diff
@@ -333,13 +341,14 @@ def perform_installation(device, boot_partition, language, mirrors):
...
 				installation.enable_service('systemd-resolved')
 
-			installation.log(f"The {archinstall.arguments.get('audio', None)} audio server will be used.", level=archinstall.LOG_LEVELS.Info)
-			if archinstall.arguments.get('audio', None) == 'pipewire':
-				print('Installing pipewire ...')
-				installation.add_additional_packages(["pipewire", "pipewire-alsa", "pipewire-docs", "pipewire-jack", "pipewire-media-session", "pipewire-pulse", "gst-plugin-pipewire", "libpulse"])
-			elif archinstall.arguments.get('audio', None) == 'pulseaudio':
-				print('Installing pulseaudio ...')
-				installation.add_additional_packages("pulseaudio")
+			if 	archinstall.arguments['audio'] != 'none':
```
Torxed, line 344: This might break, if `--audio` is never given. We should use `arguments.get()` here.

PR #232, opened 2021-04-08
removed post_install hook for kde plasma

advaithm:
🚨 PR Guidelines:

# New features *(v2.2.0)*

# Bug fixes *(v2.1.4)*

As @dylanmtaylor  using mv can break stuff during updates. Until we find a better way I think this is the best choice

# Describe your PR

this comments out the post install hook in the kde profile

# Testing

### DIFF
--- a/profiles/kde.py
+++ b/profiles/kde.py
@@ -20,16 +20,12 @@ def _prep_function(*args, **kwargs):
 		else:
 			print('Deprecated (??): xorg profile has no _prep_function() anymore')
 
-""" def _post_install(*args, **kwargs):
+def _post_install(*args, **kwargs):
 	if "nvidia" in _gfx_driver_packages:
 		print("Plasma Wayland has known compatibility issues with the proprietary Nvidia driver")
-	choice = input("Would you like plasma-wayland to be the default session [Y/n] ").lower()
-	if choice == "y":
-		installation.arch_chroot("mv /usr/share/xsessions/plasma.desktop /usr/share/xsessions/plasmax11.desktop")
-		installation.arch_chroot("mv /usr/share/wayland-sessions/plasmawayland.desktop /usr/share/wayland-sessions/plasma.desktop")
+	print("After booting, you can choose between Wayland and Xorg using the drop-down menu")
 	return True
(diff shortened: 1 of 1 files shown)


dylanmtaylor:
To clarify, as there isn't much context in this PR, I noticed that the default behavior isn't very reliable - it seems that SDDM chooses a default based on the filename (?) from what @advaithm discovered in how Fedora sets the default session here: https://src.fedoraproject.org/rpms/plasma-workspace/blob/rawhide/f/plasma-workspace.spec#_701

The issue is that if we move files (as we are not packagers, and have no control over where updates for session files get written), we might inadvertantly break updating the plasma-workspace or plasma-wayland-session packages. This could cause some weird behavior that could either require user intervention (specific to the way we install KDE, not general to Arch and therefore not listed on the manual interventions newsletter), or cause really strange behavior like two different X11 sessions in the SDDM list, one from when the system was installed, and another from when the package was updated.

As a matter of exercising an abundance of caution, we agreed that until we find a better way to set a default (and it looks like one cannot be set in sddm.conf), it is best to NOT attempt to rename files. It is easy enough for the user to select their desired session from a drop-down on first boot that removing this selection is not a big deal.

dylanmtaylor:
I checked out this commit from live media and tested it.

Checking it's the right commit:
VirtualBox_ArchLinux_07_04_2021_22_17_51

After installation (no default session question asked):
VirtualBox_ArchLinux_07_04_2021_22_42_33

Booted to a working SDDM:
VirtualBox_ArchLinux_07_04_2021_22_43_11

Notably, Wayland seemed to be the default.

advaithm:
Hmmm no idea why

dylanmtaylor:
> Hmmm no idea why

So, I have an idea. Why not just do something like this and let the user set the default themselves?

```
def _post_install(*args, **kwargs):
	print("After booting, you can choose between Wayland and Xorg using the drop-down menu")
	return True
```

advaithm:
Sure I think that's the best move for now

dylanmtaylor (approved):
Looks good to me

PR #238, opened 2021-04-08
Move some packages from awesome to desktop generic profile and add wireless/SMART tools 

### LINKED ISSUE #98: Look in to enabling SMART for drives that support it
Something like `smartctl --smart=on --offlineauto=on --saveauto=on /dev/sda` where `archinstall.hardware.detectSmart()` finds drives that support it (to extend drive lifetime if possible).

dylanmtaylor:
This moves some common packages from Awesome to desktop.

https://gitlab.archlinux.org/archlinux/archiso/-/blob/master/configs/releng/packages.x86_64 lists packages that come out of the box in the arch iso. One situation we want to avoid is laptop users not able to connect to wifi after installation, so I'm adding wifi config utilities to avoid these users having to do a reinstallation.

This also pulls in smartmontools to close #98.

### DIFF
--- a/profiles/desktop.py
+++ b/profiles/desktop.py
@@ -4,6 +4,10 @@ import archinstall, os
 
 is_top_level_profile = True
 
+# New way of defining packages for a profile, which is iterable and can be used out side
+# of the profile to get a list of "what packages will be installed".
+__packages__ = ['nano', 'openssh', 'htop', 'wget', 'iwd', 'wireless_tools', 'wpa_supplicant', 'smartmontools']
+
 def _prep_function(*args, **kwargs):
 	"""
 	Magic function called by the importing installer
@@ -14,7 +18,7 @@ def _prep_function(*args, **kwargs):
 
 	supported_desktops = ['gnome', 'kde', 'awesome']
 	desktop = archinstall.generic_select(supported_desktops, 'Select your desired desktop environment: ')
-
+	
 	# Temporarily store the selected desktop profile
 	# in a session-safe location, since this module will get reloaded
 	# the next time it gets executed.
@@ -41,7 +45,11 @@ if __name__ == 'desktop':
 	There are plenty of desktop-turn-key-solutions based on Arch Linux,
 	this is therefore just a helper to get started
 	"""
+	
+	# Install common packages for all desktop environments
+	installation.add_additional_packages(__packages__)
 
 	# TODO: Remove magic variable 'installation' and place it
 	#       in archinstall.storage or archinstall.session/archinstall.installation
 	installation.install_profile(archinstall.storage['_desktop_profile'])
+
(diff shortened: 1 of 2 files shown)


dylanmtaylor:
Commenting out the post-install hook caused indent issues, fixing that as well. Let me test and fix more before we merge this

archinstall/lib/profiles.py
```diff
@@ -177,7 +177,7 @@ def has_prep_function(self):
 					if hasattr(imported, '_prep_function'):
 						return True
 		return False
-"""
+
```
Torxed, line 180: We can leave the comments there to keep the function "out".
But the comments need to be in the correct indentation level: c5d099c

Torxed (changes requested):

archinstall/lib/profiles.py
dylanmtaylor: I'm fine if you want to make that commit and leave it commented though.

dylanmtaylor:
In case you're wondering about the wireless tools packages, this one personally bit me on a new KDE install, so I wanted to make sure I could at least connect to wifi after installing. The firmware was there, but I didn't have access to iwctl.
With this change, on a new install I can run systemctl start iwd, then iwctl and use the station wlan0 connect AP_Name to get online.

archinstall/lib/profiles.py
dylanmtaylor: Okay, I see you fixed the indentation in that commit in master, I'll drop this from the change.

dylanmtaylor:
@Torxed changed to be just one commit, and no longer uncomments the post-install hook.

profiles/desktop.py
```diff
@@ -4,6 +4,10 @@
 
 is_top_level_profile = True
 
+# New way of defining packages for a profile, which is iterable and can be used out side
+# of the profile to get a list of "what packages will be installed".
+__packages__ = ['nano', 'openssh', 'htop', 'wget', 'iwd', 'wireless_tools', 'wpa_supplicant', 'smartmontools']
```
Torxed, line 9: I wonder if we really need the wireless tools.
They are pulled in if they opt in to configure the network from the option "Copy configuration from the ISO" I believe.

dylanmtaylor: Honestly, maybe not, but I will say that I wasn't super happy that I had no way of connecting to a network on a new install and had to boot off the ISO and do a whole new installation (I mean I could have plugged in a USB ethernet adapter but that's another story)

dylanmtaylor: Anything that leaves the system in a state where a user can't easily use it without a reinstall is high-priority in my mind.

Torxed: > Anything that leaves the system in a state where a user can't easily use it without a reinstall is high-priority in my mind.

Agreed. But this is what the warning in the network-step is for I think.
I don't mind them being in desktop tho, seeing as it's optional anyway.
But might be worth re-visiting later *(and if so slim PR's like we did with KDE and Gnome)*.

I'll leave them in for now since it doesn't bloat the guided which is the main concern : )

PR #241, opened 2021-04-08
Implement Sway and use SDDM as Greeter

dylanmtaylor:
VirtualBox_ArchLinux_08_04_2021_15_05_28
VirtualBox_ArchLinux_08_04_2021_15_06_03

🚨 PR Guidelines:

# New features *(v2.2.0)*

# Bug fixes *(v2.1.4)*

Merge against `master` for bug fixes and anything that improves stability and quality of life.<br>
This excludes:
 * New functionality
 * Added complexity
 * Breaking changes

Any changes to `master` automatically gets pulled in to `torxed-v2.2.0` to avoid merge hell.

# Describe your PR

# Testing

### DIFF
--- a/profiles/applications/sway.py
+++ b/profiles/applications/sway.py
@@ -1,0 +1,3 @@
+import archinstall
+
+installation.add_additional_packages("sway sddm")
--- a/profiles/desktop.py
+++ b/profiles/desktop.py
@@ -16,7 +16,7 @@ def _prep_function(*args, **kwargs):
 	for more input before any other installer steps start.
 	"""
 
-	supported_desktops = ['gnome', 'kde', 'awesome']
+	supported_desktops = ['gnome', 'kde', 'awesome', 'sway']
 	desktop = archinstall.generic_select(supported_desktops, 'Select your desired desktop environment: ')
 	
 	# Temporarily store the selected desktop profile
--- a/profiles/sway.py
+++ b/profiles/sway.py
@@ -1,0 +1,24 @@
+# A desktop environment using "Sway"
+
+import archinstall
+
+is_top_level_profile = False
+
+def _prep_function(*args, **kwargs):
+	"""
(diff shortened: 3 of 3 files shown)


advaithm:
I don't think we should use any greeter for sway. officially sway does not support any display managers https://github.com/swaywm/sway/pull/3634#issuecomment-462779163

dylanmtaylor:
> I don't think we should use any greeter for sway. officially sway does not support any display managers [swaywm/sway#3634 (comment)](https://github.com/swaywm/sway/pull/3634#issuecomment-462779163)

It seems to work and might confuse users if they're booted to a system with no graphical login prompt. If we run into issues we can pull it out though

advaithm:
We have a postinstall hook to tell the user to run `sway` to start it

advaithm:
We also should raise a hard incompatibility error if a person uses the proprietary nvidia drivers since sway cannot run with the proprietary drivers

dylanmtaylor:
> We also should raise a hard incompatibility error if a person uses the proprietary nvidia drivers since sway cannot run with the proprietary drivers

You won't get this case right now - since there is no reliance on Xorg this profile doesn't let you pick a driver. Nvidia works with noveau so we could install that instead for sway.

advaithm:
Dude you literally can't run sway with the proprietary drivers. It has no idea how to use nvidia drivers. Also sway is wayland only. Sddm pulls xorg.

Torxed:
> We have a postinstall hook to tell the user to run `sway` to start it

I think that we could add a sway specific option to the sway profile that aks:
```
Would you like to autostart sway on login [Y/n]: 
```
And have the installer set up one of the two supported methods mentioned in [Sway#Starting *(arch wiki)*](https://wiki.archlinux.org/index.php/Sway#Starting)

advaithm:
sure

Issue #225, opened 2021-04-07
After install, prompt the user if they want to chroot into the installation


dylanmtaylor:
I think it would be a good idea to add a prompt to ask the user if they want to chroot into the newly created installation after the installation steps are executed.

For instance in pseudo-code, 

```
choice = input("Would you like to chroot into the newly created installation? [Y/n]")
if choice.lower() == 'y':
     execute('arch-chroot /mnt');
```

Torxed:
I think this is a nice touch, it explains what's about to happen and it gives an easy way to jump into the new installation.
Also add the "finished" output after it:
```python
choice = input("Would you like to chroot into the newly created installation and perform post installation configurations? [Y/n]")
if choice.lower() in ("y", ""):
    installation.log("For post installations tips, see https://wiki.archlinux.org/index.php/Installation_guide#Post-installation", fg="yellow")
    try:
        subprocess.check_call("arch-chroot /mnt", shell=True)
    except:
        pass
finally:
    installation.log("Installation completed without any errors. You may now reboot."
```

dylanmtaylor:
> I think this is a nice touch, it explains what's about to happen and it gives an easy way to jump into the new installation.
> Also add the "finished" output after it:
> 
> ```python
> choice = input("Would you like to chroot into the newly created installation and perform post installation configurations? [Y/n]")
> if choice.lower() in ("y", ""):
>     installation.log("For post installations tips, see https://wiki.archlinux.org/index.php/Installation_guide#Post-installation", fg="yellow")
>     try:
>         subprocess.check_call("arch-chroot /mnt", shell=True)
>     except:
>         pass
> finally:
>     installation.log("Installation completed without any errors. You may now reboot."
> ```

I like this a lot. I think it'd be great to add this in.

dylanmtaylor:
By the way, is there a shortened link to that? from the installer, it's not like you can copy/paste https://wiki.archlinux.org/index.php/Installation_guide#Post-installation.

Torxed:
> By the way, is there a shortened link to that? from the installer, it's not like you can copy/paste https://wiki.archlinux.org/index.php/Installation_guide#Post-installation.

Not that I know of. So printing it in full is probably the only option, I hope people would search it in their favorite search engine and find it instead of writing off the whole thing.

advaithm:
Maybe we can use a link shortener for this. Also if we do add this. I think it should be fine to change the post install hooks in i3 and sway to just print out a message stating we don't ship a default config

[...]

PR #247, opened 2021-04-09
draft: Cherry-Pick i3 Profiles and Make Common i3 Package

dylanmtaylor:
This pulls in cherry-picked original i3 commits with authorship and history from 2.2.0.

On top of that, I implemented a common i3 profile that lets you choose a configuration for i3: i3-gaps or i3-wm. This should keep the installation menu from being cluttered.

It's a draft because this is UNTESTED. Use at your own risk, but if you do test it and it works, please let me know. :)

### DIFF
--- a/profiles/applications/i3-gaps.py
+++ b/profiles/applications/i3-gaps.py
@@ -1,0 +1,2 @@
+import archinstall
+installation.add_additional_packages("i3lock i3status i3blocks i3-gaps")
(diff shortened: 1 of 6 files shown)


dylanmtaylor:
I do want to rip out this

```
	installation.log("the installation of i3 does not conatain any configuerations for the wm. In this shell you should take your time to add your desiired configueration. Exit the shell once you are done to continue the installation.", fg="yellow")
	try:
		subprocess.check_call("arch-chroot /mnt",shell=True)
	except subprocess.CallProcessError:
		return False

``` 

and put the chroot option in after every installation as mentioned in #225

dylanmtaylor:
Actually, since we don't have post-install hooks in this branch anyways, I'm going to remove the chroot prompts from i3, since I want to implement chroot globally for every profile and this will cause it to ask to chroot twice down-the-line.

PR #250, opened 2021-04-09
OOB in generic_selection

insanemal:
Out of bounds check in generic_selection is using >= on list. Lists are zero based. If you put in a value that equals the number of items in the list you get an out of bounds error. 

Removed the equals part of the test as last item in list/dictionary items is len(list)-1 not len(list)

Tested in VM and small netbook.


Torxed:
Nicely spotted!

PR #256, opened 2021-04-09
Merge cherry-picked patches for NetworkManager and PipeWire

dylanmtaylor:
This is a cherry-pick from `torxed-2.2.0` into `master` to bring in NetworkManager and PipeWire.

### DIFF
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -285,7 +285,7 @@ def perform_installation(device, boot_partition, language, mirrors):
 			print('This audio server will be used: ' + archinstall.arguments.get('audio', None))
 			if archinstall.arguments.get('audio', None) == 'pipewire':
 				print('Installing pipewire ...')
-				installation.install_profile('pipewire')
+				installation.add_additional_packages(["pipewire", "pipewire-alsa", "pipewire-docs", "pipewire-jack", "pipewire-media-session", "pipewire-pulse", "gst-plugin-pipewire", "libpulse"])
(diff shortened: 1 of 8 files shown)


examples/guided.py
```diff
@@ -160,6 +160,17 @@ def ask_user_questions():
...
+	# Ask about audio server selection if one is not already set
+	if not archinstall.arguments.get('audio', None):
+		
+		# only ask for audio server selection on a desktop profile 
+		if str(archinstall.arguments['profile']) == 'Profile(desktop)':
+			archinstall.arguments['audio'] = archinstall.ask_for_audio_selection()
+		else:
+			# packages installed by a profile may depend on audio and something may get installed anyways, not much we can do about that.
+			# we will not try to remove packages post-installation to not have audio, as that may cause multiple issues
+			archinstall.arguments['audio'] = 'none'
```
Torxed, line 172: For future reference, since this is only triggered by the `desktop` profile.
We should move this into the desktop profile itself, just to give peace of mind to anyone looking at the default standard installation proceedure. It doesn't bloat any installations, it's just a code health. But since this is a cherry pick, I don't think we should change it in this commit/PR. But I'll take this on in a few days.

Torxed (approved):

examples/guided.py
dylanmtaylor: I'm not opposed to this at all :)

PR #260, opened 2021-04-09
Switch budgie to gdm since lightdm had issues

dylanmtaylor:


dylanmtaylor:
My issue was not caused by lightdm, but VirtualBox - it was loading very slowly (black screen) but eventually came up. I thought it might be greeter related. Closing.

PR #267, opened 2021-04-10
Switch from LightDM to SDDM Everywhere

dylanmtaylor:
SDDM looks considerably more appealing out of the box than lightdm. Here's an example with XFCE4 to demonstrate what I'm talking about.

Before:
VirtualBox_arch_09_04_2021_23_11_44

After:
VirtualBox_arch_09_04_2021_23_19_40

### DIFF
--- a/profiles/applications/budgie.py
+++ b/profiles/applications/budgie.py
@@ -1,4 +1,4 @@
 import archinstall
 
 # "It is recommended also to install the gnome group, which contains applications required for the standard GNOME experience." - Arch Wiki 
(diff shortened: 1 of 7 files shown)


advaithm:
this breaks in built functionality for a lot of DEs -1 from me.

dylanmtaylor:
> this breaks in built functionality(namely theming) for a lot of DEs -1 from me.

so, sddm can't be themed in DEs other than KDE Plasma, that's true, but I'd like to know if it breaks other things.  lightdm can be made to look good but out of the box it's pretty bad. I was hoping to have some more polish out of the box.

advaithm:
look this is arch. I don't see the point in making it war more complicated for the end user when we can stick to default stuff

PR #268, opened 2021-04-10
Remove lightdm-gtk-greeter-settings

dylanmtaylor:
lightdm-gtk-greeter-settings is just a GUI to configure the lightdm GTK greeter. I don't think it's needed out of the box.


Torxed:
Doesn't feel like it's crucial to getting going or even "start using" the system.
So this is a welcome change :)

PR #275, opened 2021-04-10
GitHub Actions: Build Arch Live ISO With Each Commit Included

dylanmtaylor:
Screenshot from 2021-04-10 17-49-08

VirtualBox_arch_10_04_2021_20_02_01

So this is a really neat CI/CD feature I'm testing out - we can have GitHub automatically build us ISOs to test to avoid having to replace ArchInstall on a vanilla Arch image. I want to get it to automatically launch as well - I think this will prove to be incredibly useful when done.

This will enable the following workflow when reviewing a change/pull request:

* Download ISO
* Start up VM or Write to USB
* The specific ArchInstall commit will be automatically installed and you can run it with python -m archinstall

### DIFF
--- a/.github/workflows/iso-build.yaml
+++ b/.github/workflows/iso-build.yaml
@@ -1,0 +1,27 @@
+# This workflow will build an Arch Linux ISO file with the commit on it
+
+name: Build Arch ISO with ArchInstall Commit
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container:
+      image: archlinux:latest
+      options: --privileged
+    steps:
+    - uses: actions/checkout@v2
+    - run: pwd
+    - run: find .
+    - run: cat /etc/os-release
+    - run: mkdir -p /tmp/archlive/airootfs/root/archinstall-git; cp -r . /tmp/archlive/airootfs/root/archinstall-git
+    - run: pacman -Sy; pacman --noconfirm -S git archiso
+    - run: cp -r /usr/share/archiso/configs/releng/* /tmp/archlive
+    - run: echo -e "git\npython\npython-pip\npython-setuptools" >> /tmp/archlive/packages.x86_64 
+    - run: find /tmp/archlive
+    - run: cd /tmp/archlive;  mkarchiso -v -w work/ -o out/ ./
+    - uses: actions/upload-artifact@v2
+      with:
+        name: Arch Live ISO
+        path: /tmp/archlive/out/*.iso
--- a/.github/workflows/iso-build.yaml
+++ b/.github/workflows/iso-build.yaml
@@ -16,6 +16,7 @@ jobs:
     - run: find .
     - run: cat /etc/os-release
     - run: mkdir -p /tmp/archlive/airootfs/root/archinstall-git; cp -r . /tmp/archlive/airootfs/root/archinstall-git
+    - run: echo "pip uninstall archinstall -y; cd archinstall-git; python setup.py install; echo 'Type python -m archinstall to launch archinstall'" > /tmp/archlive/airootfs/root/.zprofile
     - run: pacman -Sy; pacman --noconfirm -S git archiso
     - run: cp -r /usr/share/archiso/configs/releng/* /tmp/archlive
     - run: echo -e "git\npython\npython-pip\npython-setuptools" >> /tmp/archlive/packages.x86_64 
--- a/.github/workflows/iso-build.yaml
+++ b/.github/workflows/iso-build.yaml
@@ -2,7 +2,7 @@
 
 name: Build Arch ISO with ArchInstall Commit
 
-on: [push, pull_request]
+on: pull_request
 
 jobs:
   build:


Torxed:
Just a ridiculous question before I merge.. Or two..
1. Is this allowed? Are we allowed to store 700+ MB worth of binary stuff on Github actions? @grazzolini?
2. Would every commit triggers this and would it remove the old ISO? Just making sure we're cleaning up after us.

dylanmtaylor:
> Just a ridiculous question before I merge.. Or two..
> 
> 1. Is this allowed? Are we allowed to store 700+ MB worth of binary stuff on Github actions? @grazzolini?
> 2. Would every commit triggers this and would it remove the old ISO? Just making sure we're cleaning up after us.

So, for #1 I'm not sure but they don't seem to prevent it. It's automatically deleted after 90 days by default

dylanmtaylor:
> Just a ridiculous question before I merge.. Or two..
> 
> 1. Is this allowed? Are we allowed to store 700+ MB worth of binary stuff on Github actions? @grazzolini?
> 2. Would every commit triggers this and would it remove the old ISO? Just making sure we're cleaning up after us.

It looks like on the project level we can change this to 14 days or so? Might make it so we're not 'abusing' GitHub's storage too much. It's a setting and not something you commit.

dylanmtaylor:
This wording seems to indicate they're perfectly fine with it:

About billing for GitHub Actions
GitHub Actions usage is free for both public repositories and self-hosted runners. For private repositories, each GitHub account receives a certain amount of free minutes and storage, depending on the product used with the account. Any usage beyond the included amounts is controlled by spending limits

As a legitimate public open-source project I think we'll be okay.

Torxed:
> This wording seems to indicate they're perfectly fine with it:
> 
> About billing for GitHub Actions
> GitHub Actions usage is free for both public repositories and self-hosted runners. For private repositories, each GitHub account receives a certain amount of free minutes and storage, depending on the product used with the account. Any usage beyond the included amounts is controlled by spending limits
> 
> As a legitimate public open-source project I think we'll be okay.

As long as we're not racking up a quota for Arch Linux public account some how, I'd be fine with this.
Just a note here that this project will eventually make it's way into the official GitLab when that opens up for the public, in which case we'll need to re-visit this and migrate it.

dylanmtaylor:
> > This wording seems to indicate they're perfectly fine with it:
> > About billing for GitHub Actions
> > GitHub Actions usage is free for both public repositories and self-hosted runners. For private repositories, each GitHub account receives a certain amount of free minutes and storage, depending on the product used with the account. Any usage beyond the included amounts is controlled by spending limits
> > As a legitimate public open-source project I think we'll be okay.
> 
> As long as we're not racking up a quota for Arch Linux public account some how, I'd be fine with this.
> Just a note here that this project will eventually make it's way into the official GitLab when that opens up for the public, in which case we'll need to re-visit this and migrate it.

Migration is actually really easy for this; I have a ton of experience with GitLab runners from working with them professionally at work and could make a .gitlab-ci.yaml file for this in very little time. :) We just need a worker node hosted somwehere with a privileged Docker runner (to allow for it to build).

grazzolini:
I think should be triggered only on master or when new PR's are submitted, not on each and every commit.

dylanmtaylor:
> I think should be triggered only on master or when new PR's are submitted, not on each and every commit.

You are probably right but I do not know if that will retrigger the pipeline if you amend a pull request with additional commits

grazzolini:
I've seen it being done on other projects, so I think it's possible.

dylanmtaylor:
> I've seen it being done on other projects, so I think it's possible.

Let me do a quick test. I will make a new commit to this pull request and change it so it just runs on pull requests and see if it runs with the additional commit.

dylanmtaylor:
Okay so it looks like that works. I'm fine with changing it, and it's probably okay to go in now. It's not like this can possibly break the library itself.

Torxed:
> Okay so it looks like that works. I'm fine with changing it, and it's probably okay to go in now. It's not like this can possibly break the library itself.

Instead of closing and re-opening, I wouldn't mind the PR being marked as WIP, helps out with knowing what's being worked on and what's tested and done. But it looks good with `pull_request` so I'm merging :) Consider this the first official test of the runner hehe.

dylanmtaylor:
> > Okay so it looks like that works. I'm fine with changing it, and it's probably okay to go in now. It's not like this can possibly break the library itself.
> 
> Instead of closing and re-opening, I wouldn't mind the PR being marked as WIP, helps out with knowing what's being worked on and what's tested and done. But it looks good with `pull_request` so I'm merging :) Consider this the first official test of the runner hehe.

Didn't mean to close and re-open, these buttons are too close together to try to reply on mobile and it doesn't offer confirmation. :(

Torxed:
I did just realize something tho, and that is: Does every PR trigger this *before* being merged? Wouldn't that mean if I spawn two PR's, the last one will overwrite the ISO? I guess this isn't an issue, but a bit weird for testing purposes. Not knowing which ISO i get.

dylanmtaylor:
> I did just realize something tho, and that is: Does every PR trigger this _before_ being merged? Wouldn't that mean if I spawn two PR's, the last one will overwrite the ISO? I guess this isn't an issue, but a bit weird for testing purposes. Not knowing which ISO i get.

It's not an overwrite. Each has it's OWN iso file :)

Torxed:
> > I did just realize something tho, and that is: Does every PR trigger this _before_ being merged? Wouldn't that mean if I spawn two PR's, the last one will overwrite the ISO? I guess this isn't an issue, but a bit weird for testing purposes. Not knowing which ISO i get.
> 
> It's not an overwrite. Each has it's OWN iso file :)

Awesome! Again, this is some what of black magic to me.
But that is great to hear, and I'm glad it's not on each commit heh.

PR #281, opened 2021-04-11
If user does not change keyboard language, log it.

dylanmtaylor:


Torxed (approved):
This looks really good, you could throw in a `fg="yellow"` here if you want.

dylanmtaylor:
> This looks really good, you could throw in a `fg="yellow"` here if you want.

I'm not sure if the console on my VM is too small or something but I'm not seeing it in the console, hence WIP.

dylanmtaylor:
Yeah, this PR seems to break things including cinnamon install and getting the install logs on the final target system. Not sure why, perhaps it is not working correctly and is somehow interrupting the process?

dylanmtaylor:
Disregard the issue about this PR breaking things, it works fine and I was able to reboot afterwards. I'm going to keep an eye on the issue I experienced if I see it again though. Seems unrelated to this change.

VirtualBox_arch_11_04_2021_17_09_59

VirtualBox_arch_11_04_2021_17_12_15

PR #282, opened 2021-04-11
Torxed launchwarning

Torxed:
If UEFI is not detected on launch, the script will log and error and exit.
This will be removed once Grub and MBR is supported in master (v2.2.0)


dylanmtaylor:
VirtualBox_arch_11_04_2021_18_36_36
Needs fixed, but whatever, I'll do so in another PR

PR #285, opened 2021-04-12
Change from argon2i to argon2id

### LINKED ISSUE #269: Why is argon2i used?
Referring to: https://github.com/archlinux/archinstall/blob/master/archinstall/lib/luks.py#L73

I read that argon2id gives better security, so is there any reason against it?

dylanmtaylor:
Closes #269

### DIFF
--- a/archinstall/lib/luks.py
+++ b/archinstall/lib/luks.py
@@ -70,7 +70,7 @@ class luks2():
 			'--batch-mode',
 			'--verbose',
 			'--type', 'luks2',
-			'--pbkdf', 'argon2i',
+			'--pbkdf', 'argon2id',
 			'--hash', hash_type,
 			'--key-size', str(key_size),
 			'--iter-time', str(iter_time),


Torxed:
I would like to test this thurrowly.
> The Internet draft recommends using Argon2id except when there are reasons to prefer one of the other two modes.

Doesn't actually go in to what those exception/reasons are, and we need to make sure we know what they are before we change. Otherwise we can't definitively say that we know that we are not those exceptions.

I'd like to benchmark on a laptop, a desktop and a VM just to compare I/O on for instance btrfs which seams to be the most common.

dylanmtaylor:
> I would like to test this thurrowly.
> 
> > The Internet draft recommends using Argon2id except when there are reasons to prefer one of the other two modes.
> 
> Doesn't actually go in to what those exception/reasons are, and we need to make sure we know what they are before we change. Otherwise we can't definitively say that we know that we are not those exceptions.
> 
> I'd like to benchmark on a laptop, a desktop and a VM just to compare I/O on for instance btrfs which seams to be the most common.

ext4 is also incredibly common.

dylanmtaylor:
I feel like security is more important than performance here, if you want to do benchmarks on this, feel free.

advaithm:
I've been using argonid to hash passwords for a while now on webservers. there isn't a noticeable difference between argoni vs argonid. I'm giving the pr my +1

advaithm:
maybe we can expose the method through a config option maybe like that's set to a default method for luks in the guided install. maybe like `archinstall.arguments['harddrive'].crypto_method`

Torxed:
I've done some testing (rudimentary), I tried to do it several times and as reproducible as possible with common hardware.
For the sake of it, I ran the installer in identical VM's on a machine with identical loads etc. The only drawback in the test is that I've done it without a GPU and I did no attempts of brute forcing.. I'll leave that to the professionals pounding against the algorithm in general. But below is a time comparison of 1.3GB disk writes `/dev/zero > ./zero`:

|   argon2i  |  argon2id  |
| ------------ | -------------- |
|   2.914s   |    1.790s   |
|   2.078s   |    1.902s   |
|   1.907s   |    1.944s   |

Since people involved seam to prefer `argon2id` over `argon2i`, and I've heard no objections or counter arguments for why `argon2i` should be the default other than *"the main developer prefer argon2i" (and prefer is the key word here)*, I see no reason why we can't attempt best efforts of making it more secure.. I will however still introduce a `--PBKDF` flag in `v2.2.0` (due in a bit over a month) to override this so people can have more degrees of freedom.. But I'll merge this for now.

PR #286, opened 2021-04-12
Use generic_select for GPU drivers

### LINKED ISSUE #205: AttributeError: 'dict' object has no attribute 'index'
I got this error when typing "gnome" when asked for what profile to install. It is thrown on this line:

https://github.com/archlinux/archinstall/blob/1a09ad7bff5d799180b926df4a3611c8f7a6ea29/archinstall/lib/user_interaction.py#L254

Would a `selected_profile = options[list(options).index(selected_profile)]` solve this one?

~can't copy text from this VM~ just remembered screenshots exist

Screenshot from 2021-04-06 12-19-58

dylanmtaylor:
This allows for better code reuse and allows specification by absolute value.

proprietary:

VirtualBox_arch_11_04_2021_21_35_16

open-source:

VirtualBox_arch_11_04_2021_21_35_59

invalid (bad user input):

VirtualBox_arch_11_04_2021_21_38_41

Closes #205.


dylanmtaylor:
This is very close to working

VirtualBox_arch_11_04_2021_21_30_30

examples/guided.py
```diff
@@ -1,4 +1,4 @@
-import getpass, time, json, os
+import getpass, time, json, sys, signal, os
```
Torxed, line 1: What is this code doing here and why is [this](https://github.com/archlinux/archinstall/blob/1be8f1f8cb7e694f4cca19a8f99c43ccc1cd2c01/examples/guided.py#L11-L22) back? How did this not get overwritten by the pull from master? o0

The signal handling and countdown should have been moved into `user_interactions` o0

dylanmtaylor: I brought it back because right not it doesn't run without it.

dylanmtaylor: I am dropping this commit, but it will need to be fixed elsewhere for 2.2.0 to run.

PR #293, opened 2021-04-12
Replace os.system with subprocess.call

JakobDev:
os.system should not be used anymore.

### DIFF
--- a/archinstall/lib/locale_helpers.py
+++ b/archinstall/lib/locale_helpers.py
@@ -1,3 +1,4 @@
+import subprocess
 import os
 
 from .exceptions import *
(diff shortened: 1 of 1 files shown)


dylanmtaylor:
Is there any benefit of doing it this way? os.system() worked fine in that case.

advaithm:
@dylanmtaylor officially os.system is depreciated and can/will be removed in the future

dylanmtaylor:
> @dylanmtaylor officially os.system is depreciated and can/will be removed in the future

No objections from me. Seems like a sensible change.

Torxed:
This is what you get for starting writing the code 5 years ago.
Nice catch and there's probably more low hanging fruit hanging around that we should take a look at :)

PR #294, opened 2021-04-12
Move to flit as build tool

advaithm:
🚨 PR Guidelines:

# New features *(v2.2.0)*

This changes our build tools to flit. I've currently left out setup.py for the archinstall-git package. Ideally this should be dropped as flit generates its own setup.py file for sdist builds. 

# Testing

I've been able to build and install a version made from `python -m flit build`


Torxed:
We'll try this and see if there's any issues down the line.
I doubt it but you never know. Since it's for `2.2.0` I'm merging : )

PR #314, opened 2021-04-16
choosing a kernel 

AggamR:
I have added the ability to choose from one of the following kernels: `linux`, `linux-lts` and `linux-zen`.


AggamR:
wrong branch

examples/guided.py
```diff
@@ -176,6 +176,15 @@ def ask_user_questions():
 			# packages installed by a profile may depend on audio and something may get installed anyways, not much we can do about that.
 			# we will not try to remove packages post-installation to not have audio, as that may cause multiple issues
 			archinstall.arguments['audio'] = None
+   
+   	# Ask what kernel user wants:
```
dylanmtaylor, line 180: fix indentation

examples/guided.py
```diff
@@ -176,6 +176,15 @@ def ask_user_questions():
...
+	kernel = input("1. linux\n2. linux-lts\n3. linux-zen\nchoose a kernel of the following: ")
```
dylanmtaylor, line 181: make this use generic_select.

archinstall/lib/installer.py
```diff
@@ -34,7 +34,8 @@ class Installer():
 	:type hostname: str, optional
 
 	"""
-	def __init__(self, target, *, base_packages='base base-devel linux linux-firmware efibootmgr'):
+	def __init__(self, target, *, base_packages=' base base-devel linux-firmware efibootmgr', kernel='linux'):
```
dylanmtaylor, line 37: make it be kernels (plural) and accept an array. For instance kernels=['linux', 'linux-lts'[

examples/guided.py
dylanmtaylor: This should also have a loop that allows selecting multiple.

dylanmtaylor:
More of a question for those following this -- should we also install header files for all selected kernels? DKMS and a few other things need them to build modules.

Issue #145, opened 2021-04-02
Option to change the kernel package used during installation


matteoturini:
I would like to use linux-zen!

Torxed:
I'll deliver on your request some how, this is something a lot of people seam to want.
But the guided installer might keep it simple - and adding custom (potentially unsupported kernels) will be optional.

So I'll have to separate the "official installer" which is guided and these some what exotic features :) But i'll come!

matteoturini:
I think it would be nice to make that an option for the programmable install! I agree, the default should be kept simple.

grazzolini:
@Torxed since the user has the liberty to install whichever kernel they want on the current method of installation, following the installation guide, I think adding a feature to list the official kernels (linux, linux-lts, linux-zen and linux-hardened) while keeping linux as the default option if the person simply press enter, would be a nice addition to have. I don't see the list of kernels growing anytime soon.

dylanmtaylor:
This would be an awesome feature to have. One thought -- I think it may be desirable to have linux AND linux-lts on the same system (in case the newer kernel breaks), and have a bootloader menu for each. This is how I have my current installation now, as an example of this:

```
➜  ~ cat /boot/loader/entries/arch.conf 
title Arch Linux
linux /vmlinuz-linux
initrd /intel-ucode.img
initrd /initramfs-linux.img
options rd.luks.name=1ea54b52-e005-4438-afd9-e690e690bec4=cryptlvm root=/dev/vg0/root resume=/dev/vg0/swap rd.luks.options=discard i915.fastboot=1 quiet splash loglevel=3 rd.udev.log_priority=3 vt.global_cursor_default=0 rw
➜  ~ cat /boot/loader/entries/archlts.conf
title Arch Linux LTS
linux /vmlinuz-linux-lts
initrd /intel-ucode.img
initrd /initramfs-linux-lts.img
options rd.luks.name=1ea54b52-e005-4438-afd9-e690e690bec4=cryptlvm root=/dev/vg0/root resume=/dev/vg0/swap rd.luks.options=discard i915.fastboot=1 quiet rw
```

dylanmtaylor:
Maybe have kernels be an array that is iterated over to install each one as a separate bootloader entry?

dylanmtaylor:
Guided installation mockup:

```
0: linux
1: linux-lts
2: linux-zen
3: linux-hardened
[...]

PR #315, opened 2021-04-16
selecting a kernel

AggamR:
added ability to select kernel: linux, linux-lts, linux-zen


dylanmtaylor:
Can you rebase this off 2.2.0 so there are not 45 commits in it and do a force push please? It makes it easier to review when I don't have to see unrelated changes.

dylanmtaylor:
This is a great change that we definitely need, just trying to make it easier to follow the pull request. Also, there's some comments on this in #145 - we should be able to select multiple and have a choice between them at boot time.

examples/guided.py
```diff
@@ -171,7 +175,16 @@ def ask_user_questions():
 		else:
 			# packages installed by a profile may depend on audio and something may get installed anyways, not much we can do about that.
 			# we will not try to remove packages post-installation to not have audio, as that may cause multiple issues
-			archinstall.arguments['audio'] = 'none'
+			archinstall.arguments['audio'] = None
+   
+   	# Ask what kernel user wants:
```
dylanmtaylor, line 180: fix indentation here

examples/guided.py
```diff
@@ -171,7 +175,16 @@ def ask_user_questions():
...
+	kernel = input("1. linux\n2. linux-lts\n3. linux-zen\nchoose a kernel of the following: ")
```
dylanmtaylor, line 181: Use generic_select instead of an input, passing in an array of choices. Make it loop so we can have multiple selected as discussed in #145.

archinstall/lib/installer.py
```diff
@@ -34,7 +34,8 @@ class Installer():
 	:type hostname: str, optional
 
 	"""
-	def __init__(self, target, *, base_packages='base base-devel linux linux-firmware efibootmgr'):
+	def __init__(self, target, *, base_packages=' base base-devel linux-firmware efibootmgr', kernel='linux'):
```
dylanmtaylor, line 37: This should be changed to be kernels (plural) and accept an array instead of a string, so that you can do ['linux', 'linux-lts']. The user should be able to select one at boot, and each should be given a bootloader entry for both GRUB and systemd-boot depending on installation type.

examples/guided.py
```diff
@@ -3,6 +3,10 @@
 from archinstall.lib.hardware import hasUEFI
 from archinstall.lib.profiles import Profile
 
+if hasUEFI() is False:
```
dylanmtaylor, line 6: This commit should NOT be in 2.2.0. Which is why rebasing these commits is so important. This breaks MBR/GRUB in the 2.2.0 branch

examples/guided.py
```diff
@@ -268,43 +281,52 @@ def perform_installation(mountpoint):
 	Only requirement is that the block devices are
 	formatted and setup prior to entering this function.
 	"""
-	with archinstall.Installer(mountpoint) as installation:
+	with archinstall.Installer(mountpoint, kernel=archinstall.arguments['kernel']) as installation:
```
dylanmtaylor, line 284: Also should be plural here

guided.py
```diff
@@ -0,0 +1,359 @@
...
+		
+		# only ask for audio server selection on a desktop profile 
+		if str(archinstall.arguments['profile']) == 'Profile(desktop)':
+			archinstall.arguments['audio'] = archinstall.ask_for_audio_selection()
+		else:
+			# packages installed by a profile may depend on audio and something may get installed anyways, not much we can do about that.
+			# we will not try to remove packages post-installation to not have audio, as that may cause multiple issues
+			archinstall.arguments['audio'] = None
+   
+   	# Ask what kernel user wants:
```
dylanmtaylor, line 180: Move the actual choice logic to the user_interaction library and call it as a function.

examples/guided.py
Torxed: I should pull in master and fix this, this is bound to happen more times. Give me an hour.

advaithm: yeah this breaks my implementation of grub/bios support. would you mine rebasing to  the 2.2.0 branch to fix this?

archinstall/lib/installer.py
advaithm: I believe it should be fine to have multiple kernels without any changes to the GRUB installation setup as `grub-mkconfig` should pick up other kernels automatically

dylanmtaylor: Systemd-boot needs an entry for each though

Torxed:
I updated your branch @AggamR, the branch was behind with 193 commits and it made the PR look a bit wonky with 18 file changes that would overwrite some of the history. I'll take a look at this again in a day or so, just need to wrap up `v2.1.4` today and tomorrow : )

AggamR:
> I updated your branch @AggamR, the branch was behind with 193 commits and it made the PR look a bit wonky with 18 file changes that would overwrite some of the history. I'll take a look at this again in a day or so, just need to wrap up `v2.1.4` today and tomorrow : )

Ok, thanks! I see that I have some fixes to make, so I'll get to work on that as well.

guided.py
AggamR: Sorry, it shows me code I didn't even write:
```
			# we will not try to remove packages post-installation to not have audio, as that may cause multiple issues
			archinstall.arguments['audio'] = None

   	# Ask what kernel user wants:
```

what code are you referring to?

dylanmtaylor: The block from 180 onwards that has the user selecting a kernel. That choice would be best in user_interaction.py. it also should use the generic_select function

archinstall/lib/user_interaction.py
```diff
@@ -263,7 +265,7 @@ def select_disk(dict_o_disks):
 		elif drive.isdigit():
 			drive = int(drive)
 			if drive >= len(drives):
-				raise DiskError(f'Selected option "{drive}" is out of range')
+				raise Dis1ror(f'Selected option "{drive}" is out of range')
```
dylanmtaylor, line 270: This seems like a mistake?

dylanmtaylor: @Torxed if you're fine with this, could you correct this line and merge this in? We can further clean up and refine this code to make it up to standards before 2.2.0 is released? Would like to see this feature land for 2.2.0. It looks good enough for testing at least.

Torxed: I'll get on it soon, merging in to master in ~8h :)

guided.py
AggamR: > That choice would be best in user_interaction.py.

so do I put all of that code in `user_interaction.py` as a function and call it in guided.py?

archinstall/lib/installer.py
Torxed: I'm creating an issue for this so we don't forget about this discussion @dylanmtaylor

Torxed:
Mergign this in now, but I did re-work and remove the kernel selection loop. As I believe pressing enter and skipping the step should default to the default kernel without any fuss. Meaning I don't like that the user have to write "continue" to break out of a loop, it disrupts the flow throughout the installer.

With that said, I think we should implement a `multi_generic_select()` that can allow the user to select multiple options, and if the user enters out without any options a `default=<option>` argument should be implemented. Allowing the selection process to return back the default (or none) :)

PR #317, opened 2021-04-17
In guided install Non-UEFI installs need to use grub-install

insanemal:
In guided install non-UEFI installs need to use grub-install as its the only supported bootloader that doesn't require UEFI
As such added a simple "hasUEFI" test around the add_bootloader call to set bootloader to grub-install as currently it is setup to default to systemd-boot.

Added grub to the pacstrap id non-UFFI is detected

Switched to tabs from spaces (PEP-8 recommends spaces)


insanemal:
Conflicts look to be places my editor automatically trimmed whitespace char's on otherwise empty lines.

Torxed:
@advaithm Didn't we already implement this?

insanemal:
When I tried this branch on bios it bailed on bootloader install because it didn't have UEFI. 

The default action of add_bootloader when called without setting the bootloader parameter tries to install systemd-boot and that is setup to bail if you don't have UEFI. 

And in the guided.py the bootloader parameter doesn't get set. This sets it. 

The other code dealt with using MBR not GPT. 

I'm happy to be wrong, then I've got to give you logs because it definitely didn't install grub instead. 

Also the point it calls add_bootloader grub wouldn't be installed anyway and the grub setup routine doesn't install grub. Bootloader install appears to happen before installing extra packages. 

Again I'm happy to be wrong but currently it doesn't work for me without this.

advaithm:
@Torxed we did implement this but it looks like it got removed at some point

insanemal:
Ok. Well at least this started the required conversation

Torxed:
> @Torxed we did implement this but it looks like it got removed at some point

Super weird, I've manually verified all the commits and PR's, I'll find your PR and merge it back in. Together with some changes here. Give me some time to do some merge hell : )

Torxed:
Corrected some merge issues, added a todo but overall this should enable Grub in non-efi.

PR #325, opened 2021-04-19
Rework of the generic_select function and fixes for various input checks

SecondThundeR:
# Describe your PR
This PR is a rework of `generic_select` that adds general stability for unexpected user inputs.
In addition, this PR adds some fixes and improvements in other places where there could be unexpected behavior.

> I have attached a changelog for each commit. You can find them in the commit description
> Also, I left a few comments where I was in doubt about the changes.

# Testing

Tested on VirtualBox 6.1 with multiple scenarios, inputs and checks

### DIFF
--- a/archinstall/lib/user_interaction.py
+++ b/archinstall/lib/user_interaction.py
@@ -351,9 +351,10 @@ def select_profile(options):
 
 		selected_profile = generic_select(profiles, 'Enter a pre-programmed profile name if you want to install one: ',
 										 options_output=False)
-		return Profile(None, selected_profile)
-
-	raise RequirementError("Selecting profiles require a least one profile to be given as an option.")
+		if selected_profile:
+			return Profile(None, selected_profile)
+	else:
+		raise RequirementError("Selecting profiles require a least one profile to be given as an option.")
 
 def select_language(options, show_only_country_codes=True):
 	"""
--- a/archinstall/lib/user_interaction.py
+++ b/archinstall/lib/user_interaction.py
@@ -189,9 +189,9 @@ def ask_to_configure_network():
 						fg='red'
 					)
 
-			gateway = input('Enter your gateway (router) IP address or leave blank for none: ').strip()
 			# Implemented new check for correct gateway IP address
 			while 1:
+				gateway = input('Enter your gateway (router) IP address or leave blank for none: ').strip()
 				try:
 					if len(gateway) == 0:
 						gateway = None
--- a/archinstall/lib/user_interaction.py
+++ b/archinstall/lib/user_interaction.py
@@ -1,5 +1,5 @@
 import getpass, pathlib, os, shutil, re
-import sys, time, signal
+import sys, time, signal, ipaddress
 from .exceptions import *
 from .profiles import Profile
 from .locale_helpers import search_keyboard_layout
@@ -164,13 +164,25 @@ def ask_to_configure_network():
 	if nic and nic != 'Copy ISO network configuration to installation':
 		if nic == 'Use NetworkManager to control and manage your internet connection':
 			return {'nic': nic,'NetworkManager':True}
-		mode = generic_select(['DHCP (auto detect)', 'IP (static)'], f"Select which mode to configure for {nic}: ")
-		if mode == 'IP (static)':
+
+		# Current workaround:
+		# For selecting modes without entering text within brackets,
+		# printing out this part separate from options, passed in
+		# `generic_select`
+		modes = ['DHCP (auto detect)', 'IP (static)']
+		for index, mode in enumerate(modes):
+			print(f"{index}: {mode}")
+
+		mode = generic_select(['DHCP', 'IP'], f"Select which mode to configure for {nic} or leave blank for DHCP: ",
+							 options_output=False)
+		if mode == 'IP':
 			while 1:
 				ip = input(f"Enter the IP and subnet for {nic} (example: 192.168.0.5/24): ").strip()
-				if ip:
+				# Implemented new check for correct IP/subnet input
+				try:
+					ipaddress.ip_interface(ip)
 					break
-				else:
+				except ValueError:
 					log(
 						"You need to enter a valid IP in IP-config mode.",
 						level=LOG_LEVELS.Warning,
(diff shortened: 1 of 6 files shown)


archinstall/lib/user_interaction.py
```diff
@@ -154,19 +154,35 @@ def ask_to_configure_network():
...
+		# Current workaround:
+		# For selecting modes without entering text within brackets,
+		# printing out this part separate from options, passed in
+		# `generic_select`
+		modes = ['DHCP (auto detect)', 'IP (static)']
+		for index, mode in enumerate(modes):
+			print(f"{index}: {mode}")
+
+		mode = generic_select(['DHCP', 'IP'], f"Select which mode to configure for {nic} or leave blank for DHCP: ",
+							 options_output=False)
```
Torxed, line 177: We probably need to pass `sort=False` here if we want to print our own list, and pass another list to `generic_select` but skip output for that call. Because Generic select could change the order and then this magic "print but use another list" won't match on indexes any longer.

archinstall/lib/user_interaction.py
```diff
@@ -206,40 +242,69 @@ def ask_for_main_filesystem_format():
...
-	elif selected_option in options:
-		pass # We gave a correct absolute value
-	else:
-		raise RequirementError(f'Selected option "{selected_option}" does not exist in available options: {options}')
+	# Checking if options are different from `list` or `dict`
+	if type(options) not in [list, dict]:
+		log(" * It looks like there are something wrong with provided options. Maybe it's time to open an issue on GitHub! * ", fg='red')
+		log(" * Here are the link: https://github.com/archlinux/archinstall/issues * ", fg='yellow')
+		raise RequirementError("generic_select() requires list or dictionary as options.")
+	if type(options) == dict: options = sorted(list(options.values()))  # To allow only `list` and `dict`, converting values of options and sorting them here. Therefore, now we can only provide the dictionary itself
```
Torxed, line 268: We have to be careful calling `sorted()`, this might mess up the magic mentioned above.

Torxed (changes requested):
Really good re-work in general.
Just one minor thing and then I'll do a couple of test runs to verify that the logic holds up. But I really like this one!

SecondThundeR:
Hi! I agree that I need to somehow carefully work with sorting and I will start working on the fix this afternoon. Thanks for the feedback!

archinstall/lib/user_interaction.py
```diff
@@ -176,6 +192,25 @@ def ask_to_configure_network():
...
+			# 		else:
+			# 			ipaddress.ip_address(gateway)
+			# 		break
+			# 	except ValueError:
+			# 		log(
+			# 			"You need to enter a valid gateway (router) IP address.",
+			# 			level=LOG_LEVELS.Warning,
+			# 			fg='red'
+			# 		)
+
```
SecondThundeR, line 213: Should I uncomment this change because I'm still unsure if we are checking for regular IP and not for IP/subnet

Torxed (approved):
Looks good!

archinstall/lib/user_interaction.py
Torxed: The gateway needs to be asked if static IP was chosen.
Otherwise there will be no internet for the poor user :)

SecondThundeR:
If there are no more objections and additions, I think it's done

archinstall/lib/user_interaction.py
```diff
@@ -206,40 +236,72 @@ def ask_for_main_filesystem_format():
...
-		selected_option = options[selected_option]
-	elif selected_option in options:
-		pass # We gave a correct absolute value
-	else:
-		raise RequirementError(f'Selected option "{selected_option}" does not exist in available options: {options}')
+	# Checking if options are different from `list` or `dict`
+	if type(options) not in [list, dict]:
+		log(" * It looks like there are something wrong with provided options. Maybe it's time to open an issue on GitHub! * ", fg='red')
+		log(" * Here are the link: https://github.com/archlinux/archinstall/issues * ", fg='yellow')
+		raise RequirementError("generic_select() requires list or dictionary as options.")
```
dylanmtaylor, line 261: Here are the link -> here is the link

dylanmtaylor: Also I don't know if we want to use GitHub here because I think it is going to move from GitHub to Arch's GitLab.

dylanmtaylor: It looks like there are something wrong is also grammatically incorrect. 

Perhaps change the wording to something like this:

The provided input is invalid, if you believe that this is an incorrect, please consider reporting a bug.

archinstall/lib/user_interaction.py
```diff
@@ -334,12 +380,18 @@ def select_language(options, show_only_country_codes=True):
...
+		languages_length = len(languages)
+
+		print(f' -- You can enter ? ({languages_length - 2}) or help ({languages_length - 1}) to search for more languages, or skip to use US layout --')
+		selected_language = generic_select(languages, 'Select one of the above keyboard languages (by number or full name): ',
+										  options_output=False)
 		
-		if len(selected_language.strip()) == 0:
+		if not selected_language:
 			return DEFAULT_KEYBOARD_LANGUAGE
-		elif selected_language.lower() in ('?', 'help'):
```
dylanmtaylor, line 342: By getting rid of lower here, Help and HELP are no longer considered valid. I'm not sure if I like this change

archinstall/lib/user_interaction.py
SecondThundeR: Typos are my bad :(, and I was thinking to make something like this https://github.com/archlinux/archinstall/blob/master/examples/guided.py#L111

But, if this part isn't needed, then of course we can remove it

archinstall/lib/user_interaction.py
SecondThundeR: Because of the workaround on line 385, the user can only enter '?' and "help", as I thought that the user would enter only such values or simply select these items in numbers

archinstall/lib/user_interaction.py
dylanmtaylor: I like the thought behind this, but I think it might be an issue as eventually I was told we will move development OFF of GitHub (it will become a read-only mirror) and it will transition to https://gitlab.archlinux.org/archlinux -- so if people write out installers with this, when it moves it will no longer be correct.

SecondThundeR:
At the moment, I haven't exactly figured out the best way to handle 'help' and '?' without adding them to list. It's possible to add yet another parameter for `generic_select()`, which can help to handle this for current and future functions with "Help" menu, but this move can bloat current implementation

dylanmtaylor:
I feel this should be an improvement for 2.2.0 since it's so close to the final 2.1.4 release and this code does make some larger changes.

archinstall/lib/user_interaction.py
Torxed: It's fine, we'll do `grep -Er "github"` and change it in one `sed` or `awk` onliner :) But you are correct.

Torxed:
2021-04-21-163115_1280x987_scrot

I'll work on fixing this later tonight before merging :)

PR #339, opened 2021-04-22
Have timezone re-prompt on invalid input

### LINKED ISSUE #338: Timezone field does not have validation in latest master (2.1.4)
Entering, for instance, US/Easter does not provide an error and opportunity to enter a valid result as intended (in this case US/Eastern). This results in a null timezone in the config that is installed.

dylanmtaylor:
Closes #338

### DIFF
--- a/archinstall/lib/user_interaction.py
+++ b/archinstall/lib/user_interaction.py
@@ -130,17 +130,19 @@ def ask_for_additional_users(prompt='Any additional users to install (leave blan
 	return users, super_users
 
 def ask_for_a_timezone():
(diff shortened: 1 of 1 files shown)


dylanmtaylor:
We could put an 
```
			exists = True
```
in the validated condition but since it immediately returns this is not necessary as the value is just discarded. The while exists is more readable than a while True or while 1, so I chose to code it that way.

dylanmtaylor:
Tested and working.

VirtualBox_Arch_22_04_2021_15_42_50

PR #343, opened 2021-04-22
While somewhat nice to have, archlinux-wallpaper isn't needed in a minimall install

dylanmtaylor:
Simple commit to remove this from default packages on lxqt to save some time/space


Torxed:
What's the default wallpaper if there is no package installed for it? Just a blank background?

dylanmtaylor:
> What's the default wallpaper if there is no package installed for it? Just a blank background?

believe it or not, no. There is one shipped with lxqt.

dylanmtaylor:
This is the current default lxqt wallpaper

VirtualBox_Arch_23_04_2021_09_22_18

PR #344, opened 2021-04-22
Misc. cleanup

SecondThundeR:
# Describe your PR

This is a polishing PR after pushing final 2.1.4 tag
All changes can be found in commits description

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -90,14 +90,14 @@ class Installer():
 		self.genfstab()
 
 		if not (missing_steps := self.post_install_check()):
-			self.log('Installation completed without any errors. You may now reboot.', bg='black', fg='green', level=logging.INFO)
+			self.log('Installation completed without any errors. You may now reboot.', fg='green', level=logging.INFO)
 			self.sync_log_to_install_medium()
 
 			return True
 		else:
-			self.log('Some required steps were not successfully installed/configured before leaving the installer:', bg='black', fg='red', level=logging.WARNING)
+			self.log('Some required steps were not successfully installed/configured before leaving the installer:', fg='red', level=logging.WARNING)
 			for step in missing_steps:
-				self.log(f' - {step}', bg='black', fg='red', level=logging.WARNING)
+				self.log(f' - {step}', fg='red', level=logging.WARNING)
             
 			self.log(f"Detailed error logs can be found at: {storage['LOG_PATH']}", level=logging.WARNING)
 			self.log(f"Submit this zip file as an issue to https://github.com/archlinux/archinstall/issues", level=logging.WARNING)
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -79,14 +79,14 @@ class Installer():
 		self.genfstab()
 
 		if not (missing_steps := self.post_install_check()):
-			self.log('Installation completed without any errors. You may now reboot.', fg='green', level=LOG_LEVELS.Info)
+			self.log('Installation completed without any errors. You may now reboot.', bg='black', fg='green', level=LOG_LEVELS.Info)
 			self.sync_log_to_install_medium()
 
 			return True
 		else:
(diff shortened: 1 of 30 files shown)


archinstall/lib/user_interaction.py
```diff
@@ -201,8 +195,7 @@ def ask_to_configure_network():
 					break
 				except ValueError:
 					log(
-						"You need to enter a valid gateway (router) IP address.",
-						level=LOG_LEVELS.Warning,
+						'You need to enter a valid gateway (router) IP address',
```
dylanmtaylor, line 198: Why remove the periods after sentences?

SecondThundeR: In my opinion, this was not necessary for the logs, since it looks somehow strange, but I can leave them for such sentences, where they do not contain ` * ... * `

SecondThundeR:
Still wondering if this PR is relevant today or I can close it for now?

dylanmtaylor:
> Still wondering if this PR is relevant today or I can close it for now?

I don't think that there is anything in this that is needed, it mostly looks like style changes

SecondThundeR:
At least the first two commits are good for further improving the input experience and removing unnecessary `else` in `generic_select`, but I think the third commit can be dropped if it's redundant

archinstall/lib/installer.py
```diff
@@ -168,7 +168,7 @@ def set_timezone(self, zone, *args, **kwargs):
 			return True
 		else:
 			self.log(
-				f"Time zone {zone} does not exist, continuing with system default.",
```
Torxed: @dylanmtaylor Is it `Time zone` or `Timezone`? IIRC most english words are spelled apart and I think this is one of those cases?

archinstall/lib/user_interaction.py
```diff
@@ -22,11 +22,7 @@ def get_longest_option(options):
 def check_for_correct_username(username):
 	if re.match(r'^[a-z_][a-z0-9_-]*\$?$', username) and len(username) <= 32:
 		return True
-	log(
-		"The username you entered is invalid. Try again",
-		level=LOG_LEVELS.Warning,
```
Torxed: Removing the `level=LOG_LEVELS.Warning` here will cause `systemd-journal` output to not be properly color coded. Perhaps this is a non-issue? Not sure how many people check the journal logs for errors :P

dylanmtaylor:
Looks good to me

Torxed (commented):
I like this cleanup, will merge as soon as I have confirmation on the language thing.

archinstall/lib/user_interaction.py
dylanmtaylor: I mean I check them for errors all the time when troubleshooting servers ;) As far as color coding goes, it does make it easier to spot it.

archinstall/lib/installer.py
dylanmtaylor: I have sen both used. I think Time zone may be correct, as that is what English Wikipedia uses, and they tend to know what they are doing ;)

dylanmtaylor: "There seem to be three spellings: timezone, time-zone, and time zone.

The Oxford English Dictionary defines time-zone,[^1] but curiously, also uses the time zone spelling.[^2] At the same time, timezone is the favored spelling in the US (unless talking about multiple time zones).[^3]

All three seem to be equally correct, but I would use time zone, except for a singular usage in the US for an American audience, in which case I would use timezone. Personally, I feel time-zone is a bit outdated."

https://english.stackexchange.com/a/3937

archinstall/lib/user_interaction.py
SecondThundeR: As I can see through the code, other logs with warnings like this didn't have a `level` parameter

Torxed: Ah, yea that's probably my bad on omitting the level in the other places hehe.

SecondThundeR:
Maybe it's better somehow to squash three last commits because of `Merge branch ...`?

Issue #333, opened 2021-04-22
Add support for loading installer args from a config file


l4zy0n3:
Support for running an unattended installation from a config file would be great as I saw that the installer is storing args as a dict, which is spitted out as JSON in logs. We could use the same JSON as a config file to run the installation.

l4zy0n3:
I have already started working on this, will raise a PR if this seems a valid case

Torxed:
Sounds awesome, but just to avoid double work, check out this discussion: https://github.com/archlinux/archinstall/discussions/150

Issue #351, opened 2021-04-24
Add error handling for non-systemd instances (such as docker)


Torxed:
As per discussion in the conversation https://github.com/archlinux/archinstall/discussions/347, running in a docker `systemd` and/or `efi` is not always present. In those scenarios it would be nice if we didn't hard crash during normal operation.

As mentioned in the comment https://github.com/archlinux/archinstall/discussions/347#discussioncomment-652837, this will not affect the guided installer - this is an API improvement to support wider .dotfiles and similar runners.

dylanmtaylor:
Who exactly is running an installer in docker? That's typically not how docker is used at all, it's a container, not a virtual machine. What's the use-case here?

advaithm:
I have to agree with @dylanmtaylor supporting containers is something that has not legitimate use case. along with this, containers can not support systemd since they startup in non standard ways and don't even have a kernel. if other distros such as artix want to have archinstall support their system they can fork and maintain their own versions. supporting multiple inits is something that will cause a lot of issues for maintainability

Torxed:
After discussing a bit, we'll table this for now.
Mainly because Arch Linux mainly supports `systemd`, so we'll focus on that as well until further notice.
We've also increasingly integrated systemd services such as `localectl` and `machinectl` for locales and container execution, so it would be hard to get those things working without regression.

PR #352, opened 2021-04-24
Containerization and  basic support for installing from a config file

l4zy0n3:
Created a Dockerfile #347
Added basic support for installing from a config file #333 #150
@Torxed @pyfisch @Didr


dylanmtaylor:
Please rebase this onto 2.2.0 so that only relevant commits are in the change. Also, consider squashing the 'Update Dockerfile' commits into one so history makes more sense.

Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM l4zy/archlinux:latest
```
dylanmtaylor, line 1: Why is this not the main archlinux image, but rather a fork? We should not use anything unofficial. Strong -1 here.

dylanmtaylor:
I'm not understanding the benefit here, why would you want to run ArchInstall inside a container, and is this some sort of use case we'd want to support? As far as using Arch inside Docker there are existing images, and you can just use custom container images to get what you need in it. Is the use case in question to bind-mount a volume onto a container's /mnt and run it that way? It seems more problematic than it is worth and I don't see many people using this one. I think this is a -1 from me.

dylanmtaylor:
Installing from a configuration file is a great feature, and would be good to have, to clarify. Docker, however, is a questionable value proposition. It's fine for *building* ArchInstall, but for running ArchInstall within Docker to bootstrap a new system? I can't see that being desirable, and if it has issues with systemd then introducing non-systemd paths in the code for this use-case is a strong -1 for me.

Dockerfile
advaithm: this is a private image, we can't even audit. I can't support this

archinstall/__init__.py
```diff
@@ -14,49 +14,68 @@
...
+        except Exception as e:
+            print(e)
+    return {}
+
+
+arguments = initialize_arguments()
 positionals = []
+
 for arg in sys.argv[1:]:
-	if '--' == arg[:2]:
```
advaithm, line 26: why are we messing with indents so much?

advaithm (commented):
I can not support merging this, as it brings  in a large number of changes that may or may not introduce bugs. along with the fact docker lacks systemd which we require.

Dockerfile
l4zy0n3: The main archlinux image did not have systemd on it, this image simply contains systemd on top of main archlinux image
https://hub.docker.com/repository/registry-1.docker.io/l4zy/archlinux/builds/21b85dbe-c21a-4454-85de-284d3c9a5108
image
This was done because, as of now archinstall does not run on systemd less machines #351

dylanmtaylor: Still a strong -1. Using ANYTHING not officially shipped by the Arch Linux project is a blocker IMO. AUR is also out of the question.

advaithm: agreed, -1. we can't support this

advaithm:
yeah if this straight up install arch on the host system or on a mounted volume. this is a big no from me as this can be abused in a insane number of ways

l4zy0n3:
> I'm not understanding the benefit here, why would you want to run ArchInstall inside a container, and is this some sort of use case we'd want to support? As far as using Arch inside Docker there are existing images, and you can just use custom container images to get what you need in it. Is the use case in question to bind-mount a volume onto a container's /mnt and run it that way? It seems more problematic than it is worth and I don't see many people using this one. I think this is a -1 from me.



> yeah if this straight up installs arch on the host system or on a mounted volume. this is a big no from me as this can be abused in a insane number of ways

The requirement from my end was, triggering archinstall from different distros/OSs, docker was just the instrument in this, I totally support that archlinux base image must be used, I am willing to remove the Dockerfile for now until we don't have support for systemd less systems

l4zy0n3:
> > I'm not understanding the benefit here, why would you want to run ArchInstall inside a container, and is this some sort of use case we'd want to support? As far as using Arch inside Docker there are existing images, and you can just use custom container images to get what you need in it. Is the use case in question to bind-mount a volume onto a container's /mnt and run it that way? It seems more problematic than it is worth and I don't see many people using this one. I think this is a -1 from me.
> 
> > yeah if this straight up installs arch on the host system or on a mounted volume. this is a big no from me as this can be abused in a insane number of ways
> 
> The requirement from my end was, triggering archinstall from different distros/OSs, docker was just the instrument in this, I totally support that archlinux base image must be used, I am willing to remove the Dockerfile for now until we don't have support for systemd less systems

I could just maintain the Dockerfile in my own repo instead till then

advaithm:
to make the long story short. I don't see any way to have systemdless archinstall and I can not safely support a docker container that runs archinstall to install on a host system, it can cause irreparable damage and a lot of undocumented things will have to be mounted. I have no issue with a config file to automate the choices.

dylanmtaylor:
Arch Linux only officially has support for systemd. Because this project is the official installer for Arch Linux I have very strong feelings about introducing any sort of code paths or logic that install it in a way that does not use systemd commands or results in an installation that is "systemd free". I don't want to get into the benefits or downsides of using systemd because that is a rabbit hole I don't care to go down, but personally systemd meets my needs just fine, and if you have issues of systemd there are other distributions that ship alternative init systems that could likely better meet your needs, like Artix or Devuan.

dylanmtaylor:
I would like to second the opinion that having a way to import a config file, perhaps JSON format or something, with the configuration choices already populated would definitely be a huge win for us

advaithm:
> Arch Linux only officially has support for systemd. Because this project is the official installer for Arch Linux I have very strong feelings about introducing any sort of code paths or logic that install it in a way that does not use systemd commands or results in an installation that is "systemd free". I don't want to get into the benefits or downsides of using systemd because that is a rabbit hole I don't care to go down, but personally systemd meets my needs just fine, and if you have issues of systemd there are other distributions that ship alternative init systems that could likely better meet your needs, like Artix or Devuan.

I agree with this, personally I wouldn't mind submitting patches to fix issues and the such if Artix plans to fork this for their distro.

>I would like to second the opinion that having a way to import a config file, perhaps JSON format or something, with the configuration choices already populated would definitely be a huge win for us

having a whole other example for this is probably not the best option. I'll look into having guided have an option to load the data from a json file

dylanmtaylor:
> having a whole other example for this is probably not the best option. I'll look into having guided have an option to load the data from a json file

Agreed, there may be a way to reuse the existing code and have a function that ingests the config file and skips the prompts if they are already provided.

Dockerfile
Torxed: We would need to figure out a way around this. Could we submit an official patch upstream? Is there a reason why they're not running EFI?

Torxed:
I think at least two things need to be fixed before I can merge this:
 1. Investigate and see why the official image doesn't use systemd (is it size reasons perhaps), if there's no good reasons why not then perhaps submit a patch/suggestion to include it
 2. Convert all spaces (entire guided for instance) to tabs

I'm merging in `torxed-v2.2.0` to `master` today, I've been holding off on it for a while to get these changes in.
But I think I'll have to merge into master first as I see specifically the docker change could take a while.

If we can split the config-json-stuff into a separate PR (against master then), that would be awesome while we wait for the docker people to answer some of our questions.

Torxed:
Ah.. When I deleted the branch the PR gets auto-closed.
Just to be clear: I still want and welcome these changes! Create a new PR and I'd be happy to merge *(but perhaps split the config stuff in one PR and the docker in a separate)*

advaithm:
Systemd does not work in containers as intended since they do not start like conventional Linux systems. I do not know any containers that include systemd for this reason

PR #362, opened 2021-04-27
Added a mini curses class and generic-multi-select

Torxed:
It can do some simple tricks to iterate over menu options and indicate which ones are chosen using generic_multi_select().

test


dylanmtaylor:
I like it.

SecondThundeR:
Nice improvement for selection experience, keep it up!

PR #366, opened 2021-04-28
Bug fixes for kernel select and grub-install

advaithm:
🚨 PR Guidelines:

# Describe your PR

This fixes issues, with grub-install always failing and selecting kernel being broken. this also installs nvidia dkms when using linux-zen or lts. this will raise a hardware incompatibility error when using the nvidia drivers with sway since sway can not run with them  

# Testing

Tested on kvm/qemu bios

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -40,8 +40,8 @@ class Installer():
 	:type hostname: str, optional
 
 	"""
-	def __init__(self, target, *, base_packages='base base-devel linux linux-firmware', kernels='linux'):
-		base_packages = base_packages + kernels.replace(',', ' ')
+	def __init__(self, target, *, base_packages='base base-devel linux-firmware', kernels='linux'):
+		kernels = kernels.split(",")
 		self.target = target
 		self.init_time = time.strftime('%Y-%m-%d_%H-%M-%S')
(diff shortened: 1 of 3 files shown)


archinstall/lib/installer.py
```diff
@@ -52,6 +51,8 @@ def __init__(self, target, *, base_packages='base base-devel linux linux-firmwar
 		}
 		
 		self.base_packages = base_packages.split(' ') if type(base_packages) is str else base_packages
+		for kernel in kernels:
+			self.base_packages.append(kernel)
```
dylanmtaylor, line 55: I do like this approach a lot more

archinstall/lib/installer.py
```diff
@@ -363,8 +364,10 @@ def add_bootloader(self, bootloader='systemd-bootctl'):
 				boot_partition = partition
 			elif partition.mountpoint == self.target:
 				root_partition = partition
-
-		self.log(f'Adding bootloader {bootloader} to {boot_partition}', level=logging.INFO)
+		if hasUEFI():
```
dylanmtaylor, line 367: I'm confused why we are even checking hasUEFI here if we do the same self.log regardless...?

dylanmtaylor: Ohhhhhh it's root partition or boot partition. Nevermind this comment.

PR #369, opened 2021-04-28
WIP plugin support

advaithm:
🚨 PR Guidelines:

# Describe your PR
WIP Plugin support, currently we don't have a way for users to use plugins. this is still pretty wip 
# Testing
untested will probably break

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -66,12 +66,12 @@ class Installer():
 		if len(plugins)>0:
 			for plugin in plugins:
 				plugin = load_plugin(plugins)
-				if plugin is not None:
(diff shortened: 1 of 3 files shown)


dylanmtaylor:
Can you make an example plugin as a test and demonstrate how this would be used?

dylanmtaylor:
Probably needs a lot of doc updates before it goes in as well.

advaithm:
yeah this is very wip, not sure if it even works

Torxed:
> yeah this is very wip, not sure if it even works

That's the spirit! :D

advaithm:
this is better implemented in  #393

PR #371, opened 2021-04-28
Avoid some word wrapping on smaller consoles like VirtualBox.

dylanmtaylor:
Not really needed, but this bothered me more than it should have. 

VirtualBox_Arch BIOS_28_04_2021_08_58_12


dylanmtaylor:
After:

VirtualBox_Arch BIOS_28_04_2021_09_00_06

PR #374, opened 2021-04-28
Cleaning up packages.

Torxed:
installer now relies on __packages__ definition. Which will work with external libs to more easily gather packages used by installer and profiles.


dylanmtaylor:
I made https://github.com/archlinux/archinstall/pull/378 as a complimentary change to this one; it has a packages definition in each profile and I eliminated application profiles that didn't do any sort of configuration work.

PR #379, opened 2021-04-28
Update generic_multi_select and generic_select

### LINKED ISSUE #402: Installation crashes if graphics card driver selection was skipped
In current master branch, there is an issue that causes the installation to crash when the user skips choosing a graphics card driver (Error example: `KeyError: None`)
**Solution:** Disable empty selection for graphics card driver selection

> Currently, pending PR #379 will close that issue once merged
> Fix commit: https://github.com/archlinux/archinstall/pull/379/commits/5f1156f80e4038a52719f4ba01c87eae4f0a8660

SecondThundeR:
# Describe your PR

This PR introduces some changes to `generic_multi_select` and `generic_select`, which also make them more unified among themselves *(Full changelogs can be found in commit description)*

In addition, the installation crash with an empty selection of the graphics card driver has been fixed (Closes #402)

### DIFF
--- a/archinstall/lib/user_interaction.py
+++ b/archinstall/lib/user_interaction.py
@@ -99,7 +99,7 @@ def print_large_list(options, padding=5, margin_bottom=0, separator=': '):
 	return column, row
 
 
-def generic_multi_select(options, text="Select one or more of the options above (leave blank to continue): ", sort=False, default=None, allow_empty=False):
+def generic_multi_select(options, text="Select one or more of the options above (leave blank to continue): ", sort=True, default=None, allow_empty=False):
 	# Checking if the options are different from `list` or `dict` or if they are empty
 	if type(options) not in [list, dict]:
 		log(f" * Generic multi-select doesn't support ({type(options)}) as type of options * ", fg='red')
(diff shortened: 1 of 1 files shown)


SecondThundeR:
At the moment, I have marked this PR as WIP due to two questions:
- Is it possible to have several kernels at once?
- At the moment there is a problem with the logs in multi-selection *(They are not shown at all after entering the option. Can be reproduced on the current master branch)*
- Is it worth resetting the selection to the default option if the user has removed all values ​​from the list?
  > (For example, having this https://github.com/archlinux/archinstall/blob/master/archinstall/lib/user_interaction.py#L111 makes https://github.com/archlinux/archinstall/blob/master/archinstall/lib/user_interaction.py#L130 pointless, as it will never be checked, since the list will always be non-empty, if the default value is not None)

Look forward to comments on this, where I could possibly be wrong

archinstall/lib/user_interaction.py
```diff
@@ -99,18 +99,29 @@ def print_large_list(options, padding=5, margin_bottom=0, separator=': '):
...
 
 	section = MiniCurses(get_terminal_width(), len(options))
 
 	selected_options = []
 
-	while True:
-		if len(selected_options) <= 0 and default and default in options:
-			selected_options.append(default)
+	if not selected_options and default in options:
+		selected_options.append(default)
```
Torxed, line 122: The reason why these are *inside* the loop, is to avoid confusion on empty selections resulting in the `default` value. This will graphically show that `default` has to be selected if nothing else is. Otherwise you could unmark everything, press enter, and something got selected anyway going against what's visible to the user during selection.

Torxed (changes requested):

SecondThundeR:
I updated my last comment. So the last issue with that is still:
> At the moment there is a problem with the logs in multi-selection
At the moment, these changes are good for further work with this feature to prevent possible accidental problems.

Torxed:
I'll run through this tomorrow, but I like the look of it!

SecondThundeR:
New commit addresses #402 which adds support for future default option of #418

PR #381, opened 2021-04-29
Fixes #350 hopefully. This reverts an old hotfix to make systemd-boot…

### LINKED ISSUE #350: Does not boot after installation
Hi, after completing installation with this installer rebooting brings me back to the installation menu.  Removing the iso file leaves me with no bootable device.  If I install manually this does not occur.

For reference I am installing on a UEFI system (XPS 17).

Torxed:
… work with dualboot variables. This **NEEDS** to be tested together with a Windows installation dual-boot setup, because this is where variable writing caused issues before.

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -397,8 +397,11 @@ class Installer():
 				for line in loader_data:
 					if line[:8] == 'default ':
 						loader.write(f'default {self.init_time}\n')
+					elif line[:8] == '#timeout' and 'timeout 5' not in loader_data:
+						# We add in the default timeout to support dual-boot
+						loader.write(f"{line[1:]}\n")
 					else:
-						loader.write(f"{line}")
+						loader.write(f"{line}\n")
 
 			## For some reason, blkid and /dev/disk/by-uuid are not getting along well.
 			## And blkid is wrong in terms of LUKS.
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -377,7 +377,7 @@ class Installer():
 			# And in which case we should do some clean up.
 
 			# Install the boot loader
-			sys_command(f'/usr/bin/arch-chroot {self.target} bootctl --no-variables --path=/boot install')
+			sys_command(f'/usr/bin/arch-chroot {self.target} bootctl --path=/boot install')
 
 			# Modify or create a loader.conf
 			if os.path.isfile(f'{self.target}/boot/loader/loader.conf'):


Redecorating:
On MacBooks with the T2 chip, writing to nvram causes a system crash. The fix for this is to mount efivars as read only:
```
mount efivarfs /sys/firmware/efi/efivars/ -o ro,remount -t efivarfs
```
But this makes bootctl without `--no-variables` fail like this (instead of making apple's t2 chip panic and crash the computer):
```
>  bootctl --path=/mnt install
Created "/mnt/EFI/systemd".
Created "/mnt/loader".
Created "/mnt/loader/entries".
Created "/mnt/EFI/Linux".
Copied "/usr/lib/systemd/boot/efi/systemd-bootx64.efi" to "/mnt/EFI/systemd/systemd-bootx64.efi".
Copied "/usr/lib/systemd/boot/efi/systemd-bootx64.efi" to "/mnt/EFI/BOOT/BOOTX64.EFI".
Random seed file /mnt/loader/random-seed successfully written (512 bytes).
Failed to write 'LoaderSystemToken' EFI variable: Read-only file system
```
Could running bootctl install with the --no-variables be added as a fallback if normal bootctl install fails?
If this is too specific of an issue (I don't know if you want archinstall to get filled with hardware specific quirks like this), or hard to implement, another option would be to have `sed -i 's/bootctl --/bootctl --no-variables --/' /lib/python3.9/site-packages/archinstall/lib/installer.py` run by people on T2 MacBooks, before archinstall is run (I don't know if there'd be anyone that needs this who wouldn't be using [this profile I mentioned on the discord a while ago](https://github.com/Redecorating/archinstall-mbp/blob/main/apple-t2.py) so If I put it in there it should cover anyone who needs it).

Torxed:
> Could running bootctl install with the --no-variables be added as a fallback if normal bootctl install fails?

First of all, thank you so much for testing this PR.
I knew that this specific PR would cause some issues some where, just not where.

I think running a fallback to running with `--no-variables` is a neat solution when it crashes/fails.
We already have some hardware detection, but minimizing the need for it is always preferred from my side of things, less to maintain ^^

I'll research if there's a "best practice" way to detect if `efivars` is writable, `ls /sys/firmware/efi/efivars` something something is probably the right way. Or something like `efivar --list` and `efivar --print --name=<name>` followed by a test `efivar --write --name` to test-write.. (spitballing ideas here)

Redecorating:
I think you could use `mount|grep efivarfs|grep rw` to check if its mounted as rw.

Torxed:
Tested, but there's another issue in the pipes before I can confirm dual-boot working: 
2021-05-12-144336_1280x987_scrot

Torxed:
> On MacBooks with the T2 chip, writing to nvram causes a system crash. The fix for this is to mount efivars as read only:
> 
> ```
> mount efivarfs /sys/firmware/efi/efivars/ -o ro,remount -t efivarfs
> ```
> 
> But this makes bootctl without `--no-variables` fail like this (instead of making apple's t2 chip panic and crash the computer):
> 
> ```
> >  bootctl --path=/mnt install
> Created "/mnt/EFI/systemd".
> Created "/mnt/loader".
> Created "/mnt/loader/entries".
> Created "/mnt/EFI/Linux".
> Copied "/usr/lib/systemd/boot/efi/systemd-bootx64.efi" to "/mnt/EFI/systemd/systemd-bootx64.efi".
> Copied "/usr/lib/systemd/boot/efi/systemd-bootx64.efi" to "/mnt/EFI/BOOT/BOOTX64.EFI".
> Random seed file /mnt/loader/random-seed successfully written (512 bytes).
> Failed to write 'LoaderSystemToken' EFI variable: Read-only file system
> ```
> 
> Could running bootctl install with the --no-variables be added as a fallback if normal bootctl install fails?
> If this is too specific of an issue (I don't know if you want archinstall to get filled with hardware specific quirks like this), or hard to implement, another option would be to have `sed -i 's/bootctl --/bootctl --no-variables --/' /lib/python3.9/site-packages/archinstall/lib/installer.py` run by people on T2 MacBooks, before archinstall is run (I don't know if there'd be anyone that needs this who wouldn't be using [this profile I mentioned on the discord a while ago](https://github.com/Redecorating/archinstall-mbp/blob/main/apple-t2.py) so If I put it in there it should cover anyone who needs it).

I've added in this PR and the fix, but I also added (in `master`) a fallback if it fails, hopefully this addresses the Apple issue a bit at least. But we'll work more on that to make it more robust.

https://github.com/archlinux/archinstall/blob/0df6eced77ce54177a9a38f1d6d2bc3fb8f91fc9/archinstall/lib/installer.py#L385-L386

PR #387, opened 2021-04-29
Add Apache HTTPD Application Profile

dylanmtaylor:
Similar to PostgreSQL or Nginx this is intended to improve using ArchInstall as a library and won't be exposed via guided.


l4zy0n3:
We could have one for nginx too

dylanmtaylor:
> We could have one for nginx too

See the other PR.

PR #391, opened 2021-04-29
Implement server top-level profile

dylanmtaylor:
This will implement a server profile that installs a minimal system and uses multi_generic_select functionality to select application profiles (servers) desired on the target system.


dylanmtaylor:
VirtualBox_Arch_29_04_2021_10_24_15

Rada8732 (approved):

PR #393, opened 2021-04-29
Adding in plugin support and on_<event> calls in installation.

Torxed:
Adding `on_pacstrap` hook for `installation()` that plugins can implement as a first test-hook.

A plugin example can be found in https://github.com/Torxed/archinstall-aur/blob/main/archinstall-aur.py:
```python
import archinstall

__version__ = 0.1

class Plugin():
	def on_pacstrap(self, packages :list) -> list:
		print(f"Identifying AUR packages in package list: {packages}")

		# ... do AUR work on the packages ...

		# Returns a curated list of packages that exludes any AUR packages.
		# This allows installation.pacstrap() to contain AUR packages,
		# but won't handle them or try to install them since we remove those here.
		return packages

def dummy_example(*args, **kwargs):
	pass

# Example function injection
archinstall.plugin_function = dummy_example
```

Plugins does not require to initialize themselves, it's handled via the plugin loader logic.
An additional `load_plugin(path)` function has been added for loading plugins via path instead of having to have them installed separately.

The following `on_<event>` hooks have been implemented:

* on_pacstrap
  ---------------
  This will enable plugins to get an early call before packages are installed.
  This way the plugin can curate the list of packages based on validity, mirror (adding custom mirrors for instance) etc.
* on_mirrors
  ---------------
  This enables a pre-set-of-mirrors call, so that the plugin can curate the mirrorlist.
  How that mirror list is curated is up to the plugin, as long as it's a valid list of mirrors.
* on_timezone
  ---------------
  This would allow plugins to change/update the timezone or perform other tasks on the installation based of the zone given.
* on_service
  ---------------
  After a service has been activated the service name is given to the plugin for further work if needed.
* on_configure_nic
  ---------------
  This hook allows plugins to change/update the systemd-network INI configuration before it's set in the installation. For instance, DNS servers or gateways can be added, or custom enterprise flags can be set in the config. Or the plugin could do work based on the information found.
* on_mkinitcpio
  ---------------
  This hook is called before `mkinitcpio` is executed, and **if a return value is given by the plugin, `mkinitcpio` will not be called**. This allows the plugin to both override the use of mkinitcpio but also tweak the variables sent to mkinitcpio.
* on_install
  ---------------
  After a successful install, this hook is called to let plugins do post-configuration of the base OS.
* on_add_bootloader
  ---------------
  This hook will interrupt the installation and configuration of supported bootloaders if a return value was given by the plugin on this call. This enables plugins to interrupt the installation and install a custom boot loader that would otherwise be unsupported, like `syslinux` for instance.


Torxed:
I'm transferring in @phisch's message from [discord](https://discord.com/channels/726808923662712894/829454166991568927/844930191921905704):

> you should make it so you can also pass a plugin to archinstall inside a separate install script. something like:
> ```python
> #!/usr/bin/env python3
> import archinstall
> 
> archinstall.load_plugin('https://github.com/Torxed/archinstall-aur/blob/main/archinstall-aur.py') # replace with release url
> ```
> i'd be way too lazy to type a plugin dependency when launching an install script that depends on a plugin

Torxed:
Maybe useful: https://setuptools.readthedocs.io/en/latest/pkg_resources.html#entry-points

Torxed:
Swapped to Python's built-in entry point system for dealing with plugins.
Also added the @phisch suggestion of adding a `load_plugin` which supports both local and remote files *(`import_via_path()` and `localize_path()` are support functions to `load_plugin()`)*.

PR #398, opened 2021-04-29
Partially revert hasUEFI grub/efibootmgr change to fix GRUB

dylanmtaylor:
🚨 PR Guidelines:

# New features *(v2.2.0)*

Any patch work to existing verions will have to create a new branch against the tagged versions.

# Describe your PR

# Testing

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -51,6 +51,10 @@ class Installer():
 		self.base_packages = base_packages.split(' ') if type(base_packages) is str else base_packages
 		for kernel in kernels:
 			self.base_packages.append(kernel)
+		if hasUEFI():
+			self.base_packages.append("efibootmgr")
+		else:
+			self.base_packages.append("grub")
 
 		self.post_base_install = []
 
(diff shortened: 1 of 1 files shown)


archinstall/lib/installer.py
```diff
@@ -51,6 +51,10 @@ def __init__(self, target, *, base_packages=__packages__[:3], kernels=['linux'])
 		self.base_packages = base_packages.split(' ') if type(base_packages) is str else base_packages
 		for kernel in kernels:
 			self.base_packages.append(kernel)
+		if hasUEFI():
+			self.base_packages.append("efibootmgr")
+		else:
+			self.base_packages.append("grub")
```
Torxed, line 57: I don't like this revert, it doesn't make sense to guess and strap in packages that may or may not be used. I fail to see how this would fix the problem as well seeing as `grub` will never be installed but we'll call `grub-install` later in the code. I think we should patch line 433 instead with:
```python
self.pacstrap('grub')
if hasUEFI():
    self.pacstrap('efibootmgr')`

Torxed (changes requested):

archinstall/lib/installer.py
dylanmtaylor: Changed in latest commit.

PR #401, opened 2021-04-30
Adding kernel parametrization

Torxed:
This PR will add support for custom kernel parameters from outside sources.


Torxed:
2021-05-14-180452_1536x1179_scrot

Tested and works so I'll merge it in.
There's still some stuff that might be nice to have, like `on_mkinitcpio` hooks, but for now that can be achieved by profiles or dotfiles simply adding stuff to `installation.KERNEL_PARAMS` and then running `installation.mkinitcpio('-P')`.

PR #404, opened 2021-05-01
Python Linter/Syntax Checking Action and Various PEP 8 Fixes

dylanmtaylor:
https://github.com/marketplace/actions/python-syntax-checker

### DIFF
--- a/.github/workflows/lint-python.yaml
+++ b/.github/workflows/lint-python.yaml
@@ -1,0 +1,21 @@
+on: [push, pull_request]
+name: Lint Python and Find Syntax Errors
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    container:
+      image: archlinux:latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: pacman --noconfirm -Syu python python-pip
+      - run: python -m pip install --upgrade pip
+      - run: pip install flake8 pytest
+      - name: Lint with flake8
+        run: |
+          # stop the build if there are Python syntax errors or undefined names
+          flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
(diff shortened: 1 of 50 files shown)


dylanmtaylor:
The commit I made to run the code formatter reduced the flake8 issues from over 4000 to something like 573. It's a noisy commit, but if we _maintain_ a consistent format going forward, it will keep diffs smaller and follow a more standard coding format.

SecondThundeR:
According from CONTRIBUTING.md:
> Archinstall uses tabs instead of spaces simply to make it easier for non-IDE developers to navigate the code (Tab display-width should be equal to 4 spaces). Exception to the rule are comments that need fine-tuned indentation for documentation purposes.

Your commits changed all tabs to spaces, which is invalid according to the quote above *(Interestingly, the tabs are left only in the comments)*

> Blank lines before/after imports and functions are not followed and discouraged. One space is commonly used in archinstall.

Your commits added two spaces between functions, imports and etc.

> Multiple Imports on the same line is allowed, but more than five imports should be avoided on any given line. This simply saves up some space at the top of the file (for non-IDE developers) and will not be enforced.

It is good, of course, that each import is on a new line, but this increases the number of lines and in places where there are many imports, we have a lot of lines at the beginning of the file

> String quotes follow PEP8, the exception being when creating formatted strings, double-quoted strings are preferred but not required on the outer edges (Example: f"Welcome {name}" rather than f'Welcome {name}').

Double-quoted strings should be used only for f-strings. But this moment is controversial, so an additional conversation with Torxed is needed

=====

In conclusion, addition of linter is a very good, but are these changes, mentioned above, worth it? Maybe it's better to add these ignore rules to fit the project guidelines

Torxed:
I don't mind the imports and spaces that much, but the  tabs are not going away and the string quotas is debatable.

PR #408, opened 2021-05-01
Fix #406

### LINKED ISSUE #406: _gfx_driver_packages not defined when choosing sway option (current master build)
VirtualBox_Arch_30_04_2021_23_01_50

dylanmtaylor:
Closes #406 

🚨 PR Guidelines:

# New features *(v2.2.0)*

Any patch work to existing verions will have to create a new branch against the tagged versions.

# Describe your PR

# Testing

### DIFF
--- a/profiles/sway.py
+++ b/profiles/sway.py
@@ -4,7 +4,19 @@ import archinstall
 
 is_top_level_profile = False
 
-__packages__ = ["sway", "swaylock", "swayidle", "waybar", "dmenu", "light", "grim", "slurp", "pavucontrol", "alacritty"]
+__packages__ = [
+	"sway",
+	"swaylock",
+	"swayidle",
+	"waybar",
+	"dmenu",
+	"light",
+	"grim",
+	"slurp",
+	"pavucontrol",
+	"alacritty",
+]
+
 
 def _prep_function(*args, **kwargs):
 	"""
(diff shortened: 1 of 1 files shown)


dylanmtaylor:
Seems to work from some quick testing

profiles/sway.py
```diff
@@ -13,13 +13,13 @@ def _prep_function(*args, **kwargs):
 	other code in this stage. So it's a safe way to ask the user
 	for more input before any other installer steps start.
 	"""
-	if "nvidia" in _gfx_driver_packages:
+	__builtins__['_gfx_driver_packages'] = archinstall.select_driver()
```
Torxed, line 16: This driver will never be installed unless `xorg` is also installed. Which I can't see that sway does?

Torxed (changes requested):

profiles/sway.py
dylanmtaylor: Wayland uses hardware acceleration as far as I know. Graphics drivers provide that and I believe the same ones are used for Wayland and X

dylanmtaylor:
Need to test that before merging it.

dylanmtaylor:
Tested to work.
VirtualBox_Arch_01_05_2021_09_55_50

PR #420, opened 2021-05-06
add dmenu dependency to i3 profile

satai:
`dmenu` is configured as the default launcher (🐧-SPACE key)
in the default i3 configuration.

The same dependency exists for the same reason in the `sway` profile


dylanmtaylor:
Sounds sensible.

Torxed:
Sorry for the delay, had a busy week. Cheers for the contribution! :)

PR #422, opened 2021-05-08
fix error when there are to many options to print and calculation spaces

arieboven:
# Describe your PR

When there are to many options to print which does not fit on the screen there is an error with undefined variables. 
Also the calculation of the spacing does not take into account the index and separator. This can lead to rows that are too long.
I have made some changes that fixes these errors

# Testing

Tested in terminal and virtualbox.

### DIFF
--- a/archinstall/lib/user_interaction.py
+++ b/archinstall/lib/user_interaction.py
@@ -83,16 +83,18 @@ def get_password(prompt="Enter a password: "):
 def print_large_list(options, padding=5, margin_bottom=0, separator=': '):
 	highest_index_number_length = len(str(len(options)))
 	longest_line = highest_index_number_length + len(separator) + get_longest_option(options) + padding
+	spaces_without_option = longest_line - (len(separator) + highest_index_number_length)
 	max_num_of_columns = get_terminal_width() // longest_line
 	max_options_in_cells = max_num_of_columns * (get_terminal_height()-margin_bottom)
 
 	if (len(options) > max_options_in_cells):
 		for index, option in enumerate(options):
 			print(f"{index}: {option}")
+		return 1, index
 	else:
 		for row in range(0, (get_terminal_height()-margin_bottom)):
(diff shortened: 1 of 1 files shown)


Torxed:
Could you take a screenshot of the issue so we know what issue this solves?
Because [this line](https://github.com/archlinux/archinstall/pull/422/files#diff-991834f970d8b0af702afcd0581ed8edc8fd87c0aec535af2d375e04b0518a43R85) takes into account the separator for instance. Reading the code it's probably true that the index isn't taken into account tho :)

arieboven:
Sure, it is easier to see in screenshots, maybe I should have added them in the first place.

I just made a long list so it is easier to show.

| Current state | Pull request | Description |
| -----|------|------|
|  |  | When there are to many options to fit on the screen it just prints each option on a separated row. The error is that when this happens the for loop does not execute which means the variables `column` and `row` are not defined so it throws an error. |
|  |  | With the calculation of the spaces it does not take into account the index and separator length. This means that the rows can get to long, when it is close to using the maximum columns. Which means that it does not fit anymore on the screen as one row.  |

I hope this makes it somewhat better to understand.

Torxed:
> Sure, it is easier to see in screenshots, maybe I should have added them in the first place.
> 
> I just made a long list so it is easier to show.
> 
> Current state	Pull request	Description
> 		When there are to many options to fit on the screen it just prints each option on a separated row. The error is that when this happens the for loop does not execute which means the variables `column` and `row` are not defined so it throws an error.
> 		With the calculation of the spaces it does not take into account the index and separator length. This means that the rows can get to long, when it is close to using the maximum columns. Which means that it does not fit anymore on the screen as one row.
> I hope this makes it somewhat better to understand.

This! This is the stuff! Thank you! I'll run the attached ISO just to verify, but looks really good!

Torxed:
Great work! Thank you for both pointing it out and fixing it, much appreciated!

PR #425, opened 2021-05-10
Auto retry on failure for bad internet.

clh021:
🚨 PR Guidelines:

# New features *(v2.2.0)*

Auto retry on failure for bad internet.

# Testing

Testing on virtualbox.

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -128,13 +128,19 @@ class Installer():
 		if type(packages[0]) in (list, tuple): packages = packages[0]
 		self.log(f'Installing packages: {packages}', level=logging.INFO)
 
-		if (sync_mirrors := sys_command('/usr/bin/pacman -Syy')).exit_code == 0:
-			if (pacstrap := sys_command(f'/usr/bin/pacstrap {self.target} {" ".join(packages)}', **kwargs)).exit_code == 0:
-				return True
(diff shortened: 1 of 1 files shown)


dylanmtaylor:
This feels like a small bandaid instead of a refind fix. We should not do a fixed number of retries automatically (5 in this case). Instead, we should detect that an installation has failed due to networking, show what failed specifically, and prompt users "Would you like to retry package installation? [Y/n] " or similar.

Torxed:
> This feels like a small bandaid instead of a refind fix. We should not do a fixed number of retries automatically (5 in this case). Instead, we should detect that an installation has failed due to networking, show what failed specifically, and prompt users "Would you like to retry package installation? [Y/n] " or similar. If it's an unattended installation, defaulting to 'y' for that and trying a few times might make sense though.

I agree with this one.
The logic should be in each individual network step, they should raise an exception when there's no possibly way to achieve the functions goal. Which we should handle in `guided` and `unattended` respectively.

The unattended installer I wouldn't mind if it re-tried a couple of times before erroring out completely with a log message tho.

Torxed:
Since we haven't heard anything for a month on this PR and the suggested way forward can't be met, we'll open a separate PR to solve this issue. Because the issue is still relevant - just the approach that needs to be refined in terms of solving it :)

PR #426, opened 2021-05-11
Re-work partitioning logic, introducing more granularity (and stability)

Torxed:
The following PR will introduce *some* breaking changes to the work-flow of `guided`. This should be a non-issue since backwards comparability of the guided installer isn't really a thing - due to the fact that it uses the shipped version of archinstall.

That said, API calls should be unaffected *(I believe, might double down on this)*.

New features
------------------

 * Partitioning in guided should now introduce more granular options (while still giving some "easy options")
 * Partitioning has been made more robust by separating partitions from partitioning logic. Which means partitions are partitions, and partitioning-stuff is kept separate, this avoids confusion in cache-logic, overall code complexity is reduced and a more clear line between the different objects and features.

The guided profile now has a `disk_layouts` section to it's configuration, which replaces the old `harddrive` layout. It allows for multiple block devices and multiple partitions. Each with their own individual flags. Here is an example of the structure which could technically be loaded through a parameter for re-running layouts:

```python
{
	"/dev/sda": {             # Block Device level
		"wipe": False,    # Safety flags
		"partitions" : [  # Affected  / New partitions
			{
				"PARTUUID" : "654bb317-1b73-4339-9a00-7222792f4ba9", # If existing partition
				"wipe" : False, # Safety flags
				"boot" : True,  # Safety flags / new flags
				"ESP" : True,   # Safety flags / new flags
				"mountpoint" : "/mnt/boot"
			}
		]
	},
	"/dev/sdb" : {
		"wipe" : true,
		"partitions" : [
			{
				# No PARTUUID required here since it's a new partition
				"type" : "primary", # parted options
				"size" : "100%",
				"filesystem" : {
					"encrypted" : True, # TODO: Not sure about this here
					"format": "btrfs",  # mkfs options
				},
				"mountpoint" : "/mnt"
			}
		]
	}
}
```
TODO:
---------

This in turn can be shipped into `archinstall.Filesystem(desired_disk_layout)` which will convert the desired layout into actionable results. Essentially making `archinstall.Filesystem()` a bit easier to work with.


examples/guided.py
```diff
@@ -417,4 +345,6 @@ def perform_installation(mountpoint):
 	else:
 		archinstall.arguments['profile'] = None
 
-perform_installation_steps()
+ask_user_questions()
```
l4zy0n3, line 348: this is already covered in the **if** section, we can remove this line

Torxed:
Re-worked the logic a bit in `guided.py` and `Installer()` to accommodate the new partitioning layout.
It now properly executes:
 * Wipes the drive if it's marked for wiping (properly clears any old partitions)
 * Creates new partitions and keeps track of them by UUID rather than size and order
 * Formats the partition based on the discovered UUID and not by order of appearance
 * `Installer()` now has a `mount_ordered_layout()` that takes the same layout structure as `Filesystem()` does, and mounts them in sorted order, meaning `/` gets mounted first, then followed by `/boot` and `/home` for instance. And any sub-mount of those gets mounted and so on.
Then hands off to the installer:
output

Torxed:
Remaining things:
 * Test multiple hard-drive selections
 * Large drives
 * NVMe drives
 * Encryption
 * Guide the user through advanced partition layout creations where the choices are free

NVMe drives are partially tested through a VM on a NVMe drive, but should be tested on real hardware as well.

examples/guided.py
```diff
@@ -339,7 +231,14 @@ def perform_installation(mountpoint):
 	Only requirement is that the block devices are
 	formatted and setup prior to entering this function.
 	"""
+
+
 	with archinstall.Installer(mountpoint, kernels=archinstall.arguments.get('kernels', 'linux')) as installation:
+		# Mount all the drives to the desired mountpoint
+		# This *can* be done outside of the installation, but the installer can deal with it.
+		for drive in archinstall.arguments['harddrives']:
+			installation.mount_ordered_layout(archinstall.storage['disk_layouts'][drive])
```
Torxed, line 240: There might be mount points that overlap between drives, causting them to be mounted in the incorrect order, `mount_ordered_layout()` should take all drives into account.

Torxed:
Remaining things:
 * Encryption
 * Guide the user through advanced partition layout creations where the choices are free

weilbith:
Hey 👋 
Looking forward to see this feature merged!

Just as a possibly interesting note: you might want to consider how [systemd discoverable partitions](https://systemd.io/DISCOVERABLE_PARTITIONS/) could be included here. But that would also kinda affect the `fstab` file generation (in fact all these partitions don't need an entry).
I currently plan to write a custom profile and plugin to achieve this feature. But it would be amazing if `archinstall` would have native support for it.

PS:
While already on it, you might be interested into [repart.d](https://www.freedesktop.org/software/systemd/man/repart.d.html) (plus `growfs`). But I just want to mention it, in case it might be interesting for you to include into your thoughts while working on this feature.

Torxed:
> Hey 
> Looking forward to see this feature merged!
> 
> Just as a possibly interesting note: you might want to consider how [systemd discoverable partitions](https://systemd.io/DISCOVERABLE_PARTITIONS/) could be included here. But that would also kinda affect the `fstab` file generation (in fact all these partitions don't need an entry).
> I currently plan to write a custom profile and plugin to achieve this feature. But it would be amazing if `archinstall` would have native support for it.
> 
> PS:
> While already on it, you might also be interested into [repart.d](https://www.freedesktop.org/software/systemd/man/repart.d.html) (plus `growfs`). But I just want to mention it, in case it might be interesting for you to include into your thoughts while working on this feature.

I've been considering it.
But since I have very little knowledge about systemd's auto-magic stuff, even basic configuration would take a while to wrap my head around. But if systemd can help us with this, it would for sure be a welcome change. Once this is done, that may be the next step in the partitioning road-map.

Humm, [repart.d](https://www.freedesktop.org/software/systemd/man/repart.d.html) sounds like a thing a lot of cloud providers use to grow volumes easily. I think this PR is growing so large that introducing new features might be sketchy. But I would love a separate issue here on github for supporting [repart.d](https://www.freedesktop.org/software/systemd/man/repart.d.html) to open up the discussion and see what people think.

If you want feel free to open a ticket for `systemd mounts to replace fstab` as well, if you don't I will tomorrow hehe.

dylanmtaylor:
> Hey 
> Looking forward to see this feature merged!
> 
> Just as a possibly interesting note: you might want to consider how [systemd discoverable partitions](https://systemd.io/DISCOVERABLE_PARTITIONS/) could be included here. But that would also kinda affect the `fstab` file generation (in fact all these partitions don't need an entry).
> I currently plan to write a custom profile and plugin to achieve this feature. But it would be amazing if `archinstall` would have native support for it.
> 
> PS:
> While already on it, you might also be interested into [repart.d](https://www.freedesktop.org/software/systemd/man/repart.d.html) (plus `growfs`). But I just want to mention it, in case it might be interesting for you to include into your thoughts while working on this feature.

If you can come up with better, cleaner ways of doing things, we're definitely open to pull requests and it'd benefit more people to have it in the installer itself rather than having a plugin for something fundamental like partitioning :)

archinstall/lib/disk.py
```diff
@@ -14,10 +13,166 @@
...
 
-# import ctypes
-# import ctypes.util
-# libc = ctypes.CDLL(ctypes.util.find_library('c'), use_errno=True)
-# libc.mount.argtypes = (ctypes.c_char_p, ctypes.c_char_p, ctypes.c_char_p, ctypes.c_ulong, ctypes.c_char_p)
+def valid_fs_type(fstype :str) -> bool:
+	# https://www.gnu.org/software/parted/manual/html_node/mkpart.html
+
+	return fstype in [
+		"ext2",
```
dylanmtaylor, line 20: Why just ext2 and not ext3 or ext4 here?

dylanmtaylor: This should work: 

sudo parted -a opt /dev/sda mkpart primary ext4 0% 100%

Torxed: `ext4` gets ignored I presume as it's not a [supported filesystem](https://www.gnu.org/software/parted/manual/html_node/mkpart.html).

dylanmtaylor:
VirtualBox_Arch_13_06_2021_21_45_07
I greatly dislike that this does not show block device capacity. Also, none and the Arch installer should be filtered out.

dylanmtaylor:
VirtualBox_Arch_13_06_2021_21_47_17
We need to better explain how the "best effort layout looks on the disk. Maybe make it work like this:

Options:
* Suggest a best effort layout (this will display the layout and confirm before asking non-partitioning questions)
* Fully manual

After either option, display it and ask "Would you like to use this layout?" or such.

dylanmtaylor:
Ran into this issue testing this pull request out on a VM I had already used (recommended layout):

VirtualBox_Arch_13_06_2021_21_52_36

VirtualBox_Arch_13_06_2021_21_53_43

Torxed:
There's little to no point in testing this until I mark it as ready for review.
There will be quite a few bugs along the line and yes, information regarding BlockDevices will be printed.
Just not something I focused on before getting even the basic functionality done.

dylanmtaylor:
I feel like it wouldn't be too hard to solve #563 as part of this change too.

Torxed:
Getting ready for general testing of this PR pretty soon.
I'd like to add a `validate_partition_layout_against_arguments` or something similar. Currently if partitions is re-used as is, and only mountpoints are given. The intended use is that partitions should remain the same but get formatted as-is. This becomes a problem if no password was given in the disk-encryption-password-stage, but a encrypted partition is re-used.. Currently the logic will ask for user input, but it will do so after configuration is confirmed and partitioning starts.
Adding a pre-run-check for partition layouts should solve this.
I've made it so that `--silent` won't trigger the above logic, but instead error out.. Not clean either, but it's a start.

I've tested most use cases now and I think the bare minimum is there.
Some phrasing needs to be re-worked and some error handling for the above, but as I said, functionality is mostly there now.

I can wipe a drive, I can re-use partitions. Encryption works. Have to test skipping the step all together tho as a last test.

Torxed:
I think I'm ready for reviews on this one.
The default partition layouts seams to be robust enough to be tested.
The guided partitioning is still a bit wonky - but testable for bugs and improvements.

I don't think it's really ready for merging in to `master`. Keep this in mind :)

dylanmtaylor:
> I think I'm ready for reviews on this one.
> The default partition layouts seams to be robust enough to be tested.
> The guided partitioning is still a bit wonky - but testable for bugs and improvements.
> 
> I don't think it's really ready for merging in to `master`. Keep this in mind :)

I almost am of the opinion that this massive change should be merged into master and fixes should be made against that, using master as a 'dev' branch of sorts.

Torxed:
I'm note entirely sure I nailed the `load_config()` and `ask_user_questions()` part regarding disk actions. But did as good as I could and will have to test master properly to ensure no edge cases pop up.

PR #428, opened 2021-05-12
A example mypy workflow

advaithm:
This will fail 100% of the time currently due to things archinstall does a bit weirdly


Torxed:
Just dropping a comment mentioning #430 will partially address this.

Torxed:
I assumed it was safe to merge in the latest changes from master, which I did through your master.
Your master contained some things such as plugin support as well as secure boot, tried to revert those but that also reverted the latest upstream `master` changes.

I apologize but I think I messed up this PR a bit. I'll wait with the merge of this one until we properly sort out the issues hehe.

PR #430, opened 2021-05-12
BC: Replaced the magic __builtin__ global variable for installation

Torxed:
This will break existing profiles if anyone has created their own profiles.
However, it will be a more reliably and robust change for the future.

This replaces `__builtin__['installation']` with `archinstall.storage['installation_session']`


Torxed:
screenshot

Appears to be holding up.
Haven't tested every single profile, but the imports of them all appears to hold up.

Issue #266, opened 2021-04-10
Grub and os-prober option (for dual boot)


jatin-cbs:
Currently this installer has systemd-boot. However grub has more options and also has support for os-prober (as an optional dependency) to detect other OS installed and dual boot which is a case for many users. 

1. Adding option to use grub instead of systemd-boot will also help in solving #42

2. Dual boot support in the grub menu (grub2-mkconfig will add entries for other operating systems it can find. That will be done based on the output of the os-prober tool. )  os-prober is an optional dependency for grub which fits in this case.

Torxed:
Thank you for the suggestion, this is implemented in #113 and a duplicate of #42.
The current roadmap/plan is to implement this into v2.2.0 which is due in roughly a month.

The next release which should be coming in two weeks time will focus on bug fixes and some quality of life changes to the current code without adding to much new complexity. But it's coming, so hang tight! :)

jatin-cbs:
@Torxed  there is nothing about os-prober. I opened the issue for dual boot support as well.

Foxboron:
With the recent set of CVEs for grub os-prober is going to be disabled by default from 2.06 and on. It shouldn't be used with `archinstall` and people ought to figure out a better way to support dual boot sadly.

Torxed:
I'm currently designing some new API calls that will allow us to detect other installations some what reliably.
This is done so that we can detect other installations across disks and not re-create the `/boot` partition if there's on available with enough space left over for our boot binaries.

Would this work as a workaround @jatin-cbs? It's not `os-prober`, but it's similar and will be an API call in this library.
Which the guided installer can use in-place of `os-prober` with either grub or systemd.

Torxed:
I've done some extensive testing on the dual-boot support and mainly because #381.
As it stands currently, there's a regression bug that causes formatting to be executed twice for some reason (still investigating why), the first one behaves as expected. But the second formatting ignores any safety flags and safety checks - which means dual-boot is currently broken.

Unless you use "Use `/mnt` as is" option, all other options including *"keep partitions layout"* are currently broken in terms of dual-boot support.

Torxed:
Would it be possible to install `os-prober` before we install grub via `arch-chroot` to detect other operating system and let `grub-mkconfig` detect it, and before we exit the installer/live ISO we uninstall it so that it doesn't remain and cause security issues?

If `os-proper` is required constantly, then we have an issue.

Torxed:
[...]

PR #434, opened 2021-05-14
Fixes dual-boot and /boot getting wiped accidentally

Torxed:
Currently, as mentioned in #434 - Dual-boot is broken unless *"Use `/mnt` as is"* option is being used.
Any other formatting option will wipe the `/boot` partition and any boot loaders with it.

This was discovered in #381 and mentioned in #266.


Torxed:
2021-05-14-201110_615x171_scrot

`/boot` is no longer explicitly touched.

PR #436, opened 2021-05-14
Fixes #292: Removed dupe formatter

### LINKED ISSUE #292: Duplicate formatting on master
Looks like partitions are getting formatted twice. It's not breaking anything but might be worth fixing eventually.

Torxed:
There should be no reason to call `.format()` here, since the steps above take care of all formatting.


Torxed:
image

This removes the dupe formatting if disk encryption is omitted.

PR #437, opened 2021-05-15
Don't list __init__.py files in profile list

dylanmtaylor:
Seeing `__init__` in the list is a bug; don't list these. Tested and working :)


dylanmtaylor:
Need to test this one before merging

PR #438, opened 2021-05-15
Fix microcode not getting added to base_packages before install

dylanmtaylor:
We were appending after the pacstrap so the changes were not taking effect.


dylanmtaylor:
Tested and seems ready to go in.

VirtualBox_Arch_14_05_2021_22_32_44

PR #442, opened 2021-05-15
rename package name

marcuxyz:
The `gpicview-gtk3` package was renamed for just  `gpicview`.

image

https://archlinux.org/packages/community/x86_64/gpicview/

this is causing an error when i try install the arch linux through of `archinstall`


Torxed:
Thanks for spotting this and submitting a PR! :)

Issue #107, opened 2021-02-07
Setting keyboard layout doesn't translate over to X11


Torxed:
`installation.set_keyboard_language` does not translate over to desktop environments as discussed with @phisch.
Traditionally I would have created something in '/etc/X11/xinit/xinit.rc` with `setxkbmap se` or similar.

But something like:
```
$ localectl set-keymap "" && localectl set-keymap de
$ localectl set-keymap de && localectl set-x11-keymap de
```
Would work, we need to investigate how this translates over to `sddm`, `gdm` and `xinit`.

---

To get the currently supported list, I suggest using `localectl list-keymaps` and `localectl list-x11-keymap-layouts`.
As pointed out by @phisch:
> VC and X11 keymaps are also different standards

The "languages" if you will don't follow the same naming convention, making it ridiculously hard to make a one-shot-function to set these things. Either we'll need to ask the user two times for language selection, or manually keep a translation table up to date *(I see this backfiring in the future)*. We would only need to ask users twice if they opted in for a desktop environment tho, which most new users probably would do.

Any suggestions on how to solve this as cleanly as possible are welcome.

grazzolini:
localectl sets things on /etc/X11/xorg.conf.d as well as it deals with /etc/vsoncole.conf and /etc/locale.conf. In my opinion is the cleanest solution to manage keyboard layouts and locales in a consistent way. We have to keep in mind that localectl doesn't touch any wayland setting, but usually those are compositor dependent.

phisch:
Not exactly sure about the specific case, but i ran into an issue on a system when running `arch_chroot(f'localectl set-x11-keymap {KEYMAP}')` because apparently the keymap was not available. Running the same command after the installation worked fine though. So i am not sure if something other than installing xorg-server is necessary for this to work properly.

grazzolini:
> Not exactly sure about the specific case, but i ran into an issue on a system when running `arch_chroot(f'localectl set-x11-keymap {KEYMAP}')` because apparently the keymap was not available. Running the same command after the installation worked fine though. So i am not sure if something other than installing xorg-server is necessary for this to work properly.

I think the issue here is that X must be installed before running this command. So, localectl stuff should be ran after package installation.

phisch:
Yeah i thought so too, but i ran into this issue after installing the x11 packages.
Reverted it until i figure out why it causes an issue.
https://github.com/phisch/dotfiles/commit/16feb270ecf197c6435ccf17da8ff0945d4c2f07

[...]

PR #443, opened 2021-05-15
Moving to localectl for all locale related activities

Torxed:
This should solve a lot of inconsistencies in terms of both listing but also setting locales.
This should also translate well over to X11 (but not wayland), see https://github.com/archlinux/archinstall/issues/107#issuecomment-775262644 for more information.


Torxed:
One issue remains, and that is `X11` is not available on the ISO, so performing keyboard locale selection based off `localectl --no-pager list-x11-keymap-layouts` won't work. Because:

2021-05-15-180605_1285x99_scrot

dylanmtaylor:
> One issue remains, and that is `X11` is not available on the ISO, so performing keyboard locale selection based off `localectl --no-pager list-x11-keymap-layouts` won't work. Because:
> 
> https://github.com/archlinux/archinstall/blob/b4fcc97bf182ce2406a907dff49ce2dbbdab8a41/archinstall/lib/locale_helpers.py#L13-L15
> 
> 2021-05-15-180605_1285x99_scrot

Can we leverage arch-chroot somehow to run the locale command on the target? It'd be halfway through the install though which is ... not ideal.

Torxed:
> > One issue remains, and that is `X11` is not available on the ISO, so performing keyboard locale selection based off `localectl --no-pager list-x11-keymap-layouts` won't work. Because:
> > https://github.com/archlinux/archinstall/blob/b4fcc97bf182ce2406a907dff49ce2dbbdab8a41/archinstall/lib/locale_helpers.py#L13-L15
> > 
> > 2021-05-15-180605_1285x99_scrot
> 
> Can we leverage arch-chroot somehow to run the locale command on the target? It'd be halfway through the install though which is ... not ideal.

Since we don't need to list the locales, if what @phisch mentioned holds up:

> This somehow sets the keymap for both x11 and vc, BUT from my testing, this only works after the locale has been set to an empty string. So, running localectl set-keymap "" && localectl set-keymap "de".

Then we'd only have to set the locale using `localectl` during installation, and if `x11` packages are installed at the time of executing, both should get set. But this means locale should be set as late as possible to ensure profiles have been installed *(which are responsible for `x11` packages)*.

archinstall/lib/locale_helpers.py
```diff
@@ -26,12 +22,27 @@ def verify_keyboard_layout(layout):
...
 	return False
 
+def verify_x11_keyboard_layout(layout):
+	for language in list_x11_keyboard_languages():
+		if layout.lower() == language.lower():
+			return True
+	return False
 
-def search_keyboard_layout(layout_filter):
+def search_keyboard_layout(layout):
```
dylanmtaylor, line 31: would prefer to not undo this change, since linter doesn't like shadowing filter.

.gitignore
```diff
@@ -20,7 +20,7 @@ SAFETY_LOCK
 **/**.network
 **/**.target
 **/**.qcow2
-**/test.py
+**/test*.py
```
dylanmtaylor, line 23: This might bite us if we add pytest tests and happen to use this naming convention.

Torxed: Good catch!

archinstall/install.log
```diff
@@ -0,0 +1,54 @@
+Not enough permission to place log file at /var/log/archinstall/install.log, creating it in /home/anton/github/archinstall/archinstall/install.log instead.
```
dylanmtaylor, line 1: Drop this file from the PR and add an ignore for it

profiles/xorg.py
```diff
@@ -44,4 +44,4 @@ def _prep_function(*args, **kwargs):
 		else:
 			archinstall.storage['installation_session'].add_additional_packages(f"xorg-server xorg-xinit {' '.join(_gfx_driver_packages)}")
 	except:
-		archinstall.storage['installation_session'].add_additional_packages("xorg-server xorg-xinit")  # Prep didn't run, so there's no driver to install
+		archinstall.storage['installation_session'].add_additional_packages(f"xorg-server xorg-xinit")  # Prep didn't run, so there's no driver to install
```
dylanmtaylor, line 47: This should not be an f-string.

archinstall/lib/installer.py
```diff
@@ -515,11 +517,38 @@ def user_set_shell(self, user, shell):
...
+			# Setting an empty keymap first, allows the subsequent call to set layout for both console and x11.
+			self.arch_chroot('localectl set-keymap ""')
+
+			if (output := self.arch_chroot(f'localectl set-keymap {language}')).exit_code != 0:
+				raise ServiceException(f"Unable to set locale '{language}' for console: {output}")
+
+			self.log(f"Keyboard language for this installation is now set to: {language}")
 		else:
-			self.log('Keyboard language was not changed from default (no language specified).', fg="yellow", level=logging.INFO)
+			self.log(f'Keyboard language was not changed from default (no language specified).', fg="yellow", level=logging.INFO)
```
dylanmtaylor, line 535: This also should not be an f-string

dylanmtaylor:
https://github.com/archlinux/archinstall/pull/476 corrects the issues I just pointed out.

archinstall/install.log
Torxed: Thought we had it already, will do :)

Torxed:
Merged those changes. Carrying on with `archinstall.Boot()`

Torxed:
`localectl` now properly sets the `vconsole` configuration as well as `x11` configuration.
So I'm happily merging this in now! There might be some commands that still freak out over the new `SysCommand`, but we'll deal with it then.

screenshot

screenshot

Issue #153, opened 2021-04-03
Add support for specifying mount options


ilyazzz:
There should be an option to specify mount options along with the mountpoint and filesystem type when using existing partitions. This will give the option to have more non-standard filesystem setups, such as ones with btrfs subvolumes (#93) or transparent compression.

Torxed:
Would #124 potentially solve this? If not I might have misunderstood the request and might need to ask some additional followup questions.

ilyazzz:
Yes, it would. My suggestion is to have something like this when the user chooses to use an existing partition: 
```
Select a partition by number that you want to set a mount-point for (leave blank when done): 0
Enter a mount-point for Partition(path=/dev/sda2, fs=btrfs): /
Enter a valid filesystem for Parition(path=/dev/sda2, fs=btrfs) (leave blank for btrfs):
Enter additional mount options(path=/dev/sda2, fs=btrfs) (leave blank to use the defaults): compress=zstd:1
```
The `compress=zstd:1` would then get passed into the `-o` option in `mount`.

Allowing to use custom /mnt would allow for a setup such as this and more, IMO at the cost of being slightly less user-friendly.

Torxed:
> Yes, it would. My suggestion is to have something like this when the user chooses to use an existing partition:
> 
> ```
> Select a partition by number that you want to set a mount-point for (leave blank when done): 0
> Enter a mount-point for Partition(path=/dev/sda2, fs=btrfs): /
> Enter a valid filesystem for Parition(path=/dev/sda2, fs=btrfs) (leave blank for btrfs):
> Enter additional mount options(path=/dev/sda2, fs=btrfs) (leave blank to use the defaults): compress=zstd:1
> ```
> 
> The `compress=zstd:1` would then get passed into the `-o` option in `mount`.
> 
> Allowing to use custom /mnt would allow for a setup such as this and more, IMO at the cost of being slightly less user-friendly.

It's not a bad idea. And as you said this is a balance between use friendliness and improving the overall experience for those that want this. I think this is something experienced users can tweak in the installed system and doesn't necessarily have to be in the guided installer. But perhaps in a `advance` installer or something similar. Or through simply the installer relying on what's given to it in `/mnt` as mentioned in #124. But if we want #93 then perhaps this is something we'll have to bake in anyway.

I'll think about it and get back to it once the initial bug reports settle from the "launch" :) Appreciate the feedback!

jatin-cbs:
[...]

PR #450, opened 2021-05-15
propagate options when mounting a partition

charlesvdv:
Hi! 

I noticed options are never used when mounting a partition so this PR fixes this. This is a bit related to #153.


dylanmtaylor:
Thanks for the contribution, this looks good.

dylanmtaylor (approved):

PR #452, opened 2021-05-15
Switch LXQt to use LightDM to be consistent with other DEs

dylanmtaylor:


advaithm:
Lightdm is not supported by upstream LxQt lets roll this back https://github.com/lxqt/lxqt/issues/795

PR #456, opened 2021-05-16
Add mirror reachability check

### LINKED ISSUE #413: Check internet connection before formatting disk
At the moment, the installation errors out after the disk has been formatted in case there is no internet connection.

Would also be nice to be able to configure wifi properly, but that probably has to wait a bit longer.

dylanmtaylor:
This pull request attempts to solve checking for mirror access before prompting for user questions.

This should be sufficient to close #413

### DIFF
--- a/archinstall/lib/networking.py
+++ b/archinstall/lib/networking.py
@@ -26,6 +26,14 @@ def list_interfaces(skip_loopback=True):
 	return interfaces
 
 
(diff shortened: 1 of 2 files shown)


dylanmtaylor:
When network adapter disabled:

VirtualBox_Arch_15_05_2021_22_47_31

PR #462, opened 2021-05-17
Attempt to solve #372

### LINKED ISSUE #372: Can't run archinstall right after running it
If the previous run of archinstall mounted /mnt and /mnt/boot, archinstall will fail to run the second time.

Workaround: umount /mnt/boot /mnt (or just reboot)

dylanmtaylor:
If we are not using existing /mnt, we should unmount /mnt prior to formatting. I think this will close #372.

### DIFF
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -3,6 +3,7 @@ import logging
 import time
 
 import archinstall
+from archinstall.lib.general import SysCommand
 from archinstall.lib.hardware import has_uefi
 from archinstall.lib.networking import check_mirror_reachable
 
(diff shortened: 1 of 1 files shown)


Torxed:
I would say that the `.format()` step should take care of this instead.
Possibly with a `unmount_first=True` flag to that function. We could also utilize the [Partition.unmount()](https://github.com/archlinux/archinstall/blob/master/archinstall/lib/disk.py#L401-L417) function *(which also has a double definition and should be corrected)*

Torxed:
> > I would say that the `.format()` step should take care of this instead.
> > Possibly with a `unmount_first=True` flag to that function. We could also utilize the [Partition.unmount()](https://github.com/archlinux/archinstall/blob/master/archinstall/lib/disk.py#L401-L417) function _(which also has a double definition and should be corrected)_
> 
> Can partition.unmount() recursively unmount the /mnt directory?

They're sadly on partition level, so it's unaware of where it's mounted.
You would have to iterate through the partitions, which we do anyway when we format.
So it would make sense to call `.unmount()` before we call `.format()`:)

PR #463, opened 2021-05-18
Proposal: Break up ask_user_questions, particularly regarding partition logic

dylanmtaylor:
This method is very long and complex.


Torxed:
I think this change will be redundant since #426 will more or less completely re-work the entire partitioning logic and it's steps.

dylanmtaylor:
> I think this change will be redundant since #426 will more or less completely re-work the entire partitioning logic and it's steps.
> So if you don't object I think we can close this with the new changes coming soon.

Sure

PR #482, opened 2021-05-19
Added support for getting configuration from a URL

### LINKED ISSUE #480: Add ability to load a config file from a URL and use that for installation
This would compliment the feature to use a configuration file

dylanmtaylor:
This will close #480. Adds support for retrieving a configuration json file from a URL as well as a filename.

### DIFF
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -1,5 +1,8 @@
 """Arch Linux installer - guided, templates etc."""
-from argparse import ArgumentParser, FileType
+import urllib.error
+import urllib.parse
+import urllib.request
+from argparse import ArgumentParser
 
 from .lib.disk import *
 from .lib.exceptions import *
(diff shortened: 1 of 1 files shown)


dylanmtaylor:
Work in progress, but to test, you can use this:
```
python -m archinstall --config https://files.dylanmtaylor.com/config-sample.json --silent
```

l4zy0n3:
Can we try to directly parse the response as json using the requests library, that way we won't have to worry about creating temp files

archinstall/__init__.py
```diff
@@ -41,7 +45,15 @@ def initialize_arguments():
 	args = parser.parse_args()
 	if args.config is not None:
 		try:
-			config = json.load(args.config)
+			# First, let's check if this is a URL scheme instead of a filename
+			parsed_url = urllib.parse.urlparse(args.config)
+
+			if not parsed_url.scheme:  # The Profile was not a direct match on a remote URL, it must be a local file.
+				config = json.load(args.config)
```
l4zy0n3, line 52: I think we might need to create the file object here if we're not using FileType, as FileType automatically parses the argument as a file and creates a file object for us

dylanmtaylor:
> Can we try to directly parse the response as json using the requests library, that way we won't have to worry about creating temp files

Fixed in latest commit.

archinstall/__init__.py
dylanmtaylor: Yes, I think it's correct now.

dylanmtaylor:
Seems to be working now, tested with both file and URL.

PR #485, opened 2021-05-19
Implement post-installation commands

### LINKED ISSUE #479: Add ability to run commands in chroot automatically
Just adding this to my wishlist, might implement eventually.
It would be nice if we added the ability to add a command list to the declarative installation instructions which would be run against the installation via `arch-chroot` or `systemd-nspawn` like so:
```json
    "custom-commands": [
        "cd /home/devel; git clone https://aur.archlinux.org/paru.git",
        "chown -R devel:devel /home/devel/paru",
        "usermod -aG docker devel"
    ]
```

### LINKED ISSUE #486: Setting profile name in config file is currently broken
Anything I tried except profile: null did not seem to work.

VirtualBox_Arch_19_05_2021_20_00_51

dylanmtaylor:
Work in progress still, very much in early stages, but the idea is that in a silent install, instead of running chroot, we'd iterate over the command list and execute each one. This would allow for automatically customizing an installation.

This also adds a more complicated, "realistic" sample configuration file.

This would close #479. Closes #486.

To test:

```
archinstall --config examples/custom-command-sample.json
```


dylanmtaylor:
I'm going to unmark it as a draft so this can be merged if you manage to test it successfully. I haven't yet confirmed this works.

dylanmtaylor:
After switching from systemd-nspawn to arch-chroot this is now functional. 
VirtualBox_Arch_20_05_2021_15_26_04

@Torxed ready to merge if you're fine with this; we can always optimize later.

dylanmtaylor:
Works great with the latest commit as well. Now you can see the output.

VirtualBox_Arch_20_05_2021_15_55_08

examples/guided.py
```diff
@@ -381,6 +377,16 @@ def perform_installation(mountpoint):
 						archinstall.log(' * Profile\'s post configuration requirements was not fulfilled.', fg='red')
 						exit(1)
 
+		# If the user provided custom commands to be run post-installation, execute them now.
+		if len(archinstall.arguments['custom-commands']):
+			for index, command in enumerate(archinstall.arguments['custom-commands']):
```
Torxed, line 382: Could we move this into a helper function, just to keep guided as small as possible?

Torxed (changes requested):

examples/guided.py
dylanmtaylor: Which library would you like the function to be in? It's definitely doable but this is a very small amount of code to extract into a function.

dylanmtaylor: Done.

dylanmtaylor:
Still working with the latest commit.

VirtualBox_Arch_20_05_2021_17_28_49

dylanmtaylor:
Tested and still working.

VirtualBox_Arch_20_05_2021_18_13_07

Torxed:
Great work from everyone on this one!

PR #488, opened 2021-05-20
look for default profiles in correct location

SimonPe:
The default `profiles` directory is a sibling of the `lib` directory, not of `storage.py` itself.

`archinstall` (installed trough pacman) kept looking for profiles under `/usr/lib/python3.9/site-packages/archinstall/lib/profiles/` while the profiles are actually located under `/usr/lib/python3.9/site-packages/archinstall/profiles/`

### DIFF
--- a/archinstall/lib/storage.py
+++ b/archinstall/lib/storage.py
@@ -11,7 +11,7 @@ storage = {
 	'PROFILE_PATH': [
 		'./profiles',
 		'~/.config/archinstall/profiles',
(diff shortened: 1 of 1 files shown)


dylanmtaylor:
Nice catch, +1

Torxed:
Awesome find! If memory serves me right, then this used to work before we changed the packaging and some other stuff in terms of how modules were executed (which starting point it had). Either way, awesome work and thank you for your contribution!

PR #490, opened 2021-05-20
Fix #489

### LINKED ISSUE #489: from .systemd import Boot is the function set_keyboard_language instead of the file
This means that Boot undeclared in set_x11_keyboard_language. We need to move this to the top of the file with other imports.

dylanmtaylor:
Closes #489

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -3,6 +3,7 @@ from .hardware import *
 from .locale_helpers import verify_x11_keyboard_layout
 from .mirrors import *
 from .storage import storage
+from .systemd import Boot
 from .user_interaction import *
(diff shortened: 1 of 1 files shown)


Torxed:
Problem here tho is that this causes a circular import because:
https://github.com/archlinux/archinstall/blob/ad8a07843944404d1d009a7a527d90e87b7ba9c6/archinstall/lib/installer.py#L6
https://github.com/archlinux/archinstall/blob/ad8a07843944404d1d009a7a527d90e87b7ba9c6/archinstall/lib/systemd.py#L4

dylanmtaylor:
You're probably right, we need a better way to solve this.

PR #491, opened 2021-05-20
fixed desktop install from config

l4zy0n3:
# Describe your PR

Desktop installation from config works now, tested with gnome

# Testing

Tested with live ISO on a VM, gnome works

#486

### DIFF
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -179,8 +179,6 @@ def ask_user_questions():
 	# Ask for archinstall-specific profiles (such as desktop environments etc)
 	if not archinstall.arguments.get('profile', None):
 		archinstall.arguments['profile'] = archinstall.select_profile(archinstall.list_profiles(filter_top_level_profiles=True))
-	else:
-		archinstall.arguments['profile'] = archinstall.list_profiles()[archinstall.arguments['profile']]
 
 	# Check the potentially selected profiles preparations to get early checks if some additional questions are needed.
 	if archinstall.arguments['profile'] and archinstall.arguments['profile'].has_prep_function():
@@ -410,8 +408,6 @@ else:
 	# Temporarily disabling keep_partitions if config file is loaded
 	archinstall.arguments['harddrive'].keep_partitions = False
 	# Temporary workaround to make Desktop Environments work
-	archinstall.storage['_desktop_profile'] = archinstall.arguments.get('desktop', None)
-	if archinstall.arguments.get('profile', None):
-		archinstall.arguments['profile'] = Profile(installer=None, path=archinstall.arguments['profile']['path'])
+	archinstall.arguments['profile'] = archinstall.Profile(None, archinstall.arguments.get('desktop', None))
 
 perform_installation_steps()
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -408,6 +408,9 @@ else:
 	# Temporarily disabling keep_partitions if config file is loaded
 	archinstall.arguments['harddrive'].keep_partitions = False
 	# Temporary workaround to make Desktop Environments work
-	archinstall.arguments['profile'] = archinstall.Profile(None, archinstall.arguments.get('profile', None))
+	if archinstall.arguments.get('profile', None) is not None:
+		archinstall.arguments['profile'] = archinstall.Profile(None, archinstall.arguments.get('profile', None))
+	else:
+		archinstall.arguments['profile'] = None
 
 perform_installation_steps()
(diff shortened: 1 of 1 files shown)


examples/guided.py
```diff
@@ -410,8 +408,6 @@ def perform_installation(mountpoint):
 	# Temporarily disabling keep_partitions if config file is loaded
 	archinstall.arguments['harddrive'].keep_partitions = False
 	# Temporary workaround to make Desktop Environments work
-	archinstall.storage['_desktop_profile'] = archinstall.arguments.get('desktop', None)
-	if archinstall.arguments.get('profile', None):
-		archinstall.arguments['profile'] = Profile(installer=None, path=archinstall.arguments['profile']['path'])
+	archinstall.arguments['profile'] = archinstall.Profile(None, archinstall.arguments.get('desktop', None))
```
Torxed, line 411: Wouldn't this overwrite any `--desktop` argument from line #L181?

l4zy0n3: #L181 will be unreachable if config is passed, as ask_user_questions won't be called

Torxed: I meant to write `--profile`.
Lets say you do `python -m archinstall guided --profile awesome`, on #L183 that would have resulted in the profile being sourced in `archinstall.arguments['profile']`, but now that will be overwritten by potentially something in `--desktop`.

Scenario:
```
python -m archinstall guided --profile=minimal
```
Initially, that would be:
```python
archinstall.arguments['profile'] == archinstall.Profile(None, 'minimal')
```
But this change *(I think)*  would mean that `archinstall.arguments['profile'] == None` because `archinstall.arguments.get('desktop', None)` was never set so it will inject `None` into `archinstall.arguments['profile']` on #L411?

l4zy0n3: Okay, I understand now, but as per my understanding this line would be only triggered if and only if config file is loaded, otherwise this would be unreachable. Please correct me if I'm wrong.

l4zy0n3: `profile` key will be ignored if present in config file

Torxed: Ah you're right, my bad! That piece of logic is still a bit new to me.

So final question from my end would be why we're putting `desktop` argument into `profile` argument?
This will cause this to be triggered every time: https://github.com/archlinux/archinstall/blob/bbb4599165a644bbd81b085fb3210cd0e497d503/profiles/desktop.py#L51-L52

This is problem an issue created outside of this PR, but I noticed it here.
The issue of #486 is probably more to do with this not being converted to an actual profile object: https://github.com/archlinux/archinstall/blob/c2f3b9e2088bfd5ed1c6a5f80bab28277653bb8a/examples/guided.py#L415

```python
if archinstall.arguments.get('profile', None):
	archinstall.arguments['profile'] = archinstall.list_profiles()[archinstall.arguments['profile']]
	archinstall.arguments['profile'] = Profile(installer=None, path=archinstall.arguments['profile']['path'])
```
This would solve it.
Since in `ask_user_questions()` this is handled, but in the config loading it isn't.

Torxed (changes requested):

examples/guided.py
```diff
@@ -410,8 +408,6 @@ def perform_installation(mountpoint):
 	# Temporarily disabling keep_partitions if config file is loaded
 	archinstall.arguments['harddrive'].keep_partitions = False
 	# Temporary workaround to make Desktop Environments work
-	archinstall.storage['_desktop_profile'] = archinstall.arguments.get('desktop', None)
-	if archinstall.arguments.get('profile', None):
-		archinstall.arguments['profile'] = Profile(installer=None, path=archinstall.arguments['profile']['path'])
```
l4zy0n3, line 415: Okay, so the reason to remove this logic was that the user would have to provide the actual path to the profile in the config file, which is not very user friendly, the new logic simply picks up a key `desktop` from the config file and uses it to install whatever profile is specified, but yeah this is very confusing, I'll update the PR with your suggested changes, I missed that the path can actually be just the profile name, my bad.

l4zy0n3: Actually, just renaming "desktop" to "profile" will also work, AFAIU

l4zy0n3: Just for the sake of keeping the UX simple, i'll just rename the key to "profile" for now, because asking users to put in
`profile: {
 path: ""
}
`
would be a lot messier

we could keep it as simple as
`profile: <name>`

Torxed: The user doesn't have to provide an actual path, we have path-lookup mechanisms, so any short version of profiles would work. We support the following from the user currently:
 * Partial
 * Absolute
 * Remote
 I'm guessing even when loading from a configuration, this should be the case as well :) But first implementation can be limited, that's fine :)

l4zy0n3: tested with all three

dylanmtaylor:
All of these changes are in #485 - might be easier to just close this and merge that one in.

PR #493, opened 2021-05-20
Change position of linter badge

SecondThundeR:
Not the most important and cool change, but usually any badges are placed under the main heading rather than under the project logo *(it also looks more nicer)*

image


dylanmtaylor:
Fine with me. +1

PR #494, opened 2021-05-20
Use pure Python code in filter_mirrors_by_region

JakobDev:
filter_mirrors_by_region uses shell commands to download and edit the mirrorlist. This simple task doesn't need any shell commands, so I had replaced it with Python code. I think that shell commands should only be used if there is no easy way to do that in Python.

### DIFF
--- a/archinstall/lib/mirrors.py
+++ b/archinstall/lib/mirrors.py
@@ -5,7 +5,7 @@ from .general import *
 from .output import log
 
 
-def filter_mirrors_by_region(regions, destination='/etc/pacman.d/mirrorlist', tmp_dir='/root', *args, **kwargs):
+def filter_mirrors_by_region(regions, destination='/etc/pacman.d/mirrorlist', *args, **kwargs):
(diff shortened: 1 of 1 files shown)


dylanmtaylor:
I like this. Can you please make the user-agent 'ArchInstall' to be consistent with #484?

archinstall/lib/mirrors.py
```diff
@@ -5,7 +5,7 @@
 from .output import log
 
 
-def filter_mirrors_by_region(regions, destination='/etc/pacman.d/mirrorlist', tmp_dir='/root', *args, **kwargs):
+def filter_mirrors_by_region(regions, destination='/etc/pacman.d/mirrorlist', *args, **kwargs):
 	"""
```
Torxed, line 9: Nicely done, this old legacy code really needed to change. One question tho, is the tmp_dir really used or can it be dropped? :)

Torxed (changes requested):

PR #498, opened 2021-05-21
Add ability to provide an array of services to enable in config file

### LINKED ISSUE #496: Add ability to specify services to enable in the configuration file
A thought - in addition to users giving a package list, what about a service list too? This could be used to iterate over and run `systemctl enable {service}` within a chroot environment. 

Possible use-case;

```
    "packages": ["docker", "git", "nginx", "wget", "zsh"],
    "services": ["docker", "nginx"]
```

dylanmtaylor:
This is similar to the last PR but much less complicated since the functionality exists already and is just being called.

Closes #496


dylanmtaylor:
Probably works fine, but I'm going to test this install soon, will report back with results.

dylanmtaylor:
VirtualBox_Arch_21_05_2021_09_25_22

PR #499, opened 2021-05-21
[WIP] Partitioning rework config

l4zy0n3:
Reworking fetching partition info from config file


l4zy0n3:
Formatting disks is broken as of now, installer directly jumps to installing packages

PR #505, opened 2021-05-22
Introduces a NTP option in guided

### LINKED ISSUE #297: Have NTP enabled as an install-time option
Having an accurate clock is a huge quality of life improvement. Maybe ask right after timezone and enable relevant systemd units if the user selects yes.

Torxed:
This feature might require more work in the future.
Especially to work with `systemd`'s time management.

But I think this is good enough for now and fixes #297


dylanmtaylor:
The NTP wording isn't bad. NTP should be capitalized since it's an acronym. Maybe these strings would be better?

Would you like to use automatic time synchronization (NTP) with the default time servers? [Y/n]

Hardware time and other post-configuration steps might be required in order for NTP to work. For more information, please check the Arch wiki.

Issue #466, opened 2021-05-18
TODO: Rewrite driver selection logic to not use __builtins__.


dylanmtaylor:
https://docs.python.org/3/reference/executionmodel.html#builtins-and-restricted-execution

CPython implementation detail: Users should not touch __builtins__; it is strictly an implementation detail. Users wanting to override values in the builtins namespace should import the __builtin__ (no 's') module and modify its attributes appropriately. The namespace for a module is automatically created the first time a module is imported.

This affects the following profiles files: `xorg.py` and `sway.py`

Torxed:
We moved away from `__builtin__` magic for the installation instance. So this would be a great time to remove any remaining magic calls since they will be breaking changes.

PR #513, opened 2021-05-24
Removed builtins and added gfx_package key in config file

l4zy0n3:
1. Tested with kde, sway profiles ( install from config)
2. builtins are replaced by archinstall.storage as suggested by @Torxed


dylanmtaylor:
This issue closes #466

PR #519, opened 2021-05-27
Change ordering of example profiles

dylanmtaylor:
I  think that it is better alphabetically, and gnome for better or worse is the default on most distros. gnome/kde are also what 90% of people are likely to use. I know sway or tiling window managers in general are more confusing to new users too. I don't know how I feel about them being first.


Torxed:
I think this makes sense :)

PR #520, opened 2021-05-27
Add descriptions for top-level profiles

dylanmtaylor:
This is for 2.3.0 future development. These strings will be displayed in the profile selection.


Torxed:
Awesome work! Wasn't sure the idea would get lost in the comments hehe

dylanmtaylor:
> Awesome work! Wasn't sure the idea would get lost in the comments hehe

probably safe to merge this too, not like it'll break anything.

PR #521, opened 2021-05-27
Updated custom-command-sample.json with newly added keys

l4zy0n3:


examples/custom-command-sample.json
```diff
@@ -16,14 +17,11 @@
         "linux"
     ],
     "keyboard-language": "us",
-    "mirror-region": {
-        "Worldwide": {
-            "https://mirror.rackspace.com/archlinux/$repo/os/$arch": true
-        }
-    },
+    "mirror-region": "Worldwide",
```
Torxed, line 20: Is this a valid configuration? I thought it had to be the sub-dict with `<country>:  { "url" : true, ... }` as per the output of archinstall?

l4zy0n3: I actually changed it to this for easier mirror selection in config

Torxed: Humm, and how would the installer know which URL's to set?
Maybe this works, I haven't tested tbh :P

l4zy0n3: #509 ,  I actually changed it to set the URL using 

```python
archinstall.arguments['mirror-region'] = {selected_region: archinstall.list_mirrors()["<mirror region key from config file/ arguments>"]}
```

This change was merely to make the config file easier to use and a little less cluttered, as users can add remove mirrors later, if that makes sense, this is not an important change, I have tested this, seems to work fine for me, however, if this is not inline with the design of archinstall, we could just revert it back,.

Torxed: Ah, for the config I think this is fine for now. In the future it would be nice if both config and not supported multiple regions as well as custom mirrors. I removed the custom mirror feature a while back but in 2.3 if would be nice if we could bring it back :)

l4zy0n3: #526 both dict and string are valid now, for both mirror-region and profile ^_^

PR #526, opened 2021-05-28
Write user configuration JSON to a file in logs directory

### LINKED ISSUE #525: Add feature to write out config to file in logs dir w/ credentials stripped
I was thinking about adding this feature. Probably should go here:

https://github.com/archlinux/archinstall/blob/4bb644bf5b8ce70bc08c76deeeb6fe71a3c36c1a/examples/guided.py#L261

dylanmtaylor:
Closes #525 - untested, so it's a draft.

### DIFF
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -260,7 +260,7 @@ def perform_installation_steps():
 	archinstall.log("-- Guided template chosen (with below config) --", level=logging.DEBUG)
 	user_configuration = json.dumps(archinstall.arguments, indent=4, sort_keys=True, cls=archinstall.JSON)
 	archinstall.log(user_configuration, level=logging.INFO)
-	with open("/var/log/archinstall/user_configuration.json") as config_file:
+	with open("/var/log/archinstall/user_configuration.json", "w") as config_file:
 		config_file.write(user_configuration)
 	print()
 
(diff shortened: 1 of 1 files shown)


l4zy0n3:
image
We need to find a way to accept both dict and string as valid mirror-region values, otherwise this won't work

l4zy0n3:
also, seeing a weird # character when doing a cat on the file
image

l4zy0n3:
> image
> We need to find a way to accept both dict and string as valid mirror-region values, otherwise this won't work

image
same issue with profile too

l4zy0n3:
> Closes #525 - untested, so it's a draft.

Tested with both dict and string in chroot, generated config successfully installs. ^_^

Torxed:
> image
> We need to find a way to accept both dict and string as valid mirror-region values, otherwise this won't work

If we can't get it working, I suggest the original as it's more flexible and granular. A single string is cleaner, but function over simplicity in this case.

l4zy0n3:
> > image
> > We need to find a way to accept both dict and string as valid mirror-region values, otherwise this won't work
> 
> If we can't get it working, I suggest the original as it's more flexible and granular. A single string is cleaner, but function over simplicity in this case.

Both dict and string are working now

l4zy0n3:
> > Closes #525 - untested, so it's a draft.
> 
> Tested with both dict and string in chroot, generated config successfully installs. ^_^

Tested

Torxed:
Then I'll bring this into the tag!

PR #529, opened 2021-05-28
Make this template usable to users of more platforms

dylanmtaylor:
In case someone is trying to edit this as a template, this will work for more people.

### DIFF
--- a/examples/custom-command-sample.json
+++ b/examples/custom-command-sample.json
@@ -8,7 +8,7 @@
     ],
     "!encryption-password": "supersecret",
     "filesystem": "btrfs",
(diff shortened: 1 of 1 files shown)


Torxed:
Fair change tbh. I think if possible, that we should omit the harddrive as well so people don't accidentally wipe `/dev/sda`, and that `config` should ask for the drive. I think this could be done if `config` only populates `arguments` and nothing else. That way normal "skipping questions" logic would apply, but those missing from config would be asked.

l4zy0n3:
We could merge all config changes into ask_user_questions() so that if any key is not provided and silent is disabled prompts could be shown to select values

PR #531, opened 2021-05-30
Fix typo in error msg

Kibouo:
The failure error message for formatting a disk with MBR mentioned GPT instead.


Torxed:
Nice catch! :) Thanks for submitting a PR to fix it!

PR #534, opened 2021-05-31
Fixing hardcoded kernel in boot-config

### LINKED ISSUE #532: Default linux kernel is hardcoded in systemd-boot entry
## Summary

Pull request #315 added the ability to select a different kernel other than the default, but it seems to result in an unbootable system, since the code that generates the systemd-boot entry is hardcoded to load the default kernel:

https://github.com/archlinux/archinstall/blob/7b1096bfa6a5d042b38341b68f857dbe4f150bb3/archinstall/lib/installer.py#L471

## How to reproduce

I wrote a JSON config for archinstall with the following key in it:

    "kernels": ["linux-zen"],

After installing and rebooting, systemd-boot showed no entries available to boot. I dropped to an UEFI shell and found that the generated entry was trying to load the default linux kernel instead of linux-zen:

    # Created by: archinstall
    # Created on: 2021-05-30_21-00-01
    title Arch Linux
    linux /vmlinuz-linux
    initrd /initramfs-linux.img
    options root=PARTUUID=e6393745-d916-4176-a7c3-315b3457092c rw intel_pstate=no_hwp

The linux-zen files were present in the EFI system partition, and manually editing the entry to add the -zen suffix made the system bootable.

## Installation log

[install.log](https://github.com/archlinux/archinstall/files/6566917/install.log)

Torxed:
This should fix issue with hardcoded kernel and this will fix #532

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -40,6 +40,7 @@ class Installer:
 			base_packages = __packages__[:3]
 		if kernels is None:
 			kernels = ['linux']
+		self.kernels = kernels
 		self.target = target
 		self.init_time = time.strftime('%Y-%m-%d_%H-%M-%S')
 		self.milliseconds = int(str(time.time()).split('.')[1])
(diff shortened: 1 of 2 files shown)


Torxed:
screenshot

Verified that at least two of the kernels boot with this fix.

PR #537, opened 2021-06-01
Introduces the use of HTTP mirrors additionally

### LINKED ISSUE #517: Allow non-HTTPS mirrors
Hello there, as discussed on the matrix chat, I'm opening this issue to gather comments and feedback from the wider community, as well as any possible security implications this change might have.

**Description**
Archinstall currently only allows HTTPS mirrors see here[[1](https://github.com/archlinux/archinstall/blob/master/archinstall/lib/mirrors.py#L82)], which works for most users, but not all countries/regions have HTTPS mirrors, for example Brazil doesn't have any HTTPS mirror, but plenty of very fast HTTP ones.

**Reasons to consider allowing non-HTTPS:**

* HTTP is more widely available.
* HTTP allows for transparent caching at the ISP or LAN, allowing for even smaller network footprint. And while setting up a local mirror isn't hard, a dumb transparent cache is much more generic and useful in more situations.

**Reasons to NOT allow non-HTTPS mirrors**

* HTTPS can act as another layer of security, for example see this[[2](https://lists.debian.org/debian-security-announce/2019/msg00010.html)] Debian/apt bug.
* Using HTTP leaks which packages you are installing, which can be seen as a privacy problem.
* Bootstrapping an outdated pacman-keyring via HTTP is unsafe, although this should (AFAIK) only happen if your .ISO is severely outdated, and is not automatic.

**My opinion on this situation**

While I'm not an Archlinux dev or pacman expert, I believe all packages (`*.tar.zstd`) are signed (`*.sig`), and the public key of these signatures is stored in pacman's keyring (located at `/etc/pacman.d/gnupg`), so even if some Evil Attacker tampers with anything you download, it should be fine, because pacman should check the signature against the locally trusted keyring.

Torxed:
This fixes #517

### DIFF
--- a/archinstall/lib/general.py
+++ b/archinstall/lib/general.py
@@ -2,14 +2,41 @@ import hashlib
 import json
 import logging
 import os
-import pty
 import shlex
 import subprocess
 import sys
 import time
 from datetime import datetime, date
-from select import epoll, EPOLLIN, EPOLLHUP
 from typing import Union
+try:
+	from select import epoll, EPOLLIN, EPOLLHUP
+except:
+	import select
+	EPOLLIN = 0
+	EPOLLHUP = 0
+	class epoll():
+		""" #!if windows
+		Create a epoll() implementation that simulates the epoll() behavior.
+		This so that the rest of the code doesn't need to worry weither we're using select() or epoll().
+		"""
+		def __init__(self):
+			self.sockets = {}
+			self.monitoring = {}
+
+		def unregister(self, fileno, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


dylanmtaylor:
Is there a way we can make sure that HTTPS is preferred?

Torxed:
> Is there a way we can make sure that HTTPS is preferred?

Humm, good question.
I could make archinstall sort the resulting mirrors based on a harcoded order like `["http", "https"]` and if no match I'll just put it at the end.

PR #538, opened 2021-06-01
Attempt to fix issue regarding broken output

### LINKED ISSUE #536: Output looks broken with pacman 6
Since `pacman` v6 was released, it introduced paralell downloads among other things (color enhancement and other good stuff).
Due to how we show "progressbars" in archinstall, the output looks broken and scrambled.

**The installs work fine!**

It's jut that the outlook looks garbage because: https://github.com/archlinux/archinstall/blob/615d0d38dd577a054151abcebf32b3a30b98e3fd/archinstall/lib/general.py#L212-L223

Torxed:
Since pacman and some other commands these days write multiple lines and goes back and fourth, it's not reliable to say the "peaked" line is a single line. And if it's not, it will make the output look like garbage.

So instead, we'll write any output - as is - and let the command deal with fancy printing.

This fixes #536.

### DIFF
--- a/archinstall/lib/general.py
+++ b/archinstall/lib/general.py
@@ -202,27 +202,6 @@ class SysCommandWorker:
 				except UnicodeDecodeError:
 					return False
 
-			output = output.strip('\r\n ')
-			if len(output) <= 0:
-				return False
-
-			from .user_interaction import get_terminal_width
-
-			# Move back to the beginning of the terminal
-			sys.stdout.flush()
-			sys.stdout.write("\033[%dG" % 0)
-			sys.stdout.flush()
-
-			# Clear the line
-			sys.stdout.write(" " * get_terminal_width())
-			sys.stdout.flush()
-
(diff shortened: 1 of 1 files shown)


dylanmtaylor:
Has this been tested? Might be safe to merge this in

Torxed:
> Has this been tested? Might be safe to merge this in

Not yet sadly. On it tomorrow.

dylanmtaylor:
Just tested this change and it looks considerably better :) @Torxed this is ready to merge in IMO
VirtualBox_Arch_08_06_2021_21_53_48
VirtualBox_Arch_08_06_2021_21_55_04

dylanmtaylor (approved):

phisch:
Is it possible to add this for chroot commands? I have a plugin that uses chroot commands which run for quite a while, and only getting the output at the end is the same issue as why peaked was introduced. Being able to see what's happening would be super nice!

Issue #541, opened 2021-06-02
No such file or directory: '/mnt/boot/loader/loader.conf'


HackBug:
Hey all, I'm trying out the arch installer for the first time today to reformat and refresh a system I've been using for a while. Process went smoothly until it began the installation, and then gave the error in the title immediately after installing efivar during the post-transaction hooks. I'm including my log file.

I'm capable of installing on my own so this isn't an urgent issue, but I'm not really sure what went wrong. It might have something to do with the fact I'm running the image off of ventoy, but I'm not 100% sure why that would be the case. I might be able to test without ventoy later.

[install.log](https://github.com/archlinux/archinstall/files/6587566/install.log)

Torxed:
This is a *weird* issue that's been popping up today. Thank you for reporting! Any information would help here to be honest. And the install lo should help!

HackBug:
So on a second try it worked again after two changes, not sure if either are relevant, but here's what I did (after a reboot as well)

1) Deleted all partitions and set the drive to gpt (didn't create any partitions myself)
2) Did not include any extra packages

Everything else was identical as far as I can tell.

HackBug:
I take it back: I apparently (accidentally) booted without UEFI into the ventoy image. The installation itself seems pretty unwilling to boot, so I'm going to try installing one more time in UEFI mode.

dylanmtaylor:
> I take it back: I apparently (accidentally) booted without UEFI into the ventoy image. The installation itself seems pretty unwilling to boot, so I'm going to try installing one more time in UEFI mode.

Does the Arch ISO work with Ventoy? Wow. I didn't realize that worked...

HackBug:
I've already reset it, but it would end on a blank black screen with blinking cursor. I'm pretty confident this was an issue with my bios, though. This particularly computer has been finicky with UEFI in the past.

HackBug:
So I've had a successful reinstall now, with a nearly identical setup. This time I used --debug, and here's the log. Hope it helps.
[install.log](https://github.com/archlinux/archinstall/files/6587818/install.log)

dylanmtaylor:
Sadly, that install just shows that it worked correctly. The installation log from when it did not work would have been more helpful, but unfortunately, it's not possible to save that off after the fact.

HackBug:
Sorry it wasn't broken again, I guess 🤣.

Torxed:
```
[...]

PR #542, opened 2021-06-02
Add some safeguards to create directories before writing files

dylanmtaylor:
Also has some minor formatting changes made by my IDE automatically. This might help with some of the errors we've seen from recent bug reports.

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -2,8 +2,8 @@ from .disk import *
 from .hardware import *
 from .locale_helpers import verify_keyboard_layout, verify_x11_keyboard_layout
 from .mirrors import *
-from .storage import storage
 from .plugins import plugins
+from .storage import storage
 from .user_interaction import *
 
 # Any package that the Installer() is responsible for (optional and the default ones)
@@ -53,7 +53,6 @@ class Installer:
 		self.base_packages = base_packages.split(' ') if type(base_packages) is str else base_packages
 		for kernel in kernels:
 			self.base_packages.append(kernel)
-
 
 		self.post_base_install = []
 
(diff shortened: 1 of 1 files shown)


archinstall/lib/installer.py
```diff
@@ -1,9 +1,9 @@
 from .disk import *
 from .hardware import *
-from .locale_helpers import verify_keyboard_layout, verify_x11_keyboard_layout
+from .locale_helpers import verify_x11_keyboard_layout
```
Torxed, line 3: Can we please stop removing imports that are being used?

archinstall/lib/installer.py
```diff
@@ -442,6 +441,10 @@ def add_bootloader(self, bootloader='systemd-bootctl'):
 				# Fallback, try creating the boot loader without touching the EFI variables
 				SysCommand(f'/usr/bin/arch-chroot {self.target} bootctl --no-variables --path=/boot install')
 
+			# Ensure that the /boot/loader directory exists before we try to create files in it
+			if not os.path.exists(f'{self.target}/boot/loader'):
+				os.makedirs(f'{self.target}/boot/loader')
+
```
Torxed, line 447: `bootctl` creates these directories. If it doesn't we should report it upstream and keep this is a patch with a `# TODO: Remove once <upstream> is fixed`.

Torxed (changes requested):

archinstall/lib/installer.py
dylanmtaylor: There's clearly something wrong with PyCharm if it's removing used imports when I tell it to optimize.

Trying to figure this out still

archinstall/lib/installer.py
dylanmtaylor: See #541 - in the case of these two installs by Steller in our Discord these dirs did NOT seem to be present.

image
image

dylanmtaylor: This was using an existing /mnt by the way.

Torxed: I'm worried that `/boot` might have been incorrectly mounted, and `/` mounting over the location in some way.
Or out-of-disk size wize on `/boot` to create folders (more unlikely). I would love to get more debugging in to really know why this happens rather than trying to fix it hot-patch style.

But that said, introducing this *shouldn't* hurt, and we can remove it later when we have all the debugging in place to catch these things.

PR #548, opened 2021-06-03
Try to solve issue with mklabel while partitions are mounted.

### LINKED ISSUE #547: Archinstall does not properly unmount all partitions from a drive before running parted -s [drive] mklabel
VirtualBox_Arch_03_06_2021_18_54_31

dylanmtaylor:
Closes #547


dylanmtaylor:
Testing this ISO now.

PR #551, opened 2021-06-05
adds python-sphinx in make dependency and manual

m1ten:
# Describe your PR

Added a make dependency (`python-sphinx`) in PKGBUILD and `man archinstall` support.

# Testing

```sh
$ makepkg -i
$ archinstall
$ man archinstall
```


m1ten:
also switch to using flit instead of setup.py

Torxed (approved):
Looks good to me, good idea to bring in the man pages!

PR #554, opened 2021-06-07
Create pkg_build.yaml

m1ten:
# Describe your PR

The purpose of `pkg_build.yaml` is to build a package (and not the whole ISO) using Github Actions, for every commit and pull request.

# Testing

Not needed for archinstall. I will test the packaging building on Github Actions.

### DIFF
--- a/.github/workflows/pkg_build.yaml
+++ b/.github/workflows/pkg_build.yaml
@@ -1,0 +1,26 @@
+# This workflow will build an ArchInstall Package with the commit on it
+
+name: Build ArchInstall Package with ArchInstall Commit
+
+on:
+  push:
+    branches:
+      - master
+      - main # In case we adopt this convention in the future
+  release:
+    types:
+      - created
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container:
+      image: archlinux:latest
+      options: --privileged
+    steps:
+      - uses: actions/checkout@v2
+      - run: makepkg
+      - uses: actions/upload-artifact@v2
+        with:
+          name: ArchInstall Package
+          path: ./*.tar.gz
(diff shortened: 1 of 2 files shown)


.github/workflows/pkg_build.yaml
```diff
@@ -0,0 +1,26 @@
...
+  release:
+    types:
+      - created
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container:
+      image: archlinux:latest
+      options: --privileged
```
Torxed, line 19: Is this required, I've seen it being misused quite a lot. Best practices say to not use it, but if there's a reason for having it I don't really object. As long as we're mindful of the usage.

dylanmtaylor: What this is doing: Launching a VM running the latest LTS version of Ubuntu and within this VM, it's launching an Arch Linux container that has privileged access to the host VM (this is a Docker flag). It's needed for ISO build and because this build is isolated it doesn't hurt to have this.

m1ten: Is there a better way to do this? I just copied it from ISO Build. 

The purpose of `pkg_build.yaml` is to build a package every commit for people who nee the package to test and not the whole ISO.

Torxed: Ah, so the privileged flag is ultimately running inside a container so there's no risk for anyone running this.

dylanmtaylor: Since this is running a known script - building the ISO inside a container it's fine. The container is ephemeral anyway. Privileged is a risk if you allow arbitrary code execution and somebody can use it to escape the container to the host by doing something like mounting the host's filesystem. In the case of GitHub Actions it's not a risk to us.

dylanmtaylor:
Privileged flag is not running it as root. It doesn't need sudo. It's already root within the container.

m1ten:
I don't think this is needed.

Issue #556, opened 2021-06-07
Selected gfx_driver and profile is missing in configuration


Flame-0:
**Description**
`gfx_driver` and `profile` is stored in `archinstall.storage`, but it is not showing in configuration.

**Example Final Result**
```json
{
    "!root-password": "******",
    "audio": "pulseaudio",
    "bootloader": "grub-install",
    "filesystem": "btrfs",
    "harddrive": {
        "path": "/dev/sdb"
    },
    "hostname": "arch",
    "kernels": [
        "linux"
    ],
    "keyboard-language": "us",
    "mirror-region": {
        "Worldwide": {
            "https://mirror.rackspace.com/archlinux/$repo/os/$arch": true
        }
    },
    "nic": "Copy ISO network configuration to installation",
    "ntp": true,
    "packages": [],
    "profile": {
        "path": "/usr/lib/python3.9/site-packages/archinstall/profiles/desktop.py"
    },
    "script": "guided",
    "superusers": {
        "user0": {
            "!password": "******"
        }
    },
    "sys-encoding": "utf-8",
    "sys-language": "en_US",
    "timezone": "US/Eastern",
    "users": {}
}
```

**Example Expected Result** (`"gfx_driver": "AMD / ATI (open-source)"` and `"profile": "sway"` added)
```json
{
    "!root-password": "******",
    "audio": "pulseaudio",
    "bootloader": "grub-install",
    "filesystem": "btrfs",
    "gfx_driver": "AMD / ATI (open-source)",
    "harddrive": {
        "path": "/dev/sdb"
    },
    "hostname": "arch",
    "kernels": [
        "linux"
    ],
    "keyboard-language": "us",
    "mirror-region": {
        "Worldwide": {
            "https://mirror.rackspace.com/archlinux/$repo/os/$arch": true
        }
    },
    "nic": "Copy ISO network configuration to installation",
    "ntp": true,
    "packages": [],
    "profile": {
        "profile": "sway",
        "path": "/usr/lib/python3.9/site-packages/archinstall/profiles/desktop.py"
    },
    "script": "guided",
    "superusers": {
        "user0": {
            "!password": "******"
        }
    },
    "sys-encoding": "utf-8",
    "sys-language": "en_US",
    "timezone": "US/Eastern",
    "users": {}
}
```

l4zy0n3:
`profile` is set to `desktop` if installing without config, as ask_user_questions is run, possible solution is to change the logic to set the resolved path to the actual desktop environment profile

PR #557, opened 2021-06-07
Setting proper keys in exported config

l4zy0n3:
Addresses #556

### DIFF
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -264,7 +264,10 @@ def perform_installation_steps():
 	with open("/var/log/archinstall/user_configuration.json", "w") as config_file:
 		config_file.write(user_configuration)
 	print()
-
+	
+	if archinstall.arguments.get('dry_run'):
+		exit(0)
+	
 	if not archinstall.arguments.get('silent'):
 		input('Press Enter to continue.')
 
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -7,7 +7,7 @@ import archinstall
 from archinstall.lib.general import run_custom_user_commands
 from archinstall.lib.hardware import *
 from archinstall.lib.networking import check_mirror_reachable
-from archinstall.lib.profiles import Profile
+from archinstall.lib.profiles import Profile, is_desktop_profile
 
 if archinstall.arguments.get('help'):
 	print("See `man archinstall` for help.")
@@ -212,8 +212,9 @@ def ask_user_questions():
 
 	# Ask about audio server selection if one is not already set
 	if not archinstall.arguments.get('audio', None):
-		archinstall.arguments['audio'] = archinstall.ask_for_audio_selection()
-	
+		# The argument to ask_for_audio_selection lets the library know if it's a desktop profile
+		archinstall.arguments['audio'] = archinstall.ask_for_audio_selection(is_desktop_profile(archinstall.arguments['profile']))
+
 	# Ask for preferred kernel:
 	if not archinstall.arguments.get("kernels", None):
 		kernels = ["linux", "linux-lts", "linux-zen", "linux-hardened"]
@@ -264,10 +265,10 @@ def perform_installation_steps():
 	with open("/var/log/archinstall/user_configuration.json", "w") as config_file:
 		config_file.write(user_configuration)
 	print()
-	
+
 	if archinstall.arguments.get('dry_run'):
 		exit(0)
-	
+
 	if not archinstall.arguments.get('silent'):
 		input('Press Enter to continue.')
 
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -212,14 +212,8 @@ def ask_user_questions():
 
 	# Ask about audio server selection if one is not already set
 	if not archinstall.arguments.get('audio', None):
-		# only ask for audio server selection on a desktop profile
-		if str(archinstall.arguments['profile']) == 'Profile(desktop)':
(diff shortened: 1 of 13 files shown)


profiles/desktop.py
```diff
@@ -54,6 +54,8 @@ def _prep_function(*args, **kwargs):
 		archinstall.storage['_desktop_profile'] = desktop
 
 	profile = archinstall.Profile(None, desktop)
+	# Set the resolved profile path to the actual desktop environment
+	archinstall.arguments['profile'] = profile
```
Torxed, line 58: This is how `ask_user_questions` does it. So should work in theory.

l4zy0n3: ask_user_questions always sets the profile to desktop, so when the config is exported it shows up as
```json
"profile": {
    "path": "/usr/lib/python3.9/site-packages/archinstall/profiles/desktop.py"
}
```

l4zy0n3: > Ah, that's probably an issue in the `Profile().json()` function which overwrites `profile`. Because it should say `awesome` if awesome was chosen for instance. But what I more meant was that the code in `guided` does the same thing as the modification to `--config` is doing:
> 
>
 https://github.com/archlinux/archinstall/blob/1a254700256784c2954fe42e35019b0680236886/examples/guided.py#L204

true, this issue is only there when selecting from prompt, as select_profile returns desktop

>
>
 https://github.com/archlinux/archinstall/blob/1a254700256784c2954fe42e35019b0680236886/examples/guided.py#L202

also, we might need to change
>
>
 https://github.com/archlinux/archinstall/blob/1a254700256784c2954fe42e35019b0680236886/examples/guided.py#L216
if we set profile to a specific DE, otherwise audio install won't be prompted

Torxed: Agreed, this issue is probably a bit complex as we could fix multiple profile support and this would solve it.
But this PR addresses it for now :)

l4zy0n3: @Torxed , I'm not sure if this logic is even necessary, as users might want an audioserver on a minimal system

examples/guided.py
```diff
@@ -212,14 +212,8 @@ def ask_user_questions():
...
 	if not archinstall.arguments.get('audio', None):
-		# only ask for audio server selection on a desktop profile
-		if str(archinstall.arguments['profile']) == 'Profile(desktop)':
-			archinstall.arguments['audio'] = archinstall.ask_for_audio_selection()
-		else:
-			# packages installed by a profile may depend on audio and something may get installed anyways, not much we can do about that.
-			# we will not try to remove packages post-installation to not have audio, as that may cause multiple issues
-			archinstall.arguments['audio'] = None
-
+		archinstall.arguments['audio'] = archinstall.ask_for_audio_selection()
```
dylanmtaylor, line 215: So this might cause issues - since this was originally written for desktop, it'll prompt for audio on any install incl. minimal, and audio selection is an either/or for pulseaudio or pipewire right now. The rest of this change I'm fine with, but -1 on this unless we somehow fundamentally change audio selection since you'd have audio on a min. install this way no matter what.

dylanmtaylor: Perhaps we should ask for audio if the selected profile is a desktop profile OR the selected profile it is in the list of desktop profiles in desktop.py

dylanmtaylor: To keep that clean, we may need to implement a is_desktop_profile(profile_name) -> bool helper function or something

l4zy0n3: This is just a prompt right, the user could leave this blank, but yeah +1 for the helper function

dylanmtaylor: That's a possibility, but I think that is might be better to just rework this logic so we display it for just desktop profiles for now, since those will have audio anyways. Part of the issue with 'none' in a DE is that packages will get pulled in by deps anyways and you'll end up with audio even if you select none. But pipewire is a 'replacement' for pulseaudio. Hence, it makes sense the way it is now.

Torxed: Maybe we can have (like we do with `top_level_profile`) a meta variable that specifies what the profile supports, for instance any desktop profile (or desktop itself) could have:
```
import archinstall

__supports__ = ['audio', 'desktop', 'xorg']
```
Or something similar, and we could create the helper function you mentioned dylan, but instead of just doing `is_desktop_profile` we could return the supported elements of the profile?

dylanmtaylor: All desktop profiles support audio, so I'm not sure if this makes sense

dylanmtaylor: Actually, audio can work without a DE at all if you're adventurous

Torxed: Absolutely, which is why profiles should define that they support `audio`, which could enable the question "Which audio-server do you want?". I just saw it as a way to trigger the question by using the meta-definitions of the profiles : )

dylanmtaylor: All profiles support audio. There's no value in specifying this in my opinion. Even on a server install you might want it if you're trying to use it for audio streaming or something like that.

dylanmtaylor: I thought about this a bit more and here's what I came up with:

Write the is_desktop_profile function
If it's a desktop, do a generic select of the set pipewire, pulseaudio. Otherwise, do a generic select of the set pipewire, pulseaudio, none. If it's already set via config, suppress the prompt.

dylanmtaylor:
Once #575 is merged, just merge the master back into this pull request and I think we'll be good to go

dylanmtaylor (approved):

PR #558, opened 2021-06-07
Add Cutefish profile 

JakobDev:
Installs the Cutefish DE

### DIFF
--- a/profiles/cutefish.py
+++ b/profiles/cutefish.py
@@ -16,6 +16,7 @@ __packages__ = [
 	"cutefish-statusbar",
 	"cutefish-wallpapers",
 	"cutefish-calculator",
+	"noto-fonts",
 	"konsole",
 	"sddm"
 ]
--- a/profiles/cutefish.py
+++ b/profiles/cutefish.py
@@ -1,0 +1,51 @@
+# A desktop environment using "Cutefish"
+
+import archinstall
+
+is_top_level_profile = False
+
+__packages__ = [
+	"cutefish-core",
+	"cutefish-dock",
+	"cutefish-filemanager",
+	"cutefish-icons",
+	"cutefish-kwin-plugins",
+	"cutefish-launcher",
+	"cutefish-qt-plugins",
+	"cutefish-settings",
+	"cutefish-statusbar",
+	"cutefish-wallpapers",
+	"cutefish-calculator",
+	"konsole",
+	"sddm"
+]
+
+
+def _prep_function(*args, **kwargs):
+	"""
+	Magic function called by the importing installer
+	before continuing any further. It also avoids executing any
+	other code in this stage. So it's a safe way to ask the user
+	for more input before any other installer steps start.
+	"""
+
+	# Cutefish requires a functional xorg installation.
+	profile = archinstall.Profile(None, "xorg")
+	with profile.load_instructions(namespace="xorg.py") as imported:
+		if hasattr(imported, "_prep_function"):
+			return imported._prep_function()
+		else:
+			print("Deprecated (??): xorg profile has no _prep_function() anymore")
+
+
+# Ensures that this code only gets executed if executed
+# through importlib.util.spec_from_file_location("cutefish", "/somewhere/cutefish.py")
+# or through conventional import cutefish
+if __name__ == "cutefish":
+	# Install dependency profiles
+	archinstall.storage["installation_session"].install_profile("xorg")
+
+	# Install the Cutefish packages
+	archinstall.storage["installation_session"].add_additional_packages(__packages__)
+
+	archinstall.storage["installation_session"].enable_service("sddm")
--- a/profiles/cutefish.py
+++ b/profiles/cutefish.py
@@ -5,17 +5,7 @@ import archinstall
 is_top_level_profile = False
 
 __packages__ = [
-	"cutefish-core",
-	"cutefish-dock",
-	"cutefish-filemanager",
-	"cutefish-icons",
-	"cutefish-kwin-plugins",
-	"cutefish-launcher",
-	"cutefish-qt-plugins",
-	"cutefish-settings",
-	"cutefish-statusbar",
-	"cutefish-wallpapers",
-	"cutefish-calculator",
+	"cutefish",
 	"noto-fonts",
 	"konsole",
 	"sddm"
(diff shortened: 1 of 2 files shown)


dylanmtaylor:
-1 because it's a niche DE and we already have plenty of DE profiles. Custom Xorg exists for a reason. I also can't find the packages by doing a Google search for Cutefish so that makes it a very strong -1 from me.

dylanmtaylor:
It took me a very long time to even find info on this DE. https://cutefishos.com/
I still feel this shouldn't go in. @Torxed what do you think?

dylanmtaylor:
I suppose it'd be okay if we add this file but do not put it in the profiles/desktop.py choice so it can be used for custom scripted installs as a compromise.

JakobDev:
It's a very new De but it's already supported by Arch with it's own [group](https://archlinux.org/groups/x86_64/cutefish/). I think archinstall, as the official installer of Arch, should support every DE that is supported by Arch. The profile is just a package list and a command to enable sddm, so it's nothing that requires a lot of work to maintain. Manjaro has also already a [Community Edition](https://twitter.com/ManjaroLinux/status/1397514171198103553) with Cutefish.

dylanmtaylor:
> I think archinstall, as the official installer of Arch, should support every DE that is supported by Arch

This isn't realistically feasible as we have limited resources. Also, things that are more flexible like openbox which require decent amounts of configuration to get set up in a way that is usable by users are best served by a fully custom installation based on the Xorg profiles. Having lots of options makes regression testing more difficult.

dylanmtaylor:
It's also apparently a fork of Cyber with very few changes, from what I'm seeing. https://getcyberos.org/#/

RyzenDew:
Add both cyber and cutefish it be good

dylanmtaylor:
Does this require SDDM or will lightdm suffice? Historically, the way we have written the profiles is that we do a lightdm installation unless the developers of the DE recommend something else. Specifically, sway doesn't officially support a greeter so we don't use one, KDE and LXQt both only support SDM officially, and GNOME wants GDM. I'm going to test an installation of this DE from the generated ISO now to validate it at least works.

dylanmtaylor:
I can validate that this does boot. 

VirtualBox_Arch_08_06_2021_21_38_32

profiles/cutefish.py
```diff
@@ -0,0 +1,42 @@
...
+
+import archinstall
+
+is_top_level_profile = False
+
+__packages__ = [
+	"cutefish",
+	"noto-fonts",
+	"konsole",
+	"sddm"
```
dylanmtaylor, line 11: Let's switch sddm to lightdm unless there's a good reason not to.

profiles/cutefish.py
```diff
@@ -0,0 +1,42 @@
...
+# through importlib.util.spec_from_file_location("cutefish", "/somewhere/cutefish.py")
+# or through conventional import cutefish
+if __name__ == "cutefish":
+	# Install dependency profiles
+	archinstall.storage["installation_session"].install_profile("xorg")
+
+	# Install the Cutefish packages
+	archinstall.storage["installation_session"].add_additional_packages(__packages__)
+
+	archinstall.storage["installation_session"].enable_service("sddm")
```
dylanmtaylor, line 42: Switch here too

RyzenDew:
yes it works best with sddm

Torxed:
> It's a very new De but it's already supported by Arch with it's own [group](https://archlinux.org/groups/x86_64/cutefish/). I think archinstall, as the official installer of Arch, should support every DE that is supported by Arch. The profile is just a package list and a command to enable sddm, so it's nothing that requires a lot of work to maintain. Manjaro has also already a [Community Edition](https://twitter.com/ManjaroLinux/status/1397514171198103553) with Cutefish.

I agree with @dylanmtaylor on this. Just because Arch Linux supports things, doesn't automatically mean we can.
This DE is in a very early stage and it adds additional testing before a release of archinstall can happen *(since we aim to test every combination and feature we have)*.

DE's is especially tricky because they have a lot of hidden dependencies like:
 * greeters
 * graphics drivers
 * Locale's
 * hardware requirements

All of which has to be tested for possible combinations and on a wire range of hardware.
I don't mind merging in this PR allowing people to script their own installations and using this profile, but for test and maintenance reasons I do not wish to keep this listed by `desktop.py` as a general profile.
I don't mind circling back once this DE has reached a more mature stage in terms of development and testing or the user base gets so big it outshadow's some of the more common ones.

Remove this line for now and I can merge it: https://github.com/archlinux/archinstall/blob/3059fa6953481af64b626ea761446682f1c374b6/profiles/desktop.py#L46

dylanmtaylor:
> yes it works best with sddm

Is this documented somewhere?

dylanmtaylor:
> > It's a very new De but it's already supported by Arch with it's own [group](https://archlinux.org/groups/x86_64/cutefish/). I think archinstall, as the official installer of Arch, should support every DE that is supported by Arch. The profile is just a package list and a command to enable sddm, so it's nothing that requires a lot of work to maintain. Manjaro has also already a [Community Edition](https://twitter.com/ManjaroLinux/status/1397514171198103553) with Cutefish.
> 
> I agree with @dylanmtaylor on this. Just because Arch Linux supports things, doesn't automatically mean we can.
> This DE is in a very early stage and it adds additional testing before a release of archinstall can happen _(since we aim to test every combination and feature we have)_.
> 
> DE's is especially tricky because they have a lot of hidden dependencies like:
> 
> * greeters
> * graphics drivers
> * Locale's
> * hardware requirements
> 
> All of which has to be tested for possible combinations and on a wire range of hardware.
> I don't mind merging in this PR allowing people to script their own installations and using this profile, but for test and maintenance reasons I do not wish to keep this listed by `desktop.py` as a general profile.
> I don't mind circling back once this DE has reached a more mature stage in terms of development and testing or the user base gets so big it outshadow's some of the more common ones.
> 
> Remove this line for now and I can merge it:
> 
> https://github.com/archlinux/archinstall/blob/3059fa6953481af64b626ea761446682f1c374b6/profiles/desktop.py#L46
> 
> ```python
> installation.install_profile('cutefish')
> ```
> 
> Will still work even after that line is removed.

@Torxed you can merge this as-is and I can make a commit to remove the line.

JakobDev:
Cutefish has now been removed from .desktop, but I think it would be nice to use the profile in the guided installer. Maybe a experimental section?

dylanmtaylor:
@JakobDev You can use it like this:

VirtualBox_Arch_10_06_2021_08_18_03

No "experimental" section needed :)

PR #561, opened 2021-06-10
Initial port of GitHub Actions to GitLab CI/CD

dylanmtaylor:
This is an attempt to port all of our GitHub workspaces to GitLab CI/CD. This makes a couple of assumptions, including the presence of a privileged docker runner tagged 'docker'. This also expects the CI/CD variables needed for flit to publish would be declared in the project's CI/CD variables.

Privileged Docker execution is allowed on the GitLab.com runners, as Docker-in-Docker uses this functionality. https://docs.gitlab.com/ee/user/gitlab_com/#linux-shared-runners

Note that the build_iso will fail on GitLab.com's shared runners due to memory constraints, exiting with code 137 which indicates excessive CPU/memory consumption.

The pytest stage is disabled because running pytest is a waste of compute time and unnecessarily fails due to lack of test cases, so we can skip this one until we actually have unit tests.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,83 @@
+# This file contains GitLab CI/CD configuration for the ArchInstall project.
+# It defines several jobs that get run when a new commit is made, and is comparable to the GitHub workflows.
+# All jons will be run in the official archlinux container image, so we will declare that here.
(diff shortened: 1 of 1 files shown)


dylanmtaylor:
Screenshot from 2021-06-09 22-12-45

PR #562, opened 2021-06-10
Only print requires root messages if effective user id is not root

dylanmtaylor:
This error message seems to confuse people. https://www.reddit.com/r/archlinux/comments/nvz86f/check_mirror_reachable_uses_pacman_sy_which/


dylanmtaylor:
Wait how can exit code even be 256? Isn't it between 0 and 255?? @Torxed is this some special value?

Issue #566, opened 2021-06-10
Proposal: add an --dry-run flag to only export the config


l4zy0n3:

Torxed:
I'm for it. Instead of the countdown `5... 4...` etc we could exit there and that would output the config :)

l4zy0n3:
Closed via #570

PR #570, opened 2021-06-10
added --dry-run flag

l4zy0n3:
Addresses #566

### DIFF
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -33,6 +33,8 @@ def initialize_arguments():
 	parser.add_argument("--creds", nargs="?", help="JSON credentials configuration file")
 	parser.add_argument("--silent", action="store_true",
 						help="Warning!!! No prompts, ignored if config is not passed")
+	parser.add_argument("--dry-run", action="store_true",
+						help="Warning!!! Only creates config, does not install")
 	parser.add_argument("--script", default="guided", nargs="?", help="Script to run for installation", type=str)
 	args, unknowns = parser.parse_known_args()
 	if args.config is not None:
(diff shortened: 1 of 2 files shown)


archinstall/__init__.py
```diff
@@ -33,6 +33,8 @@ def initialize_arguments():
 	parser.add_argument("--creds", nargs="?", help="JSON credentials configuration file")
 	parser.add_argument("--silent", action="store_true",
 						help="Warning!!! No prompts, ignored if config is not passed")
+	parser.add_argument("--dry-run", action="store_true",
+						help="Warning!!! Only creates config, does not install")
```
dylanmtaylor, line 37: I don't like the Warning!!! in this message. Can we change the help text to "Generates a configuration file and then exits instead of performing an installation"?

archinstall/__init__.py
```diff
@@ -33,6 +33,8 @@ def initialize_arguments():
 	parser.add_argument("--creds", nargs="?", help="JSON credentials configuration file")
 	parser.add_argument("--silent", action="store_true",
 						help="Warning!!! No prompts, ignored if config is not passed")
```
dylanmtaylor, line 35: 'Warning!!!' also looks out of place here now that I think about it. Let's make this help text something like "WARNING: Disables all prompts for input. If no configuration is provided, this is ignored"

archinstall/__init__.py
l4zy0n3: Sure, sounds way better

dylanmtaylor:
It'd be nice to squash this so it's not 5 commits

PR #573, opened 2021-06-10
[wip] Add unit tests for check_mirror_reachable

ninchester:
Tested locally but not on any CI


dylanmtaylor:
We have pytest in the repo already and have the CI ready to go, could you please use that?

dylanmtaylor:
Can you uncomment this block?

https://github.com/archlinux/archinstall/blob/master/.github/workflows/lint-python.yaml#L37

marcuxyz:
[suggest]

i was thinking if we should use [ward](https://ward.readthedocs.io/en/latest/) because:

- most simple
- output enhanced
- less verbose

PR #575, opened 2021-06-12
Implement is_desktop_profile helper function

dylanmtaylor:
This is tested and it works. This pull request makes a change in how we ask for audio selections. If the user chooses desktop or they choose a profile within desktop by typing that profile name directly, they would be prompted to choose between pipewire and pulseaudio. If they choose a profile that is not a desktop profile they will additionally be allowed to choose none which will be the default for that profile. None is not available for desktop profiles because of dependencies and it would not work correctly. I moved all the supported desktop profiles in the desktop file to a global variable that is being read in to determine whether or not a profile is a desktop or not.

### DIFF
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -213,7 +213,7 @@ def ask_user_questions():
 	# Ask about audio server selection if one is not already set
 	if not archinstall.arguments.get('audio', None):
 		# The argument to ask_for_audio_selection lets the library know if it's a desktop profile
-		archinstall.ask_for_audio_selection(is_desktop_profile(archinstall.arguments['profile']))
+		archinstall.arguments['audio'] = archinstall.ask_for_audio_selection(is_desktop_profile(archinstall.arguments['profile']))
 
 	# Ask for preferred kernel:
 	if not archinstall.arguments.get("kernels", None):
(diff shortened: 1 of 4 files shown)


examples/guided.py
```diff
@@ -212,13 +212,8 @@ def ask_user_questions():
...
 	if not archinstall.arguments.get('audio', None):
-		# only ask for audio server selection on a desktop profile
-		if str(archinstall.arguments['profile']) == 'Profile(desktop)':
-			archinstall.arguments['audio'] = archinstall.ask_for_audio_selection()
-		else:
-			# packages installed by a profile may depend on audio and something may get installed anyways, not much we can do about that.
-			# we will not try to remove packages post-installation to not have audio, as that may cause multiple issues
-			archinstall.arguments['audio'] = None
+		# The argument to ask_for_audio_selection lets the library know if it's a desktop profile
+		archinstall.ask_for_audio_selection(is_desktop_profile(archinstall.arguments['profile']))
```
Torxed, line 216: We're not interested in the return value?
I assume this would be the case:
```python
archinstall.arguments['audio'] = archinstall.ask_for_audio_selection(is_desktop_profile(archinstall.arguments['profile']))
```
Or why is `ask_for_audio_selection()` returning a value?

Torxed (changes requested):

examples/guided.py
dylanmtaylor: Nice catch, fixed.

PR #577, opened 2021-06-13
fix the case when there's --config but --silent is not set

mytbk:


examples/guided.py
```diff
@@ -442,7 +442,7 @@ def perform_installation(mountpoint):
 	archinstall.log(f"Arch Linux mirrors are not reachable. Please check your internet connection and the log file '{log_file}'.", level=logging.INFO, fg="red")
 	exit(1)
 
-if archinstall.arguments.get('silent', None) is None:
+if not archinstall.arguments.get('silent'):
```
Torxed, line 445: Isn't this doing exactly the same thing?
(in the original code) if `silent` is missing, it will return `None` and we check if the get() return value is `None` and we wait for input.

mytbk: This place sets arguments['silent'] when there's --config used:
https://github.com/archlinux/archinstall/blob/b9af735d858dc2400fe7070f5c11c07c29dd3734/archinstall/__init__.py#L57
However, --silent is a store_true argument, so the above sets arguments['silent'] to False instead of None:
https://github.com/archlinux/archinstall/blob/b9af735d858dc2400fe7070f5c11c07c29dd3734/archinstall/__init__.py#L34

Torxed: Thanks for clarifying this, that feature is still a bit new to me and since I don't run it on a regular basis I wasn't sure what the fix was really for : )

dylanmtaylor (approved):
Looks fine to me

l4zy0n3:
Looks fine to me, however, we might need to modify `ask_user_questions()` to not show any prompts iff config is passed

mytbk:
> Looks fine to me, however, we might need to modify `ask_user_questions()` to not show any prompts iff config is passed

A user may pass part of the options as default, and let the guided installer show prompts on other questions.

PR #578, opened 2021-06-13
created load_config() to load configuration

### LINKED ISSUE #602: Tried installing - got an error and everything halted
I recorded every step of the way

https://youtu.be/aTUaysnMQEY

It is hidden so no one but the visitors here will see it.
It is best I show you rather than typing it all out.

Hope you understand.

### LINKED ISSUE #606: KeyError '_selected_servers'
Hi, I've just tried installing with this in a fresh virt-manager VM with archiso 2021.06. The installation failed once (don't remember the exact error anymore) and when retrying with `archinstall --config /var/log/archinstall/user_configuration.json` it failed with this error:

Screenshot_20210722_201409

I have chosen server profile and selected only sshd (nr 7) in the prompt, it looks like this wasn't properly saved in the json?

Files:
[install.log](https://github.com/archlinux/archinstall/files/6864404/install.log)
[user_configuration.json](https://github.com/archlinux/archinstall/files/6864406/user_configuration.json.txt)

l4zy0n3:
Creating a `load_config()` function that shall only populate the config
Goals
1. Having a `load_config()` function which only loads the config
2. Refactoring `ask_user_questions()` to **only** prompt if config is not loaded, set default values **iff** not entered from config or from prompt, maybe separate setting default values to a separate function
3. Refactor `perform_installation_steps()` and `perform_installation()` to **only** perform installation and not modify anything else

we could have a utility function that loads key from config if present else prompts else returns a default value

Fixes #602, while condition does not work if only one partition was present before formatting
Fixes #606 added new `servers` key

### DIFF
--- a/archinstall/lib/disk.py
+++ b/archinstall/lib/disk.py
@@ -519,6 +519,8 @@ class Filesystem:
 			self.blockdevice.partition[1].allow_formatting = True
 		else:
 			# we don't need a seprate boot partition it would be a waste of space
+			if not self.parted_mklabel(self.blockdevice.device, "msdos"):
+				raise KeyError(f"Could not create a MSDOS label on {self}")
 			self.add_partition('primary', start='1MB', end='100%')
 			self.blockdevice.partition[0].filesystem = root_filesystem_type
 			log(f"Set the root partition {self.blockdevice.partition[0]} to use filesystem {root_filesystem_type}.", level=logging.DEBUG)
@@ -539,11 +541,9 @@ class Filesystem:
 
 		if partitioning:
 			start_wait = time.time()
-			while previous_partitions == self.blockdevice.partitions:
-				time.sleep(0.025)  # Let the new partition come up in the kernel
-				if time.time() - start_wait > 10:
-					raise DiskError(f"New partition never showed up after adding new partition on {self} (timeout 10 seconds).")
-
+			time.sleep(0.025)  # Let the new partition come up in the kernel
+			if time.time() - start_wait > 20:
+				raise DiskError(f"New partition never showed up after adding new partition on {self} (timeout 10 seconds).")
 			return True
 
 	def set_name(self, partition: int, name: str):
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -280,7 +280,7 @@ def perform_installation_steps():
 		config_file.write(user_configuration)
 	print()
 
-	if archinstall.arguments.get('dry_run'):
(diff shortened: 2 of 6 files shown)


l4zy0n3:
Tested on Virtualbox

Anders-1:
Please merge this.

Torxed:
Merging this in once I wrap my head around the changes (they're not many or complicated, it's just my brain needs a jog)

archinstall/lib/disk.py
```diff
@@ -539,11 +541,9 @@ def add_partition(self, partition_type, start, end, partition_format=None):
...
 		if partitioning:
 			start_wait = time.time()
-			while previous_partitions == self.blockdevice.partitions:
-				time.sleep(0.025)  # Let the new partition come up in the kernel
-				if time.time() - start_wait > 10:
-					raise DiskError(f"New partition never showed up after adding new partition on {self} (timeout 10 seconds).")
-
+			time.sleep(0.025)  # Let the new partition come up in the kernel
+			if time.time() - start_wait > 20:
+				raise DiskError(f"New partition never showed up after adding new partition on {self} (timeout 10 seconds).")
```
Torxed, line 546: Just a note to myself: We needed the while loop to compare partition layouts, but there's a bug where the length of the lists were identical so this while loop would just fly by anyway. The new #426 will hopefully handle this more gracefully. So merging this in anyway.

examples/guided.py
```diff
@@ -26,7 +26,33 @@
...
+			archinstall.arguments['mirror-region'] = {selected_region: archinstall.list_mirrors()[selected_region]}
+	if archinstall.arguments.get('sys-language', None) is not None:
+		archinstall.arguments['sys-language'] = archinstall.arguments.get('sys-language', 'en_US')
+	if archinstall.arguments.get('sys-encoding', None) is not None:
+		archinstall.arguments['sys-encoding'] = archinstall.arguments.get('sys-encoding', 'utf-8')
+	if archinstall.arguments.get('gfx_driver', None) is not None:
+		archinstall.storage['gfx_driver_packages'] = AVAILABLE_GFX_DRIVERS.get(archinstall.arguments.get('gfx_driver', None), None)
+	if archinstall.arguments.get('servers', None) is not None:
+		archinstall.storage['_selected_servers'] = archinstall.arguments.get('servers', None)
+
```
Torxed, line 55: Note to self so I don't forget: Update this section when the new partitioning JSON layout comes flying in.

Issue #586, opened 2021-06-19
Encryption does not seem to work with GRUB.


dylanmtaylor:
We have multiple users reporting issues using a GRUB bootloader with encryption enabled. I think this is related to the /etc/default/grub file. From what I'm reading on https://wiki.archlinux.org/title/Dm-crypt/Encrypting_an_entire_system#Configuring_GRUB_2 it looks like we need to have `GRUB_ENABLE_CRYPTODISK=y` and `cryptdevice=UUID...` in that config file.

Redo11:
Still broken in 01.09 archiso image

dabljues:
I can confirm, still broken on the newest (01.09) arch image

Torxed:
A few `parted` calls later, and retries on trying to retrieve PARTUUID for partitions *(I really dislike the legacy way the kernel deals with partition information)* it seams I can perform a complete installation at least with GRUB and disk encryption.

However, the boot process still needs to be fixed.
So I'm digging in to how GRUB and encryption works and will tweak the suggested parameters.

![screenshot](https://user-images.githubusercontent.com/861439/142026245-b65c755a-1d5c-4ac7-ab88-8a16b56326d6.jpg)

![screenshot](https://user-images.githubusercontent.com/861439/142026354-eb3e1ee1-6f10-4252-ae36-f5880839fe84.jpg)

Ammako:
Things to keep in mind:

- `grub-install` needs to be called *after* adding `GRUB_ENABLE_CRYPTODISK=y` to the config, otherwise core.img won't know what to do with an encrypted partition.
- Might be worth asking for separate encryption passwords during setup, one for boot and one for root. Users may not necessarily want the same for both, especially if you're using luks1 for /boot and luks2 for root.
- For ease of use, I suggest to generate a keyfile for /boot (readable only by root! set permissions appropriately), store it on root partition, and edit crypttab to use the keyfile instead of prompting for a password. This should be reasonably secure because the keyfile can never be accessed without first decrypting the root partition anyway, and it'll avoid having to enter /boot decryption password twice.

Torxed:
> Things to keep in mind:
> 
> * `grub-install` needs to be called _after_ adding `GRUB_ENABLE_CRYPTODISK=y` to the config, otherwise core.img won't know what to do with an encrypted partition.
[...]

PR #595, opened 2021-07-07
Fix encryption + grub both in UEFI and BIOS systems.

hamadmarri:
We need to have two partitions in BIOS one for boot (grub)
and the other for root (/). The format of the boot partition
is ext2 (so it is added).

If disk is chosen to be encrypted, `then /etc/default/grub` is edited
as the followings: https://wiki.archlinux.org/title/Dm-crypt/Encrypting_an_entire_system#Configuring_GRUB_2

Issue: https://github.com/archlinux/archinstall/issues/586

This work is done while working on CachyOS which is Arch based and
it is using customized archinstaller.

To test these changes you can try CachyOS installer which supports
encrypted disk with grub (https://wiki.cachyos.org/).

Hamad

🚨 PR Guidelines:

# New features *(v2.2.0)*

# Fix encryption + GRUB
https://github.com/archlinux/archinstall/issues/586<br>

We need to have two partitions in BIOS one for boot (grub)
and the other for root (/). The format of the boot partition
is ext2 (so it is added).

If disk is chosen to be encrypted, `then /etc/default/grub` is edited
as the followings: https://wiki.archlinux.org/title/Dm-crypt/Encrypting_an_entire_system#Configuring_GRUB_2

Issue: https://github.com/archlinux/archinstall/issues/586

This work is done while working on CachyOS which is Arch based and
it is using customized archinstaller.

To test these changes you can try CachyOS installer which supports
encrypted disk with grub (https://wiki.cachyos.org/).

# Testing
It has been tested with CachyOS ISO (https://wiki.cachyos.org).

Hamad

### DIFF
--- a/archinstall/lib/disk.py
+++ b/archinstall/lib/disk.py
@@ -347,6 +347,11 @@ class Partition:
 				raise DiskError(f'Could not format {path} with {filesystem} because: {b"".join(handle)}')
 			self.filesystem = 'ext4'
 
+		elif filesystem == 'ext2':
+			if (handle := SysCommand(f'/usr/bin/mkfs.ext2 -F {path}')).exit_code != 0:
+				raise DiskError(f'Could not format {path} with {filesystem} because: {b"".join(handle)}')
+			self.filesystem = 'ext2'
+
 		elif filesystem == 'xfs':
 			if (handle := SysCommand(f'/usr/bin/mkfs.xfs -f {path}')).exit_code != 0:
 				raise DiskError(f'Could not format {path} with {filesystem} because: {b"".join(handle)}')
@@ -518,12 +523,20 @@ class Filesystem:
 			self.blockdevice.partition[0].allow_formatting = True
 			self.blockdevice.partition[1].allow_formatting = True
 		else:
-			# we don't need a seprate boot partition it would be a waste of space
-			self.add_partition('primary', start='1MB', end='100%')
-			self.blockdevice.partition[0].filesystem = root_filesystem_type
-			log(f"Set the root partition {self.blockdevice.partition[0]} to use filesystem {root_filesystem_type}.", level=logging.DEBUG)
-			self.blockdevice.partition[0].target_mountpoint = '/'
+			self.add_partition('primary', start='1MiB', end='513MiB', partition_format='ext2')
+			self.set(0, 'boot on')
+			self.add_partition('primary', start='513MiB', end='100%')
+
+			self.blockdevice.partition[0].filesystem = 'ext2'
+			self.blockdevice.partition[1].filesystem = root_filesystem_type
+
+			log(f"Set the root partition {self.blockdevice.partition[1]} to use filesystem {root_filesystem_type}.", level=logging.DEBUG)
+
+			self.blockdevice.partition[0].target_mountpoint = '/boot'
+			self.blockdevice.partition[1].target_mountpoint = '/'
+
 			self.blockdevice.partition[0].allow_formatting = True
+			self.blockdevice.partition[1].allow_formatting = True
 
 	def add_partition(self, partition_type, start, end, partition_format=None):
 		log(f'Adding partition to {self.blockdevice}', level=logging.INFO)
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -322,8 +322,7 @@ def perform_installation_steps():
 			else:
 				fs.find_partition('/').mount(archinstall.storage.get('MOUNT_POINT', '/mnt'))
 
-			if has_uefi():
-				fs.find_partition('/boot').mount(archinstall.storage.get('MOUNT_POINT', '/mnt') + '/boot')
+			fs.find_partition('/boot').mount(archinstall.storage.get('MOUNT_POINT', '/mnt') + '/boot')
 
 	perform_installation(archinstall.storage.get('MOUNT_POINT', '/mnt'))
 
@@ -352,7 +351,7 @@ def perform_installation(mountpoint):
 				installation.set_mirrors(archinstall.arguments['mirror-region'])  # Set the mirrors in the installation medium
 			if archinstall.arguments["bootloader"] == "grub-install" and has_uefi():
 				installation.add_additional_packages("grub")
-			installation.add_bootloader(archinstall.arguments["bootloader"])
+			installation.add_bootloader(archinstall.arguments["harddrive"], archinstall.arguments["bootloader"])
 
 			# If user selected to copy the current ISO network configuration
 			# Perform a copy of the config
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -409,7 +409,7 @@ class Installer:
 
 		return True
 
-	def add_bootloader(self, bootloader='systemd-bootctl'):
+	def add_bootloader(self, _device, bootloader='systemd-bootctl'):
 		for plugin in plugins.values():
 			if hasattr(plugin, 'on_add_bootloader'):
 				# Allow plugins to override the boot-loader handling.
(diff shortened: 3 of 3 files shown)


archinstall/lib/disk.py
```diff
@@ -518,12 +523,20 @@ def use_entire_disk(self, root_filesystem_type='ext4'):
...
-			# we don't need a seprate boot partition it would be a waste of space
-			self.add_partition('primary', start='1MB', end='100%')
-			self.blockdevice.partition[0].filesystem = root_filesystem_type
-			log(f"Set the root partition {self.blockdevice.partition[0]} to use filesystem {root_filesystem_type}.", level=logging.DEBUG)
-			self.blockdevice.partition[0].target_mountpoint = '/'
+			self.add_partition('primary', start='1MiB', end='513MiB', partition_format='ext2')
+			self.set(0, 'boot on')
+			self.add_partition('primary', start='513MiB', end='100%')
+
+			self.blockdevice.partition[0].filesystem = 'ext2'
```
Torxed, line 530: Is `ext2` a hard dependency to get this working on GRUB? Or can we retaim `.filesystem = root_filesystem_type`?

examples/guided.py
```diff
@@ -352,7 +351,7 @@ def perform_installation(mountpoint):
 				installation.set_mirrors(archinstall.arguments['mirror-region'])  # Set the mirrors in the installation medium
 			if archinstall.arguments["bootloader"] == "grub-install" and has_uefi():
 				installation.add_additional_packages("grub")
-			installation.add_bootloader(archinstall.arguments["bootloader"])
+			installation.add_bootloader(archinstall.arguments["harddrive"], archinstall.arguments["bootloader"])
```
Torxed, line 354: The `archinstall.arguments["harddrive"]` argument should already be given to `installation` via: https://github.com/archlinux/archinstall/pull/595/files#diff-283d51110d004467e254215ad93e4dfcf0188beb22d9f878bb7088c12076dd2eR505

Is there a specific reason why we need to pass the harddrive argument once more?
I'm mainly asking because I'm re-working the logic for partitions and harddrive logics, and I'm wondering if I'll need to translate this over into my feature branch later on.

archinstall/lib/disk.py
hamadmarri: > Is `ext2` a hard dependency to get this working on GRUB? Or can we retaim `.filesystem = root_filesystem_type`?

Most articles I found about boot w/ grub mention that the format should be either ext2, ext3, or ext4. I am not sure tbh which one is better, however, I wouldn't leave the format related to users choice for root since xfs, btrfs, or other formats may not work with grub in separate boot.

https://unix.stackexchange.com/questions/175728/what-file-system-should-a-grub-2-boot-partition-use

Thanks

examples/guided.py
hamadmarri: > The `archinstall.arguments["harddrive"]` argument should already be given to `installation` via: https://github.com/archlinux/archinstall/pull/595/files#diff-283d51110d004467e254215ad93e4dfcf0188beb22d9f878bb7088c12076dd2eR505
> 
> Is there a specific reason why we need to pass the harddrive argument once more?
> I'm mainly asking because I'm re-working the logic for partitions and harddrive logics, and I'm wondering if I'll need to translate this over into my feature branch later on.

I was thinking this as well, but something weird that doesn't make this work with virtualbox. I tested with hard coded `/dev/sda` and it worked perfectly. So thus I simply pass the `archinstall.arguments["harddrive"]` to the function which holds the `/dev/sdX`. I have done this in a harry. I am sure the PR code need to be refactored. 

For this specific point, the only try worked correctly for me is:
`grub-install --target=i386-pc --recheck {_device.path}`

where `{_device.path}` = `/dev/sda`
`/boot` = `/dev/sda1`
`/` = `/dev/sda2`

Torxed: I think we should investigate why `/dev/sda` isn't detected why [real_device](https://github.com/archlinux/archinstall/blob/51eb05691a5845652a9dc583de05519f095961f3/archinstall/lib/installer.py#L491) isn't detected properly, if not we should fall back to [root_partition](https://github.com/archlinux/archinstall/blob/51eb05691a5845652a9dc583de05519f095961f3/archinstall/lib/installer.py#L491) instead of passing in more arguments.

This isn't easy to know for a new contributor, but we did extensive work to avoid passing in harddrive's as arguments to these functions because the installation should handle this within itself.

archinstall/lib/disk.py
Torxed: In that case, I think `ext4` would be better if we're going to hard code anything. Which means the user's choice of root partition filesystem will be ignored. @advaithm can probably chime in here :)

advaithm: There isn't a hard dep on the file system however if we do chose to hard code fs etx4 would be the best choice

advaithm: note for btrfs on mbr you may have to offset the start partition by 1mb

ptr1337:
Im also reinstalled today my rig, im also maintainer at cachyos (its in really early state) 

Mostly none Calmares Install with Grub and encryption failed at my nvme, also archinstall failed. 

endavouros worked 8/10 times with grub, archinstall not one time with encryption. 

Today mostly my first flawless install on my nvme.

archinstall/lib/disk.py
Ammako: Never had issues with ext4, speaking of /boot partition though keep in mind that luks2 support for grub2 is [limited](https://www.gnu.org/software/grub/manual/grub/html_node/cryptomount.html).

jatin-cbs:
I also recommend changing ext2 to ext4 for boot. Most distros recommend ext4 for Boot. With some exceptions like Kali Linux using older file systems

ptr1337:
> I also recommend changing ext2 to ext4 for boot. Most distros recommend ext4 for Boot. With some exceptions like Kali Linux using older file systems.
> filesystem = root_filesystem_type is NOT recommended due to other file systems still have regressions with boot. Even distros with Btrfs/XFS as /root and /home use ext4 as a boot filesystem.

I mostly never saw a distro which has the /boot parition using ext2 or ext4. 
Mostly i only saw for /boot vfat and then root for whatever, xfs,btrfs,f2fs,ext4

Ammako:
> I mostly never saw a distro which has the /boot parition using ext2 or ext4.

Neither Debian nor Ubuntu even *use* a /boot partition by default; they just put everything under / as ext4, unless you use UEFI then they reserve a fat32 partition for /boot/efi. Some setups will put the EFI system partition at /boot instead of /boot/efi, that's not generally what's being referred to when talking about /boot partition though.

If you pick the default encryption layout in their installers, then they'll have an unencrypted ext4 partition at /boot, with an additional fat32 EFI system partition at /boot/efi if applicable.

advaithm:
I've done some research and there isn't a fs restriction for grub. We don't need to enforce one. Could the pr be changed to reflect this?

jatin-cbs:
Yes @Ammako is correct.. the default encryption layout in fedora for example uses ext4 for boot and vfat for /boot/efi. @ptr1337  I was referring to this layout should have clarified tho !

ptr1337:
> Yes @Ammako is correct.. the default encryption layout in fedora/ubuntu for example uses ext4 for boot and vfat for /boot/efi. @ptr1337 I was referring to this layout should have clarified tho !

this pull request maybe needs overwork, but it works. 
at the archinstall its not working with grub x)

advaithm:
> Yes @Ammako is correct.. the default encryption layout in fedora/ubuntu for example uses ext4 for boot and vfat for /boot/efi. @ptr1337  I was referring to this layout should have clarified tho ! 

I think the better option is to stick with unified root and boot for bios(so ext4/btrfs etc) with a the current vfat /boot and what ever the user picks for the root partition.

Ammako:
I'm not a big fan of keeping bios boot and root together on the same partition, at least until grub is updated to properly support luks2. Otherwise you have to weaken encryption for the whole thing by using luks1 for everything. By keeping them separate you can use luks1 for bios boot and luks2 for the rest.

advaithm:
> I'm not a big fan of keeping bios boot and root together on the same partition, at least until grub is updated to properly support luks2. Otherwise you have to weaken encryption for the whole thing by using luks1 for everything. By keeping them separate you can use luks1 for bios boot and luks2 for the rest.


Afaik archinstall does unencrypted boot on uefi since systemd-boot doesn't support encrypted boot. I think it would be acceptable to support unecrypted boot on bios.

jatin-cbs:
> > I mostly never saw a distro which has the /boot parition using ext2 or ext4.
> 
> Neither Debian nor Ubuntu even _use_ a /boot partition by default; they just put everything under / as ext4, unless you use UEFI then they reserve a fat32 partition for /boot/efi. Some setups will put the EFI system partition at /boot instead of /boot/efi, that's not generally what's being referred to when talking about /boot partition though.
> 
> If you pick the default encryption layout in their installers, then they'll have an unencrypted ext4 partition at /boot, with an additional fat32 EFI system partition at /boot/efi if applicable.

In case encryption is selected -
unencrypted ext4 partition at /boot with a vfat /boot/efi is what I would recommend as Ammako said explained with others being encrypted the rest partitions or sub volumes (LUKS). This is what most distros do / recommend.

advaithm:
I believe arch defaults to kernels in /boot so it would be better to use vfat for /boot when we are on uefi

Ammako:
> unencrypted ext4 partition at /boot

As an option that's totally fine, however this PR is for encrypted boot. Grub doesn't need anything special to load unencrypted boot --> encrypted root, but the issue here is that when archinstall does an encrypted install, the proper configuration isn't being done for grub to be able to boot from the encrypted /boot.

Torxed:
Sorry for the delay in this matter, I've had some big personal matters to attend to.
I'm currently in the process of merging this in (with some modifications). It will most likely not take long, but the work has started :)

Torxed:
As I do not have write permission in CachyOS:master, I will have to force these changes into our master manually, which may or may not break traceability back to you @hamadmarri, this only means that you might not get credit for the work (not sure how github will deal with this.. But I'm going for it to avoid waiting more on this merge) : )

Issue #264, opened 2021-04-10
Wireless configuration is static after installing ArchL (with Gnome in my case)


AntiBigTech:
Dear people of Archinstall:

When I was being asked for: what configuration of the network Interface would I choose? (during the archinstall steps), I choose the first option [Save the current network configuration ~ and I was installing it whith my wifi enabled whith iwctl]...

And after the installation I noticed warious facts:

First one, I can't see the Wifi Button on the main menu of the upper bar:
![Screenshot from 2021-04-09 19-07-16](https://user-images.githubusercontent.com/82244838/114251627-531ca700-9967-11eb-989f-ea100324f24c.png)

But i wasn't asked for which Wifi networ to choose or typing the password for it. It simply "catched" my first wifi enabled, but it don't allow me to GUI search other wifi networks...

Second: It is not possible to open the network configuration from the settings window: 
![Screenshot from 2021-04-09 19-19-11](https://user-images.githubusercontent.com/82244838/114251939-890e5b00-9968-11eb-9669-2c5abdd62e5c.png)

Third: I manual installed the NetworkManager (with "$ sudo pacman -S networkmanager"), but it doesn't take effect.

It's obwious that the installer was too "Literal" whith maintaining the first config, and don't installed that service with the GUI integration with Gnome.

Thanks.

dylanmtaylor:
> First one, I can't see the Wifi Button on the main menu of the upper bar:

GNOME doesn't display this without the networkmanager package installed and the systemd service running, do

```
sudo pacman -S networkmanager
sudo systemctl enable --now NetworkManager
```

dylanmtaylor:
This is improved in the version in master, which will be in 2.1.4 by the way - there is a new option to use NetworkManager that takes care of this for you so you don't have to deal with it.

dylanmtaylor:
I thought about adding networkmanager to GNOME and KDE profiles in the past because of this specific issue, but our goal in the installer is to basically bootstrap a minimal functional desktop and let users configure it without making choices for them, and not everyone wants networkmanager. (https://github.com/archlinux/archinstall/pull/245).

AntiBigTech:
> > First one, I can't see the Wifi Button on the main menu of the upper bar:
> 
> GNOME doesn't display this without the networkmanager package installed and the systemd service running, do
> 
> ```
> sudo pacman -S networkmanager
> sudo systemctl enable --now NetworkManager
> ```

Thanks a lot! It worked perfect.

[...]

PR #603, opened 2021-07-19
Mention GNOME and KDE's soft dependency on NetworkManager in the installation process

j-james:
This PR tags onto #264 - most people don't know that NetworkManager is needed for a fully functional GNOME (and KDE) system, and might accidentally choose to just copy over the ISO network configuration.


Torxed:
Apologies for the late action on this, good catch and it's now merged!

PR #605, opened 2021-07-22
Update hardware.py

CYPT71:
Rewrite some function 

if condition is True then return true else return false, transform in return condition directly 

Also I don't understand why we need a try/except at line 151 and why we not write return condition ??

🚨 PR Guidelines:

# New features *(v2.2.0)*

# Describe your PR

# Testing


Torxed:
I tweaked the PR a little bit. Mainly to add some more error handling that is being affected by the PR. You are correct on the Try/Catch, that's from legacy code. So instead of commenting it out I removed it all together.

PR #608, opened 2021-07-24
Remove lightdm packages from Deepin profile

felixonmars:
As of deepin-session-shell 5.4.42-2 (which is in the deepin group), lightdm is added as a dependency because of lightdm-deepin-greeter. A configuration (usr/share/lightdm/lightdm.conf.d/60-deepin.conf) in startdde (also in the deepin group) should override default lightdm configuration to use lightdm-deepin-greeter instead of the gtk greeter. Thus these two packages could be removed from the profile now.

This effectively reverts #441

🚨 PR Guidelines:

# New features *(v2.2.0)*

# Describe your PR

# Testing

### DIFF
--- a/profiles/deepin.py
+++ b/profiles/deepin.py
@@ -8,6 +8,8 @@ __packages__ = [
 	"deepin",
 	"deepin-terminal",
 	"deepin-editor",
+	"lightdm",
+	"lightdm-deepin-greeter",
 ]
 
 
--- a/profiles/deepin.py
+++ b/profiles/deepin.py
@@ -8,8 +8,6 @@ __packages__ = [
 	"deepin",
 	"deepin-terminal",
 	"deepin-editor",
-	"lightdm",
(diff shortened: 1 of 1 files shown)


dylanmtaylor:
Since it's pulled in as a dependency - explicitly having it listed doesn't cause issues, and adds clarity about which greeter is used. I think it might make sense to keep it.

felixonmars:
lightdm-gtk-greeter should not be used though. I think there is no point to keep it?

dylanmtaylor:
I feel like this commit should the be changed to make lightdm-gtk-greeter into lightdm-deepin-greeter.

Torxed:
Apologies for the delay in this. As Iv'e mentioned in other places, I had to deal with some personal stuff.
But I appreciate you all working out the details in my absence and that the PR was clean and easy to merge :) It's there for now merged!

PR #613, opened 2021-08-17
fixed the git checkout non-existent branch

RustyRaptor:
It looks like the `torxed-v2.2.0` branch doesnt exist anymore and I assume its now just `v2.2.0` instead so I changed this command in Using a live ISO

🚨 PR Guidelines:

# New features *(v2.2.0)*

# Describe your PR

# Testing


Torxed:
This is correct, this was text taken from the old ways of working :) Well spotted!

PR #614, opened 2021-08-18
Fix wrong default in kernels (guide.py)

laxect:
Fix wrong default in kernels.


Torxed:
Thank you for fixing this and catching it!

PR #615, opened 2021-08-20
Native cpu detection

conqp:
🚨 PR Guidelines:

# New features *(v2.2.0)*
N/A

# Describe your PR
Refactoring of CPU detection functions.

# Testing
Tested on Intel and AMD machines.


dylanmtaylor:
This looks like a good change to me, +1

Torxed:
Nice and clean implementation. It's a clear advantage over `check_output` which causes exception and is a bit slower *(not that performance is a huge issue for us, but still, I like optimized stuff)*.

PR #637, opened 2021-09-18
Adding in options for BTRFS subvolumes

### LINKED ISSUE #93: BTRFS subvolumes to support tools like Snapper with rollbacks etc.
Just a question/feature request?

Would it be possible to have an option for BTRFS that sets up subvolumes like OpenSUSE and/or any Timeshift/Snapper compatible way?

https://rootco.de/2018-01-19-opensuse-btrfs-subvolumes/

https://endeavouros.com/docs/encrypted-installation-2/btrfsonluks-verbose-version/
https://endeavouros.com/docs/encrypted-installation-2/btrfsonluks-quick-copypaste-version/

Torxed:
This should fix #93.
It will allow for BTRFS subvolumes, and only question the user to use them if btrfs filesystem was selected for a specific partition.

This change also brings:
 * Restructured the `/lib/disk.py` into separate `/lib/disk/*.py` files *(in order to clean up the code and separate logic a bit, from being one massive file)*


Torxed:
screenshot

The default BTRFS subvolume suggestion is based on https://rootco.de/2018-01-19-opensuse-btrfs-subvolumes/:
```json
"subvolumes" : {
	"@home" : "/home",
	"@log" : "/var/log",
	"@pkgs" : "/var/cache/pacman/pkg",
	"@.snapshots" : "/.snapshots"
}
```

I don't know enough about the snapshot tools to say if this will work or not, but I consider this PR ready for review and ready to go into master. I'll resolve the conflicts with `disk.py` since there were some updated on it between this fork splitting `disk.py` into `lib/disk/*.py` to make it more maintainable.

timakro:
@Torxed I'm not happy with the current support for subvolume. Let me explain. To understand the full space of options Btrfs provides in creating subvolumes this guide is the perfect resource: https://btrfs.wiki.kernel.org/index.php/SysadminGuide#Layout It's essential to understand the 3 layouts mentioned in the guide. At the moment archinstall only supports the nested approach. And poorly at that, because:

* Currently there is no point in using a dictionary to store the subvolumes. The key is never used.
* Currently, `mount_subvolume` is unnecessary. Subvolumes are always visible under their location. They can be mounted to another place additionally though.

I'm suggesting to use the following scheme to represent the layouts. The "Flat" layout from the guide would be represented as

```json
"subvolumes": {
  "root": "/",
  "home": "/home",
  "var/www": "/var/www",
  "postgres": "/var/lib/postgresql"
}
```

The "Nested" layout is

```json
"subvolumes": {
  "home": null,
  "var": null,
  "var/www": null,
  "var/lib/postgresql": null
}
```

The second "Nested" layout is

```json
"subvolumes": {
  "root": "/",
  "root/home": null,
  "root/var": null,
  "root/var/www": null,
  "root/var/lib/postgresql": null
}
```

It should also be valid to begin the strings on the left side (keys) with a slash as those are just absolute paths. And I suppose an empty string `""` on the right side (values) should be interpreted the same as `null`.

Creating subvolumes would be as easy as just calling `btrfs subvolume create` but the order has to be figured out first so children are always created after parents. Mounting is only necessary for ones which have non-null right sides, similarly the order matters.

Torxed:
Thank you for your feeedback.
I'll try to address most of them : )

> And poorly at that, because:
> * Currently there is no point in using a dictionary to store the subvolumes. The key is never used.

The use of a dictionary here is mostly irrelevant and has little to do with BTRFS support I think, so I don't think the word *because* here accurate. The variable could be anything. The key was used for a bit in my code tweaks when passing name separate from the path, but got changed to omit the name entirely and rely on only the path. It used to be `create_subvolume(installer, name, mountpoint)`. The key acts as a placeholder and uniqueness instead to better expand BTRFS things in the future. Where it would be easy to convert the subsequent code with the structure of something like:
```json
"subvolumes" : {
	"@home" : {"path" : "/home", "mount-options" : ["...", "..."]},
	...
}
```
But it's important to know that this PR is the *first step* towards BTRFS subvolumes, not meant to be a complete support for btrfs and all its complexity. Either I skip it for this release and no one gets subvolumes - or I add partial functionality that caters to some or a lot of people. And I'm not saying this to hold the feature hostage, only to put some light on the fact that there's multiple aspects to how we release features and our use cases.

> * Currently, `mount_subvolume` is unnecessary. Subvolumes are always visible under their location. They can be mounted to another place additionally though.

I don't see the harm in keeping the function there and expand on it later. Mounting them explicitly was something that was done in [the guide mention in the issue #93](https://rootco.de/2018-01-19-opensuse-btrfs-subvolumes/). It would also make sense that this function would updated mount options/fstab arguments in the future.

> I'm suggesting to use the following scheme to represent the layouts. The "Flat" layout from the guide would be represented as

Your suggestions is more or less the same concept in terms of variable usage, or I struggle to see the difference.
The key in the dictionary is just a unique placeholder to ensure we have unique entries and could be anything. And it would be the same as setting the value to `null`. So I think we're saying the same thing, the key is rather in the interpretation of the structure.

And that's where I struggle a bit, and it's clear that I'll need to both use and understand the actual differences between *(and I'm using their terms now)* flag, nester and mixed. To me that's just different mount-points and nothing else.

> Creating subvolumes would be as easy as just calling `btrfs subvolume create` but the order has to be figured out first so children are always created after parents.

That is a valid point, and I'll get on it to implement the sorting/order, just like we have for the mount proceedure in the `Installer()` class.


> EDIT: If you want we can hang out on IRC or some other chat some time. I know Btrfs subvolumes are difficult to grasp and I think I could be of help.

I'd love to, sadly IRC is something I have to go out of my way to use as it's not part of the daily flow. And currently we have a bit of a family situation *(again)* that prohibits me to take that time. So I hang out on discord & matrix for the most part as that's where friends and family hang out. The rest of my time goes to coding, work and family. But I'd love a chat in a bit.

> EDIT2: I believe genfstab will recognize those layouts correctly.

I'll verify! : )
But most importantly, would this be a good first step PR towards adding subvolume functionality or will this break systems? I'm mainly interested in giving *some* or *most* people subvolume support using one traditional use case. And cover the rest later. Iterative beats holding back until we've reached unicorn status.

PR #639, opened 2021-09-18
Update xfce4.py

falieri:
add network support to the xfce panel


Torxed:
Thanks for contributing and making the profiles better! :)

PR #678, opened 2021-11-01
Adding support for swap on zram

### LINKED ISSUE #175: enabling swap on ZRAM
There is no SWAP by default in the guided installer, However I recommend Swap on ZRAM by default as it will automatically allocate swap on boot.    So SWAP will automatically be equal to RAM size ! This is how its in Fedora and is the default behavior since fedora 33.

Torxed:
This will fix #175 by adding support for swap on zram.
Which will be installed and setup using [zram-generator](https://github.com/archlinux/svntogit-community/blob/packages/zram-generator/trunk/PKGBUILD) and the default example of [zram-generator.conf](https://github.com/systemd/zram-generator/blob/main/zram-generator.conf.example).


uhthomas:
This is really cool, I'm looking forward to this welcome change.

I used `archinstall` a few days ago on a new system and didn't setup swap pre-install. As such, I went through the process of installing `zram-generator` and have some observations.

1. I don't think it's necessary to explicitly enable any services. It looks like `zram-generator` kinda just does it.

2. Is it wise to use the example configuration? It creates two zram devices, one of which a comment clearly states is unused.

3. It looks like the maintainers of `zram-generator` tend to prefer `/etc/systemd/zram-generator.conf` as the default config path.

The `zram-generator` configuration can and probably should be quite minimal. I've found the following works well:

```systemd
[zram0]
compression-algorithm = zstd
```

on a machine with 16GB of memory produces:

```sh
❯ zramctl
NAME       ALGORITHM DISKSIZE DATA COMPR TOTAL STREAMS MOUNTPOINT
/dev/zram0 zstd            4G   4K   63B    4K      10 [SWAP]
❯ swapon -s
Filename				Type		Size		Used		Priority
/dev/zram0                              partition	4194300		0		100
```

Although that does require `zstd`, so maybe a _really_ minimal configuration file is better suited:

```systemd
[zram0]
```

I'd be interested to hear your thoughts!

Torxed:
@uhthomas I think your judgement is really good. I assumed (naively) that the example config were the defaults more or less.

I'll commit your suggestions in a few minutes. :)

PR #684, opened 2021-11-02
Fix #682 - swap ntpd for systemd-timesyncd

### LINKED ISSUE #682: Time, Timezone and Synchronize System Clock problem!
i guess , it need "timedatectl set-ntp true" for synchronize

Torxed:
As per question/issue/suggestion in this PR will fix #682. If we opt to use systemd to handle the time sync instead of the traditional `ntpd`.

This PR is **Optional**, but submitting it for convenience if we decide to go this route.


dylanmtaylor:
I think this may be a good approach, I am in favor of using native systemd services where possible due to the great level of integration with the rest of the system, +1 from me.

archinstall/lib/installer.py
```diff
@@ -248,9 +248,15 @@ def set_timezone(self, zone, *args, **kwargs):
 
 	def activate_ntp(self):
 		self.log('Installing and activating NTP.', level=logging.INFO)
```
dylanmtaylor, line 250: This log line should be changed to "Configuring systemd-timesyncd for NTP functionality" or similar.

Torxed: I'll probably chance in guided as well to better reflect this, and rename the function name, good call!

Issue #674, opened 2021-10-22
[master] crash at disk size checking


wllacer:
Using master branch. 
Starting the process (after getting keyboard and country data) archinstall crashes with a ValueError exception.
```Select one of the above regions to download packages from (by number or full name): 57
Traceback (most recent call last):
  File "/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/werner/projects/archinstall-master/archinstall/__main__.py", line 4, in <module>
    archinstall.run_as_a_module()

  File "/home/werner/projects/archinstall-master/archinstall/__init__.py", line 103, in run_as_a_module
    script.execute()

  File "/home/werner/projects/archinstall-master/archinstall/lib/profiles.py", line 178, in execute
    self.spec.loader.exec_module(sys.modules[self.namespace])

  File "<frozen importlib._bootstrap_external>", line 850, in exec_module

  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed

  File "/home/werner/projects/archinstall-master/archinstall/examples/guided.py", line 371, in <module>
    ask_user_questions()

  File "/home/werner/projects/archinstall-master/archinstall/examples/guided.py", line 113, in ask_user_questions
    archinstall.arguments['harddrives'] = archinstall.generic_multi_select(archinstall.all_disks(),

  File "/home/werner/projects/archinstall-master/archinstall/lib/user_interaction.py", line 152, in generic_multi_select
    printed_options.append(f'{option}')

  File "/home/werner/projects/archinstall-master/archinstall/lib/disk.py", line 255, in __repr__
    return f"BlockDevice({self.device_or_backfile}, size={self.size}GB, free_space={'+'.join(part[2] for part in self.free_space)}, bus_type={self.bus_type})"

  File "/home/werner/projects/archinstall-master/archinstall/lib/disk.py", line 398, in size
    return self.convert_size_to_gb(device['size'])

  File "/home/werner/projects/archinstall-master/archinstall/lib/disk.py", line 391, in convert_size_to_gb
    return float(units.get(unit, lambda s : None)(size[:-1]))

  File "/home/werner/projects/archinstall-master/archinstall/lib/disk.py", line 385, in <lambda>
    'G' : lambda s : float(s),

ValueError: could not convert string to float: '298,1'
```

[...]

PR #695, opened 2021-11-08
Solves issue #674

wllacer:
We turned the size query of the devices to byte mode in lsblk (lsblk -b)
It avoids problems with the localized output of the lsblk utility.
Output formatting may be in need of some care (too much decimal)


wllacer:
I've added a round() function to avoid the excess of decimals. Behaviour should be OK as the result is the same as with the standard lsblk output

dylanmtaylor:
This looks like a good change, it cleans up a lot of conditional logic. Assuming this is tested to work, a huge +1 from me.

PR #704, opened 2021-11-09
Fix #689 by adding a exception to filesystem.parted() on failure

### LINKED ISSUE #689: AttributeError: 'NoneType' object has no attribute 'format' on master
It seams that `Getting mount information at location` is trying to lookup `/dev/sda1` instead of `/`, or something else is wrong.

image

Virtualbox 6.1

Torxed:
This might not entirely solve it, but it's a step along the way by adding a failsafe to `parted()` calls.
We check if the PARTUUID shows up only if parted succeeds, so maybe this solves it?


dylanmtaylor (approved):
This should result in more helpful error messages. +1

PR #706, opened 2021-11-09
Adding in storage of user supplied credentials in a separate file.

Torxed:
This separates credentials from user_configuration.json into user_credentials.json. As well as the JSON serializer will omit the credentials from the user_configuration.json by detecting ! in the dictionary keys (which is why they are important). UNSAFE_JSON will leave those keys in there, good for storing credentials in a separate file."

### DIFF
--- a/archinstall/lib/general.py
+++ b/archinstall/lib/general.py
@@ -77,6 +77,13 @@ def json_dumps(*args, **kwargs):
 
 class JsonEncoder:
 	def _encode(obj):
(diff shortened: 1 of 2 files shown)


Torxed:
> I noticed my encrpytion password in plain text in the output to screen when it was displaying config instead of ********. Will this fix that as well?

That should be fixed, but please verify.
I did forget to add it to the credentials file, which has now been corrected.

PR #707, opened 2021-11-09
Forgot to change the cls=JSON inside the recursive call of UNSAFE_JSON

Torxed:
This should make sure nested dictionaries don't have their `!<sensitive>` data exposed, even on print.
It also allows for sensitive data to be printed when using `UNSAFE_JSON` rather than `JSON` as the `cls=` trigger to `json.dumps`.

### DIFF
--- a/archinstall/lib/disk/filesystem.py
+++ b/archinstall/lib/disk/filesystem.py
@@ -94,15 +94,17 @@ class Filesystem:
 
 			if partition.get('filesystem', {}).get('format', False):
 				if partition.get('encrypted', False):
-					if not partition.get('password'):
+					if not partition.get('!password') and not storage['arguments'].get('!encryption-password'):
 						if storage['arguments'] == 'silent':
 							raise ValueError(f"Missing encryption password for {partition['device_instance']}")
 						else:
 							from ..user_interaction import get_password
-							partition['password'] = get_password(f"Enter a encryption password for {partition['device_instance']}")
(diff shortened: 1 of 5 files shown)


Torxed:
Humm, still doesn't get safed out from: https://github.com/archlinux/archinstall/blob/4b4c66caa1a5f896e44e679e4e1afef5caa657e0/examples/guided.py#L212-L213

Torxed:
Because it's stored as [password](https://github.com/archlinux/archinstall/blob/4b4c66caa1a5f896e44e679e4e1afef5caa657e0/archinstall/lib/disk/filesystem.py#L102) and not `!password`. Will find all the locations and correct.

Torxed:
Culprit: https://github.com/archlinux/archinstall/blob/cd07dafaacec84954961fbbd8d6f4a2bc997410a/archinstall/lib/disk/filesystem.py#L102

Torxed:
Still adds and outputs `password` instead of `!password`. Looking further.

Torxed:
[x] No passwords are outputted in the terminal
[x] No credentials or sensitive information in `user_configuration.json`
[x] No credentials in `user_disk_layout.json`
[  ] Credentials are stored in `user_credentials.py` *(Currently users, superusers and root pw is not stored for some reason)*
screenshot

CarAuto34:
Ty

CarAuto34:
Ty

PR #709, opened 2021-11-10
Solves issue #689. Attribute Error.'None type' has no attribute 'form…

wllacer:
…at' ...

It seems the system does not syncronus update its internal information after a partitioning.
Two places are affected. Directly on filesystem.add_partition (the uuid of the new partition isn't available after the parted command)
and blockdevice.get_partition, where the list of partitions for the iterator might not be available in the query.
The patch places both sections under controlled loops, giving the system the chance to update the information.
Should be more controlled via application parameters

Since implemented i haven't found any more glitches at that place, but a couple of cases for which both loops were triggered.

### DIFF
--- a/archinstall/lib/disk/filesystem.py
+++ b/archinstall/lib/disk/filesystem.py
@@ -167,17 +167,21 @@ class Filesystem:
 			parted_string = f'{self.blockdevice.device} mkpart {partition_type} {start} {end}'
 
 		if self.parted(parted_string):
-			start_wait = time.time()
-
-			while previous_partition_uuids == {partition.uuid for partition in self.blockdevice.partitions.values()}:
-				if time.time() - start_wait > 10:
-					raise DiskError(f"New partition never showed up after adding new partition on {self} (timeout 10 seconds).")
-				time.sleep(0.025)
-
-			# Todo: Find a better way to detect if the new UUID of the partition has showed up.
-			#       But this will address (among other issues)
-			time.sleep(float(storage['arguments'].get('disk-sleep', 2.0))) # Let the kernel catch up with quick block devices (nvme for instance)
-			return self.blockdevice.get_partition(uuid=(previous_partition_uuids ^ {partition.uuid for partition in self.blockdevice.partitions.values()}).pop())
+			count = 0
+			while count < 10:
+				new_uuid = None
+				new_uuid_set = (previous_partition_uuids ^ {partition.uuid for partition in self.blockdevice.partitions.values()})
+				if len(new_uuid_set) > 0:
+					new_uuid = new_uuid_set.pop()
+				if new_uuid:
+					return self.blockdevice.get_partition(new_uuid)
+				else:
+					count += 1
+					log(f"Could not get uuid for partition. Waiting for the {count} time",level=logging.DEBUG)
+					time.sleep(float(storage['arguments'].get('disk-sleep', 0.2)))
+			else:
+				log("Add partition exiting due to excesive wait time",level=logging.INFO)
+				raise DiskError(f"New partition never showed up after adding new partition on {self}.")
 
 	def set_name(self, partition: int, name: str):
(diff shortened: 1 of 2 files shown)


archinstall/lib/disk/filesystem.py
```diff
@@ -167,17 +167,21 @@ def add_partition(self, partition_type, start, end, partition_format=None):
...
-			# Todo: Find a better way to detect if the new UUID of the partition has showed up.
-			#       But this will address (among other issues)
-			time.sleep(float(storage['arguments'].get('disk-sleep', 2.0))) # Let the kernel catch up with quick block devices (nvme for instance)
-			return self.blockdevice.get_partition(uuid=(previous_partition_uuids ^ {partition.uuid for partition in self.blockdevice.partitions.values()}).pop())
+			count = 0
+			while count < 10:
+				new_uuid = None
+				new_uuid_set = (previous_partition_uuids ^ {partition.uuid for partition in self.blockdevice.partitions.values()})
+				if len(new_uuid_set) > 0:
+					new_uuid = new_uuid_set.pop()
```
Torxed, line 175: Are we sure the new `PARTUUID` is always at the end? Are they sorted by time/order or by name?

Torxed:
Overall, great addition. I had hope that the original `while` loop to catch and wait for partitions to show up would be enough. Sadly not the case.

archinstall/lib/disk/filesystem.py
wllacer: TBH is the part i got from the original code, so i have no idea. 
The trick is that at the start of the method the existing partition list is fetched (Blockdevices.partitions is a dynamic property), so only the  UUIDs from partitions created while the method executes should appear at new_uuid_set.
Noone expects someone doing partitioning in parallel, so the code works ... almost always.
OTOH it is something worth investigating to get the code "concurrency safe".

wllacer: Sadly not possible without a serious rewrite (and probably even then). The only chance I know to avoid collisions is to compare the partition.path, which is not available at this point of time for the just created partition.  To get the path  one should check parted print or lsblk for a partition with the same characteristics as the one created and not existing previously. As unsafe for concurrency as our actual code, as collision prone, and probably a coding nightmare. 
Better a clearly documented warning as not to try to partition in parallel (which is simply common sense) than adding that much complexity

Torxed: Worst part is that I wrote that code, and apparently I forgot I did and I forgot the trick I used. But you've correctly identified the intended behavior of that pop and I thank you for explaining it to me, even tho I created it haha.

dylanmtaylor:
I'm not a fan of sleeps, or arbitrary loop counts to check if an operation is done yet, I'd prefer a blocking operation that returns when it finishes or fails. That said, I 1000% believe this is better than it not working at all on the first try and the user having to run the install again. +1, with hopes for future improvements

Torxed:
> I'm not a fan of sleeps, or arbitrary loop counts to check if an operation is done yet, I'd prefer a blocking operation that returns when it finishes or fails. That said, I 1000% believe this is better than it not working at all on the first try and the user having to run the install again. +1, with hopes for future improvements

I agree, and this is why I had the original loop. But damn, I do not have a better solution hehe. I don't get why the kernel releases the calls before the changes have been made lower down :)

dylanmtaylor:
Would calling 'sync' help? I assume we already do this after writing partitions

dylanmtaylor:
This also looks like it might be useful: https://linux.die.net/man/8/partprobe

Torxed:
> Would calling 'sync' help? I assume we already do this after writing partitions

Not entirely sure actually. But we do in places.

dylanmtaylor:
Might as well merge this until we find something better. It's desperately needed at this point.

dylanmtaylor:
I just tested a real hardware install with this PR and it worked beautifully.

Torxed:
Merged, great work and thank you for testing Dylan!

Issue #705, opened 2021-11-09
[2.3.0-RC1] On uninitialized disk (VirtualBox), the selected partition table format 2 does not match that of BlockDevice


dylanmtaylor:
![VirtualBox_arch_09_11_2021_10_32_13](https://user-images.githubusercontent.com/277927/140954414-a5c159af-74b5-4645-9b14-06712b163c21.png)

dylanmtaylor:
I got this error twice in a row.

Torxed:
I'm looking in to this error now, after running 20+ installations with no breakage surrounding `.format()` or `PARTUUID <partuuid> never showed up` I'm confident enough that it's good enough to merge the other code. And now I can begin chasing down this one.

Torxed:
It might seam silly, but I've checked through the code and simply removed this on entry check if `format == MBR` or `format == GPT`. We don't need that check any longer. Will run a few tests with the PR and make sure it doesn't introduce anything weird I didn't think about..

PR #711, opened 2021-11-11
Adding partprobe at strategic places

Torxed:
This should combat the kernel not knowing the new partitioning layout after successful `parted` calls as well as `umount` calls.
This should allow `lsblk` to properly show the correct information, perhaps `partprobe` would be appropriate to add right before any `lsblk` call as well, but not sure.

### DIFF
--- a/archinstall/lib/disk/filesystem.py
+++ b/archinstall/lib/disk/filesystem.py
@@ -165,6 +165,11 @@ class Filesystem:
 				new_uuid_set = (previous_partition_uuids ^ {partition.uuid for partition in self.blockdevice.partitions.values()})
 				if len(new_uuid_set) > 0:
 					new_uuid = new_uuid_set.pop()
+				print('Blockdevice:', self.blockdevice)
+				print('Partitions:', self.blockdevice.partitions)
+				print('Partition set:', new_uuid_set)
+				print('New UUID:', [new_uuid])
+				print('get_partition():', self.blockdevice.get_partition)
 				if new_uuid:
 					return self.blockdevice.get_partition(new_uuid)
 				else:
--- a/archinstall/lib/disk/filesystem.py
+++ b/archinstall/lib/disk/filesystem.py
@@ -21,7 +21,7 @@ class Filesystem:
 
 	def __enter__(self, *args, **kwargs):
 		# TODO: partition_table_type is hardcoded to GPT at the moment. This has to be changed.
-		elif self.mode == self.blockdevice.partition_table_type:
+		if self.mode == self.blockdevice.partition_table_type:
 			log(f'Kept partition format {self.mode} for {self.blockdevice}', level=logging.DEBUG)
 		else:
 			raise DiskError(f'The selected partition table format {self.mode} does not match that of {self.blockdevice}.')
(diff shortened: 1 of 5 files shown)


Torxed:
screenshot

It complains that the partition never showed up after the initial partition add. Might need to revert some old code or change the current.

Torxed:
Only some times triggered:
screenshot

I think we've found the source of the issue at least, with the whole `.format()` issue before.

Torxed:
I don't get how `blockdevice.partitions` here: https://github.com/archlinux/archinstall/blob/e3e62039f9c4cbbb6f604c838cd3cd8032150565/archinstall/lib/disk/filesystem.py#L175
Can contain the new UUID, but just four lines down the line when we try to get that partition: https://github.com/archlinux/archinstall/blob/e3e62039f9c4cbbb6f604c838cd3cd8032150565/archinstall/lib/disk/filesystem.py#L179
...Using the convenience function `get_partition(new_uuid)` it fails using the same data source `blockdevice.partitions`: https://github.com/archlinux/archinstall/blob/e3e62039f9c4cbbb6f604c838cd3cd8032150565/archinstall/lib/disk/blockdevice.py#L220-L221

I really don't think the meta `__iter__` can cause a weird issue..

Torxed:
Only 1 out of  8 runs have failed. I can't seam to trigger this any longer in Qemu.

Torxed:
screenshot

Added more debug output to `get_mount_info()` because the above error.

archinstall/lib/disk/filesystem.py
```diff
@@ -20,24 +20,8 @@ def __init__(self, blockdevice, mode):
 		self.mode = mode
 
 	def __enter__(self, *args, **kwargs):
-		if self.blockdevice.keep_partitions is False:
```
dylanmtaylor, line 23: This looks like we no longer initialize the disk with a partition label in any case?

archinstall/lib/luks.py
```diff
@@ -61,6 +61,8 @@ def encrypt(self, partition, password=None, key_size=512, hash_type='sha512', it
 		with open(key_file, 'wb') as fh:
 			fh.write(password)
 
+		SysCommand(f'bash -c "partprobe"') # Might be redundant
```
dylanmtaylor, line 64: A method should probably be used to be consistent with the other source code.
```
self.partprobe()
```

dylanmtaylor:
Just tested in VirtualBox with default settings and a new drive. I ran into the crash mentioned in #705 on my first run. This is using MBR. I think I'm going to try to track this one down too as it's an almost certainty it fails on the first run on a fresh VM.
VirtualBox_arch_11_11_2021_11_18_08
.

Torxed:
@dylanmtaylor yes that part I haven't addressed yet at all. So that's expected. This issue more belongs to the fact that partitions didn't show up. I can include a fix for that too if you wish in this PR

archinstall/lib/disk/filesystem.py
Torxed: That is true, it's done in the `load_layout()` instead. To reduce the complexity and separation of duty. Because the `__enter__` was doing so many things *"for no reason"*.

archinstall/lib/luks.py
Torxed: I was considering it, but since I wasn't sure I'd be calling it in many other places I skipped it. But it would look cleaner, so I'll add it in the next commit.

PR #714, opened 2021-11-12
Removed the GPT vs MBR lookup on __enter__

### LINKED ISSUE #705: [2.3.0-RC1] On uninitialized disk (VirtualBox), the selected partition table format 2 does not match that of BlockDevice

Torxed:
it's no longer necessary to validate this on instance creation. load_layout() Uses this only to detect what partition table format it should use when wiping the drive. Other than that we only check if MBR and part numbers are > 3, that's the only use of this variable at this moment.

This will fix #705

### DIFF
--- a/archinstall/1
+++ b/archinstall/1
@@ -1,0 +1,4 @@
+grep: lib/__pycache__/disk.cpython-39.pyc: binary file matches
+grep: lib/__pycache__/installer.cpython-39.pyc: binary file matches
+grep: lib/__pycache__/user_interaction.cpython-39.pyc: binary file matches
+grep: lib/disk/__pycache__/helpers.cpython-39.pyc: binary file matches
--- a/archinstall/1
+++ b/archinstall/1
@@ -1,4 +1,0 @@
-grep: lib/__pycache__/disk.cpython-39.pyc: binary file matches
-grep: lib/__pycache__/installer.cpython-39.pyc: binary file matches
-grep: lib/__pycache__/user_interaction.cpython-39.pyc: binary file matches
-grep: lib/disk/__pycache__/helpers.cpython-39.pyc: binary file matches
--- a/archinstall/lib/disk/filesystem.py
+++ b/archinstall/lib/disk/filesystem.py
@@ -20,12 +20,6 @@ class Filesystem:
 		self.mode = mode
 
 	def __enter__(self, *args, **kwargs):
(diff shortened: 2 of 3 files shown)


dylanmtaylor:
I'm testing this now, if it works for me, I think we should merge and tag RC2

dylanmtaylor:
I downloaded this ISO to test. It definitely fixed the issue at the beginning of the install. I did run into this however after it started installing.
VirtualBox_arch_12_11_2021_10_54_33

dylanmtaylor:
Fails after GRUB install.
VirtualBox_arch_12_11_2021_12_05_09

Torxed:
@dylanmtaylor Both issues should be solved now.
It was a redundant lookup for the `boot_partition` *(came with a legacy fix for GRUB and encryption fix)* that was using the a function call incorrectly.
screenshot

Both with and without encryption (using GRUB) appears to be working.

dylanmtaylor:
> @dylanmtaylor Both issues should be solved now. It was a redundant lookup for the `boot_partition` _(came with a legacy fix for GRUB and encryption fix)_ that was using the a function call incorrectly. screenshot
> 
> Both with and without encryption (using GRUB) appears to be working.

Awesome! This is so exciting. Should we tag an RC2 now and think about targeting 2.3.0 by December?

Torxed:
> Awesome! This is so exciting. Should we tag an RC2 now and think about releasing 2.3.0 by December?

I've got one or two minor tweaks to merge/fix. But I should be able to tag RC2 before the weeks end.
After which I'll finish up any remaining empty code paths that are marked as #TODO and then it should be release party time :D

dylanmtaylor:
> > Awesome! This is so exciting. Should we tag an RC2 now and think about releasing 2.3.0 by December?
> 
> I've got one or two minor tweaks to merge/fix. But I should be able to tag RC2 before the weeks end. After which I'll finish up any remaining empty code paths that are marked as #TODO and then it should be release party time :D

I saw the one for GRUB encryption -- that has been a pain point for MBR users for ages. Personally, I use UEFI on everything but would love to see that fixed.

Torxed:
Yea that's the final big PR I'll merge in, and I think I got most of the small stuff sorted out today. So #724 will mark the RC2 publication.

PR #716, opened 2021-11-15
Add EFISTUB bootloader support

hexwell:
This PR adds the ability to use EFISTUB booting.

It has been tested locally.

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -663,9 +663,7 @@ class Installer:
 					log(f"Identifying root partition by PART-UUID on {root_partition}, looking for '{root_partition.uuid}'.", level=logging.DEBUG)
 					kernel_parameters.append(f'root=PARTUUID={root_partition.uuid} rw intel_pstate=no_hwp {" ".join(self.KERNEL_PARAMS)}')
 
-				boot_partition = find_partition_by_mountpoint(self.partitions, relative_mountpoint=f"/boot")
-
-				SysCommand(f'efibootmgr --disk {boot_partition[:-1]} --part {boot_partition[-1]} --create --label "{label}" --loader {loader} --unicode \'{" ".join(kernel_parameters)}\' --verbose')
+				SysCommand(f'efibootmgr --disk {boot_partition.path[:-1]} --part {boot_partition.path[-1]} --create --label "{label}" --loader {loader} --unicode \'{" ".join(kernel_parameters)}\' --verbose')
 
 			self.helper_flags['bootloader'] = bootloader
 		else:
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -648,7 +648,7 @@ class Installer:
 					elif vendor == "GenuineIntel":
 						kernel_parameters.append("initrd=\\intel-ucode.img")
 					else:
-						self.log("unknow cpu vendor, not adding ucode to systemd-boot config")
+						self.log("unknow cpu vendor, not adding ucode to firmware boot entry")
 
 				kernel_parameters.append(f"initrd=\\initramfs-{kernel}.img")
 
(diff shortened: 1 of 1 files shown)


archinstall/lib/installer.py
```diff
@@ -623,6 +623,49 @@ def add_bootloader(self, bootloader='systemd-bootctl'):
...
+
+				if not is_vm():
+					vendor = cpu_vendor()
+					if vendor == "AuthenticAMD":
+						kernel_parameters.append("initrd=\\amd-ucode.img")
+					elif vendor == "GenuineIntel":
+						kernel_parameters.append("initrd=\\intel-ucode.img")
+					else:
+						self.log("unknow cpu vendor, not adding ucode to systemd-boot config")
+
```
dylanmtaylor, line 652: This message references systemd-boot but this is efistub and doesn't use a bootloader

dylanmtaylor:
I'm fine with this as an advanced option you can configure via a configuration file or by using archinstall as a library, but don't think efistub should be in guided. Most users will benefit from having a bootloader. I'd validate this works if you select more than one kernel, e.g. linux-lts and linux.

dylanmtaylor:
In my opinion this should target 2.4.0 - don't want to add and test this with multiple permutations in this cycle as it'd delay 2.3.0

Torxed:
I'm also fine with this being in `--advanced` and allowing this to be part of API calls so that users can choose to script installations with this as an option.

But there will need to be some work done on `guided.py` if we're allowing it under `--advanced`, because that would mean we'd have to swap from a simple `y/N` question on GRUB to a select-one menu instead. Doable tho.

Adding this to the v2.4.0 plan, as we're too close to v2.3.0 being released with enough changes as it is : )

dylanmtaylor:
systemd-boot is already super minimal - is there enough benefit to EFISTUB to make it a mainstream option and switch guided to a multi-selection instead of a simple y/n?

Torxed:
> systemd-boot is already super minimal - is there enough benefit to EFISTUB to make it a mainstream option and switch guided to a multi-selection instead of a simple y/n?

I think guided is good enough as it is. But I'm willing *(if enough people convince me)* to allow a `ask_bootloader()` function to contain `y/N` without `--advanced` and a multi-select if `--advanced` is used *(as an example)*.

But if there's no traction for it, I'm fine leaving it at the API call : ) That's probably enough for those advanced used anyway who's willing to script and/or use a configuration file.

archinstall/lib/installer.py
hexwell: Good catch, changed in 436cabb8d057d88f97fc7f51ba1d5abd4a6b7fd7

dylanmtaylor:
Can you test and show that on real hardware if you run an install with this option with more than one kernel selected there will be multiple firmware boot entries and you can, for instance, choose between arch linux and arch linux-lts?

hexwell:
I ran an installation with:

```json
{
  "bootloader": "efistub",
  "kernels": [
    "linux",
    "linux-lts"
  ]
}
```

This is the result on my ASUS motherboard:

IMG_20211117_112009

I can confirm that both entries work.
I will run an additional test in VM later.

Torxed:
> I ran an installation with:
> 
> ```json
> {
>   "bootloader": "efistub",
>   "kernels": [
>     "linux",
>     "linux-lts"
>   ]
> }
> ```
> 
> This is the result on my ASUS motherboard:
> 
> IMG_20211117_112009
> 
> I can confirm that both entries work. I will run an additional test in VM later.

Nice, thank you for testing this.
Will it work with disk encryption?

hexwell:
> Will it work with disk encryption?

Yes, I have tested it.
The efistub firmware entry will have any kernel parameters that systemd-boot entries would have.

hexwell:
Test in VM (HyperV) succeeded as well.

Capture

Torxed:
Merged as there's no added complexity or addition to the overall installation workflow.
Guided and base platform is unaffected but adds API/config features for more advanced users.

dylanmtaylor:
This is excellent! Glad we were able to get this feature in. I don't see any potential issues this will cause, so agreed with all Torxed just said.

hexwell:
Thanks to both, keep up the amazing work!

PR #719, opened 2021-11-16
Add Noto fonts

atirut-w:
As a Linux user who visit Thai sites, I find the Noto fonts essential since some text in certain languages(notably Thai and Japanese) won't render without these.

### DIFF
--- a/profiles/desktop.py
+++ b/profiles/desktop.py
@@ -19,6 +19,9 @@ __packages__ = [
 	'wpa_supplicant',
 	'smartmontools',
 	'xdg-utils',
+	'noto-fonts-cjk',
+	'noto-fonts-emoji',
+	'noto-fonts',
 ]
 
 __supported__ = [


bebetoalves:
Excellent! Every time I had to manually install these fonts to fix the rendering 😅

dylanmtaylor:
-1 because not everyone will want this. I feel it could be seen as bloat.

Torxed:
In a good world, we could detect the keyboard layout and/or region and do some guessing work to see if these are needed.
Either we "bloat" the default package list for most parts of the world where UTF-8 and standard fonts is more than enough, or we add another question during desktop setup if the user wish to enable multiple fonts or something similar *(phrasing could be discussed)*.

I'm torn.

dylanmtaylor:
> In a good world, we could detect the keyboard layout and/or region and do some guessing work to see if these are needed. Either we "bloat" the default package list for most parts of the world where UTF-8 and standard fonts is more than enough, or we add another question during desktop setup if the user wish to enable multiple fonts or something similar _(phrasing could be discussed)_.
> 
> I'm torn.

I definitely am strongly against more questions, especially fonts. Once the user gets a base system up they can install any fonts they want. And users may want fonts other than noto. This is a rabbit hole it's not worth going down. This falls into the category of configuring an installed system, IMO. This is like saying almost everyone on kde will want archive management and screenshots so lets add ark and spectacle, for instance. Or almost everyone will need git for the AUR, so lets add that. Etc. I think our aim should be a minimal install.

Torxed:
> I definitely am strongly against more questions, especially fonts. Once the user gets a base system up they can install any fonts they want. And users may want fonts other than noto. This is a rabbit hole it's not worth going down. This falls into the category of configuring an installed system, IMO. This is like saying almost everyone on kde will want archive management and screenshots so lets add ark and spectacle, for instance. Or almost everyone will need git for the AUR, so lets add that (personally the first thing I do on an Arch system is install paru). Or maybe even 'everyone on real hardware will want to be able to do firmware updates, lets install fwupd'. Etc. I think our aim should be a minimal install.

I'm willing to agree, it's a pesky thing to have to do every install *(I too have some personal things, mainly `yay` and certain basic Python libs)*. But we all have those things and we can't cater to *all* of them. The guided process would contain more questions than the manual installation instructions at some point.

I'm not against these types of features, but they probably have to cater to "the majority" of what people do with their systems and can't contain too much flavor in opinions *(to some extent that's fine tho)*. So I'll go ahead and close this to not leave to many PR's and issues open.

*Today I learned about `fwupd` hehe.*

atirut-w:
> The guided process would contain more questions than the manual installation instructions at some point.

Would adding a warning about fonts be more viable?

Torxed:
> > The guided process would contain more questions than the manual installation instructions at some point.
> 
> Would adding a warning about fonts be more viable?

I don't think it's a breaking change to the extent where it requires a warning.
You usually just notice it in the browser and add the required font if and when. Which ties back to the individual issues.

And friendly reminders like this tend to get drowned in the output unless it's in the very end, and we should probably reserve the attention span and console space for crucial pieces of information.

I feel like a Grinch here, but I think if anything we should solve this language/regional issues in a broader aspect.
I know we have a lot of users from Asian origins, and we've had issues with keyboard locales etc, and we should tackle the whole conceptual issue rather than individual pieces.

PR #724, opened 2021-11-16
Fixing broken encryption support in GRUB

### LINKED ISSUE #586: Encryption does not seem to work with GRUB.

Torxed:
This will fix #586

* Adding in partprobe and sleep on retrieving PARTUUID on partitions.

### DIFF
--- a/archinstall/lib/disk/filesystem.py
+++ b/archinstall/lib/disk/filesystem.py
@@ -33,12 +33,18 @@ class Filesystem:
 		return True
 
 	def partuuid_to_index(self, uuid):
-		output = json.loads(SysCommand(f"lsblk --json -o+PARTUUID {self.blockdevice.device}").decode('UTF-8'))
+		for i in range(3):
+			self.partprobe()
+			output = json.loads(SysCommand(f"lsblk --json -o+PARTUUID {self.blockdevice.device}").decode('UTF-8'))
+			
+			for device in output['blockdevices']:
+				for index, partition in enumerate(device['children']):
+					if partition.get('partuuid', '').lower() == uuid:
+						return index
 
-		for device in output['blockdevices']:
-			for index, partition in enumerate(device['children']):
-				if partition['partuuid'].lower() == uuid:
-					return index
+			time.sleep(1)
+
+		raise DiskError(f"Failed to convert PARTUUID to archinstall partition index on {self.blockdevice.device}")
 
 	def load_layout(self, layout :dict):
 		from ..luks import luks2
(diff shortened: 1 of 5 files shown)


Torxed:
screenshot

Issue still remains, but installation can complete.

Torxed:
I believe it's fixed now. I'll test it in the morning.
But the start position of the `/boot` partition was too early on the disk, not allowing for enough room on BIOS+MBR systems to allow for the GRUB bootloader to be injected into the early blocks of the drive.

I've now moved the `/boot` partition default start position to 5MiB instead of 1MiB.
And I've added `exit_code` checks to the `grub-install` and `grub-mkconfig` calls so that we can catch these errors.

The following was the issue going undetected because we didn't catch bad exit codes:
image

*(Again, not 100% sure it's solved, but should be on the right track at least)*

dylanmtaylor:
I just tested the ISO generated by commit 2d7c5aa6f57fc4abb3eb5280d2211705286d32b4. It did not install successfully. 
VirtualBox_Arch_16_11_2021_20_04_27
VirtualBox_Arch_16_11_2021_20_04_45

Torxed:
Good, that's the ghost issues we had before that all of a sudden stopped coming up. And it looks like it's because it's trying to use the disk and not a partition when looking up partition order :)

Torxed:
For whatever reason, `device_instance` here points towards the wrong device: https://github.com/archlinux/archinstall/blob/a18a652bb0557c0dae7c5706893efd3a217c6598/archinstall/lib/disk/filesystem.py#L107-L108

In your first screenshot @dylanmtaylor, the partition *(for boot)* is `/dev/sda` and not `/dev/sda1`.
I get that MBR and BIOS setups don't really require a boot partition, so we should *probably* remove the need for it on that setup. But for code maintainability and stability I won't go there right now. But that still leaves the issue of `/dev/sda` being the device instance, instead of `/dev/sda1`.

Torxed:
screenshot

`["device_instance"].uuid` returns `None`. One piece of the puzzle.

Torxed:
screenshot

Works with encryption now.
I'm pretty sure most issues are resolved, I consider this ready for merging.
If anyone wants to test this and see if they can reproduce issues I'll be waiting until later tonight to merge this.

dylanmtaylor:
> screenshot
> 
> Works with encryption now. I'm pretty sure most issues are resolved, I consider this ready for merging. If anyone wants to test this and see if they can reproduce issues I'll be waiting until later tonight to merge this.

Go ahed and squash merge now, it offers considerable improvements already. I'll test it tonight.

Torxed:
Will do. There's a merge conflict that I need to fix in installer. Will do that when I'm back at a computer :)

PR #725, opened 2021-11-17
Parm mount

wllacer:
A less error prone of PR 715 (creation of mount-point parameter). I had some silly errors a times annoying to debug,  and wanted to make it more foolproof for my own benefit
I added it explicitly to the list to accept both the --parm value and --parm=value syntax
I added a synomym (accepting the name both with underscore and dash) and ignoring when no value specified

DOES NOT check the contents of the parameter

On my own accord I reordered a couple of the original lines, so all the defined parameters are processed together and then the unkown arguments are parsed. I hope it is just an esthetical change

Feel free to ignore the patch ... it is just suited to my taste, although i think it could be useful to others. Didn't I told you arguments are "black magic"? ;-)

# Testing
now this ways of calling archlinux work:
* python -m archlinux   -> install at /mnt/archlinux
* python -m archlinux --mount-point  -> idm.
* python -m archlinux --mount_point -> idm
* python -m archlinux --mount-point /mnt   -> installs at /mnt
* python -m archlinux --mount-point=/mnt  -> idm
* python -m archlinux --mount_point /mnt  -> idm
* python -m archlinux --mount_point=/mnt  -> idm


Torxed:
I think this is a good way forward, but this PR is a bit incomplete.
If we're going to do this, I'd like to do it for all options not just a single one.

So we should add mux alternatives for all options.
If you could update this PR with the same idea for all options, and re-use as much code as possible so we avoid too many `if` and `else` statements, that'd be awesome!

wllacer:
I was tempted ... i will try to include all of them. What do you mean with mux?
I will keep the unknow ones. It's an easy way to test new ones before they're set in stone

Torxed:
> I was tempted ... i will try to include all of them. What do you mean with mux? I will keep the unknow ones. It's an easy way to test new ones before they're set in stone

By mux i mean we allow many, but consolidate them into one single option so the code can keep using one of the "alternatives" for the same option. `--ex_ample` & `--ex-ample` both lead to `ex-ample` for instance. That way the users can input whatever they want - but we'll keep using one semantic.

wllacer:
"Muxing" is the easiest part. For the time being there is only one (dry-run) left. Already in my local branch.
But i will keep for now the case.
I'd need a suitable help text for the plugin parameter.
And now that we are changing disk, do you want that the unknown parameters behave as of now ( --key without a value is boolean and if not it must be explicitly --key=value) or to behave like known parameters ( --key value equals --key=value and --key w/o a value is a boolean) ? I think i did write once code to make it to behave that way
My schedule is a bit tight this days. but i hope to have an enhanced version Real Soon Now

Torxed:
Some good questions :D
 Let me sleep on it for a night and I'll see what I can think of.

wllacer:
@Torxed  Are you sure the parameter --debug still does anything, beyond writing the initial log entry ? I could find no reference to it inside the code ?

Torxed:
It activates command execution output to the terminal.

wllacer:
I've sent an updated version of the code with most of the changes we talked about
I added some internal documentation for the next time we have to fiddle with it. 
Flake8 is driving me crazy   .... How do you make a decent documentation header for a function ?

wllacer:
This is a bit of off-topic.
While getting around the parameters there are two which read json files (--config and --creds) but archinstall produces three of them (user_configuration, user_credentials and user_disk_layout). The last one is lacking a parameter, and i can not find where it is read. Shouldn't it have an argument and be read with the profile ?

wllacer:
DON'T MERGE. Wait a little.
I found the missing argument (--disk_layouts), but i have  almost no time today to handle it. Do you know if there are any more undocumented arguments ?
I will make also --dry_run and --silent incompatible. And silent will take precedence if dry_run specified in the config file (as it happens now)
The format of the config json file changes a little (includes values for all the explicit parameters and those not standard). It should be compatible (back and forth), but someone should check if there is any parameters that mustn't appear in the config file

Torxed:
> DON'T MERGE. Wait a little. I found the missing argument (--disk_layouts), but i have almost no time today to handle it. Do you know if there are any more undocumented arguments ? I will make also --dry_run and --silent incompatible. And silent will take precedence if dry_run specified in the config file (as it happens now) The format of the config json file changes a little (includes values for all the explicit parameters and those not standard). It should be compatible (back and forth), but someone should check if there is any parameters that mustn't appear in the config file

I'll wait. I'll extract a complete list of arguments. Will need it for documentation anyway ^^

wllacer:
Merge can proceed if deemed worthy
I added the --disk_layouts argument and centralized its handling at __init__.py.
The fiasco at issue #743 made me create a function to handle the reading of the JSON files and to gracefully manage the errors (no more exceptions but manual handling). Failure to load one of the files is deemed  important enough to abort the execution of archinstall.
Till new arguments appear, only one thing is left for me. I've noticed that all the config files I create do include the the dry_run argument to true (if I run that way), which gets propagated to succesive executions (same would happen with --silent).
 I'm not so sure it is a misfeature I introduced or it was there before and if it should be "sanitized" (ignored in succesive runs). I'm totally for it. @Torxed  is up to you to decide

Torxed:
The most important thing is to keep the generic argument handling.
As I and others use `storage['arguments']` with generic `--my-arg="moo"` as `storage['arguments']['my-arg']` in side-applications that relies on archinstall. Hence the weird manual argument parsing from `sys.argv`.
But it makes sense to have the guided or very least archinstall required arguments as fixed arguments.

I'll have a proper look on this before I merge, as it needs some testing.

wllacer:
This behaviour you need is/should be preserved, it's only the assignment operator which isn't needed anymore 
AFAIK the only  other behavioral changes I've introduced (barring some unknown bug :-() are
* A more user  friendly approach to errors during config file loading (gracefully exiting the aplication)
* While largely unneded, the json files can be replaced with json formatted strings. This is a carry over from the original disk_layouts processing,
* Now, every predefined argument is stored in the storage['arguments'] structure, even if not set. Formerly, only those explicitly set  went into. This might affect PR #727 , but it is (should be) otherwise neutral and version compatible

Torxed:
`v2.3.0` is out, whenever you feel ready I will begin testing on this PR and merge it in.

wllacer:
I did upgrade It this morning to avoid conflicts with the new release.
I think it's ready for testigo and reviewing

wllacer:
I've added a new algorithm to parse non defined arguments. It allows some more flexibility if needed

Torxed:
> I've added a new algorithm to parse non defined arguments. It allows some more flexibility if needed

I'll test this during the day and merge if I find no issues :) Thank you for merging in master!

wllacer:
OK. I will upload an update in a couple of minutes. For once flake8 caught an interesting error ... somewhere at Installler.mount_ordered_layout there is a *!= None*, which of course should be *is not None*   (came with the last master update):
If you like the changes, have a look at PR #810. I believe It handles a problem  I did have (and i think i resolved). OTOH my PR "destroys" the code its based on

Torxed:
Finally got around to testing some exotic parameters.
And it appears to be working just fine : ) Great work on this one, I barely recognize my own magical parser anymore hehe, which is a good thing I think :D The most important thing was for subsequent profiles to be able to have their own "magical arguments" transparently passed through into `archinstall.arguments` which seems to be working fine. For instance https://github.com/Torxed/archoffline which uses the archinstall builtin argument parser to do the work, but has its own arguments :)

PR #727, opened 2021-11-18
Created config schema

OhKannaDuh:
This PR introduces a JSON schema for creating config files against. The main goal here is to make it easier for people to write config files by defining the schema of those files. I'm not 100% that this belongs in this repo but it seems like a nice thing to have and I thought I'd just put it out there.

Will this be useful for others? I'm not sure, but I found it useful when I was messing around with `archinstall`, and testing different set ups. It sped up that process.

# Testing
### VSCode
- Register the schema and associate it with a file (`settings.json`):
```
"json.schemas": [
    {
        "fileMatch": [
            "*.archinstall.json",
        ],
        "url": "https://raw.githubusercontent.com/OhKannaDuh/archinstall/config-json-schema/schema.json"
    }
]
```
- Create a file that matches the pattern (`config.archinstall.json`)
- Start editing the file and bask in auto complete goodness

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -1,0 +1,215 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "title": "Archinstall Config",
+    "description": "A schema for the archinstall command config",
+    "type": "object",
+    "properties": {
+        "audio": {
+            "description": "Audioserver to be installed",
+            "type": "string",
+            "enum": [
+                "pipewire",
+                "pulseaudio"
+            ]
+        },
+        "bootloader": {
+            "description": "Bootloader  to be installed",
+            "type": "string",
+            "enum": [
+                "systemd-bootctl",
+                "grub-install"
+            ]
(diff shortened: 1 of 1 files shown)


Torxed:
Good idea, I think `filesystem` is not needed in the new `master` version.
And disk related things have moved out of the general config and in to `user_disk_configuration.json`.
Might be that it's still in the main config and we forgot to move it out, but even then it's optional as formatting is optional.

OhKannaDuh:
I'll have to put aside some time to read through the code and see what I can get about the schema from that. The rest will have to come from people more familiar with the program. I'll see what I can do this weekend anyway.

Torxed:
> I'll have to put aside some time to read through the code and see what I can get about the schema from that. The rest will have to come from people more familiar with the program. I'll see what I can do this weekend anyway.

When I get a moment over I can contribute to your PR, but it's on my back burner for the release.

PR #731, opened 2021-11-18
Added a log message for when xorg fails a certain step.

### LINKED ISSUE #726: Nvidia proprietary driver installation
Hello 

I tried to install my system via this installer but i can't install nvidia proprietary driver with this installer.

I think there is an option "Nvidia (proprietary)" but it still installing open source nvidia driver.
https://github.com/archlinux/archinstall/blob/0071a069080732047e11309134869f3ab40c642c/archinstall/lib/hardware.py

How can install proprietary drivers ?
Instead of gfx_driver option adding nvidia package names to packages option like that :
```"packages": ["bash-completion","linux-headers","nvidia","nvidia-settings","opencl-nvidia"]```

My config file :

```
{
  "bootloader": "systemd-bootctl",
  "audio": "pulseaudio",
  "filesystem": "ext4",
  "gfx_driver": "Nvidia (proprietary)",
  "harddrive": {
   "path": "/dev/sda"
  },
  "hostname": "pc",
  "kernels": ["linux"],
  "keyboard-language": "trq",
  "mirror-region": {},
  "nic": {
    "NetworkManager": true
  },
  "profile": "kde",
  "ntp": true,
  "packages": ["bash-completion", "linux-headers"],
  "sys-encoding": "utf-8",
  "timezone": "UTC",
  "!root-password": "****",
  "superusers": {
    "kaan": {
      "!password": "****"
      }
  }
}
```

Thank you.

Torxed:
This should fix #726

### DIFF
--- a/profiles/xorg.py
+++ b/profiles/xorg.py
@@ -1,6 +1,7 @@
 # A system with "xorg" installed
 
 import archinstall
+import logging
 
 is_top_level_profile = True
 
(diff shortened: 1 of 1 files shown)


Torxed:
I don't get how flake8 haven't detected this earlier, in one of many commits the last few weeks.
I haven't modified this source but the f-strings have been causing this for a while I think. Worth checking if there's other spots as well that flake8 doesn't catch until we touch the rows.

PR #733, opened 2021-11-18
Add `gnome-software-packagekit-plugin`

atirut-w:
This allows GNOME Software to work out of the box. I don't know why it's not a dependency, but GNOME Software is kinda borked without it.

### DIFF
--- a/profiles/gnome.py
+++ b/profiles/gnome.py
@@ -9,6 +9,7 @@ __packages__ = [
 	"gnome",
 	"gnome-tweaks",
 	"gdm",
+	"gnome-software-packagekit-plugin",
 ]
 
 


Torxed:
My lack of knowledge surrounding GNOME here, but what exactly isn't working without this?

dylanmtaylor:
> My lack of knowledge surrounding GNOME here, but what exactly isn't working without this?

GNOME Software is an application that allows you to install packages, kind of like the Ubuntu Software Center, the Apple App Store, KDE Discover, Google Play, etc. Without this, you can install packages via Terminal or use something like pamac.

Torxed:
> > My lack of knowledge surrounding GNOME here, but what exactly isn't working without this?
> 
> GNOME Software is an application that allows you to install packages, kind of like the Ubuntu Software Center, the Apple App Store, KDE Discover, Google Play, etc. Without this, you can install packages via Terminal or use something like pamac.

And this "app store" works with Arch packages natively? So there's no downside of adding this PR to make it work?

dylanmtaylor:
> > > My lack of knowledge surrounding GNOME here, but what exactly isn't working without this?
> > 
> > 
> > GNOME Software is an application that allows you to install packages, kind of like the Ubuntu Software Center, the Apple App Store, KDE Discover, Google Play, etc. Without this, you can install packages via Terminal or use something like pamac.
> 
> And this "app store" works with Arch packages natively? So there's no downside of adding this PR to make it work?

I think it needs packagekit. I personally don't use it but can appreciate that it's something a lot of users would want.

dylanmtaylor:
Packagekit basically abstracts implementation details of the OS's package manager, so yum/dnf, apt, zypper, etc. doesn't matter - it gives the application a way to say "Please just install firefox in a native way"

Torxed:
I can appreciate that users would want this to work too.
 It adds an abstraction layer that *some* people in the community might object, because it further distance the users from "how packages are installed".

But seeing as we cater to new users who want to ease their way into Arch, I'll allow this : )

PR #737, opened 2021-11-18
Adding a cmd_history.txt log under /var/log/archinstall/

Torxed:
This will create a more clear picture of which commands was executed.


Torxed:
This file will be somewhat sensitive, as it may contain passwords and should *NEVER* be posted without careful thought.
But it will greatly help with debugging commands and workflows.

Issue #527, opened 2021-05-28
How to dual boot with windows 10 in same disk.


nazar2sfive:
main disk = /dev/sda

/dev/sda1 = efi
/dev/sda2 = windows 10
/dev/sda3 = free space allocated for arch linux

how to use guided installer to choose /dev/sda3?

Torxed:
Good question, this is one of the remaining issues that I'm trying to work out in #446.
Currently there's two issues:

1. Windows only allocates ~100MB forthe `/boot` partition, in your case `/dev/sda1`. Arch Linux default initramfs is roughly 80MB or so and `mkinitcpio` intalls two initframfs, meaning there's not enough space.
2. archinstall fails to detect free space currently, so it doesn't have a way to create a partition from empty space (oddly enough I forgot this scenario in testing last version).

**Workaround:** Manually create the last partition:
```
fdisk /dev/sda
n
p
<leave rest default>
```
And once that partition is created, add any supported filesystem to it so archinstall allows you to tinker with it *(also a limitation for safety reasons)*:
```
mkfs.ext4 /dev/sda4
```
Once that's done, in archinstall you should be able to use the option *"Re-use partitions"* and select `/boot` for `/dev/sda1` and `/` for `/dev/sda5` and it should allow you to install and dual-boot. It might throw a warning that everything didn't go according to plan, but if that happens just run:
```
arch-chroot /mnt bootctl --path=/boot install
arch-chroot /mnt passwd root
```
And you should have a minimal setup at least.
This workaround won't install profiles and such, but at least give you a working system.

Once #426 is done, this should work much more smoothly, but I need to solve both those problems in order to get dual-boot working, and one issue is slightly out of our control - but we'll try to solve it some how. Resizing partitions is a bit scary so don't know if we can go down that route reliably.

mkromar:
I agree that resizing partitions is a bad idea.

A solution for a small boot partition could be to default to grub, mount ESP to /efi and leave /boot in the rootfs.

Torxed:
> I agree that resizing partitions is a bad idea.
> 
> A solution for a small ESP could be to default to grub, mount ESP to /efi and leave /boot in the rootfs.

This will introduce "unnecessary" complex logic but would solve the problem. *(unnecessary is probably the wrong word here, but couldn't come up with any other phrasing)*.
[...]

PR #738, opened 2021-11-18
Adding a warning when /boot is reasonably too small

Torxed:
Moved convert_size_to_gb() into disk/helpers.py, Added a Partition().size property meta function. Using the .size value to check if /boot is too small which will raise an exception. The only drawback is that it's done post-formatting. This in order to catch scenarios where formatting isn't used.

This will address #527

### DIFF
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -268,7 +268,7 @@ def perform_installation(mountpoint):
 		for partition in installation.partitions:
 			if partition.mountpoint == installation.target + '/boot':
 				if partition.size <= 0.25: # in GB
-					raise archinstall.DiskError(f"The /boot partition in use is not large enough to install boot loaders on properly. Please resize it using a tool like gparted and re-run the installation.")
+					raise archinstall.DiskError(f"The selected /boot partition in use is not large enough to properly install a boot loader. Please resize it to at least 256MB and re-run the installation.")
 
 		# if len(mirrors):
 		# Certain services might be running that affects the system during installation.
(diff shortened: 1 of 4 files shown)


examples/guided.py
```diff
@@ -264,6 +264,12 @@ def perform_installation(mountpoint):
 		# This *can* be done outside of the installation, but the installer can deal with it.
 		installation.mount_ordered_layout(archinstall.storage['disk_layouts'])
 
+		# Placing /boot check during installation because this will catch both re-use and wipe scenarios.
+		for partition in installation.partitions:
+			if partition.mountpoint == installation.target + '/boot':
+				if partition.size <= 0.15: # in GB
```
dylanmtaylor, line 270: 150MB is kind of small for boot. I'd probably do .25 here.

Torxed: Fixed! : )

dylanmtaylor (approved):
Looks good to me!

examples/guided.py
```diff
@@ -264,6 +264,12 @@ def perform_installation(mountpoint):
 		# This *can* be done outside of the installation, but the installer can deal with it.
 		installation.mount_ordered_layout(archinstall.storage['disk_layouts'])
 
+		# Placing /boot check during installation because this will catch both re-use and wipe scenarios.
+		for partition in installation.partitions:
+			if partition.mountpoint == installation.target + '/boot':
+				if partition.size <= 0.25: # in GB
+					raise archinstall.DiskError(f"The /boot partition in use is not large enough to install boot loaders on properly. Please resize it using a tool like gparted and re-run the installation.")
```
dylanmtaylor, line 271: Better wording: 

The /boot partition currently in use is not large enough to properly install a bootloader. Please resize it to at least 256MB and re-run the installation.

dylanmtaylor: Another thought, maybe 'currently in use' -> 'selected' would make more sense.

Torxed:
Changed the wording and merged!

PR #748, opened 2021-11-21
NTFS Root Filesystem Support

dylanmtaylor:
VirtualBox_Arch_21_11_2021_12_00_27

VirtualBox_Arch_21_11_2021_14_45_14

This is just for fun (and not really practical), but inspired by 
https://www.reddit.com/r/archlinux/comments/qwsftq/arch_linux_on_ntfs3/, I decided to see if it could be done. 

This probably shouldn't be used by anyone as anything other than as a toy.

Tested working with/without encryption on both systemd-boot (UEFI) and GRUB (BIOS) installations using VirtualBox. 

Thanks to Paragon for their excellent work on the new NTFS3 driver.

### DIFF
--- a/archinstall/lib/disk/partition.py
+++ b/archinstall/lib/disk/partition.py
@@ -341,18 +341,15 @@ class Partition:
 					raise DiskError(f'Need to format (or define) the filesystem on {self} before mounting.')
 				fs = self.filesystem
 
-			if fs == 'ntfs':
-				fs = 'ntfs3'  # Needed to use the Paragon R/W NTFS driver
-			elif fs == 'fat32':
-				fs = 'vfat'  # This is the actual type used for fat32 mounting.
+			fs_type = self.get_mount_fs_type(fs)
 
 			pathlib.Path(target).mkdir(parents=True, exist_ok=True)
 
 			try:
 				if options:
-					mnt_handle = SysCommand(f"/usr/bin/mount -t {fs} -o {options} {self.path} {target}")
+					mnt_handle = SysCommand(f"/usr/bin/mount -t {fs_type} -o {options} {self.path} {target}")
 				else:
-					mnt_handle = SysCommand(f"/usr/bin/mount -t {fs} {self.path} {target}")
+					mnt_handle = SysCommand(f"/usr/bin/mount -t {fs_type} {self.path} {target}")
 
 				# TODO: Should be redundant to check for exit_code
 				if mnt_handle.exit_code != 0:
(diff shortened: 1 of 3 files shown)


archinstall/lib/disk/partition.py
```diff
@@ -301,6 +302,13 @@ def format(self, filesystem=None, path=None, log_formatting=True, options=[]):
 				raise DiskError(f"Could not format {path} with {filesystem} because: {handle.decode('UTF-8')}")
 			self.filesystem = filesystem
 
+		elif filesystem == 'ntfs':
+			options = ['-f'] + options
+
+			if (handle := SysCommand(f"/usr/bin/mkfs.ntfs {' '.join(options)} {path}")).exit_code != 0:
```
Torxed, line 308: I guess we'd want a `-Q` option here as well for quick?

dylanmtaylor: That might be worthwhile if it speeds up partitioning, I didn't think that it was taking too long when I was doing it in my virtual machine, but it was only a 35 GB volume.

dylanmtaylor:
I've tested with the latest changes, and surprisingly it works pretty well, but I will personally stick with btrfs. :-)

Torxed:
Merging, if we decide to hide it for the general users we can filter it out from the available fstype list when running guided. But if it works as well as it does in testing, then I don't see the harm. This would allow users to more seamlessly dualboot with Windows and share files across both installations.

dylanmtaylor:
> Merging, if we decide to hide it for the general users we can filter it out from the available fstype list when running guided. But if it works as well as it does in testing, then I don't see the harm. This would allow users to more seamlessly dualboot with Windows and share files across both installations.

The biggest issue by far is the lack of fsck support, so if the filesystem becomes damaged in any way, you can't currently repair it from Linux like the others. I can't say this will never be added, but because of this, it can't recover from corruption nearly as well as other filesystems. It's also only been around since kernel 5.15, as the Paragon NTFS3 implementation is really new. This means that right now it won't boot with linux-lts, as that is < 5.15 but the others should work fine.

Torxed:
I will most likely hide this from the end users before release and leave it for "`--advanced`" stuff for now. We can wait until the -lts kernel gets the new feature.

dylanmtaylor:
> I will most likely hide this from the end users before release and leave it for "`--advanced`" stuff for now. We can wait until the -lts kernel gets the new feature.

I'm fine with that, and it's probably for the best. I didn't expect this to be something widely used, I just did it for fun to see how it worked after seeing NTFS used as a root partition on Reddit.

PR #752, opened 2021-11-22
Simplified size definition in dict. 

### LINKED ISSUE #742: Errors while partitioning disk
image
I'm using v2.3.0.RC2 version with Disk Encryption. I was getting the same thing while using master branch version before.

Torxed:
Also changed from MiB to MB and GiB to GB on places where they were used, as BlockDevice().size now returns GB by default, so no math operations needed

This will fix #742


Torxed:
screenshot

The start of `/home` should probably be `512MB`+`20GB` and not `20GB`.
But perhaps parted understands?

Torxed:
Merging, installation appears to be completing successfully.

screenshot

PR #753, opened 2021-11-22
Make NTFS an advanced option

dylanmtaylor:
As NTFS is probably not a great choice for a root filesystem, due to lack of fsck among other things, make it only appear if users use the `--advanced` flag.


dylanmtaylor:
Hmm, so without --advanced, it looks like this does NOT show NTFS, as intended, but with it I'm not even getting the prompt for a filesystem type. Leaving draft for now.

dylanmtaylor:
The latest commit fixed it, tested with and without `--advanced`, works as expected.

PR #756, opened 2021-11-23
Update filesystem.py

Firminator:
just a few typos
* label instead of labling
* make UUID consistent


dylanmtaylor (approved):
Looks like a good change, thanks!

PR #760, opened 2021-11-23
Add accessibility packages to installed system if using the on the live ISO

### LINKED ISSUE #758: Detect if espeakup.service is active in live medium and copy over
When `espeakup.service` is used on the ISO, assume that the user would want that copied over and do so.
I don't think we necessarily need to ask the question in the guided installer, as it would probably be enough to do the detection and act on it.

See discussion on the `argh-general` mailing list: https://lists.archlinux.org/pipermail/arch-general/2021-November/049405.html

dylanmtaylor:
Closes #758

### DIFF
--- a/archinstall/lib/disk/blockdevice.py
+++ b/archinstall/lib/disk/blockdevice.py
@@ -2,7 +2,6 @@ import os
 import json
 import logging
 import time
-from .helpers import convert_size_to_gb
 from ..exceptions import DiskError
 from ..output import log
 from ..general import SysCommand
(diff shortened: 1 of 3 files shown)


dylanmtaylor:
This needs tested still, but I think this is the right approach. Thoughts?

dylanmtaylor:
Hmm. ArchInstall isn't starting with this change. Hold on merging.
VirtualBox_Arch_23_11_2021_13_47_16

dylanmtaylor:
^ Reverting for testing until I figure out what's up with the circular dependency issue

dylanmtaylor:
With accessibility:
image
Without:
image


This can be merged in

Issue #764, opened 2021-11-24
Installer does not append 'btrfs-progs' into mkiniticpio


wllacer:
When installing over a btrfs filesystem btrfs-progs has to be included in binaries@mkinitcpio. Due to a mistype at Installer.py lnr 450, this is not done.
There is written
```
				if '/usr/bin/btrfs-progs' not in self.BINARIES:
					self.BINARIES.append('/usr/bin/btrfs')
```
Should be  *--Edited: the original solution was backwards*
```
				if '/usr/bin/btrfs' not in self.BINARIES:
					self.BINARIES.append('/usr/bin/btrfs')
```
I'll attach a PR

wllacer:
I have to resyncroniz my branch. Solution will wait a litte

dylanmtaylor:
> I have to resyncroniz my branch. Solution will wait a litte

This may ended up making it into the next release, Torxed is about to tag 2.3.0

Torxed:
> > I have to resyncroniz my branch. Solution will wait a litte
> 
> This may ended up making it into the next release, Torxed is about to tag 2.3.0

There's still time, I'm about 50% done with the writeup of the changelog.
So there's time : )

PR #765, opened 2021-11-24
Issue #764 Modify inclusion of self.BINARIES entry due to a mistype 

wllacer:
It's simply a oneliner

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -478,8 +478,8 @@ class Installer:
 			if partition.filesystem == 'btrfs':
 				if 'btrfs' not in self.MODULES:
 					self.MODULES.append('btrfs')
-				if '/usr/bin/btrfs-progs' not in self.BINARIES:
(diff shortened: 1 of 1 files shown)


dylanmtaylor:
@Torxed  do we have time to get this in before tagging?

dylanmtaylor (approved):

wllacer:
Hope it didn't came in ... the real solution is exactly the reverse. (not btrfs-progs, but btrfs is what we need to search for. I still can't boot in my test enviroment but comparing with the "real" one gave me the clue
I've made the update

Torxed:
Good catch, I don't know how any of the BTRFS installations have worked at all without this.. Luck?

wllacer:
No. Whatever is in the binaries is only really needed if the root FS is not available ... and then probably you are in the need of a recovery tool.
OTOH noone pays attention to the messages on mkinitcpio building ... They were too obvious

PR #773, opened 2021-11-26
Updated examples - fixes #772

### LINKED ISSUE #772: JSON configuration files
Hi. I use archinstall to provide automated installation on some servers, but with the new version, JSON has completely changed. Is there an expected date when https://github.com/archlinux/archinstall/blob/master/examples/config-sample.json and https://github.com/archlinux/archinstall/blob/master/examples/custom-command-sample.json will be updated?

Torxed:
Fix #772

### DIFF
--- a/examples/config-sample.json
+++ b/examples/config-sample.json
@@ -1,35 +1,29 @@
 {
-    "!root-password": "<root_password>",
     "audio": null,
     "bootloader": "systemd-bootctl",
-    "filesystem": "btrfs",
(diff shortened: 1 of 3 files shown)


kaan35:
I found an error in journal related with fstab. I run same command in OpenSuse, there is no same error.

`journalctl -p 4 -b`

`systemd-fstab-generator[229]: Mount point  is not a valid path, ignoring.`

What should i do ?

Thank you.

PR #777, opened 2021-11-29
XFCE profile: enabled service NetworkManager

rohandebsarkar:
Without enabling `NetworkManager` the `network-manager-applet` is not usable.

### DIFF
--- a/profiles/xfce4.py
+++ b/profiles/xfce4.py
@@ -43,4 +43,5 @@ if __name__ == 'xfce4':
 	# Install the XFCE4 packages
 	archinstall.storage['installation_session'].add_additional_packages(__packages__)
 
+	archinstall.storage['installation_session'].enable_service('NetworkManager') # NetworkManager
 	archinstall.storage['installation_session'].enable_service('lightdm')  # Light Display Manager


Torxed:
Depends, if the user opted in to use NetworkManager or not.
I personally don't like NetworkManager and would not like to see that service forced upon me.

We already do this in the guided profile, if you chose to use NetworkManager but if that isn't working we should create a Issue to track that problem and submit a PR to fix it.

Here's the relevant code, so I'll go ahead and close this: https://github.com/archlinux/archinstall/blob/22ee2d90a1d2fb6af2ab5e80235e31f9da7a0550/examples/guided.py#L301-L303

Torxed:
I might re-open this, so don't delete the branch just yet.
I just have to figure out if this is the right way to do it in that case.

dylanmtaylor:
> Network Manager is installed as a dependency of [`network-manager-applet`](https://archlinux.org/packages/extra/x86_64/network-manager-applet/).
> 
> I think since it is installed as a dependency and not chose manually, it is not enabled by `guided.py`.
> 
> After booting for the first time, the `network-manager-applet` is not usable. Therefore, for connecting to a Wi-Fi, one has to go to the terminal and enable it using `systemctl`.
> 
> Just wanted to bring that to your notice as, I thought, the `network-manager-applet` was included for ease of connecting to the internet using a GUI, since no GUI network managers are provided by default in XFCE as in other desktop environments.

Should we not install network-manager-applet` unless the user opts to use Network manager then?

Torxed:
> > Network Manager is installed as a dependency of [`network-manager-applet`](https://archlinux.org/packages/extra/x86_64/network-manager-applet/).
> > I think since it is installed as a dependency and not chose manually, it is not enabled by `guided.py`.
> > After booting for the first time, the `network-manager-applet` is not usable. Therefore, for connecting to a Wi-Fi, one has to go to the terminal and enable it using `systemctl`.
> > Just wanted to bring that to your notice as, I thought, the `network-manager-applet` was included for ease of connecting to the internet using a GUI, since no GUI network managers are provided by default in XFCE as in other desktop environments.
> 
> Should we not install network-manager-applet` unless the user opts to use Network manager then?

That sounds like a good approach! I think.. I'm not entirely sure what *"NetworkManager people"* *(misunderstand me correctly here)* want, but I would assume to get no graphical application or *"applet"* when opting out of NetworkManager, and I would suggest getting one when using NetworkManager.

To me installing the applet should be optional, and I'd like to not have it there when I pick systemd-networkd.
But I didn't think twice of it since it was such as small package and thought it would do no harm in adding it previously.

Issue #776, opened 2021-11-29
Small UI annoyance when using a prepartitioned layout


wllacer:
Using the guide script, when trying to execute over a preexisting disk layout (i.e. the target disk is parititoned and mounted BEFORE executing archinstall, we get following flow:
```
0: BlockDevice(/home/werner/vms/testimage.img, size=8.0GB, free_space=4177kB+1032kB, bus_type=None)
1: BlockDevice(/dev/sda, size=465.8GB, free_space=4177kB+7680B, bus_type=sata)
2: BlockDevice(None, size=1.0GB, free_space=, bus_type=sata)
Select one or more harddrives to use and configure (leave blank to skip this step):

0: Wipe all selected drives and use a best-effort default partition layout
1: Select what to do with each individual drive (followed by partition usage)
Select what you wish to do with the selected block devices:

Would you like to use GRUB as a bootloader instead of systemd-boot? [y/N]
```
As you will notice the question if i want to manipulate the disks still appears when it should not (The first time is one of this What the Heck moments)

Torxed:
Heh, yea that should be fixed.
We're re-working the menu system altogether in #660 and making it more polished. We'll make sure that these questions does not come up, unnecessary scare : )

wllacer:
The culprit is ln 111 at guided.py

https://github.com/archlinux/archinstall/blob/22ee2d90a1d2fb6af2ab5e80235e31f9da7a0550/examples/guided.py#L104-L113

Simply there is a check for **not None** when archinstall.arguments['harddrives'] holds an empty list.

I have a till now private change which besides solving this (and making the following code somehow more readable), has a bit of information for the user, and some "ass covering" (although i doubt a lawyer would agree). 
See the example 
```
0: BlockDevice(/home/werner/vms/testimage.img, size=8.0GB, free_space=4177kB+1032kB, bus_type=None)
1: BlockDevice(/dev/sda, size=465.8GB, free_space=4177kB+7680B, bus_type=sata)
2: BlockDevice(None, size=1.0GB, free_space=, bus_type=sata)
Select one or more harddrives to use and configure (leave blank to skip this step):

You decided to skip harddrive selection
and will use whatever drive-setup is mounted at /mnt/archinstall (experimental)
WARNING: Archinstall won't check the suitability of this setup
Do you wish to continue ? [Y/n]

Would you like to use GRUB as a bootloader instead of systemd-boot? [y/N]
```
When the user says no, i end the execution gracefully with an exit(1)
[...]

PR #778, opened 2021-11-29
better handling the skip partitioning option.

wllacer:
plus sending some warnings to the user
Solves the issue #776 (when installing on a prepartitioned disk, archinstall still ask for partitioning)
I've added some informational and warning messages. And the chance to get out


dylanmtaylor:
I'm not sure if we should go out of our way to reduce function complexity just to satisfy the linter - there should be a maintenance benefit as well. Not sure I like this change.

wllacer:
I'd rather revert the function split. It felt too artificial. 
Assuming you can pull it despite the check failure, the problem is that, probably from now on, Flake8 will most probably fail at any change on guided.py. 
Awaiting instructions

Torxed:
> I'd rather revert the function split. It felt too artificial. Assuming you can pull it despite the check failure, the problem is that, probably from now on, Flake8 will most probably fail at any change on guided.py. Awaiting instructions

I do have the power to overrule the linter hehe.
But we might have to add a exception in the `.flake8` file, I'd like to get all notifications from this *(mainly to go all in or not at all)*. But I do get flooded by notifications when the linter goes nuts and we leave things in to "ignore" the linter.

But I can fix the linting issues. I've started to kind of like the damn thing..

wllacer:
Done

wllacer:
I was still using the old call to get the disks

examples/guided.py
```diff
@@ -96,20 +96,27 @@ def ask_user_questions():
...
-	if archinstall.arguments['harddrives'] and archinstall.arguments.get('!encryption-password', None):
-		# If no partitions was marked as encrypted, but a password was supplied and we have some disks to format..
-		# Then we need to identify which partitions to encrypt. This will default to / (root).
-		if len(list(archinstall.encrypted_partitions(archinstall.storage['disk_layouts']))) == 0:
-			archinstall.storage['disk_layouts'] = archinstall.select_encrypted_partitions(archinstall.storage['disk_layouts'], archinstall.arguments['!encryption-password'])
+	# we skip the customary .get('harddrives',None) 'cause we are pretty certain that at this point it contains at least none (behaviour has changed from previous version, where it had an empty list. Shouls be compatible with my code
+	if not archinstall.arguments['harddrives']:
+		archinstall.log("You decided to skip harddrive selection",fg="red",level=logging.INFO)
+		archinstall.log(f"and will use whatever drive-setup is mounted at {archinstall.storage['MOUNT_POINT']} (experimental)",fg="red",level=logging.INFO)
+		archinstall.log("WARNING: Archinstall won't check the suitability of this setup",fg="red",level=logging.INFO)
```
Torxed, line 103: I'm about to merge this PR.
This might be me being picky, but could we change these three lines to have a `fg="yellow"` instead? Just to reserve red for errors and really really really crucial information? :)

Torxed (changes requested):

examples/guided.py
wllacer: Just done. I even think it shows better that way

wllacer:
As you are doing New Year's cleaning ... the changes you requested are done aeons ago. And this is probably my less problematic PR.
Happy New Year

Torxed:
> As you are doing New Year's cleaning ... the changes you requested are done aeons ago. And this is probably my less problematic PR. Happy New Year

It sure is the most straight forward, had to wrap my head around the new `if / else` logic and took a day hehe.
But I finally figured out what we used `harddrives` for and why this will work without the `if harddrives and not disk_layouts`.

Issue #781, opened 2021-12-02
Consider changing the BTRFS subvolume layout


KhalilSantana:
Hello,

I'm very happy to see that v2.3.X includes BTRFS subvolume support as a beta feature! This was a major blocker for me and now it's (mostly) gone.

However, I would like to comment on the current subvolume layout used by Archinstall's guided mode, and why (and how) I think it could be improved.

# The Current Layout

For reference this is the current layout used by guided mode as of 2021-12-1 using archinstall v2.3.0:
```
# btrfs subvol list -t /
ID	gen	top level	path	
--	---	---------	----	
256	21	5		home
258	21	5		var/log
259	17	5		var/cache/pacman/pkg
260	10	5		.snapshots
264	14	5		var/lib/portables
265	15	5		var/lib/machines
```

# How BTRFS handles subvolumes, and snapshots

(this section exists mostly to add context, skip it if you are familiar with BTRFS)

In BTRFS, each subvolume has an ID and a path, relative to the toplevel (a.k.a subvolid=5), these subvolumes can be mounted anywhere, and can be used for many things just like folders, however, the main differentiator between a regular folder and a subvol is that the later has special powers: atomic snapshots, etc.

Snapshots are just a clone of that filesystem root (subvol) at that time, and it shares extents with the snapped subvol, this is one of reasons snaps are so useful: they are fast and small.

However, there are some important quirks/behaviors that BTRFS currently does:

* Rule 0:  You cannot replace the toplevel (subvolid=5, subvol=/) with another subvol.
* Rule 1: `btrfs subvol snap` **won't recurse**  into nested subvolumes. So you cannot create atomic snaps of nested subvolumes and their parents at once.
* Rule 2: As `btrfs subvol snap` doesn't recurse, the resulting subvolume (the snapshot) won't contain ANY of it's nested subvols.

With this in mind, I hope the next section becomes easy to understand.

# What's wrong with the current layout?

The current layout basically prohibits using snapshots as a rollback feature, why?

* Per Rule 0, you cannot rollback (without lots of jiggling mv's and rm's) the root filesystem as it's been mounted at the toplevel subvol.
* Per Rule 1 and 2, even if you manually rollback using methods, like using the (often confusing) `btrfs sub set-default`, your homedir will be empty! You'd need to manually snap RW the old home subvol into the right place.

In short the main issues with the current subvol layout are:
* The root filesystem using the toplevel subvolume directly.
* The home dir0 being a nested subvolume.

[...]

Issue #718, opened 2021-11-15
[master] Crashes trying to use an already existing btrf layout. Does not find the root mount point


wllacer:
I have a somewhat peculiar btrfs layout (a flat one in the sense of `https://btrfs.wiki.kernel.org/index.php/SysadminGuide`). For reasons, i decided it was the one that suited me, but it is not set in stone
layout via findmnt
```

└─/mnt                              /dev/loop0p2[/root]
                                                       btrfs      rw,noatime,ssd,space_cache=v2,subvolid=256,subvol=/root
  ├─/mnt/boot                       /dev/loop0p1       vfat       rw,relatime,fmask=0022,dmask=0022,codepage=437,iocharset=ascii,shortname=mixed,utf8,errors=remount-ro
  ├─/mnt/home                       /dev/loop0p2[/home]
  │                                                    btrfs      rw,noatime,ssd,space_cache=v2,subvolid=257,subvol=/home
  ├─/mnt/opt                        /dev/loop0p2[/opt] btrfs      rw,noatime,ssd,space_cache=v2,subvolid=259,subvol=/opt
  ├─/mnt/srv                        /dev/loop0p2[/srv] btrfs      rw,noatime,ssd,space_cache=v2,subvolid=260,subvol=/srv
  └─/mnt/var                        /dev/loop0p2[/var] btrfs      rw,noatime,ssd,space_cache=v2,subvolid=261,subvol=/var
```
Note that the root dir is also a subvolume.
Using this structure i run archinstall NOT doing anything disk related (for the time being i'm content to have my own partitioning/mounting script). When the system ought to pacstrap the base directories, archinstall abends with the following messages
```
Getting mount information for device path /dev/loop0p2[/root]
Could not get PARTUUID for /dev/loop0p2[/root]: lsblk: /dev/loop0p2[/root]: no es un dispositivo de bloques
{
   "blockdevices": [

   ]
}

[!] A log file has been created here: /var/log/archinstall/install.log
    Please submit this issue (and file) to https://github.com/archlinux/archinstall/issues
Traceback (most recent call last):
...
  File "/home/werner/projects/archinstall/archinstall/lib/disk/helpers.py", line 168, in get_partitions_in_use
    mounts.append(Partition(target['source'], None, filesystem=target.get('fstype', None), mountpoint=target['target']))
  File "/home/werner/projects/archinstall/archinstall/lib/disk/partition.py", line 41, in __init__
[...]

PR #787, opened 2021-12-06
Btrfs2

wllacer:
Contains all the changes to make btrfs subvolumes work correctly in archinstall (thus solving #718 and paving the way for things like #781 .
I hope you'all cant test it and get whatever is missing / wrong. 
Things missing, because I need help in this areas
* An UI to define subvolumes. As it is now, subvolume definition has to be done by disk_layouts file. 
* ~have only tested it in non encripted environments. I feel this path is still incomplete~
* ~Only tested with systemd-bootctl. For unexplained reasons I've been unable to boot wih Btrfs, Grub and loop files, together. The code for grub needs changes for sure~
* ~I haven't still implemented code for handling btrfs mount options, although it is needed and has some twists. Real Soon Now~ (done)

The faster way of testing would be to download my branch **wllacer/archinstall/btrfs2** , and use the script only_hd, which only manages partition creation and mounting

*Edit 20/12 Grub works without any manual intervention under VM. The problem seems to be with loop files and grub*

### DIFF
--- a/archinstall/lib/disk/btrfs.py
+++ b/archinstall/lib/disk/btrfs.py
@@ -144,7 +144,8 @@ def manage_btrfs_subvolumes(installation, partition :dict, mountpoints :dict, su
 				# * size. When the OS queries all the subvolumes share the same size as the full partititon
 				# * uuid. All the subvolumes on a partition share the same uuid
 				fake_partition['device_instance'] = Partition(f"{partition['device_instance'].path}[/{name}]",partition['device_instance'].size,partition['device_instance'].uuid)
-				# we reset this attribute, which holds where the partition is actually mounted. Remember, the physical partition is mounted at this moment and therefore has the value '/'. If i don't reset it, process will abort as "already mounted' .
+				# we reset this attribute, which holds where the partition is actually mounted. Remember, the physical partition is mounted at this moment and therefore has the value '/'.
+				# If i don't reset it, process will abort as "already mounted' .
 				# TODO It works for this purpose, but the fact that this bevahiour can happed, should make think twice
 				fake_partition['device_instance'].mountpoint = None
 				#
--- a/archinstall/lib/disk/btrfs.py
+++ b/archinstall/lib/disk/btrfs.py
@@ -90,6 +90,7 @@ def manage_btrfs_subvolumes(installation, partition :dict, mountpoints :dict, su
 	"""
 	# first we mount the basic partition
 	# TODO get rid of the need to use the installation object
+	# TODO For encrypted devices we need some special processing prior to it
 	installation.mount(partition['device_instance'],"/")
 	# We process each of the pairs <subvolume name: mount point | None | mount info dict>
 	# th mount info dict has an entry for the path of the mountpoint (named 'mountpoint') and 'options' which is a list
@@ -123,6 +124,17 @@ def manage_btrfs_subvolumes(installation, partition :dict, mountpoints :dict, su
 					raise DiskError(f"Could not set  nodatacow attribute at {installation.target}/{name}: {cmd}")
 				# entry is deleted so nodatacow doesn't propagate to the mount options
 				del mount_options[mount_options.index('nodatacow')]
+			# Make the compress processing now
+			# it is not an options which can be established by subvolume (but for whole file systems), and can be
+			# set up via a simple attribute change in a directory (if empty). And here the directories are brand new
+			# in this way only zstd compression is activaded
+			# TODO WARNING it is not clear if it should be a standard feature, so it might need to be deactivated
+			if 'compress' in mount_options:
+				if (cmd := SysCommand(f"chattr +c {installation.target}/{name}")).exit_code != 0:
+					raise DiskError(f"Could not set compress attribute at {installation.target}/{name}: {cmd}")
+				# entry is deleted so nodatacow doesn't propagate to the mount options
+				del mount_options[mount_options.index('compress')]
+			# END compress processing.
 			# we do not mount if THE basic partition will be mounted or if we exclude explicitly this subvolume
 			if not partition['mountpoint'] and location is not None:
 				# we begin to create a fake partition entry. First we copy the original -the one that corresponds to
--- a/archinstall/lib/disk/btrfs.py
+++ b/archinstall/lib/disk/btrfs.py
@@ -82,16 +82,12 @@ def manage_btrfs_subvolumes(installation, partition :dict, mountpoints :dict, su
 	* the partition dictionary entry which represents the physical partition
 	* mountpoinst, the dictionary which contains all the partititon to be mounted
 	* subvolumes is the dictionary with the names of the subvolumes and its location
-	First we mount the partition as root
-	Then we create all the subvolumes inside btrfs as demandd
+	We expect the partition has been mounted as / , and it to be unmounted after the processing
+	Then we create all the subvolumes inside btrfs as demand
 	We clone then, both the partition dictionary and the object inside it and adapt it to the subvolume needs
 	Then we add it them to the mountpoints dictionary to be processed as "normal" partitions
-	At the end we unmount the "real" partition to further process
-	"""
-	# first we mount the basic partition
-	# TODO get rid of the need to use the installation object
 	# TODO For encrypted devices we need some special processing prior to it
-	installation.mount(partition['device_instance'],"/")
+	"""
 	# We process each of the pairs <subvolume name: mount point | None | mount info dict>
 	# th mount info dict has an entry for the path of the mountpoint (named 'mountpoint') and 'options' which is a list
 	# of mount options (or similar used by brtfs)
@@ -165,11 +161,7 @@ def manage_btrfs_subvolumes(installation, partition :dict, mountpoints :dict, su
 				# as "normal" ones
 				mountpoints[fake_partition['mountpoint']] = fake_partition
 		except Exception as e:
-			# every exception unmounts the physical volume. Otherwise we let the system in an unstable state
-			partition['device_instance'].unmount()
 			raise e
 	# if the physical partition has been selected to be mounted, we include it at the list. Remmeber, all the above treatement won't happen except the creation of the subvolume
 	if partition['mountpoint']:
 		mountpoints[partition['mountpoint']] = partition
-	# correct end implies the unmounting
-	partition['device_instance'].unmount()
--- a/archinstall/lib/disk/btrfs.py
+++ b/archinstall/lib/disk/btrfs.py
@@ -101,7 +101,7 @@ def manage_btrfs_subvolumes(installation, partition :dict, mountpoints :dict, su
 				name = name[1:]
 			# renormalize the right hand.
 			location = None
-			mount_options = None
+			mount_options = []
 			# no contents, so it is not to be mounted
 			if not right_hand:
 				location = None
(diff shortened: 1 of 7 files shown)


wllacer:
This is one example of the partition part of disk_layouts.json of btrfs partition with subvolumes
**WARNING** *don't use this example as a pattern for anything. It is just a sample for testing syntactic features, and contains several errors and atrocities*
```json
            {
                "btrfs": {
                    "subvolumes": {
                        "@":"/",
                        "@.snapshots": null,
                        "@home": "/home",
                        "@log": "/var/log",
                        "@pkgs": "/var/cache/pacman/pkg",
                        "@/home": "/home2",
                        "@/log": "/var/log2",
                        "@/pkgs": "/var/cache/pacman/pkg",
                        "root": null,
                        "home": null,
                        "sieso/total":null,
                        "/sieso/parcial":"/opt"
                    }
                },
                "encrypted": false,
                "filesystem": {
                    "format": "btrfs"
                },
                "format": true,
                "mountpoint": null,
                "size": "100%",
                "start": "518MiB",
                "type": "primary"
            }

```
## Special entries
### The first interesting entry is **mountpoint**
It can contain either None (null) or a string representing a mount point in the target system .
If we give some value to it , we will mount the *whole partition* there, and will ignore any other mount directive (*cf. infra*) information inside the entry. This is the current behaviour. All subvolume directories will be physically found under this point in the target system, with the **name** we gave them (the key of the dictionary)[¹]
The rationale for this behaviour is simple. If we mount the full partition all subvolumes are accessible in the *file system tree*. To mount them again is innecesary and can be the cause of problems. In this setup, subvolumes are exclusively for internal btrfs use (snapshoting mainly)
If we let this argument null, the next section applies

### The second point of interest is the dictionary **[btrfs][subvolumes]**
It is a series of entries **subvolume name:mount point**, where *mount point* can be null. As you can see, this last follows the usual UNIX filesystem notation. As subvolumes are materialized as directories, the subvolume name is also a directory name , but has some semantic twists:
* In our implementation [²] the names are relative to the "root" of the filesystem in the partition as their starting point. This is not necesarily equal to "/" in the operating system . Better asume it is its own hierarchy.
* The at sign **@** is a mere convention to indicate where the OS root partition will be. It's also customary that its siblings also start with the at sign. *Caveat* Some tools assume this convention as a standard
* The slash is a directory level delimiter, as in the OS. If you look at the example **@home** is a sibling of @ (*flat structure*); but **@/home** creates a subvolume home inside @ (*nested subvolumes*).
When naming a subvolume remember that it is also used as identifier with the partiton name, during the mount operation and the boot procedure.

## What the current implementation is missing.
We miss a place where to document btrfs and mount options for subvolumes. As probably is need to have backward compatibility  we change the structure as:
**subvolume name: None | mountpoint | options_structure **
Most of the options will be routed to the mount procedure, but AFAIK *nodatacow* might need specially handling.

## When to define a mount point and when not.
Some tips. We have no plan to implement them as rules, for the time being.
* Don't define mountpoints to nested subvolumes when a parent has a mount poi nt. If you need some convoluted structure, better use symbolic links
* If you need the subvolume only for backup reasons (include or exclude from snapshots), there is no need to mount, but be certain that the sname is nested within a subvolume that reflects the OS hierarchy. Supose you want to exclude /var/log/archinstall from the snapshot, and /var/log is the @varlog subvolume; you need to name the subvolume @varlog/archinstall
* If you need special btrfs features for your subvolume (e.g. compression, nodatacow,...), it usually better/the only way to separatelly mount from its parent.
* If you have the need to be able to switch rapidly certain areas of the system (soipecially from snapshots) to isolate them in mounted subvolumes makes it a very fast operation.


[1] Actually, the current implementation does not behaves that way. It uses the content of the item. We feel it is an error as we prefer to have separted namespaces for subvolumes and the FSH
[2] We create all the subvolumes from the root of the partition. In general usage, subvolumes are created relative to the current directory where is executed. Absolute adresses are from the "/" mountpoint of the OS image.

Torxed:
A bit of a big change and I saw some things that looked strange *(it looked like some `master` changes were undone)*. I'll have a proper look when I have some time to sit down in peace a and quiet and look at this! Might take a few days.

wllacer:
I hope not. 
What i am working on this patch is more or less orthogonal to the most important changes on the master branch lately (encription and TUI).
There is a small chance of collision on Installer.add_bootloader  (lines 635 passim in my version) I moved part of the management of the options line  to a variable defined before the selection of encrypted or normal root partition. As it is written now, my patch only affects the behavior of the NON encripted part, and far as I can see i haven't altered the behaviour with encryption (if not is a bug)
I believe lines 645-648 should be moved after line 638, but as i have not the slightest experience with encrypted partitions i would prefer the opinion of someone more knowledgable in this concrete point.

https://github.com/wllacer/archinstall/blob/a105a0ff00bddd27306c839ab72a9fcb9fea9642/archinstall/lib/installer.py#L635-L650

The biggest departure from master now is the script, which is a descendent from the original guided.py, prior to the TUI fication,

wllacer:
Most of the changes in Partition.py and disk/helper.py are due to the same reason. For subvolumes the path the system exposes is /dev/partition[/subvolume]. I've created two new properties (device_path and bind_name) which separates its components and have changed all uses of **Partition.path** regarding lsblk and findmnt to **Partititon.device_path** (which for normal partitions would be equal to Partition.path). For the few cases outside a partition instance i've created a function which does the same (split_bind_path).
The changes in **get_mount_info** are of the same kind, but have to use a bit of a list comprehension trick, because I had to guarantee unicity and correction in the return(lines 156-157 in my version)
The changes in the installer are more serious. For one i had to add the "rootflags=subvolume=" option to the options line of the bootloader (as I recall, only got to work for systemd-bootctl ).
On the other hand, i had to create the subvolumes and made them able to be mounted. The relevant code is in the **manage_btrf_subvolumes** at disk/btrf.py. It is the dirtiest piece of code of the whole bunch. First I had to mount the whole partition first and then operate (and unmount it at the end) to be able to create the subvolumes. To mount them what i did was to extented the *mounpoints* dictionary used in **Installer.mount_orderded_layout**, creating a "fake partition" for each subvolume which has to be mounted (with the corresponding data massage and creation of a Partition instance "ad hoc") and adding it to mountpoints. After this the general procedure of the installer can mount and manage the subvolumes as "partitions". And it works fine ;-). I admit is ugly as sin, but it is the most minimal intervention possible, while being still maintenable and basically non invasive(well i ought to better document the routine). One of this days I'll have to make a proposal for a different object model for the disk part that avoids this "incoherence". But i can understand that it is not perhaps the best moment now

wllacer:
I've added an hipercommented version of the manage_btrfs_subvolumes function as it is probably the most "hard" part of my processing.
IN between I've managed to add an options entry for the btrfs subvolume. Instead of having only the mountpoint, i process a dictionary with optionally the mountpoint and a list of possible mount options, like in this example
```
                "btrfs": {
                    "subvolumes": {
                        "root": "/",
                        "home": "/home",
                        "var": {
                            "mountpoint":"/var",
                            "options":["space_cache=v2","nodatacow"]
                        }
                    }
                },
```
The *nodatacow* option has an special treatment (it shouldn't be a mount option), and the rest of the options are functionally available, but I hit a strange behaviour at the OS level, i'm currently investigating *(in mount options take no efect if the system type is specified, at least for btrfs ¿?)*

wllacer:
Now there are a couple of known conflicts with master. I have accepted the changes in master,  using the new properties i defined in partition, and the patch for path (in complex situations) i have for the same situations when not inside a partition.
Preliminary checks don't show problems ... let's hope

timakro:
This is exactly what I was asking for when the original Btrfs PR was merged. Great work!

timakro:
As for the default, it's probably best to go with what other major distros are doing. ([source](https://github.com/archlinux/archinstall/issues/781#issue-1069019909))

> * Ubuntu's installer uses the flat layout, with @ and @home, mounted at / and /home, respectively
> * Manjaro's installer uses the flat layout, identical to the Ubuntu one.
> * Fedora's Installer uses the flat layout, with a different naming scheme (root and home AFAIK)

And [make home optional with a separate prompt](https://github.com/archlinux/archinstall/pull/801).

Being able to fully configure this on the command line would be even more awesome. Maybe offer

1. `@` and `@home`
2. Only `@`
3. Custom

Not sure how "custom" would work. Does archinstall have an option to custom configure regular partitions which could serve as inspiration? Otherwise gdisk/fdisk can serve as inspiration.

wllacer:
@timakro Sadly it is still incomplete. The biggest missing piece (aside from UI) is the integration with encryption handling.  As the code stays now it is incompatible with the current implementation. AFAIK I think i have the sore spot(s) located, but i'm still researching  on the solution. A lot has to do with my own lack of experience with encryption, so any help would be very very welcome

wllacer:
We are now able to use an encrypted btrfs partition with subvolume mounts.
Only tested with systemd-bootctl. And still not checked for compatibility.
For the time being, it uses no keyfiles, just passwords
Code is just good (tricky) enough to work

ShaunTheQuietGamer:
> As for the default, it's probably best to go with what other major distros are doing. ([source](https://github.com/archlinux/archinstall/issues/781#issue-1069019909))

Mentioning this here so it's seen before a decision is made on this - If you see my comment on the aforementioned issue, that is not the only layout used by major distros, opensuse uses a different one that is the only compatible layout for full snapper support, which is one of the two main snapshot programs / wrappers, the other being timeshift. (I'd say snapper has more features out of the box, and it's CLI only, with a separate GUI for people to install if they want.) `snap-pac` also integrates snapper in into pacman for pre/post snapshots, and I'm not sure timeshift has something like that that's not abandoned. Long winded way of saying that opensuse was the one that really mainstreamed btrfs when it made in default for all installations, and has used it as a core component of their distro for a while, rather than an optional feature, and for that reason, I think it should be considered for being default as much as the other layout in that issue, whether as the only default, or as an option alongside the original suggestion.

(Something else to look at is possibly adding an extra Y/N if the opensuse layout is selected to also install and set up snapper and snap-pac with `snapper rollback` functioning.)



_Sorry if this whole comment is out of line, or if this isn't the right place, I'm entirely new to the.. contributor? side of github, and I know there's usually some sort of etiquette with these kind of things. I'm happy to help test and answer any questions I can on what would have to be done for it to function correctly. I barely know any python, or I would do it myself. If I did managed to code something that worked, it would almost certainly be below the standards for being officially added._

Torxed:
I'm running and debugging this now :) It's a major change so it's going to take me some time.
But this is the last remaining fix before v2.3.1 is released. This PR is made against `master` so I'll have to pull it in to master first and then force it into `v2.3.1-dev`, which is fine because otherwise I'd had to do it the other way around anyway. But it will take some time :) Major work on this, good job!

Torxed:
screenshot

I couldn't push the latest changes from `master` into this branch for some reason.

Torxed:
Tested and it seems to work.
The default layout is still preserved in a "simple state", which is nice since it doesn't add complexity there, but allows for more complicated layouts. We can consider moving the default layout after this.

wllacer:
I'm glad to see it merged.  There are still some open issues i'm working now, but they'll come next year

Torxed:
> I'm glad to see it merged. There are still some open issues i'm working now, but they'll come next year

Really appreciated! Heh, next year sounds so far away ^^ Happy new year btw!

PR #792, opened 2021-12-10
fix logic: could not detect root **or** boot

MichaelMai2000:
This PR contains a simple logic fix upon root and boot directory detection.


Torxed:
Nice catch! Still on the run, so can't test major stuff. But this one was easy to verify manually :) Thanks for submitting the PR!

PR #793, opened 2021-12-10
Update installer.py (#1)

OneLongneck:
Added "--removable" after "--bootloader-id=GRUB" on Line 669, because it would throw:
"Could not prepare Boot variable:
Interrupted system call
grub-install: error efibootmgr failed to register the boot entry: Input/output error"
without it on my laptop (with every distro I tried installing)

🚨 PR Guidelines:

# New features *(v2.2.0)*

# Describe your PR

# Testing

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -666,7 +666,7 @@ class Installer:
 			log(f"GRUB uses {boot_partition.path} as the boot partition.", level=logging.INFO)
 			if has_uefi():
 				self.pacstrap('efibootmgr') # TODO: Do we need? Yes, but remove from minimal_installation() instead?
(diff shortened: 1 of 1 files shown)


Torxed:
There's a warning on the arch wiki:
> If dual booting with Windows, be aware Windows usually places an EFI executable there, but its only purpose is to recreate the UEFI boot entry for Windows.

We've had similar issues on `systemd-boot` *(#350, https://github.com/systemd/systemd/issues/13603)* and in most cases it's a firmware issue. The way we solved it was similar as well, we added `--no-variables`. But that in of itself is just a bandaid and causes other issues. So I'm wondering if perhaps we can find a way to fix the reason why you need `--removable` instead?

Torxed:
Me saying this is an understatement, bu you are awesome!
Spent 18 hours trying to figure out grub+uefi+[btrfs/subvolumes/luks] issues.

All I had to do, was add this. So I'm merging this, and pulling it into `v2.3.1-dev` as well! Thank you!

PR #794, opened 2021-12-11
Fix partition layout

HTV04:
This PR fixes the partition layout to remove unnecessary unallocated space and use MiB instead of MB.


archinstall/lib/disk/user_guides.py
```diff
@@ -36,7 +36,7 @@ def suggest_single_disk_layout(block_device, default_filesystem=None, advanced_o
 	layout[block_device.path]['partitions'].append({
 		# Root
 		"type" : "primary",
-		"start" : "518MB",
+		"start" : "512MiB",
```
Torxed, line 39: Archinstall uses [MB not MiB](https://github.com/archlinux/archinstall/issues/674#issuecomment-952837370) in all it's internal calculations, and `parted` supports it as well. So we uniform all numbers to MB. The code to do the bytes -> GB function was introduced [here](https://github.com/archlinux/archinstall/commit/6d0c55e5d610abd199c1dc0430576535dab894b7).

archinstall/lib/disk/user_guides.py
```diff
@@ -122,8 +122,8 @@ def suggest_multi_disk_layout(block_devices, default_filesystem=None, advanced_o
 	layout[root_device.path]['partitions'].append({
 		# Boot
 		"type" : "primary",
-		"start" : "5MB",
-		"size" : "513MB",
+		"start" : "0",
```
Torxed, line 125: This will break GRUB, it needs somewhere between 2-5MB in order to install to the boot sector of a harddrive.
This picture taken from https://www.akadia.com/services/dual_boot.html describes it pretty nicely:
dual_boot_disk_layout

archinstall/lib/disk/user_guides.py
```diff
@@ -150,7 +150,7 @@ def suggest_multi_disk_layout(block_devices, default_filesystem=None, advanced_o
 		"type" : "primary",
 		"encrypted" : False,
 		"format" : True,
-		"start" : "5MB",
+		"start" : "0",
```
Torxed, line 153: Same as above.

Torxed (changes requested):
Some of these changes are breaking changes.

archinstall/lib/disk/user_guides.py
HTV04: Is there a way to check for a BIOS within the user_guided.py script? I don't believe it needs the 512 MB boot partition like the UEFI does, and UEFI doesn't need the 5 MB even with GRUB, so there could be a part of the script that leaves 5 MB for GRUB on the BIOS and makes a 512 MB partition for the UEFI.

HTV04: Fixed, script now accounts for BIOS and UEFI setups and leaves 2 MB unallocated space as recommended by https://wiki.archlinux.org/title/GRUB#Master_Boot_Record_(MBR)_specific_instructions

archinstall/lib/disk/user_guides.py
HTV04: Discussed on Discord, archinstall appears to use MiB/GiB for calculations instead of MB/GB. parted also uses separate conversion values for MiB/GiB and MB/GB https://www.gnu.org/software/parted/manual/parted.html#unit

archinstall/lib/disk/user_guides.py
HTV04: Yep, did some digging and I found out that hardware.py had that convenient function.

Before you made that post I used it in 01749ec, which added a small if statement that changes some variables if there is a UEFI in order to make the most optimal layout. Should I remove this?

Torxed: You can leave it in i think. I'll have a look at the overall PR and it it doesn't add too much complexity it should be fine :)

Torxed:
I will test this and the other PR you've submitted regarding `/home` some time during tomorrow most likely.
And if they look good I'll smash them into `master` : )

Torxed:
I've tweaked the partition size to be 512MB instead of 513. Since `size` is positional from `start` we don't need to account for `start` being `1MiB` in order to create a `512MiB` large partition.

HTV04:
> I've tweaked the partition size to be 512MB instead of 513. Since `size` is positional from `start` we don't need to account for `start` being `1MiB` in order to create a `512MiB` large partition.

I did that intentionally, because `size` seems to be an `end`, not a partition size like the name suggests.

In my testing, I noticed that in the case of /boot, if `size` is 512MiB it will be 1 MiB too short. I think that the changes you made in the last commit will also reintroduce the unallocated space.

I think the best solution here is to either rename the `size` variable to `end`, which is more accurate to its current functionality, or have `size` be interpreted by taking its number and subtracting it from `start`, which would require extra trickery since the unit (i.e. MiB) is also attached to the number in its string.

Torxed:
Ah bugger, so we'll have to revert my attempt at fixing things >_<

HTV04:
That sounds good, I like that it's more true to parted too.

By the way, I noticed that the function for handling contents of partition "lists" interprets `size` as an `end` variable lol

PR #795, opened 2021-12-12
Added delay to partprobe

willyt3hwhale:
Added delay so that partprobe has time to finish.
Otherwise information such as partuuid wont be readable with lsblk.


willyt3hwhale:
Only fixed in one source file, new PR with full fix incoming

PR #802, opened 2021-12-15
Changed lsblk to blkid where possible

Torxed:
There's a at least two more places where I'd like to swap out `lsblk`.
However, due to possibly adding complexity I'll avoid it.
Instead I bumped up the sleep and moved it to a more strategic place in two places.

`partuuid_to_index` gets a higher sleep, which leaves `partprobe` to be quick in the many places we use it.
And the helper `partprobe()` *(not the class methods)* also got a heavy sleep. As it's mainly used in `convert_device_to_uuid()` which is also used rarely.

This should speed things up, and safely grab the `PARTUUID` and `UUID` in the places we need it to.
Other alternatives are:
 * [sfdisk](https://linux.die.net/man/8/sfdisk)
 * [partx](https://man7.org/linux/man-pages/man8/partx.8.html)
 * [findfs](https://man7.org/linux/man-pages/man8/findfs.8.html)
 * [hdparm](https://man7.org/linux/man-pages/man8/hdparm.8.html)

I've tested `sfdisk` as well as `blkid` and both seem to have a 100% success rate of grabbing the `PARTUUID` no matter how quickly I attempted to grab it right after `parted` commands.

For historical purposes, this was the command I used to test it:
```
dd if=/dev/zero of=/dev/loop0 bs=4096 count=10 && \
partprobe /dev/loop0 && \
parted /dev/loop0 mklabel gpt && \
python test.py && \
echo $(sfdisk --part-uuid /dev/loop0 1 && lsblk -o+PARTUUID /dev/loop0p1) >> result.txt
```
[test.py](https://github.com/dcantrell/pyparted/tree/master/examples) being an example script using [pyparted](https://github.com/dcantrell/pyparted). And the loop device was set up using:
```
# truncate -s 20G testimage.img
# losetup -fP ./testimage.img
```

### DIFF
--- a/archinstall/lib/disk/blockdevice.py
+++ b/archinstall/lib/disk/blockdevice.py
@@ -150,8 +150,7 @@ class BlockDevice:
 		This is more reliable than relying on /dev/disk/by-partuuid as
 		it doesn't seam to be able to detect md raid partitions.
 		"""
-		for partition in json.loads(SysCommand(f'lsblk -J -o+UUID {self.path}').decode('UTF-8'))['blockdevices']:
-			return partition.get('uuid', None)
+		return SysCommand(f'blkid -s PTUUID -o value {self.path}').decode('UTF-8')
 
 	@property
 	def size(self):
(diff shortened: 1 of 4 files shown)


dylanmtaylor:
Anything that we can do to avoid that race condition that we seem to hit all the time is welcome.

PR #803, opened 2021-12-18
Update documentation

remLse:
Hi there! I'm interested in your project and read the documentation, fixed some typos along the way. I also added a description on how to build the docs locally for others who may want to contribute to the docs for the first time.

I didn't notice at first that the table for the json profile specification is able to scroll to the right, revealing more information. I added a note about that so the info doesn't go unnoticed.

Was it correct to create the PR against master? Or should I change it to another branch?

Thanks for your amazing work on this project!

### DIFF
--- a/docs/installing/guided.rst
+++ b/docs/installing/guided.rst
@@ -1,7 +1,10 @@
+.. _guided_installation:
+.. autofunction:: guided_installation
+
 Guided installation
 ===================
 
-This is the default scripted installation you'll encounter on the official Arch Linux Archinstall package as well as the unofficial ISO found on `https://archlinux.life <https://archlinux.life>`_. It will guide your through a very basic installation of Arch Linux.
+This is the default scripted installation you'll encounter on the official Arch Linux Archinstall package as well as the unofficial ISO found on `https://archlinux.life <https://archlinux.life>`_. It will guide you through a very basic installation of Arch Linux.
 
 The installer has two pre-requisites:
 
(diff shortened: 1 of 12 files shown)


docs/installing/guided.rst
```diff
@@ -91,61 +94,66 @@ To run it, execute archinstall as a Python module:
...
+| gfx_driver           | - "VMware / VirtualBox (open-source)"               | Graphics Drivers to install                                                          | No                                            |
+|                      | - "Nvidia"                                          |                                                                                      |                                               |
+|                      | - "Intel (open-source)"                             |                                                                                      |                                               |
+|                      | - "AMD / ATI (open-source)"                         |                                                                                      |                                               |
+|                      | - "All open-source (default)"                       |                                                                                      |                                               |
```
remLse, line 116: This diff here is a bit unruly, a lot of changes to white space. The only substantive change I made to the table was to put all possible values for `gfx_driver` on separate lines. This makes the column "Values/Description" much shorter and doesn't hide the "Description" column anymore. The forth column "Required" is still hidden though. I added a note at the bottom of the table about that. Obviously this change makes the row for `gfx_driver` much taller, let me know if you think that's worse than before.

dylanmtaylor:
Documentation changes are always welcome, so thank you for that.  Master is the correct branch to make the change against.

Torxed:
I agree with dylan, huge thank you for this one!

Issue #799, opened 2021-12-13
Missing step to add user


SoongVilda:
I went 2 times through archinstall, but I only saw question for hostname.
Where is question (step) for adding user?

Version ISO
archlinux-2021.12.01-x86_64

SoongVilda:
That sounds weird, you be able use sudo, because you have superuser (root password), but user is also need it.
So I must miss root password and then I will be prompt to set user?

dylanmtaylor:
> That sounds weird, you be able use sudo, because you have superuser (root password), but user is also need it. So I must miss root password and then I will be prompt to set user?

That appears to be how the current behavior works. @Torxed -  is this something we want to change?  So you can set the password and the password of the supervisor through the guided installer?

SoongVilda:
Version 2.2.0 was pretty intuitive and follows simple way.

- Encryption
- hostname
- root password 
- user (name & password) 
- question for more user
done

I believe these are the most used options and perhaps you always have to configure them manually anyway, automation through archinstall would shorten the installation.

But latest version 2.3.0 without user? It's confusing

Torxed:
The workflow here should be:

 1. Do you want root? No -> Force at least one superuser -> Do you want to add additional users? -> Sudo?
 2. Do you want root? Yes -> Do you want to add a user? -> Should that user be a super-user (sudo)?

Somewhere along the way during the last version that workflow changed and apparently lost some flexability.
Recommendation for now is to set a root password and then create your other users manually.
It's a bit of an inconvenience but at least is not a breaking issue. But I'll fix it in v2.3.1 this week.

Torxed:
I'll have to apologize because I fixed this before seeing this PR.
So I'll close this but wanted to note that this was a welcome PR and there was nothing wrong with it, so I'm not closing it because it was bad or anything. Apologies!

*(See 2419c6b and aa21cd4)*

PR #804, opened 2021-12-20
Fix: Missing step to add additional users #799

rohandebsarkar:
Fixes issue #799


Torxed:
As mentioned in the issue, I apologies for not seeing this PR prior to fixing it. I was in a bit of a rush to solve some issues before tagging a release candidate for v2.3.1 so this got fixed in the rush of things.
The PR would have solved it perfectly, and the only tweak I did really was to consolidate your `if / else` into:
```python
archinstall.arguments['!superusers'] = {**archinstall.arguments.get('!superusers', {}), **superusers}
```
A lazy grab using `** .get("!superusers", {})`.

Again, apologies and thank you for submitting a PR to fix an issue!

PR #660, opened 2021-10-12
Add simple menu for better UX

svartkanin:
This PR adds the simple menu (https://github.com/IngoMeyer441/simple-term-menu) as a new selection menu to enhance the UX of the guided user flow. 

It has been suggested in issue https://github.com/archlinux/archinstall/issues/655 and this PR changes most selection menus to use the new layout. 

The only part where it still uses the old one is during manual partitioning. The reason for that is that I think there are some bugs in the partitioning code and it kept exploding when I was going through the flow in master. Therefore, I decided to leave that one as is and fix the bugs first and only after switch it over to the new layout as well. 

*Edit: This will fix #173 (I can't link earlier issues to newer PR's because the issues was created before the move from Torxed/archinstall to archlinux/archinstall)*

PR #806, opened 2021-12-21
Global menu

svartkanin:
This is a follow-up PR of the previously merged new interface menu (https://github.com/archlinux/archinstall/pull/660).

Bare in mind that this is still WIP (there are some minor things left to fix) but I wanted to raise it to get some feedback on it. 

@Torxed as we discussed in the previous PR we wanted to introduce an overall menu as well where previously selected options could be changed again. 
I've added a global menu now with pre-set defaults as they're used in the single menus, so that the user doesn't have to modify anything they're happy with in the first place. 

In addition, the defaults get overridden by any value that has been specified via a config (basically anything that has been stored in the `archinstall.arguments` container. This way those values are already set and displayed and can, if needed, also be changed again. 

I'll update the PR to fix the remaining things but it is mostly functioning and ready to play around with


svartkanin:
@Torxed I've updated the last bits and pieces and it should be ready for a review now

examples/guided.py
```diff
@@ -70,124 +71,59 @@ def ask_user_questions():
...
 		will we continue with the actual installation steps.
 	"""
-	if not archinstall.arguments.get('keyboard-layout', None):
-		archinstall.arguments['keyboard-layout'] = archinstall.select_language()
 
-	# Before continuing, set the preferred keyboard layout/language in the current terminal.
-	# This will just help the user with the next following questions.
-	if len(archinstall.arguments['keyboard-layout']):
-		archinstall.set_keyboard_language(archinstall.arguments['keyboard-layout'])
+	global_menu = GlobalMenu()
```
Torxed, line 75: You can omit the `from archinstall.lib.menu.selection_menu import GlobalMenu` and use `archinstall.GlobalMenu` as long as you expose the class in: https://github.com/archlinux/archinstall/blob/577a060e792a82e1e93822c9a16b4093f94c003c/archinstall/__init__.py#L23
This will make the guided example a bit more cleaner. As it serves as an example of how to write an installer not having to dig through the inners of archinstall and learn all the paths to stuff :)

Torxed (changes requested):
I only have one minor tweak, then I'm ready to merge :) Solid work on this one, guided became a lot cleaner :)

examples/guided.py
svartkanin: @Torxed awesome thanks for the review! 
I've updated the import

Torxed: Great work! We've modified guided a bit in regards to `"!users"` and `"harddrives"` section, could you pull that in to the global menu? I think that would help with the merge conflict against master that is right now.

svartkanin:
I realized the conflicts only after, I think I had the user changes already in there, working on the harddrive now

svartkanin:
@Torxed I fixed the conflicts now (hopefully correctly ;)) and updated minor dependency things in the menu options

Torxed:
I'll run a last test probably tomorrow and if everything works (which I'm sure it will) this will fly into `master` and we have ourselves a new menu system :) Great work so far!

wllacer:
@svartkanin Good work. It looks very beautiful now.
After playing a while with it, i have some suggestions. Where do you want that i send them, here or at your repository ?. But two merit to be here
**A bug report:**. If you are working with merge have a look at some changes at guided.load_config where the disk_layouts file is loaded. Since a couple of days, this is not needed anymore (it's loaded at __init__.py and with a seemly less error prone code). There, testing your branch, did reproduce a well know bug, we were already rid.
**A wish**, please don't delete -still- the old user interfaces. Some workflows might not be ready to be changed on short notice.

svartkanin:
@wllacer feel free to leave any feedback here in this thread, I'll try to address it when time allows

Torxed:
I'm ready to merge this. I'll probably hold off on #846 until this is merged as I would probably poke around in too many places where this interacts.

svartkanin:
@Torxed sounds good!

Torxed:
Great work! Any remaining issues can be done in a new PR :) We probably need to tweak a bit :)

svartkanin:
@Torxed Fantastic :partying_face: , please feel free to assign any tweaks to me if you like, I'm happy to take them on since I got the context :)

Torxed:
Will do with pleasure! hehe

Issue #779, opened 2021-11-29
Wrong "path completion" when using the --disk_layouts flag


NotNorom:
Given a directory like `/root`:
```
ls
.rw-r--r-- 6.7k root 28 Nov 15:34 config.json
.rw-r--r--  786 root 28 Nov 00:30 partitions.json
```

When running the program with the `disk_layouts` flag like this:
```
archinstall --config config.json --dry-run --disk_layouts=partitions.json
```

The program will return an error:
```
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/archinstall/examples/guided.py", line 62, in load_config
    archinstall.storage['disk_layouts'] = json.loads(archinstall.arguments['disk_layouts'])
  File "/usr/lib/python3.9/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.9/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.9/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/bin/archinstall", line 33, in <module>
    sys.exit(load_entry_point('archinstall==2.3.0', 'console_scripts', 'archinstall')())
  File "/usr/lib/python3.9/site-packages/archinstall/__init__.py", line 111, in run_as_a_module
    script.execute()
  File "/usr/lib/python3.9/site-packages/archinstall/lib/profiles.py", line 180, in execute
    self.spec.loader.exec_module(sys.modules[self.namespace])
  File "<frozen importlib._bootstrap_external>", line 850, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/usr/lib/python3.9/site-packages/archinstall/examples/guided.py", line 383, in <module>
    load_config()
  File "/usr/lib/python3.9/site-packages/archinstall/examples/guided.py", line 64, in load_config
    raise ValueError("--disk_layouts=<json> needs either a JSON file or a JSON string given with a valid disk layout.")
ValueError: --disk_layouts=<json> needs either a JSON file or a JSON string given with a valid disk layout.
```

[...]

PR #810, opened 2021-12-25
Fix disk layout argument #779

svartkanin:
This PR  addresses the bug reported at https://github.com/archlinux/archinstall/issues/779. 

There seems to have been an issue with setting the CWD to the scripts path which means that the any files specified cannot be resolved from the original CWD anymore. 
I have tested the flow by removing the explicit setting of the CWD and it still seems to work fine. 
Also the check for the disk layout seemed to have contained some redundant code which got simplified as well.


wllacer:
@svartkanin  Can you test it with the master branch as it is now? I think I addressed a similar issue and the changes are now upstream

svartkanin:
@wllacer thanks for that your PR fixed the original problem as well which means this PR is no longer necessary :)

Torxed:
Great work both of you!

Issue #805, opened 2021-12-21
Regenerate initramfs at the end


rednaz1337:
It would be great if this script regenerated the initramfs at the end. When you use an encrypted rootfs and set the keyboard layout to a non-english one, the keyboard layout is still set to english in the initramfs where you need to enter your password. This makes it difficult to enter the password correctly. It can be fixed by manually running `mkinitcpio -P` in the chroot environment at the end, but it would be great if archinstall did this automatically. It's not a huge problem, but it would be nice to have.

Torxed:
Humm, we set the keyboard locale before generating initramfs, or so I thought.
But if we don't then we should absolutely implement this : )

Torxed:
Fixed in c4f80b9 and should be released soon in v2.3.1.

Issue #783, opened 2021-12-02
'<' not supported between instances of 'NoneType' and 'str'


hkaancaliskan:
I was trying to dual boot with windows.
/dev/nvme0n1p1 -> EFI -> 1GB
/dev/nvme0n1p2 -> MSR
/dev/nvme0n1p3 -> Windows 11
/dev/nvme0n1p4 -> Newly created 20GB Btrfs

In archinstall just mounted EFI as /boot and btrfs as /.
2021.12.01 archiso
I forgot to take logfile but if necessary i can try again and take log.
![20211202_104351](https://user-images.githubusercontent.com/17825767/144379982-146579ea-be70-421a-a8f4-4bcbe51bc094.jpg)
![20211202_104330](https://user-images.githubusercontent.com/17825767/144379996-9353577b-3b55-4989-9bb2-a797120d68f6.jpg)

Edit: I saw something right now. In the second photo it shows PARTUUID for efi in "Re-using partition_instance" line but it shows as None in the next line.

Torxed:
Thank you for posting this issue. I'll investigate and see what could cause this. But as you mentioned the device instance being re-used is probably the issue if the previous instance was `None` for some reason. I think it has to do with re-using partitions all together. So I think I know where to look, I'll come back to you if I really need the `install.log` in this instance.

ndfr623:
Same here, except on a completely separate drive due to the Windows installer's bad habit of appropriating EFI partitions.
![IMG_20211206_114839731](https://user-images.githubusercontent.com/69735179/144845083-ddc86efd-4f78-43ab-a939-a4a2016edfa6.jpg)

flechtwerkflechtwerk:
I get exactly the same error message when using archinstall. The partition selected for /boot is the Windows 10 EFI partition on one hard drive, the / partition is a newly formated ext4 partition on another drive. I did not create a log file but I can redo the install and create one if it would be helpful.

wllacer:
This issue has happens un installer.ordered_mount*.
For some reason one partition element has no future mountpoint assigned, and It crashes all the process. I had a lot of them during my take on btrfs

danboid:
I've just been hit by this bug trying to install arch alongside W10.

TheDrawingCoder-Gamer:
same issue, trying to install arch alongside W10 as well

phaitonican:
Same error

paweljonik:
[...]

Issue #782, opened 2021-12-02
ZRAM swap issue from reddit


psyfry:
A user posted about an issue with archinstall on Reddit, not sure if they have a github account, so copying it here for them.

===================================
**Archinstall script on latest 2021.12.01 iso is a bit broken**

So decided to use archinstall script again for giggles and ran into a bunch of issues:

When not selecting zram swap and selecting the option to optimally format it is called 0 now,it still creates 3 partitions for some weird reason and one of them for a whooping 20GB and ads like 3 free space partitions in between,looks like this in cfdisk:

free space

/dev/sda1

free space

/dev/sda2

free space

/dev/sda3

2. Also there is no option to create root and user with superuser(sudo) privileges anymore only one or another which is not secure/weird decision?

3. Still present from previous version,even if you don't choose to install NVIDIA drivers you still get the weird Wayland session on top of X11 in KDE Plasma,which is obviously borked on NVIDIA.

4. Once the zram swap partition or whatever it is,is created the entire root/main user directory is put on these 20GB so you are out of luck if you try to install a bunch of dependencies/packages that are more than 20GB in size

Anyway, it is all curable by a manual Arch Linux installation,but still the previous version of archinstall was better,this one is weird.Hope this helps anyone who decides to use archinstall to speed things up,that is not the case anymore.

==============================

Another user's response to above post:
I don't have the zram issue, but I do have the free space thing. 5MB free, 500MB FAT partition, 5MB free, then 1000GB BTRFS. Also agree with the user thing, weird decision to remove that. I made a post about how to deal with that, but it ended up getting downvoted because I may or may not have introduced a small security issue by giving root to anyone who knows the root password... I was a bit slow to figure out that I needed to uncomment out the sudo and wheel group before adding a user to the groups.

Torxed:
Good feedback. Did the user by any chance post the log files? Especially the install.log?

fkortsagin:
Hello!
archinstall.log pastebin
https://pastebin.com/ACBsGyXy
pacman log in case it is needed:
https://pastebin.com/ZxgWqngV
And we will probably need to open another issue for the swap space being used as main root/sudoer /mnt point and not using the home directory.Low disk space:
https://ibb.co/3YB1fR1
cfdisk result after archinstall the mnt point is located on the 18 GB instead of 70GB Home,same results on HW testing:
https://ibb.co/LCNfKHW

fkortsagin:
[...]

PR #818, opened 2021-12-31
V2.3.1 fix free space and #783

Torxed:
This fixes:
 * #783
 * #782
 * #805 

This eliminates *(hopefully)* the `0.5GB` windows between partitions.
It's still not perfect, as I'll need to create a `xB -> xiB` converter for parted.

But this also fixes a sorting issue when re-using partitions and partition ID's.
`PARTUUID` should no longer be able to be `None`.


Torxed:
Wrong target branch, >_<

PR #821, opened 2021-12-31
Adding a Pipewire application profile

Torxed:
This to better manage the pipewire setup process and minimize guided a bit. This also adds the concept of `@archinstall.plugin` decorators to add a plugin in run-time. Which pipewire uses to detect user creation and enable the pipewire-pulse service for new users.


Torxed:
screenshot

screenshot

PR #827, opened 2022-01-02
Define an standard locale for all the commands executed

wllacer:
This one liner makes a standard locale for everything executed as system commands inside Archinstall.
We have selected the C locale, ascii and english texts, as we believe it is always available inside Archlinux.
As it is coded it takes precedence over any possible change specified thru the SysCommand* interfaces.

### DIFF
--- a/archinstall/lib/general.py
+++ b/archinstall/lib/general.py
@@ -185,6 +185,8 @@ class SysCommandWorker:
 		self.callbacks = callbacks
 		self.peak_output = peak_output
 		self.environment_vars = environment_vars
+		# define the standard locale for command outputs. For now the C ascii one"
+		self.environment_vars['LC_ALL'] = "C"
 		self.logfile = logfile
 		self.working_directory = working_directory
 
@@ -364,7 +366,7 @@ class SysCommand:
 		peak_output :Optional[bool] = False,
 		environment_vars :Optional[Dict[str, Any]] = None,
 		working_directory :Optional[str] = './'):
-	
+
 		_callbacks = {}
 		if callbacks:
 			for hook, func in callbacks.items():
--- a/archinstall/lib/general.py
+++ b/archinstall/lib/general.py
@@ -184,9 +184,8 @@ class SysCommandWorker:
 		self.cmd = cmd
 		self.callbacks = callbacks
 		self.peak_output = peak_output
-		self.environment_vars = environment_vars
-		# define the standard locale for command outputs. For now the C ascii one"
-		self.environment_vars['LC_ALL'] = "C"
+		# define the standard locale for command outputs. For now the C ascii one. Can be overriden
+		self.environment_vars = { 'LC_ALL':'C' , **environment_vars }
 		self.logfile = logfile
 		self.working_directory = working_directory
 
--- a/archinstall/lib/general.py
+++ b/archinstall/lib/general.py
@@ -185,7 +185,7 @@ class SysCommandWorker:
 		self.callbacks = callbacks
 		self.peak_output = peak_output
 		# define the standard locale for command outputs. For now the C ascii one. Can be overriden
-		self.environment_vars = { 'LC_ALL':'C' , **environment_vars }
+		self.environment_vars = {'LC_ALL':'C' , **environment_vars}
 		self.logfile = logfile
 		self.working_directory = working_directory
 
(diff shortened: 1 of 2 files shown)


archinstall/lib/general.py
```diff
@@ -185,6 +185,8 @@ def __init__(self,
 		self.callbacks = callbacks
 		self.peak_output = peak_output
 		self.environment_vars = environment_vars
+		# define the standard locale for command outputs. For now the C ascii one"
+		self.environment_vars['LC_ALL'] = "C"
```
Torxed, line 189: I'm thinking this might be bad if the users use `SysCommand(Worker)` to execute things in their profiles, and it reverts their global locale. I think we'll have to internally set `LC_ALL=C` like you've done in `luks.py` to avoid this.

Torxed (changes requested):

archinstall/lib/general.py
wllacer: No. The default i set in general.py (LC_ALL=C), only affects commands executed INSIDE Archinstall, so we have a guaranteed stable environment. It does not touch user's own environment.
To make matters easier for a developer, i now allow to change the locale on a command by command base (as i did in Luks ). This is a change i just uploaded, my previous version

wllacer:
@Torxed . I'm not sure i make my response to your change request correctly.
Putting the LC_ALL inside the SysCommandWorker initializacion guarantees a stable interface inside Archinstall, but has no bearing outside it, as it is done on a command by command basis, and only for those executed thru the SysCommand* interface (if someone would use a os.exec* call inside Archinstall code, would see their locale, whichever it might be).
On the other hand, with today's change, i corrected the precedence. If some author needs a different locale (or doesn't trust the default) can change it thru the *environment_vars* parameter in the SysCommand* call (the luks.py one might be taken as an example )  Yesterday's version ignored it, but now this choice is respected

Torxed:
> @Torxed . I'm not sure i make my response to your change request correctly. Putting the LC_ALL inside the SysCommandWorker initializacion guarantees a stable interface inside Archinstall, but has no bearing outside it, as it is done on a command by command basis, and only for those executed thru the SysCommand* interface (if someone would use a os.exec* call inside Archinstall code, would see their locale, whichever it might be). On the other hand, with today's change, i corrected the precedence. If some author needs a different locale (or doesn't trust the default) can change it thru the _environment_vars_ parameter in the SysCommand* call (the luks.py one might be taken as an example ) Yesterday's version ignored it, but now this choice is respected

That is true, but take this use case as an example where archinstall is used as a tool in another project: https://github.com/Torxed/archoffline/blob/f63831d06bf9fed8ffe7c6ed534f4539b0b7f820/offline.py#L339

Wouldn't this change we're discuss make it so the locale for that pacman call always were `LC_ALL=C` and make that output ignore the systems locale if one was set?

Torxed:
I guess to solve both of our use cases and not overrule the users system locale, we could initiate the environment variables here: https://github.com/archlinux/archinstall/blob/4f50d5ab38b5884ae99e4a088913a71babf95e46/archinstall/lib/general.py#L172

With something like:
```python
if not environment_vars:
	environment_vars = dict(os.environ)
```

wllacer:
os.environ is alredy injected in all SysCommand* calls
https://github.com/archlinux/archinstall/blob/93aa4aa6cc88462cf2cf4797644086e6879724bf/archinstall/lib/general.py#L340-L343
My solution is too restrictive regarding localization efforts, is true. Probably what we need is some kind of global parameter at **storage** which can be switched at the start of (or even during) the app execution; kind of (virtual code)
```
		# define the standard locale for command outputs. For now the C ascii one. Can be overriden
		self.environment_vars = {'LC_ALL':storage.get('DEFAULT_LOCALE','C') , **environment_vars}
```
I will have a look at your counterexamples, and see what is feasible.

Torxed:
We can fine tune this once it's in master, so merged it :)

wllacer:
Sorry for the delayed answer.  You're right, this is a thing (real case collision) which is better solved with experience. I'm afraid we have made it, even now, too strict. But I need some time to explore alternatives. (localization IS a world unto itself)

Torxed:
No worries :) I've been busy myself and I'll never expect people to prioritize this over anything else :) If and when you have time, all help is appreciated no matter how delayed it is :)

Great work on identifying and coming up with a fix for this. We've gotten lucky for quite some time hehe.

Issue #830, opened 2022-01-02
httpS problems if date is wrong in the installer


ilario:
Using archinstall on a system with a wrong date/clock fails at the mirror selection step.
This happens because the https certificate used for downloading the mirrors list gets recognized as invalid when comparing its validity date and with the (wrong) system date.

The resulting error message in my case was this:

`Could not fetch an active mirror-list: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: certificate is not yet valid (_ssl.c:1129)>`

Maybe this was related to #682 and #297

Torxed:
This is expected. It's a security measurement in the HTTPS protocol. And any tool you run will have the same issue (wget, curl, archinstall, python.. etc).

Anytime you want to go out on the internet, even on your desktop client (windows, mac or linux) you will have to first ensure that the time of your BIOS and or OS (including Arch Linux live ISO) is correct.

There's not much archinstall can do here, so I'll close this unless suggestions to improve the UI to notify some how, but I'm not sure how we'd detect this prior to going out on the internet to check the time.

ilario:
Yep, I know it is normal to happen. Check out #831 for a proposed solution (as it's the first time I use archinstall, this could well be a very bad way to fix, I don't know).

Torxed:
Apologies, until I read your PR I thought this was more of a support ticket on how HTTPS works.
I now understand your intentions more clear and I merged your PR. It was a solid 10/10 fix I would say! : ) Good job!

PR #831, opened 2022-01-02
examples/guided: if selected, activate NTP also in installer system

ilario:
This is a proposed solution for #830

Activating the time synchronization (NTP) for the installation system [is part of the installation guide](https://wiki.archlinux.org/title/Installation_guide#Update_the_system_clock).

Here I modified #505 for activating NTP as soon as the user selects it for the target system.
And moved the question about NTP before the mirror selection, which is the first step that requires a valid date.

### DIFF
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -69,6 +69,8 @@ def ask_user_questions():
 		archinstall.arguments['ntp'] = input("Would you like to use automatic time synchronization (NTP) with the default time servers? [Y/n]: ").strip().lower() in ('y', 'yes', '')
 		if archinstall.arguments['ntp']:
 			archinstall.log("Hardware time and other post-configuration steps might be required in order for NTP to work. For more information, please check the Arch wiki.", fg="yellow")
+			from archinstall.lib.general import SysCommand
+			SysCommand(f'timedatectl set-ntp true')
 
 	# Set which region to download packages from during the installation
 	if not archinstall.arguments.get('mirror-region', None):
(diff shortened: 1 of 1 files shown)


ilario:
Maybe before passing to the mirror list retrieval, the code should wait for NTP to do its thing... I'm not sure if there is a nice way to do that, otherwise I can add a cycle checking the output of `timedatectl`. But I think that most of the users, when they get to this situation, would ctrl+c and re-run archinstall (and by that time NTP should be happy).

Torxed:
I'm not against enabling NTP before/after choosing locale for instance.
That would probably solve the HTTPS error mentioned in #830. Just like the locale sets it both for the live medium and the installation, the ntp question could too. I don't see the harm in merging this.

Torxed:
Thank you for this PR, really quick work!

ilario:
Thank you indeed! Also for the improvements to the code!

svartkanin:
@Torxed should we possibly move this solution over to the new menu style as well? 
We re-implemented the question now in the `guided.py` and have it in the `user_interaction.py` with `ask_ntp`. 
Currently in the new selection menu NTP is set to `True` by default, which we could just apply automatically by not asking the question at all at the beginning, but rather opted out in case someone explicitly turns it off.

Torxed:
I think that's a good thing to do.
There's a few other options that still use the old methods, which I think we should carry over to the new menu system as well.
I can't remember from the top of my head which ones it was, but I think it had to do with btrfs subvolumes and the question about "standard subvolumes".

svartkanin:
@Torxed I opened a PR for it https://github.com/archlinux/archinstall/pull/874

Issue #835, opened 2022-01-03
(Query)  mount options in disk_layout ?


wllacer:
Although all the infraestructure for executing mount command with options is in place (i'm using them now) i haven't found a formal way to specify them with the partition dictionary (the disk layout json). 
The usual way of mounting (via Installer.mount_ordered_layout) didn't used any (they do now, i put them), and I haven't yet found any other use of them, so existing code is of no help.
Anyone is using them it any other related project? and how? or did it have a planned, but never implemented place ?
As I need/wish them for **btrfs**, I've defined two new entries for the partition dictionary, **format_options** and **mount_options**. Before merging the code, any suggestions ?

I found an *options* keys , but it used is for filesystem creation (*mkfs*) , so they correspond to the first new key (compatibility code in place)
https://github.com/archlinux/archinstall/blob/3474176e641eed9ced37c918b157fd35c07dc388/archinstall/lib/disk/filesystem.py#L124

Torxed:
They are used by people when modifying user disk layouts like so *(And I'm spitballing from the top of my head now)*:
```json
[
    {
        "/dev/sda": {
            "partitions": [
                {
                    "type": "primary",
                    "format": true,
                    "mountpoint": "/",
                    "options": ["casefold", "something-else"],
                    "filesystem": {
                        "format": "ext4"
                    }
                }
            ]
        }
    }
]
```

The placement of the options is a bit out of order if you ask me. I don't know why I placed it there to begin with. But the above example will make ext4 case insensitive for instance. And I would have liked the options to live under the `"filesystem": { ... }` block instead.

The full history can be found here: https://github.com/archlinux/archinstall/issues/380

wllacer:
@torxed. 
1) Makes a lot of sense: There seems to be absolutely no problem to put the "format_options" were you want: under *filesystem* Just one line of code to touch, and already "backward compatible". The change is already done on my branch (btrfs-encrypt)
I just assume this is an advanced option, and for the time being implies editing the json file.
2) Any place where you would like to see the *mount_options*? I have the very specific problem with btrfs that i have to handle two sets of options, one for the whole partition (although specified at OS level by mount point) and one for each subvolume (sometimes it will be a mount option, sometimes not). This i why i choosed *mount_options* below partition (and general for all FS), and the *options* key in the new subvolume dictionary . The first is straightforward to change if desired, the second, not serious but not trivial.
[...]

Issue #813, opened 2021-12-27
TypeError "<" not supported between instances of type "str" and "NoneType"


bebetoalves:
I tried to do manual partitioning, as automatic partitioning doesn't offer an option to not create the "home" partition. And it was like this:

![20211227_094417](https://user-images.githubusercontent.com/13764984/147473693-c773e06a-ca38-4add-addf-f4191e85483e.jpg)
![20211227_094515](https://user-images.githubusercontent.com/13764984/147473694-57559e89-f85f-40b2-87c6-82f7d0e085b2.jpg)

However, when formatting the partitions this error occurred:

![20211227_094638](https://user-images.githubusercontent.com/13764984/147473762-4b2d87b1-38c9-4546-8c0d-fd56ac1bbafd.jpg)

Torxed:
The main reason being `/boot` doesn't have a `mountpoint` set.
I'm not sure why, so I would need to know which steps you took to end up here.

The error itself, was fixed in #783 with the same issue title. But again, your particular issue is a bit different but I won't know why it happened without some more information.

bebetoalves:
> The main reason being `/boot` doesn't have a `mountpoint` set. I'm not sure why, so I would need to know which steps you took to end up here.
> 
> The error itself, was fixed in #783 with the same issue title. But again, your particular issue is a bit different but I won't know why it happened without some more information.

The only different thing I did was not set the 500MB partition to the /boot mount point because I ran into an error, already reported in #812 . I did the same process using the stable version of archinstall and it worked.

wllacer:
Did anyone tested it with the new 2.3.1.dev  or master?  
The reason behind is that somehow unmountable partitions slipped in the list of partitions to be mounted, and a sort routine freaked out. @Torxed has corrected it in the above mentioned versions (just a couple of days ago)

Torxed:
> Did anyone tested it with the new 2.3.1.dev or master? The reason behind is that somehow unmountable partitions slipped in the list of partitions to be mounted, and a sort routine freaked out. @Torxed has corrected it in the above mentioned versions (just a couple of days ago)

I think it's fixed. The thing that spook me a bit and why I haven't closed this issue.
Is because the installer continued without a `/boot` which would break the installation further down the line. I'm still wondering how that could happen.

wllacer:
Unless It is a btrfs partition, once the process enters
mount_ordered_layout partition['mountpoint'] will not be touched (Believe
[...]

PR #838, opened 2022-01-04
Btrfs II

wllacer:
For master:
This is the second installement of the btrfs patches, with something more general, complementing PR #787 
Features:
1) For all filesystem types, we have created two new keys in the layout. Former options are mantained for backward compatibility. (see issue #835)
```json
{
    "/dev/loop0": {
        "partitions": [
            {
                ...
                "encrypted": false,
                "filesystem": {
                    "format": "btrfs",
                    "format_options":[                   for options to be used during the formatting of the fs (mkfs)
                        "--label liquindoy"
                    ],
                    "mount_options":[                  for options to be used at the mount of the partition/subvolume (mount)
                       "noatime",
                        "compress=lzo"
                    ]
                },
                "format": true,
                "mountpoint": null,
            }
        ],
        "wipe": true
    }
}
```
2) Now **btrfs** partitions (and subvolumes) can be encrypted and managed like any other filesystem via passwords and key files for not root partitions. Naturally we respect the upstream limitation than on **btrfs**, encryption is for full partitions (all the subvolumes inside, without exception)

3)  In **btrfs**, a certain level of compression (*zstd*) can be defined on the subvolume level, and more sophisticated modes at the partition level, we now allow both levels of definition, only that on the subvolume level, instead of being defined at mount point, it is handled via extended attributes.

# Testing
Till it is merged, you can test it at  my branch `https://github.com/wllacer/archinstall/tree/btrfs-encrypt`
Remember there is a script **only-hd** which only handles partitioning and mounting, not proceeding with the real arch installation.

# Limitations
There is still no UI. All use has to be via disk layout files


wllacer:
A few technical details.
Most of the processing was streamlined when we discovered that we could subtitute the luks generated Partition object (unlocked device) for the normal Partition object at the directory entry partition['device_instance']. This way no special processing was needed during the mount phase for encrypted  partitions / subvolume mountpoints and their counterparts in the clear. It also allowed for some simplification of the manage_btrfs_subvolumes function.

This led to a total overhaul of Installer.mount_ordered_partition. Now this method is based on four loops thru a list of partitions dictionaries.:
* The first one open the encrypted partitions and creates a list of those which need the creation of a keyfile. As the creation of the object is instanced with *auto_unmount= false*, they are kept open as long as we need them
* The second one selects the btrfs partitions, creates the subvolumes and adds the new "partition dictionaries" needed for all the subvolumes which will be mounted
*  The third one selects all the entries which will be mounted and does the mounting. There is where we have put the support for mount options. Btw. it solves again issue #813, but with some different coding style as the solution from @Torxed two days ago (we use no more the old mountpoints dictionary).
* The forth and last is only for the unlocked devices which need the creation of keyfiles. Putting it at the end avoids some errors if the keyfile was created before the / directory had been mounted (not depending on the UI, but only from user written json files has this drawback)
One of the advantages we see with this approach is that it allows easy extensions in case other filesystems behave as divergent of the age old standard as **btrfs**
I like the code, but will accept that perhaps i overplayed a bit the use of list comprehensions here

wllacer:
A full example of a disk layout file for btrfs without encryption
```json
{
    "/dev/loop0": {
        "partitions": [
            {
                "boot": true,
                "encrypted": false,
                "filesystem": {
                    "format": "fat32"
                },
                "format": true,
                "mountpoint": "/boot",
                "size": "513MiB",
                "start": "5MiB",
                "type": "primary"
            },
            {
                "btrfs": {
                    "subvolumes": {
                        "root": "/",
                        "home":{
                            "mountpoint":"/home",
                            "options":["compress"]
                        },
                        "var": {
                            "mountpoint":"/var",
                            "options":["nodatacow"]
                        },
                        "var/cache/pacman":null,
                        "var/lib/pacman": null,
                        ".snapshots": null
                    }
                },
                "encrypted": false,
                "filesystem": {
                    "format": "btrfs",
                    "format_options":[
                        "--label liquindoy"
                    ],
                    "mount_options":[
                        "noatime",
                        "compress=lzo"
                    ]
                },
                "format": true,
                "mountpoint": null,
                "size": "100%",
                "start": "518MiB",
                "type": "primary"
            }
        ],
        "wipe": true
    }
}
```
And another one with encryption (and two partitions)
```json
{
    "/dev/loop0": {
        "partitions": [
            {
                "boot": true,
                "encrypted": false,
                "filesystem": {
                    "format": "fat32"
                },
                "format": true,
                "mountpoint": "/boot",
                "size": "513MiB",
                "start": "5MiB",
                "type": "primary"
            },
            {
                "btrfs": {
                    "subvolumes": {
                        "root": "/",
                        "var": {
                            "mountpoint": "/var",
                            "options": [
                                "space_cache=v2",
                                "nodatacow"
                            ]
                        }
                    }
                },
                "encrypted": true,
                "filesystem": {
                    "format": "btrfs"
                },
                "format": true,
                "generate-encryption-key-file": true,            this partition includes the / mountpoint. key file will NOT be generated
                "mountpoint": null,
                "size": "4092MiB",
                "start": "518MiB",
                "type": "primary"
            },
            {
                "btrfs": {
                    "subvolumes": {
                        "home": {
                            "mountpoint": "/home",
                            "options": [
                                "space_cache=v2",
                                "compress",
                                "noatime"
                            ]
                        },
                        "opt": {
                            "mountpoint": "/opt",
                            "options": [
                                "space_cache=v2",
                                "noatime"
                            ]
                        },
                        "srv": {
                            "mountpoint": "/srv",
                            "options": [
                                "space_cache=v2",
                                "noatime"
                            ]
                        }
                    }
                },
                "encrypted": true,
                "filesystem": {
                    "format": "btrfs"
                },
                "format": true,
                "generate-encryption-key-file": true,
                "mountpoint": null,
                "size": "100%",
                "start": "4610MiB",
                "type": "primary"
            }
        ],
        "wipe": true
    }
}
```

Torxed:
This is some solid work to improve the life of the btrfs handling!

wllacer:
Having a look at my own "to-do" list, i recalled that there is an aditional bootloader option not in the general execution called **efistub**.  Most probably it needs some "btrfs care". My question is which priority this point could has for Archinstall as a project?

Torxed:
> Having a look at my own "to-do" list, i recalled that there is an aditional bootloader option not in the general execution called **efistub**. Most probably it needs some "btrfs care". My question is which priority this point has for Archinstall as a project?

Very little as of now. It's hidden behind `--advanced` or through careful manuall addition in the JSON syntax.
More info can be found here #716. So you can safely ignore that for now, as users using it will be aware of any shortcomings.

wllacer:
> > Having a look at my own "to-do" list, i recalled that there is an aditional bootloader option not in the general execution called **efistub**. Most probably it needs some "btrfs care". My question is which priority this point has for Archinstall as a project?
> 
> Very little as of now. It's hidden behind `--advanced` or through careful manuall addition in the JSON syntax. More info can be found here #716. So you can safely ignore that for now, as users using it will be aware of any shortcomings.
Thanks for the input
For what i've read in the arch wiki, and seen in the code, changes seem to be pretty trivial, (I'd say, copy the ones i did for systemd-bootctl); but the testing might not be. S,  implementing it will not be forgotten, but put on "low prio"

Torxed:
Great work on this whole btrfs fixes.
I'll run some tests against master as a whole. Nothing appears to break but even if it does we have plenty of time to fix things before `v2.4.0` : ) So merging!

Issue #789, opened 2021-12-06
Error when using generated disk layout


asininemonkey:
Getting the following error when using dry run generated configuration files:

```
$ sudo archinstall --dry-run --config user_configuration.json --disk_layouts user_disk_layout.json
Testing connectivity to the Arch Linux mirrors ...
Traceback (most recent call last):
  File "/usr/bin/archinstall", line 33, in <module>
    sys.exit(load_entry_point('archinstall==2.3.0', 'console_scripts', 'archinstall')())
  File "/usr/lib/python3.9/site-packages/archinstall/__init__.py", line 111, in run_as_a_module
    script.execute()
  File "/usr/lib/python3.9/site-packages/archinstall/lib/profiles.py", line 180, in execute
    self.spec.loader.exec_module(sys.modules[self.namespace])
  File "<frozen importlib._bootstrap_external>", line 850, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/usr/lib/python3.9/site-packages/archinstall/examples/guided.py", line 383, in <module>
    load_config()
  File "/usr/lib/python3.9/site-packages/archinstall/examples/guided.py", line 54, in load_config
    if (dl_path := pathlib.Path(archinstall.arguments['disk_layouts'])).exists() and str(dl_path).endswith('.json'):
  File "/usr/lib/python3.9/pathlib.py", line 1082, in __new__
    self = cls._from_parts(args, init=False)
  File "/usr/lib/python3.9/pathlib.py", line 707, in _from_parts
    drv, root, parts = self._parse_args(args)
  File "/usr/lib/python3.9/pathlib.py", line 691, in _parse_args
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not bool
```

Not really seeing anything that stands out:

```
{
    "audio": "pipewire",
    "bootloader": "systemd-bootctl",
    "disk_layout": true,
    "dry-run": true,
    "harddrives": [
        "/dev/nvme0n1"
    ],
    "hostname": "intel-nuc",
    "kernels": [
        "linux"
    ],
    "keyboard-layout": "uk",
    "mirror-region": {
        "United Kingdom": {
            "https://archlinux.uk.mirror.allworldit.com/archlinux/$repo/os/$arch": true,
            "https://lon.mirror.rackspace.com/archlinux/$repo/os/$arch": true,
            "https://mirror.bytemark.co.uk/archlinux/$repo/os/$arch": true,
[...]

PR #844, opened 2022-01-06
Upgrade script only_hd with some changes at master

wllacer:
I forgot that *only_hd.py* was already upstream, and made all the needed changes to follow *master* in the *btrfs II* branch. As a result, #789 has reappeared (but only here)

- No treatement of disk_layouts_file
- Ordered exit

### DIFF
--- a/examples/only_hd.py
+++ b/examples/only_hd.py
@@ -4,7 +4,6 @@ import os
 import pathlib
 
 import archinstall
-import glob
 
 def load_mirror():
 	if archinstall.arguments.get('mirror-region', None) is not None:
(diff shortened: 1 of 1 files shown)


Torxed:
Do you mind fixing this:
```
./archinstall/lib/simple_menu.py:61:1: F841 local variable 'e' is assigned to but never used
except ImportError as e:
```
I can't modify that file since it's not part of the PR.
But simply remove the `as e` and we should be fine :)

wllacer:
I Saw the lint error and wondered about It ...
Not at home till late in the evening. Will look into It.

PR #853, opened 2022-01-08
partition.py: Use exit code for mkfs.fat exception

rgoncalvesse:
When using archinstall on an existing Arch Linux installation, (e.g. for migrating the current system on a new drive), no exception is raised if mkfs.vfat is missing in the base install (the dosfstools package is not installed).

NOTE: I have just tested this change with the context described above.


Torxed:
It's better to use the exit codes anyway, using output data is unreliable at best.
This was an old surviving code from the old days before the installer even became a library.
Thank you for cleaning up the code!

PR #854, opened 2022-01-08
Rename Partition attribute "size" to "end"

HTV04:
Internally, the "size" attribute of a Partition is treated as an end point. This PR fixes this as discussed in #794.

The changes from 7e959cd were made as a result of this issue and are reverted in this PR.

PR is currently drafted because I'm not sure I replaced all references of "size" to "end" yet.

### DIFF
--- a/archinstall/lib/disk/filesystem.py
+++ b/archinstall/lib/disk/filesystem.py
@@ -7,7 +7,7 @@ from typing import Optional, Dict, Any, TYPE_CHECKING
 # https://stackoverflow.com/a/39757388/929999
 if TYPE_CHECKING:
 	from .blockdevice import BlockDevice
-	
+
 from .partition import Partition
 from .validators import valid_fs_type
 from ..exceptions import DiskError
@@ -78,7 +78,7 @@ class Filesystem:
 				print("Adding partition....")
 				partition['device_instance'] = self.add_partition(partition.get('type', 'primary'),
 																	start=partition.get('start', '1MiB'), # TODO: Revisit sane block starts (4MB for memorycards for instance)
-																	end=partition.get('size', '100%'),
+																	end=partition.get('end', '100%'),
 																	partition_format=partition.get('filesystem', {}).get('format', 'btrfs'))
 				# TODO: device_instance some times become None
 				# print('Device instance:', partition['device_instance'])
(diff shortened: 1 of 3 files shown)


Torxed:
Would it be possible to add backwards compatibility here? Because this would surely break a lot of user-stored configurations.
At least in the `.get()` statements.

vgavro:
> Would it be possible to add backwards compatibility here?
I don't think that backward compatibility is the case here, it's like "stop calling black white, but with backward compatibility". 
1. They configs SHOULD break, because this field is not size. I had to reformat my SSD three times before I got that, and some who has stored configs may not realize that (or checked) because it's pretty common to have root and home (without var).
2. There is NO POSSIBILITY to put disk config in last released version (at least there is no such option in --help), only in master.

wllacer:
I've been bitten with this in some new development. Not for tomorrow, but, how's the status of this ? Seems stuck in draft status since forever.
As long as we never use the `size` string again, this can be made compatible very easily ... only that i need exactly that attribute. Any suggestion for a new name? My `size` was to be _size in sectors_, and i was about to create another (provisionally called `sizeG` as _size_as_entered_

wllacer:
Forget about the "As long as we never ..."  We can get full backwards compatibilty without compromising future development
What we need is to incorporate, with this change, the __version__ attribute into the user_disk_layout.json, as we do now at the config file
Currently the version is not written at disk_layout_file, but it is just changing  line 66 to the corresponding copy of the lines 58-62
https://github.com/archlinux/archinstall/blob/ac0162aba7405f87f17cfd99c99f7d540482ee6c/archinstall/lib/configuration.py#L57-L67
Then, at https://github.com/archlinux/archinstall/blob/ac0162aba7405f87f17cfd99c99f7d540482ee6c/archinstall/__init__.py#L222-L235
just add before lnr 235 (pseudocode, not indented properly, should align with the `format` test)
``` 
if `size` in partition and layout_storage.get('version','2.3.3') < '2.4':
     partition['end'] = partition['size']
     del partition['size']
```

wllacer:
Here are the changes i proposed (with slight changes due to reality) in patch format
[0001-add-a-layout_version-entry-for-disk_layout.json.patch.txt](https://github.com/archlinux/archinstall/files/8350732/0001-add-a-layout_version-entry-for-disk_layout.json.patch.txt)
The version entry at disk_layout is called 'layout_version' and is deleted after being processed, so it is regenerated every time and will not interfere with later usage of the file

Torxed:
> Here are the changes i proposed

I can merge that in, give me a few minutes :)

svartkanin:
This PR is quite old and outdated. I'll close it for now, but if you would like re-raise an updated one please feel free to do so and we'll have a look

PR #856, opened 2022-01-09
Created a save_user_configurations

### LINKED ISSUE #855: Credentials don't get written out to file, if installation is interrupted before perform_installation() is called
image

Torxed:
This should avoid accidental placement of individual json outputs in a way that some doesn't get saved before installation continues etc.

This fixes #855


Torxed:
Configurations are now properly saved before `Press Enter to continue.`
screenshot

PR #857, opened 2022-01-09
Restore generic_select function

wllacer:
We have restored the generic_select and generic_multi_select functions which were missing. Adapted to the new interface

### DIFF
--- a/archinstall/lib/menu/menu.py
+++ b/archinstall/lib/menu/menu.py
@@ -1,15 +1,19 @@
 from archinstall.lib.menu.simple_menu import TerminalMenu
+from ..exceptions import RequirementError
+from ..output import log
+
+from copy import copy
 
 
 class Menu(TerminalMenu):
-	def __init__(self, title, options, skip=True, multi=False, default_option=None, sort=True):
+	def __init__(self, title, p_options, skip=True, multi=False, default_option=None, sort=True):
 		"""
 		Creates a new menu
 
 		:param title: Text that will be displayed above the menu
 		:type title: str
 
-		:param options: Options to be displayed in the menu to chose from;
+		:param p_options: Options to be displayed in the menu to chose from;
 		if dict is specified then the keys of such will be used as options
 		:type options: list, dict
 
@@ -25,6 +29,18 @@ class Menu(TerminalMenu):
 		:param sort: Indicate if the options should be sorted alphabetically before displaying
 		:type sort: bool
 		"""
+		# we guarantee the inmutability of the options outside the class
+		options = copy(p_options)
+
+		# Checking if the options are different from `list` or `dict` or if they are empty
+		if not isinstance(options, (list,tuple, dict)):
+			log(f" * Menu doesn't support ({type(options)}) as type of options * ", fg='red')
+			log(" * If problem persists, please create an issue on https://github.com/archlinux/archinstall/issues * ", fg='yellow')
+			raise RequirementError("Menu.__init__() requires list or dictionary as options.")
+		if not options:
+			log(" * Menu didn't find any options to choose from * ", fg='red')
+			log(" * If problem persists, please create an issue on https://github.com/archlinux/archinstall/issues * ", fg='yellow')
+			raise RequirementError('Menu.__init__() requires at least one option to proceed.')
 
 		if isinstance(options, dict):
 			options = list(options)
--- a/archinstall/lib/menu/menu.py
+++ b/archinstall/lib/menu/menu.py
@@ -30,10 +30,14 @@ class Menu(TerminalMenu):
 		:type sort: bool
 		"""
 		# we guarantee the inmutability of the options outside the class.
-		# but if somebody sends a dictionary.keys() or dictionary.values(), without putting it inside a list stmt, like
-		# list(dictionary.keys) the copy will fail, so we have to check the types (the syntax type({}.keys()) is the only
-		# way to denote a dicts_keys object (as the value part) and do the conversion ourselves
-		options = copy(list(p_options) if isinstance(p_options,(type({}.keys()),type({}.values()))) else p_options)
+		# an unknown number of iterables (.keys(),.values(),generator,...) can't be directly copied, in this case
+		# we recourse to make them lists before, but thru an exceptions
+		# this is the old code, which is not maintenable with more types
+		# options = copy(list(p_options) if isinstance(p_options,(type({}.keys()),type({}.values()))) else p_options)
+		try:
+			options = copy(p_options)
+		except TypeError:
+			options = copy(list(p_options))
 		# Checking if the options are different from `list` or `dict` or if they are empty
 		if not isinstance(options, (list,tuple, dict)):
 			log(f" * Menu doesn't support ({type(options)}) as type of options * ", fg='red')
--- a/archinstall/lib/menu/menu.py
+++ b/archinstall/lib/menu/menu.py
@@ -29,9 +29,11 @@ class Menu(TerminalMenu):
 		:param sort: Indicate if the options should be sorted alphabetically before displaying
 		:type sort: bool
 		"""
-		# we guarantee the inmutability of the options outside the class
-		options = copy(p_options)
-
+		# we guarantee the inmutability of the options outside the class.
+		# but if somebody sends a dictionary.keys() or dictionary.values(), without putting it inside a list stmt, like
+		# list(dictionary.keys) the copy will fail, so we have to check the types (the syntax type({}.keys()) is the only
+		# way to denote a dicts_keys object (as the value part) and do the conversion ourselves
+		options = copy(list(p_options) if isinstance(p_options,(type({}.keys()),type({}.values()))) else p_options)
 		# Checking if the options are different from `list` or `dict` or if they are empty
 		if not isinstance(options, (list,tuple, dict)):
 			log(f" * Menu doesn't support ({type(options)}) as type of options * ", fg='red')
--- a/archinstall/lib/menu/menu.py
+++ b/archinstall/lib/menu/menu.py
@@ -2,7 +2,7 @@ from archinstall.lib.menu.simple_menu import TerminalMenu
 from ..exceptions import RequirementError
 from ..output import log
 
-from copy import copy
+from collections.abc import Iterable
 import sys
 import logging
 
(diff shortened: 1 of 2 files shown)


wllacer:
@svartkanin As it is the first time i work upon your code, have a look please, if i haven't messed anything

archinstall/lib/user_interaction.py
```diff
@@ -939,3 +939,86 @@ def select_locale_enc(default):
...
+		sort=sort
+	).run()
+	# we return the original objects, not the strings
+	if not selected_option:     # empty
+		return selected_option
+	elif isinstance(selected_option,list):  # for multi True
+		selected_option = list(map(lambda x: options[soptions.index(x)],selected_option))
+	else:                                 # for multi False
+		selected_option = options[soptions.index(selected_option)]
+	print(selected_option)
```
svartkanin, line 1009: This should probably be removed

archinstall/lib/user_interaction.py
```diff
@@ -939,3 +939,86 @@ def select_locale_enc(default):
...
+		third option
+	When the user has entered the option correctly,
+	this function returns an item from list, a string, or None
+
+	Default value must be in the list of options
+	"""
+	# I (shallow) copy the options object. This is an input only parameter
+	options = copy(p_options)
+	# Checking if the options are different from `list` or `dict` or if they are empty
+	if type(options) not in [list, dict]:
```
svartkanin, line 967: I wonder if it wouldn't be better to move those validations of menu parameters into the menu class since it's the same for every call essentially

archinstall/lib/user_interaction.py
```diff
@@ -939,3 +939,86 @@ def select_locale_enc(default):
 	).run()
 
 	return selected_locale
+
+def generic_select(p_options :Union[list,dict],
+				input_text :str = "Select one of the above by index or absolute value: ",
```
svartkanin, line 944: Can we change the "above" to "below" because the options are under the text :) (some for the multi select function)

svartkanin (changes requested):
Great work on this! I left some comments for minor things

archinstall/lib/user_interaction.py
wllacer: The print stmt. of course. 
The rest sadly can't be removed. In the old select* anything could be thrown in as menu entries, and therefore selected. With the Menu  interface only "string or byte like" can be handled. This code, ugly as it is, and a few lines before the call, is needed for in-place replacement. It retrieves and returns the original object instead of the textual menu option.
OTOH it has to handle single and multiple response, which makes it even more convoluted

archinstall/lib/user_interaction.py
wllacer: Haven't checked if they already are. I lifted them directly from the old version (with the exception of the copy, which is my particular obsession).
Will have a look

archinstall/lib/user_interaction.py
wllacer: You're totally right ;-) And, there is no index anymore (I grew with IBM's CUA and miss them)

archinstall/lib/user_interaction.py
svartkanin: Sorry I just meant the "print" and "exit" statements :)

wllacer: Oooops ...

archinstall/lib/user_interaction.py
wllacer:
```python
	# I (shallow) copy the options object. This is an input only parameter
	options = copy(p_options)
```
I need it very much propagated (see below about sort), but it needs to stay here (I do some specific preprocessing)
 ```python
	# Checking if the options are different from `list` or `dict` or if they are empty
	if type(options) not in [list, dict]:
		log(f" * Generic select doesn't support ({type(options)}) as type of options * ", fg='red')
		log(" * If problem persists, please create an issue on https://github.com/archlinux/archinstall/issues * ", fg='yellow')
		raise RequirementError("generic_select() requires list or dictionary as options.")
	if not options:
		log(" * Generic select didn't find any options to choose from * ", fg='red')
		log(" * If problem persists, please create an issue on https://github.com/archlinux/archinstall/issues * ", fg='yellow')
		raise RequirementError('generic_select() requires at least one option to proceed.')
```
both can be moved to menu, but add some new dependencies (log and the exception). What do you think ?
```python
	# After passing the checks, function continues to work
	if type(options) == dict:
		# To allow only `list` and `dict`, converting values of options here.
		# Therefore, now we can only provide the dictionary itself
		options = list(options.values())
```
This is exactly the opposite of what Menu does (it uses the keys as entries by default). Must stay here
```python
	# check that the default value is in the list
	if default and default not in options:
		log(f" * Default option {default} not in the list of options ", fg='red')
		log(" * If problem persists, please create an issue on https://github.com/archlinux/archinstall/issues * ", fg='yellow'
		raise RequirementError('generic_select() requires the default to be on the list.')
```
AFAIK Menu does not make any assumption about it, which i'd liked to do. But it helps me to make a cleaner code ;-)
Don't think it should be propagated, in general. Even here, it could eventually disappear
```python 
	if sort:
		# As we pass only list and dict (converted to list), we can skip converting to list
		options = sorted(options)
```
Ooops, the whole sort can create very difficult to trace bugs, unless inmutability of the lists can be guaranteed, and even then as the sort here and the one at menu are not against the same object type.

The next upload will have the copy and sort questions "sorted out". I wait to your feedback about pushing depenedencies into menu

wllacer:
I pushed a new version, addressing all the comments by @svartkanin. Incluiding the checks at *menu.py*

wllacer:
**Errors found**. 
The copy statement at menu, gives sometimes (at select_mirror_regions) an exception:
```
 File "/home/werner/projects/archinstall/archinstall/lib/menu/menu.py", line 33, in __init__
    options = copy(p_options)
  File "/usr/lib/python3.10/copy.py", line 92, in copy
    rv = reductor(4)
TypeError: cannot pickle 'dict_keys' object

```
and also an exception at select_profile
```
  File "/home/werner/projects/archinstall/archinstall/lib/user_interaction.py", line 790, in select_profile
    selection = Menu(title=title, options=options.keys()).run()
TypeError: Menu.__init__() got an unexpected keyword argument 'options'
```
I stumbled on them doing some related tests. Will have a look on them. The second is obvious ... I changed the name of the parameter. I didn't expect the options to be called that way... Ouch.

Torxed:
First one is "easy" too:
```python
options = copy(list(p_options))
```
I don't know why on this green earth cPython didn't implement `__dict__` or `__getstate__` so that it's pickable.
There's also other instances where `dict_keys` behave *oddly*. So we'll need to safely convert any `.keys()` to an actual `list(.keys())`. It reduces performance, but overall that's nothing for this project since we don't lack speed as of now : )

wllacer:
> First one is "easy" too:
> 
> ```python
> options = copy(list(p_options) if type(p_options) == dict_keys else p_options)
> ```
> 
> I don't know why on this green earth cPython didn't implement `__dict__` or `__getstate__` so that it's pickable. There's also other instances where `dict_keys` behave _oddly_. So we'll need to safely convert any `.keys()` to an actual `list(.keys())`. It reduces performance, but overall that's nothing for this project since we don't lack speed as of now : )

Python is wonderful, except when it isn't
```
  File "/home/werner/projects/archinstall/archinstall/lib/menu/menu.py", line 34, in __init__
    options = copy(list(p_options) if type(dict_keys) else p_options)
NameError: name 'dict_keys' is not defined

```
The correct answer (a lot of google went into this) is 
```
		options = copy(list(p_options) if type({}.keys()) else p_options)
```
although the more general solution would be
```
		options = copy(list(p_options) if isinstance(p_options,(type({}.keys()),type({}.values()))) else p_options)
```
I was tempted to leave it as it was, and let Archinstall crash whenever someone forget the list() function, and flog the offender afterwards ... but being able to write the above, definitively makes me an entrant to the "obfuscated python contest of 2022" ;-)

wllacer:
Uploaded a corrected version. (btw. I updated a few .keys and .values without the list around the menu :-(

wllacer:
Hope this change ends the saga of "unpickable types".
 The message is just a flicker on the screen, but at least it will appear at install.log
A warning, the same code could go into *generic_select*. I've decided against, simply because the crash should alert the user that he should use a list. A bit mean, but ...

archinstall/lib/user_interaction.py
```diff
@@ -939,3 +939,83 @@ def select_locale_enc(default):
...
+	this function returns an item from list, a string, or None
+
+	Default value must be in the list of options
+	sort will be handled by Menu()
+	"""
+	# I (shallow) copy the options object. This is an input only parameter
+	options = copy(p_options)
+	# Checking if the options are different from `list` or `dict` or if they are empty
+	# duplicate of menu, but i need to check it prior as i do some type dependent processing
+	if not isinstance(options, (list,tuple, dict)):
```
svartkanin, line 969: Since we moved these checks now into the menu class I think we can savely remove them from here now :)

Torxed:
I'm fine with implementing
```python
options = copy(list(p_options) if isinstance(p_options,(type({}.keys()),type({}.values()))) else p_options)
```
But I'm wondering out of curuisity if there's a pythonic way of detecting and converting in a more generic way.
I'm wondering if they just didn't think of people calling `sort` and `copy` etc on the object or if this is a design choice.

`dict_keys` isn't a stdtype either >_<

archinstall/lib/user_interaction.py
wllacer: I still need to check, in the module, if the options is an iterable, to be able to do the map processing further on. And the copy ... Might be if I do 
```python
if type(p_options) == dict:
    options =list(p_options.values)
else:
    options = list(p_options)
```
I solve almost everything. Well, if it isn't an iterable it will crash, but we can live with it. It should only happen during development time.
Have to do a couple of tests

wllacer:
@Torxed. You missed the last installment of the saga ... A third type appeared ( *generator"). See https://github.com/archlinux/archinstall/pull/857#issuecomment-1009459909. This is becoming too barroque (as the type trick doesn't work) to be maintenable
The pythonic way is to keep it simple and better ask for forgiveness ... so the way would be simply let a TypeError exception, as in the comment above.
I have to test the other solution i outlined talking to @svartkanin ( https://github.com/archlinux/archinstall/pull/857#discussion_r78204733): proceed to the list() function directly, making the copy -or whatever- implicit. Most probably,this is the Way

wllacer:
Uploaded a corrected version. Now it allows at generic_select default values without restriction.
Finally i settled with the simpler solution ... doing the list () function upfront, it copies the iterable (so whatever happens the parameter isn't touched) and handles all iterable types. Strings are considered as such, but as each character would correspond to an entry, the mistake (if such is) will be apparent.
I only keep a verification -and the resulting exception- and it is to check if they are Iterables. Seems to work on all the types the copy failed. I've mantained the verification in both modules -when checking i prefer to be exhaustive in each call of the stack-. I've kept the exception as it was already there;  although, as it is an exception that should only happen during development time; I think no user exception needs to be raised. But this is a discussion for another day (after 2.4 perhaps)

svartkanin:
Aren't the checks in the generic function and the menu the same? For maintainability reasons I'd probably vote for keeping only the one in the menu unless there's a good reason to have them in both?

wllacer:
There is only one remaining common check; the one which checks if p_options is an Iterable. I grew up with the standard that parameter checks have to be done -if needed- everywhere where they are used locally. I simply follow this convention ... It is practical if, somehow, the backends (the menu engine) are changed.
If i omit the check in *generic_select* an equivalent system exception would spring on the ``` options = list(p_options) ``` statement, and never reach Menu() 
In Archlinux,  an user exception -or check- in this case, can be useful as they are reflected (thru the log() statements) in the **install.log**; which the system ones aren't. Should never happen outside development, but ,as a general measure, is useful in debugging user problems.

archinstall/lib/menu/menu.py
```diff
@@ -35,24 +35,25 @@ def __init__(self, title, p_options, skip=True, multi=False, default_option=None
...
-		# Checking if the options are different from `list` or `dict` or if they are empty
-		if not isinstance(options, (list,tuple, dict)):
-			log(f" * Menu doesn't support ({type(options)}) as type of options * ", fg='red')
-			log(" * If problem persists, please create an issue on https://github.com/archlinux/archinstall/issues * ", fg='yellow')
-			raise RequirementError("Menu.__init__() requires list or dictionary as options.")
+		# We check that the options are iterable. If not we abort. Else we copy them to lists
+		# it options is a dictionary we use the values as entries of the list
+		# if options is a string object, each character becomes an entry
+		# if options is a list, we implictily build a copy to mantain immutability
+		if not isinstance(p_options,Iterable):
```
Torxed, line 42: Clean way!

svartkanin:
Thanks for the clarification :) 
I'm happy with changes otherwise, great job!

Torxed:
Let me know when you consider this ready for merging :)

wllacer:
The original intent of providing a compatibility layer, i believe, is
achieved. Haven't had any new issue I think it is mergeable right now.

Torxed:
Any future issues can be solved in a new commit :) great work, you two are quick on this menu system :)

PR #863, opened 2022-01-14
Changed the default btrfs layout to have @ as /

Torxed:
This should be inline with https://github.com/archlinux/archinstall/pull/787 which was also merged into `v2.3.1` branch.
And follows the suggested layout here: https://github.com/archlinux/archinstall/issues/781


Torxed:
screenshot

Tested it and it mounts, decrypt and works.

PR #874, opened 2022-01-16
Migrate old input to new menu

svartkanin:
In this PR the NTP will be automatically set to true in the current system since that is also the default value from the menu. In case the value is changed by the user it will be immediately set in the system as well. 
 
In addition, some input methods have been migrated to the new menu as well


wllacer (commented):
I've been merging your PR with mine  (#866), i i have had a problem.
I've seen you put in a couple of places`from build.lib.archinstall import Menu` 
It fails.
The solution  is absolutely simple:
* Since a couple of commits the incluide is already at  \__init__.py, so if archinstall is imported, just call `archinstall.Menu`, and it will do the trick (f.i. at guided.py)
* For user_guides.py use
   `from ..menu.menu import Menu`
and it works now

For the time being, i plan to mantain both branches -my original and the one patched- separated, so if you change anything here, please warn me. I don't know what @Torxed will say, but having equivalent changes in two PR seems a bit weird. I'd prefer yours will be merged first. Mine, while in good shape, is lots more debatable and, perhaps worth some enhancements.

And to end, a question. The menu option for NTP executes the system command just after selection (I've done the same, in one test, but using the new *exec_func* hook). But in the menu, shouldn't it go to the new system instead to the host ?

svartkanin:
@wllacer yes that was my bad, pycharm aeems to get constantly confused what the Right import path is and I was too lazy so far to fix it properly. Thanks for pointing it out though I'll fix that!

svartkanin:
Regarding the ntp, as far as I understand from the related issues was some problems when selecting mirrors so that should all be on the host rather than the new system. I've just converted the existing code over so there shouldn't have been any logical changes

Torxed:
Yepp `archinstall.Menu` :) and we moved the NTP option first, and if selected Yes we run NTP on the live ISO and after/during installation we also enable NTP on the installation :)

svartkanin:
@Torxed I've fixed the imports and this should be ready now :)

profiles/sway.py
```diff
@@ -1,7 +1,7 @@
 # A desktop environment using "Sway"
+from archinstall import Menu
```
Torxed, line 2: Technically you don't even need this import if you do `archinstall.Menu()` in the code :) saves one import

svartkanin: True that :) 
I've removed the imports now and access it directly with `archinstall.Menu`

Torxed:
Awesome, if someone could verify that the affected profiles install correctly I'd be happy to merge. I can't verify until sunday :)

Torxed:
Working on this and I'm about to merge it.
Just need to fix some imports and make sure it runs : )

svartkanin:
@Torxed thanks for resolving all the conflicts!

Torxed:
> @Torxed thanks for resolving all the conflicts!

No worries :) Great way for me to learn the code.
I did some minor tweaks and changes too it too, like optimizing dictionary population in places.

I'll have this merged within the hour :)

Torxed:
Great work on this one! I have a few additional ideas that would bring this even further, for instance if it could remember last menu-entry-position when coming back to the menu after setting an option. Would save some `down down down enter` things :)

But this is really great work so far! Really like the look and color choices!

svartkanin:
Thanks @Torxed ! Regarding keeping a memory of the current selection and jumping to that one again after exiting a sub-menu, that should be fairly simple to accomplish in the archinstall code. I just saw that the simple_menu has a `preselect` argument which we might be able to use, I'll have a look at that!

Issue #880, opened 2022-01-20
Enable periodic trim


uhthomas:
[Periodic trim](https://wiki.archlinux.org/title/Solid_state_drive#Periodic_TRIM) may improve performance and increase the longevity of SSDs. [Continuous trim](https://wiki.archlinux.org/title/Solid_state_drive#Continuous_TRIM) is also an option, but is strongly discouraged for reasons explained in the associated article.

Periodic trim can be enabled with the `fstrim.timer` systemd unit. There should be no need to selectively enable the service as trimming has adverse effect on HDDs or other unsupported devices. Many distros enable periodic trim by default, see the [fedora changelog](https://fedoraproject.org/wiki/Changes/EnableFSTrimTimer) for example.

I believe this is a sensible default, and poses no obvious drawbacks or concerns given its widespread use and the infrequency of the timer.

dylanmtaylor:
I agree that it's probably a good idea idea to enable periodic trim on all supported filesystems. Definitely worth looking into. This looks like it would introduce additional packages but should result in better SSD performance and longevity.

uhthomas:
Glad you think it's a good idea!

It looks like the `fstrim.timer` systemd unit is owned by [`util-linux`](https://archlinux.org/packages/core/x86_64/util-linux/), a dependency of [`base`](https://archlinux.org/packages/core/any/base/) and systemd. Consequently, no additional packages are needed to enable periodic trimming.

```zsh
❯ pacman -Qo /usr/lib/systemd/system/fstrim.timer
/usr/lib/systemd/system/fstrim.timer is owned by util-linux 2.37.2-5
❯ pacman -Qi util-linux
Name            : util-linux
Version         : 2.37.2-5
Description     : Miscellaneous system utilities for Linux
Architecture    : x86_64
URL             : https://github.com/karelzak/util-linux
Licenses        : GPL2
Groups          : None
Provides        : rfkill  hardlink
Depends On      : pam  shadow  coreutils  systemd-libs  libsystemd.so=0-64  libudev.so=1-64  libcap-ng  libxcrypt  libcrypt.so=2-64
                  util-linux-libs  libmagic.so=1-64  libncursesw.so=6-64  libreadline.so=8-64
Optional Deps   : python: python bindings to libmount [installed]
                  words: default dictionary for look
Required By     : base  fakeroot  mkinitcpio  systemd  tracker3  zeromq
Optional For    : None
Conflicts With  : rfkill  hardlink
[...]

PR #882, opened 2022-01-20
Enable periodic TRIM

uhthomas:
Fixes https://github.com/archlinux/archinstall/issues/880

Would it be preferable to make this configurable? I can't envision a scenario where it would be desirable to have periodic TRIM disabled.


Torxed:
Could we add a comment/link/reference in the code as to why we're enabling it? For historical purposes

dylanmtaylor:
> Fixes #880
> 
> Would it be preferable to make this configurable? I can't envision a scenario where it would be desirable to have periodic TRIM disabled.

Maybe allow it to be disabled if the advanced flag is set. Otherwise I think it is a very same default and it's probably overwhelming to most users to offer the choice. Almost every other distribution does it by default.

uhthomas:
Sounds good to me!

I've added a comment which briefly explains periodic TRIM and links to the original issue.

Torxed:
Thank you for both suggesting and submitting a PR for this :) great work!

Issue #881, opened 2022-01-20
Disable zswap


uhthomas:
Archinstall uses [zram](https://www.kernel.org/doc/html/latest/admin-guide/blockdev/zram.html) for the swap device, which is great and the right thing to do. However, [zswap](https://wiki.archlinux.org/title/zswap) should be disabled when using zram.

> Tip: Since it is enabled by default, disable zswap when you use zram to avoid it acting as a swap cache in front of zram. Having both enabled also results in incorrect zramctl(8) statistics as zram remains mostly unused; this is because zswap intercepts and compresses memory pages being swapped out before they can reach zram.

https://wiki.archlinux.org/title/Improving_performance#zram_or_zswap

Torxed:
This feels borderline on being a post-installation thing. We follow the installation guidelines as much as possible, the only things we do on purpose is stuff that can't be changed post install. Like filesystem compression and stufff.

I'm not against this tho as it seems sensible. But I'm wondering why this isn't default?

uhthomas:
I imagine that zswap was enabled by default when using a "real" swap device was more common. It's only recently that zram seems to have become the recommended swap device and unfortunately they conflict in subtle ways.

I think it's sensible to disable zswap when using zram due to its undesirable behaviour. Zswap should be left alone in all other instances however.

PR #883, opened 2022-01-20
Disable zswap

uhthomas:
Disable zswap when using zram.

The swap device is now initialised before the bootloader, inline with the
installation guide.

Fixes https://github.com/archlinux/archinstall/issues/881


Torxed:
Same here, great and swift work! :) Thank you for improving the project!

PR #886, opened 2022-01-22
Command locales (second batch)

wllacer:
Flexibilize the definition of execution locale for OS commands executed via the SysCommand* interface:
* Defined a storage argument which holds the default
   ```python
	'CMD_LOCALE':{'LC_ALL':'C'}, # default locale for execution commands. Can be overriden with set_cmd_locale()
	'CMD_LOCALE_DEFAULT':{'LC_ALL':'C'}, # should be the same as the former. Not be used except in reset_cmd_loc
   ```
* Added functions to
* * unset the program own locales
* * reset to the program default locales
* * set a specific locale
* * A decorator to execute functions in the host locale environment and one to execute in the C environment
 See this sample code
``` python
import archinstall
from pprint import pprint

@archinstall.host_locale_environ
def lsblk(pepe):
	print(pepe)
	agay=[]
	for line in archinstall.SysCommand('lsblk'):
		agay.append(line.decode('utf-8').strip())
	return agay

@archinstall.c_locale_environ
def lsblk_it(pepe):
	print(pepe)
	agay=[]
	for line in archinstall.SysCommand('lsblk'):
		yield(line.decode('utf-8').strip())

# default status (LC_ALL=C)
for line in archinstall.SysCommand('lsblk'):
	print(line.decode('utf-8').strip())
# local environ
archinstall.unset_cmd_locale()
for line in archinstall.SysCommand('lsblk'):
	print(line.decode('utf-8').strip())
# we return to default
archinstall.reset_cmd_locale()
for line in archinstall.SysCommand('lsblk'):
	print(line.decode('utf-8').strip())
# we reset the chosen locales to its native status
archinstall.set_cmd_locale()
print(archinstall.storage['CMD_LOCALE'])
# we change some selected locales
archinstall.reset_cmd_locale()
archinstall.set_cmd_locale(numbers='es_ES.utf8',charset=None)
print(archinstall.storage['CMD_LOCALE'])
for line in archinstall.SysCommand('lsblk'):
	print(line.decode('utf-8').strip())
#using decorators
archinstall.reset_cmd_locale()
for line in lsblk(1):
	print(line)
archinstall.set_cmd_locale(general='es_ES.utf8')
for line in lsblk_it(1):
	print(line)

```


Torxed:
Looks pretty nice! I'll merge it tomorrow : )

wllacer:
Profiting that tt wasn't still merged i added a new decorator to execute a function always under the C locale (named *c_locale_environ*). I've renamed the other decorator to *host_locale_environ*
Sample code has been edited

Torxed:
Testing and merging this within the hour :)

Torxed:
I'm bringing in some other fixes that needed to get done before I could test this PR.
Most of them are unrelated to this particular code. So far everything looks good. Almost managed to run through at least one installation.

Torxed:
Stuck at setting key-map. Again, I don't think that's to do with this PR necessarily, but I can't get past it.

screenshot

Torxed:
If I run it manually, it works.
So something's up with our code.

screenshot

dylanmtaylor:
> If I run it manually, it works. So something's up with our code.
> 
> screenshot

Might be related to the "" argument at the end. I wonder how bash is interpreting this compared to the code. That positional argument might be handled as two double quotes instead of an empty positional argument or something.

Torxed:
So I've tracked down the issue, I think.
And it's this: https://github.com/archlinux/archinstall/blob/b0dac26524ececf324674b05d36ff794edeb7aff/archinstall/lib/general.py#L356

And by that, I mean anything that gets printed after `pty.fork()` appears to execute **as is**.
archinstall

Torxed:
This is the output of: https://github.com/archlinux/archinstall/blob/8a4eb590c706541c6d01b91ee4d0e19e41d6cec6/archinstall/lib/packages/packages.py#L102-L106

screenshot

Torxed:
Is it maybe because the child file descriptor, and the parent process share a common `stdout`?
If so this would mean that the following code will add any output by the parent process to the `SysCommand` trace log and subsequently a lot of result data: https://github.com/archlinux/archinstall/blob/8a4eb590c706541c6d01b91ee4d0e19e41d6cec6/archinstall/lib/general.py#L325

Torxed:
Nope, different pid and different stdouts. But somehow the output of the parent pollutes the forked process input..

Torxed:
Looks like I'll need to re-think some error handling in `SysCommand` due to how the forking works.
As anything written to `sys.stdout` ends up in the child fork as well between `pty.fork()` and `os.execve()` some how: https://discord.com/channels/267624335836053506/470884583684964352/938463662472060938

dylanmtaylor:
I see lots of noise on this PR, are we still stuck on something?

Torxed:
Yepp, can't set locale.

Torxed:
Sorry for the noise and spam, great work on this.
I'll revert some changes in `master` later, namely the use of `os.system` with no exit code checks.

Issue #841, opened 2022-01-05
RFC: load_config into __init__.py


wllacer:
When i updated the parameter handling, i moved the treatment of the disk_layouts file from *guided.load_config()* to *__init__,py*. It was simply a matter of putting all the config files handling in the same place, and the most logical one, because it will be always be executed whenever you use Archinstall.

Since then, i'm toying with the idea of moving everything from load_config() there also.

 On the positive side:
*  we can ensure that the **global** objects *archinstall.storage*, and *archinstall.arguments* are fully initialized when we access to them, 
* no user written script (or 3rd party application) misses some, which could be needed.
*  cost in memory and processing, seems basically trivial

On the negative side
* __init__.py grows too big and with more need of care, for such a critical piece of code
*  might be, we open the way to imprudent handling and unpredictable side-effects

What I don't know
*  What would be the consequences for 3rd party applications which use Archinstall ?
* Will it have an impact on the workflow in other scripts ?
 
What do you think of it ?

Torxed:
As long as the processing of the arguments are optional and allow us to continue to use archinstall without `disk_layouts`, `config` and `creds` for instance. Some times I just want to do:
```
import archinstall
archinstall.list_profiles()
```
Without any parsing or loading of arguments. And just use it as a library.
So as long as that is possible, I'm more than fine with this change in a PR to tinker with :)

PR #887, opened 2022-01-22
Moving load_config from guided to \__init__.py

wllacer:
As outlined at PR #841  we moved the contents of the load_config function from the scripts to \__init__.py.
We think that there is the proper place (as of now, the routine handles basically the conversion of certain contents of the config file from strings to archinstall objects. 
We haven't seen any negative effect till now, and less code to mantain on the scripts
A sample config file which would be processed by it is attached

[user_841.json.txt](https://github.com/archlinux/archinstall/files/7919363/user_841.json.txt)


examples/guided.py
```diff
@@ -22,34 +22,6 @@
 # For support reasons, we'll log the disk layout pre installation to match against post-installation layout
 archinstall.log(f"Disk states before installing: {archinstall.disk_layouts()}", level=logging.DEBUG)
 
-def load_config():
-	if archinstall.arguments.get('harddrives', None) is not None:
-		if type(archinstall.arguments['harddrives']) is str:
-			archinstall.arguments['harddrives'] = archinstall.arguments['harddrives'].split(',')
```
Torxed, line 28: Where did these logics go? Are they baked in some where else? :)

Torxed: Ah nvm, I just did `ctrl+f` and looked for it, but it's changed slightly to `arguments['harddrives'] = arguments['harddrives'].split(',')` heh. My bad!

wllacer: It did cost me a bit to understand what it was about. Probably some legacy config files to convert ... This lines are there since i started with the code

Torxed:
We might have some issues from this, but we'll sort those out in a new PR.
It's easier to keep track of things when the PR's are a bit smaller so patch PR's in case something goes bonkers is better for now.

Issue #873, opened 2022-01-16
Unless run with --advanced (on master), KeyError sys-language


panchoh:
Hi!

Trying `3d30a4eb28bd84713571c52622cb88b6458076a5 @ master`, on real hardware.

If I simply run `archinstall`, it eventually dies with a `KeyError`, regarding a missing `sys-language`.
After examining the code, I realized that the registration of default values for both `sys-language` and `sys-locale` is dependent on `archinstall` being called with the `--advanced` parameter.

(So, as a workaround, invoke `archinstall --advanced`).

I reckon that the registration of the default values should happen even if `--advanced` is not supplied.

Wonderful project, thanks!!!

![photo_2022-01-16_11-13-12](https://user-images.githubusercontent.com/471059/149655772-4d71f4fa-fd5c-4fc4-8dd6-d796d0cda2f6.jpg)

wllacer:
Running master branch?
I have the bug pinned down, but still not solved in full. I hope to have it today and a corresponding PR uploaded. I'll explain later

wllacer:
@panchoh  Gracias por notificar el problema. 
The problem is that the new menu infraestructure, sometimes, if an item was not selected didn't transfer its default value. I tried to solve it in the patch you were testing, but my conditions were too restrictive, trying to keep the config area as leanest as possible. Thanks to your report, we have been able to get, i hope, this problem from the table

Torxed:
Merged the code, it got me thinking where we store the arguments. Especially when we dump the config at the end as you mention, we don't want duplicate information and we don't want to mix creds, disk layout and config :)

wllacer:
It was a good idea to centralize all config information at  *arguments* and let *storage* only for "environment variables".
The problem of "printing" the config info as three separate files, is that the solution is either:
* have an internal structure *arguments['config']* for config data, *arguments['disk_layout']* ... for disk_layout and *arguments['creds']* for credentials. I hate the idea  to migrate it, and clashes with the new menu implementation (it expects a flat structure)
* Not to "print" directly from *arguments* but do it from intermediate areas. A bit of not-so-elegant code at the script(s), and probably some rearranging, but a lot less impact, on the short term.

wllacer:
More elegant than what i have now in the kitchen, but basically the  same. 
My solution loops only once thru the arguments dict, the former has to loop once per area.
[...]

PR #888, opened 2022-01-22
Routine to properly print and save config data

wllacer:
As outlined at issue #873 (in comments) there were some issues with the printing and saving of the config data ( aka *archinstall.arguments*) as some of them were duplicated.
We have prepared a routine which print and/or saves the data in json format, with the desired separations (general, disk layout and credentials -which aren't printed-)
`def output_configs(show :bool = True, save :bool = True):`
it's use is simple
```python
from archinstall.examples.commons import output_configs
....
`output_configs(show=False if archinstall.arguments.get('silent') else True)
```
What can be more controversial is where i did put it. As it was a function which is (and could be by others) easyly reused, i've put them in a module (called *commons*) at the *examples* directory. Why there?
It is a function which accesses the whole *arguments* dictionary, and uses a number of archinstall.* calls. The second can be solved, but the first implies that it can not go into most of the modules at *lib* directory without the risk of creating a circular dependency, or needing an explicit parameter for the dictionary, in what is a very specific function ONLY for this structure.
And is a function which a script developer would have some use to customize.

I have also with the new menu some similar issues (where to put code which ought to use the archinstall.* notation, and are third party developer customizable). They should go into a module which can do an ìmport archinstall` and can, with reason and some easy be user customizable. The obvious place in this case, now is at *examples*; but i don't know if in other situation (a great deal of code menu related) it would not be a better idea to have a sibling directory to  *lib* as the default, from which developers can overload in their own structure.


examples/guided.py
```diff
@@ -1,9 +1,9 @@
-import json
 import logging
 import os
 import time
 
 import archinstall
+from archinstall.examples.commons import output_configs
```
Torxed, line 6: I would say cross-importing examples is strictly forbidden.
The examples should only serve as an example of *"how would you use archinstall"* type of thing.

So I would say this is a blocker and I won't be able to merge this PR sadly.
If we move the logic of saving configuration to `lib/environment/configuration.py` or similar, that would work. We could also move the loading of configuration to that file.

I don't mind making those changes if you wish, but feel free to change if you have time and/or want to :)

Torxed (changes requested):

wllacer:
I knew this was about to happen ...
I'll try to get first an *lib* allocatable version of the output routine.
There is too much need for It.
And then we can look for a place for code like this ...

Torxed:
Sounds good, great work tho! I'll wait for new commits to this PR before merging, you do not need to open a new one. And no rush, I know you're busy at the moment :)

wllacer:
A bit late tonight ;-) It was not that complex to change, and got a quiet moment. With a litte effort and some more parameters, it could be a general solution for split printing dictionaries ... The location is just provisional, but not a roadblock anymore.
BTW. Have a look at the elements which go to user_credentials.json. I'm missing the `!root_password` argument. But it wasn't there before either.

Torxed:
`!root-password` is omitted, and only inserted if the user chose to set one during earlier steps. Maybe we should localize all those arguments close to each other so that it's easier to find the logic. Good work on this tho, merged! : )

Issue #807, opened 2021-12-21
Make this installer translatable, please


roxfr:
Hello,
Could you make this installer translatable, please?
Thank you.
Regards,
Roxfr

Torxed:
I don't mind this and I'll try to get on it for v2.4.0. Will be a fun one to solve neatly.

svartkanin:
@Torxed I'd be happy to pick this one up. Maybe we can have a discussion around it how to best do it, I have some ideas but just to double check since it will be quiet a significant change

Torxed:
> @Torxed I'd be happy to pick this one up. Maybe we can have a discussion around it how to best do it, I have some ideas but just to double check since it will be quiet a significant change

Go for it, we could use this issue, a discussion topic or a PM on Discord.
I'll be occupied for a bit tho : )

svartkanin:
My idea was to use the `gettext` module and create a new `locals` folder with all the translations in it. 
Then we can use them in the code with `text = _('Select harddrive')`

Torxed:
> gettext

That seems to be the best practice way, and it's a builtin. I very much like the idea! :)
If you've used that one before, feel free to start on it. I can help with replacing more strings once the skeleton is there.

svartkanin:
@Torxed I reckon we can close this?

Torxed:
Correct, translation has been added. Currently we have English, German and Spanish is on the way in a PR.

PR #893, opened 2022-01-23
Provide nationalization 

svartkanin:
@Torxed this is a first draft for the nationalization #807 for Archinstall. I've included a README.md in the `locales` folder as well to provide some information on a how to add new translations. 

New languages can be added simply by creating a new folder with the proper language abbrevation (see list `languages.json` if unsure).  
Run the following command to create a new template for a language
```
    mkdir -p <abbr>/LC_MESSAGES/ && touch <abbr>/LC_MESSAGES/base.po
```

After that run the script `./locales_generator.sh` it will automatically populate the new `base.po` file with the strings that 
need to be translated into the new language.  
For example the `base.po` might contain something like the following now 
```
#: lib/user_interaction.py:82
msgid "Do you really want to abort?"
msgstr ""
```

The `msgid` is the identifier of the string in the code as well as the default text to be displayed, meaning that if no
translation is provided for a language then this is the text that is going to be shown. 

To provide a translation for the language, simply write the translation in the `msgstr` part

```
#: lib/user_interaction.py:82
msgid "Do you really want to abort?"
msgstr "Wollen sie wirklich abbrechen?"
```

After that run the script once more `./locales_generator.sh` and it will auto-generate the `base.mo` file with the included translations.

### How things work under the hood:

In the code we mark strings that should be translated with a wrapper `_("translate me")` the `_` is a hacky thing that gets added to the Python builtins so it can be used everywhere. The only catch is that it can't be used in places that would run that code before it gets actually added to the builtins (maybe this can be improved). 

When running the `locales_generator.sh` it will search the code for all the occurances of the `_(...)` and put those strings into the `locales/base.pot` file. This is the template file that is used to create new language files.

The translations for a language are also located in the `locales/<abbr>/LC_MESSAGES/base.po` files from which the script generates `base.mo` files that are then used by `gettext` to provide the translations. 

The class `DeferredTranslation` was needed because `gettext` does not out of the box cater for changing translation on the fly. That means as soon as a translation is set one can't change that one any more while the program is running. With this "hacky" solutuion of the `DeferredTranslation` we create our own string wrapper and postpone the interpretation of the translation until it's actually applied, that way we can change the translation on the fly.


archinstall/lib/menu/selection_menu.py
```diff
@@ -1,5 +1,7 @@
 import sys
 
+from archinstall import Translation
```
Torxed, line 3: We should fix all these imports. They should be relative from /lib/ and not import archinstall, otherwise they would use the wrong version if both installed and git cloned and used "locally"

svartkanin:
@wllacer this will most likely cause some conflicts with the your work on the flexible menu system. I'm happy to help out out merging things into your branch as soon as things get finalized

wllacer:
> @wllacer this will most likely cause some conflicts with the your work on the flexible menu system. I'm happy to help out out merging things into your branch as soon as things get finalized
Great work. If I've seen correctly the conflict is limited to one place (the definition of the the menu options). It is -still- manageable.  And don't worry if you go after *user_interaction.py* next. There are no conflicts with me there. I have to check how it would impact my new code. And start preparing the spanish translation ...

archinstall/lib/menu/selection_menu.py
svartkanin: I've added the import for `Translation` in the __init__.py now as we do with others, is that what you meant?

Torxed: We have to do: https://github.com/archlinux/archinstall/blob/e65d9723bb72e2cdfe9935af63a4873414537943/archinstall/lib/user_interaction.py#L31

Otherwise this will fail if use as modules, git clones and other weird use cases : ) I suspect it's your IDE wanting you to import it like this because you have it installed as a module and it auto-detects this path.

Torxed:
Great work on this. I probably couldn't have solved it in a more transparent way myself :)

erickisos:
So, now can I create a PR to add a Spanish version? Amazing!!

svartkanin:
@erickisos Yes we can! There's a readme `archinstall/locales/README.md` that describes how to do it :)

erickisos:
Just in case, I checked that README but I didn't find the `./locales_generator.sh` script :(

svartkanin:
Just you're absulutely right, I just realized that I forgot to push this in the PR :( 
I'll raise a new PR with some updates for the nationalization and include it as well. If you wanna get started already this is the script 

```
#!/bin/bash

cd $(dirname "$0")/..

find . -type f -iname "*.py" | xargs xgettext -j --omit-header -d base -o locales/base.pot

for file in $(find locales/ -name "base.po"); do
	echo "Updating: $file"
	path=$(dirname $file)
	msgmerge --quiet --width 512 --update $file locales/base.pot
	msgfmt -o $path/base.mo $file
done
```
You can just copy past that into a file in the `archinstall/locales/` directory and run it with `bash locales_generator.sh`

erickisos:
Great, thank you @svartkanin , also, just pointing out in case someone else would want to help, we can use Poedit too to generate the needed files without "touching code".

svartkanin:
Rhanks for pointing that out, I didn't know about it, will have a look :)

wllacer:
@svartkanin  The integration of the nationalization code with my take on the menus, has worked (almost) flawless.
If i've understood correctly you activate the framework with a call
```
		self._translation = Translation.load_nationalization()
```
and then activate a language via 
```
		language = select_archinstall_language(default_lang)
		self._translation.activate(language)

```
Is the first step to be executed only once per session or for every class/module hierarchy you instantiate ?

svartkanin:
@wllacer the first step is a one time thing, I'm working on an update for it (hopefully ready tomorrow) where I moved that into the __init__.py actually

wllacer:
Thanks.
I think putting It at \__init__ is the most sensible thing to do.

PR #895, opened 2022-01-24
disk layout: allow to omit partition "start" option to start from previous partition end

vgavro:
disk layout: allow to omit partition "start" option to start from previous partition end.

Actually this is almost only possible solution not to have sector gaps between partitions, so I guess example configs should be also updated (not with this PR, though)

### DIFF
--- a/archinstall/lib/disk/filesystem.py
+++ b/archinstall/lib/disk/filesystem.py
@@ -18,7 +18,10 @@ from ..storage import storage
 GPT = 0b00000001
 MBR = 0b00000010
 
-DEFAULT_PARTITION_START = '1MiB' # TODO: Revisit sane block starts (4MB for memorycards for instance)
+# A sane default is 5MiB, that allows for plenty of buffer for GRUB on MBR
+# but also 4MiB for memory cards for instance. And another 1MiB to avoid issues.
+# (we've been pestered by disk issues since the start, so please let this be here for a few versions)
+DEFAULT_PARTITION_START = '5MiB'
 
 class Filesystem:
 	# TODO:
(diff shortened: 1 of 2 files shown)


vgavro:
Also note that `partition.end` property is actually size in sectors (which was also noted here https://github.com/archlinux/archinstall/pull/794#issuecomment-1007323704)
IMHO all this properties should be refactored - for example `size` is converted to GB for some reason, while `start` units is sectors.
I guess good refactoring would be:
start, end, size, sector_size should always return bytes and be cached (only sector_size is actually return bytes for now), conversion to other units should be in upper functions.

Torxed:
IIRC we have to work with `GiB`units and not `GB`.
We need a centralized function to unify all units to one unit. I think we're getting lucky in parted as it treats `GB` as `GiB` *(kinda?)*.

We have another PR that is trying to modify/fix this too I believe.
I just woke up and came back from vacation, so give me a few hours to re-read the code and see if I misunderstood anything here. Otherwise I'd be happy to merge during the day.

vgavro:
If you mean this https://github.com/archlinux/archinstall/pull/794/ - it affects only "auto-partitioning" suggested by interface (even commit message states that it just tweaks numbers), but manual partitining will still have this problem (first of all - you can't guess start in GB or GiB not to have gaps without knowing sector_size, and even if you know sector_size - you should write 'end' in bytes in such case. The thing that it's working for auto-partitioning now - is only because 512 sector_size, and manually calculated 'end' which luckily match bytes in GiB - with other sizes it may not much).
Perfectly your partition config should look something like: `[{'size': '10GB'}, {'size': '5GB'}, {'size': '100%'}]` and start/end should be detected by archinstall, but for now (while size is not implemented) at least it could be `[{'end': '10GB'}, {'end': '15GB'}, {'end': '100%'}]` after https://github.com/archlinux/archinstall/pull/854 merge..

Torxed:
Great work on this. Apologies for the delay in merging this, I've been occupied by both the release of `v2.3.1` and work, as well as trying to chime in to other arch projects.

PR #896, opened 2022-01-24
FIX: disk layout: partprobe should be called and checked only for target device

vgavro:
I was using archinstaller from usb flash, which gave error (even after reboot) on partprobe for usb flash for some reason.
archinstaller was failing on creating disk layout because of this, even if target device was not affected at all.
Obviously this is wrong, this PR fixes this.

### DIFF
--- a/archinstall/lib/disk/filesystem.py
+++ b/archinstall/lib/disk/filesystem.py
@@ -139,7 +139,7 @@ class Filesystem:
 				return partition
 
 	def partprobe(self) -> bool:
-		return SysCommand(f'bash -c "partprobe"').exit_code == 0
+		return SysCommand(f'bash -c "partprobe {self.blockdevice.device}"').exit_code == 0
 
 	def raw_parted(self, string: str) -> SysCommand:
 		if (cmd_handle := SysCommand(f'/usr/bin/parted -s {string}')).exit_code != 0:
--- a/archinstall/lib/disk/filesystem.py
+++ b/archinstall/lib/disk/filesystem.py
@@ -139,7 +139,7 @@ class Filesystem:
 				return partition
 
 	def partprobe(self) -> bool:
-		return SysCommand(f'bash -c "partprobe {self.blockdevice.device}"').exit_code == 0
+		return SysCommand(f'partprobe {self.blockdevice.device}').exit_code == 0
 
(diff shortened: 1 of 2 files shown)


vgavro:
Also - is there any good reason to call partprobe inside another bash process? I guess this should be explicitly explained in comment or removed at all.

Torxed:
To my knowledge there's no exlicity reason for calling `bash -c "partprobe"`, we can probably call `/usr/bin/partprobe` directly yes.

dylanmtaylor:
> To my knowledge there's no exlicity reason for calling `bash -c "partprobe"`, we can probably call `/usr/bin/partprobe` directly yes.

I think that would work just fine.

vgavro:
removed unnecessary `bash -c ".."`. Also I've just found out that in `blockdevice.py` `partprobe` was called without `bash -c` and with explicit device path (like this PR do in `partition.py` and `filesystem.py`).

Also i've just found out some thing for crypted devices: If we call `self.blockdevice.partprobe()` - it will be called on crypted device path
https://github.com/archlinux/archinstall/blob/8ebdaa706d3facaef60f5a7286d4356f6477cea7/archinstall/lib/disk/blockdevice.py#L226

 but with this PR (`partprobe self.blockdevice.device`) - is called on parent kernel device path
https://github.com/archlinux/archinstall/blob/8ebdaa706d3facaef60f5a7286d4356f6477cea7/archinstall/lib/disk/blockdevice.py#L109
I'm not sure which one is right.
Like I see from `filesystem.py` - `parted` is called on parent kernel device name https://github.com/archlinux/archinstall/blob/8ebdaa706d3facaef60f5a7286d4356f6477cea7/archinstall/lib/disk/filesystem.py#L178
So either this is another bug (with parted), or this PR is OK. Anyway my suggestion is merge this PR, and if this is really problem - new PR that affects both parted and partprobe may be opened.

Torxed:
Thank you for submitting this.
While working on a seemingly unrelated issue https://github.com/archlinux/archinstall/issues/689 it turns out:
```
Could not execute partprobe: Warning: Unable to open /dev/sr0 read-write (Read-only file system).  /dev/sr0 has been opened read-only.
```
Caused `add_partition` to error out, and return `None` causing bad `device_instance` objects.
Hopefully this solves that too.

Issue #894, opened 2022-01-23
KeyError: 'disk_layouts'


uhthomas:
There is no log, sorry. I selected the default disk layout with ext4.

Using the head of master.

![image](https://user-images.githubusercontent.com/9749173/150680292-b2f980a5-b7fd-4ca1-b45b-6595878d3852.jpeg)

uhthomas:
I suspect this is because the new UI allows the encryption key to be set after specifying the disk layout.

wllacer:
No, it was a problem with a change of the location of some variables at archinstall.arguments vs archinstall.storage. Was solved at some development branches but not at master
PR on its way

Torxed:
PR merged, should be fixed now. I'm doing a lot of testing on master today.
And as always, `master` is expected to be a bit unstable. But testing with it is highly appreciated! :)

wllacer:
The PR #898 contains a little error (an exit() i left out from testing. PR #899 should solve it, or just go to selection_menu.py
https://github.com/archlinux/archinstall/blob/bb98594054939e469d07f704e70f597a8b379657/archinstall/lib/menu/selection_menu.py#L243-L245
And delete the exit() statement at line 245.  Sorry for the mistake

Torxed:
> The PR #898 contains a little error (an exit() i left out from testing. PR #899 should solve it, or just go to selection_menu.py
> 
> https://github.com/archlinux/archinstall/blob/bb98594054939e469d07f704e70f597a8b379657/archinstall/lib/menu/selection_menu.py#L243-L245
> 
> 
> And delete the exit() statement at line 245. Sorry for the mistake

Quick catch! merged :)

PR #898, opened 2022-01-24
move references to storage['disk_layouts'] in the new menu

wllacer:
(See issue #894) some references to disk_layouts, formerly at archinstall.storage now moved to archinstall.arguments, as per the new interface.
Moved self._post_processing() to the very last place of the execution loop


wllacer:
@Torxed. It had an exit() left in a place it shouldn't. I'll send an updated copy in a couple of minutes

PR #900, opened 2022-01-25
Fixed unmount bug in Filesystem.parted_mklabel  (e.g., /dev/loop0p1)

comrumino:
To reproduce the issue, we follow the steps in [Without a Live ISO Image](https://github.com/archlinux/archinstall#without-a-live-iso-image)
```shell
# truncate -s 20G testimage.img
# losetup -fP ./testimage.img
# losetup -a | grep "testimage.img" | awk -F ":" '{print $1}'
# pip install --upgrade archinstall
# python -m archinstall --script guided
```
Rerun the final step and observe that the partitions `/dev/loop0p*` are never unmounted. The proposed changes use the existing blockdevice class to discover defined partitions and their interface to safely unmount.

### DIFF
--- a/archinstall/lib/disk/filesystem.py
+++ b/archinstall/lib/disk/filesystem.py
@@ -204,12 +204,14 @@ class Filesystem:
 
 	def parted_mklabel(self, device: str, disk_label: str):
 		log(f"Creating a new partition label on {device}", level=logging.INFO, fg="yellow")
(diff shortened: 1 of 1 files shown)


Torxed:
Thank you for submitting this PR :)
I'll verify shortly that it doesn't cause any unexpected side effects and merge it.

Just want to get #902 merged in before I touch this branch as I want those changes to go in here too.

PR #902, opened 2022-01-25
Fixing the 'device_instance' being None in some partition places

Torxed:
One minor issue was that `add_partition` was defined to return `-> None`.
Another issue might be uuid comparison using small/capital letters when comparing, so unified that.
Added a bunch of debugging and placed an exception in case `add_partition` doesn't return anything at all.

### DIFF
--- a/archinstall/lib/disk/filesystem.py
+++ b/archinstall/lib/disk/filesystem.py
@@ -221,7 +221,7 @@ class Filesystem:
 		# TODO: This should never be able to happen
 		log(f"Could not find the new PARTUUID after adding the partition.", level=logging.ERROR, fg="red")
 		log(f"Previous partitions: {previous_partition_uuids}", level=logging.ERROR, fg="red")
-		log(f"New partitions: {new_uuid_set}", level=logging.ERROR, fg="red")
+		log(f"New partitions: {(previous_partition_uuids ^ {partition.uuid for partition in self.blockdevice.partitions.values()})}", level=logging.ERROR, fg="red")
 		raise DiskError(f"Could not add partition using: {parted_string}")
 
 	def set_name(self, partition: int, name: str) -> bool:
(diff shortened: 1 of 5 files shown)


archinstall/lib/disk/filesystem.py
```diff
@@ -181,31 +189,41 @@ def add_partition(self, partition_type :str, start :str, end :str, partition_for
...
+						log(f'Blockdevice: {self.blockdevice}', level=logging.ERROR, fg="red")
+						log(f'Partitions: {self.blockdevice.partitions}', level=logging.ERROR, fg="red")
+						log(f'Partition set: {new_uuid_set}', level=logging.ERROR, fg="red")
+						log(f'New UUID: {[new_uuid]}', level=logging.ERROR, fg="red")
+						log(f'get_partition(): {self.blockdevice.get_partition}', level=logging.ERROR, fg="red")
 						raise err
 				else:
 					count += 1
-					log(f"Could not get UUID for partition. Waiting for the {count} time",level=logging.DEBUG)
+					log(f"Could not get UUID for partition. Waiting for the {count} time",level=logging.WARNING, fg="yellow")
```
dylanmtaylor, line 215: Can we make the text something like 'Retry attempt 1 of 10.' instead of Waiting for the {count} time"? It's more grammatically correct since we're not doing '1st', '2nd', etc.

Torxed:
Found the possible culprit: `filesystem.partprobe` some times couldn't execute properly for different reasons.
This was passed silently allowing for `add_partition` (who calls `partprobe`) to return `None` when that happened.

Added an exception to `add_partition` so that it can't return nothing, but also added more debugging in case something like this ever happens again.

Torxed:
We also had to change and make sure `Partition()` objects are properly configured with a `BlockDevice()` so that subsequent calls to `.size` and similar attributes doesn't break.

PR #903, opened 2022-01-25
Make text for getting filesystems more grammatically correct

dylanmtaylor:
Use 'Retry attempt {count} of 10.' instead of Waiting for the {count} time". It's more grammatically correct since we're not doing '1st', '2nd', etc.


Torxed:
Thanks! : ) Pulled it into #902 as well which should hopefully fix `NoneType` on `device_instance`.

PR #905, opened 2022-01-25
Fixing 781 and it's btrfs mountpoints and kernel parameters

Torxed:
-  [X] Before running the equivalent of pacstrap, create the subvols on the toplevel
-  [x] Umount the toplevel and mount the rootfs subvol (`/@`) into `/mnt/archinstall`
-  [x] Mount other subvolumes such as `/@home`into their place inside the `/mnt/archinstall`
-  [X] Run `pacstrap` in `/mnt/archinstall`
-  [X] Run `genfstab -U >> /mnt/archinstall/etc/fstab` (the currently generated /etc/fstab is incorrect, but it should fix itself after solving the items above).
-  [x] Make sure the kernel cmdline has the `rootflags=subvol=/@` parameter, so it mounts the correct subvolume as root without having to use `btrfs subvolume set-default`. To accomplish this, configure the bootloader to pass that flag to the Linux kernel, example for systemd-boot:

```
title   Arch Linux
linux   /vmlinuz-linux
initrd  /intel-ucode.img
initrd  /initramfs-linux.img
options  rootflags=subvol=/@
```
## Tested
- [x] `systemd-boot` with subvolumes
- [x] `systemd-boot` + `luks2` with subvolumes
- [x] `grub` in BIOS with subvolumes
- [x] `grub` in BIOS + `luks2` with subvolumes
- [x] `grub` in UEFI with subvolumes
- [x] `grub` in UEFI + `luks2` with subvolumes

### DIFF
--- a/archinstall/lib/disk/blockdevice.py
+++ b/archinstall/lib/disk/blockdevice.py
@@ -13,7 +13,10 @@ class BlockDevice:
 			from .helpers import all_disks
 			# If we don't give any information, we need to auto-fill it.
 			# Otherwise any subsequent usage will break.
-			info = all_disks()[path].info
+			try:
+				info = all_disks()[path].info
+			except KeyError:
+				raise KeyError(f"Could not find {path} in {all_disks()} for some reason.")
 
 		self.path = path
 		self.info = info
--- a/archinstall/lib/disk/blockdevice.py
+++ b/archinstall/lib/disk/blockdevice.py
@@ -241,7 +241,7 @@ class BlockDevice:
 				count += 1
 		else:
 			log(f"Could not find {uuid} in disk after 5 retries",level=logging.INFO)
-			print(f"Cache: {self.part_cache}")
-			print(f"Partitions: {self.partitions.items()}")
-			print(f"UUID: {[uuid]}")
+			log(f"Cache: {self.part_cache}", level=logging.DEBUG)
+			log(f"Partitions: {self.partitions.items()}", level=logging.DEBUG)
+			log(f"UUID: {[uuid]}", level=logging.DEBUG)
 			raise DiskError(f"New partition {uuid} never showed up after adding new partition on {self}")
--- a/archinstall/lib/disk/btrfs.py
+++ b/archinstall/lib/disk/btrfs.py
@@ -102,16 +102,22 @@ def manage_btrfs_subvolumes(installation, partition :dict, mountpoints :dict, su
 		# no contents, so it is not to be mounted
 		if not right_hand:
 			location = None
+
 		# just a string. per backward compatibility the mount point
-		elif isinstance(right_hand,str):
+		elif isinstance(right_hand, str):
 			location = right_hand
 		# a dict. two elements 'mountpoint' (obvious) and and a mount options list ¿?
-		elif isinstance(right_hand,dict):
+		elif isinstance(right_hand, dict):
 			location = right_hand.get('mountpoint',None)
-			mount_options = right_hand.get('options',[])
+			mount_options = right_hand.get('options', [])
+
+		if not mount_options or any(['subvol=' in x for x in mount_options]) is False:
+			mount_options = [f'subvol=@{location}']
+
+		mountpoints[location] = {'partition': partition, 'mount_options' : mount_options}
 
 		# we create the subvolume
-		create_subvolume(installation,name)
+		create_subvolume(installation, name)
 		# Make the nodatacow processing now
 		# It will be the main cause of creation of subvolumes which are not to be mounted
 		# it is not an options which can be established by subvolume (but for whole file systems), and can be
@@ -176,10 +182,8 @@ def manage_btrfs_subvolumes(installation, partition :dict, mountpoints :dict, su
 			
 			# Well, now that this "fake partition" is ready, we add it to the list of the ones which are to be mounted,
 			# as "normal" ones
-			mountpoints[fake_partition['mountpoint']] = {'partition': partition, 'mount_options' : [f"subvol=@{fake_partition['mountpoint']}"]}
 
-	# if the physical partition has been selected to be mounted, we include it at the list. Remmeber, all the above treatement won't happen except the creation of the subvolume
-	if partition['mountpoint']:
-		mountpoints[partition['mountpoint']] = {'partition': partition, 'mount_options' : [f"subvol=@{partition['mountpoint']}"]}
+	if partition['mountpoint'] and partition.get('subvolumes', False) is False:
+		mountpoints[partition['mountpoint']] = {'partition': partition}
 
 	return mountpoints
--- a/archinstall/lib/disk/btrfs.py
+++ b/archinstall/lib/disk/btrfs.py
@@ -92,87 +92,94 @@ def manage_btrfs_subvolumes(installation, partition :dict, mountpoints :dict, su
 	# th mount info dict has an entry for the path of the mountpoint (named 'mountpoint') and 'options' which is a list
 	# of mount options (or similar used by brtfs)
 	for name, right_hand in subvolumes.items():
-		try:
-			# we normalize the subvolume name (getting rid of slash at the start if exists. In our implemenation has no semantic load - every subvolume is created from the top of the hierarchy- and simplifies its further use
-			if name.startswith('/'):
-				name = name[1:]
-			# renormalize the right hand.
+		# we normalize the subvolume name (getting rid of slash at the start if exists. In our implemenation has no semantic load - every subvolume is created from the top of the hierarchy- and simplifies its further use
+		if name.startswith('/'):
+			name = name[1:]
+
+		# renormalize the right hand.
+		location = None
+		mount_options = []
+		# no contents, so it is not to be mounted
+		if not right_hand:
 			location = None
-			mount_options = []
-			# no contents, so it is not to be mounted
-			if not right_hand:
-				location = None
-			# just a string. per backward compatibility the mount point
-			elif isinstance(right_hand,str):
-				location = right_hand
-			# a dict. two elements 'mountpoint' (obvious) and and a mount options list ¿?
-			elif isinstance(right_hand,dict):
-				location = right_hand.get('mountpoint',None)
-				mount_options = right_hand.get('options',[])
-			# we create the subvolume
-			create_subvolume(installation,name)
-			# Make the nodatacow processing now
-			# It will be the main cause of creation of subvolumes which are not to be mounted
-			# it is not an options which can be established by subvolume (but for whole file systems), and can be
-			# set up via a simple attribute change in a directory (if empty). And here the directories are brand new
-			if 'nodatacow' in mount_options:
-				if (cmd := SysCommand(f"chattr +C {installation.target}/{name}")).exit_code != 0:
-					raise DiskError(f"Could not set  nodatacow attribute at {installation.target}/{name}: {cmd}")
-				# entry is deleted so nodatacow doesn't propagate to the mount options
-				del mount_options[mount_options.index('nodatacow')]
-			# Make the compress processing now
-			# it is not an options which can be established by subvolume (but for whole file systems), and can be
-			# set up via a simple attribute change in a directory (if empty). And here the directories are brand new
-			# in this way only zstd compression is activaded
-			# TODO WARNING it is not clear if it should be a standard feature, so it might need to be deactivated
-			if 'compress' in mount_options:
-				if (cmd := SysCommand(f"chattr +c {installation.target}/{name}")).exit_code != 0:
-					raise DiskError(f"Could not set compress attribute at {installation.target}/{name}: {cmd}")
-				# entry is deleted so nodatacow doesn't propagate to the mount options
-				del mount_options[mount_options.index('compress')]
-			# END compress processing.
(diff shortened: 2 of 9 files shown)


Torxed:
Current state is that it installs.
But the mountpoints are still not there:
```
[root@bigrigv2 archinstall]# findmnt --real
TARGET                    SOURCE               FSTYPE      OPTIONS
└─/mnt/archinstall        /dev/loop0p2[/@]     btrfs       rw,relatime,ssd,space_cache=v2,subvolid=256,subvol=/@
  └─/mnt/archinstall/boot /dev/loop0p1         vfat        rw,relatime,fmask=0022,dmask=0022,codepage=437,iocharset=ascii,shortname=mixed,utf8,errors=remount-ro
```
```
[root@bigrigv2 archinstall]# ls /mnt/archinstall/
bin  boot  dev	etc  home  lib	lib64  mnt  opt  proc  root  run  sbin	srv  sys  tmp  usr  var
```

wllacer (commented):
From my phone.
I think you don't need the mountpoints.update at line 213 (and others) of installer.py
In the worst case, i've seen you ported *manage_btrs_subvolumes* from máster. IIRC i made It to behave slightly diferent. That could imply you need to Port *mount_ordered_partitions* also. And this is a lap of faith.
I'd rather revert this patch and try with #906 only

Torxed:
In #906 the subvolumes doesn't get mounted when I tried. Or so I'm 90% sure of.
I've just now got `v2.3.1-dev` to a state where it completes an installation and the default subvolume layout is supported and works.

There's still some questionable code, will sleep on it before I merge this.
For instance, a subvolume definition of `partition = Partition(target['source'], partition.real_device, filesystem=target.get('fstype', None), mountpoint=target['target'])` in `get_partitions_in_use` breaks because it will try to auto-mount a device that is potentially already mounted. I put a `auto_mount = False` in the class and the call for now.

As this logic only appear to happen surrounding btrfs subvolumes, and we've fixed this in `master`.
I would deem this stable enough and a improvement to `v2.3.0`. I don't think people can go to crazy with subvolumes using `v2.3.1`, but at least it will have a better default layout and that works. + all the other disk related issues that got fixed too.

The PR is ready for review and testing, but again, keep in mind:
 * It's for the default btrfs layout in #781 first and foremost

Torxed:
This is the current state of the code in `v2.3.1-dev` -fix branch:
```
[root@bigrigv2 archinstall]# findmnt --real
TARGET                                    SOURCE                     FSTYPE      OPTIONS
└─/mnt/archinstall                        /dev/loop0p2[/@]           btrfs       rw,relatime,ssd,space_cache=v2,subvolid=256,subvol=/@
  ├─/mnt/archinstall/boot                 /dev/loop0p1               vfat        rw,relatime,fmask=0022,dmask=0022,codepage=437,iocharset=ascii,shortname=mixed,utf8,errors=remount-ro
  ├─/mnt/archinstall/.snapshots           /dev/loop0p2[/@.snapshots] btrfs       rw,relatime,ssd,space_cache=v2,subvolid=257,subvol=/@.snapshots
  ├─/mnt/archinstall/home                 /dev/loop0p2[/@home]       btrfs       rw,relatime,ssd,space_cache=v2,subvolid=258,subvol=/@home
  ├─/mnt/archinstall/var/cache/pacman/pkg /dev/loop0p2[/@pkg]        btrfs       rw,relatime,ssd,space_cache=v2,subvolid=260,subvol=/@pkg
  └─/mnt/archinstall/var/log              /dev/loop0p2[/@log]        btrfs       rw,relatime,ssd,space_cache=v2,subvolid=259,subvol=/@log
```
Not sure this is enough, I'm missing the `@home` when doing `ls -l /mnt/archinstall`

wllacer:
> This is the current state of the code in `v2.3.1-dev` -fix branch:
> 
> ```
> [root@bigrigv2 archinstall]# findmnt --real
> TARGET                                    SOURCE                     FSTYPE      OPTIONS
> └─/mnt/archinstall                        /dev/loop0p2[/@]           btrfs       rw,relatime,ssd,space_cache=v2,subvolid=256,subvol=/@
>   ├─/mnt/archinstall/boot                 /dev/loop0p1               vfat        rw,relatime,fmask=0022,dmask=0022,codepage=437,iocharset=ascii,shortname=mixed,utf8,errors=remount-ro
>   ├─/mnt/archinstall/.snapshots           /dev/loop0p2[/@.snapshots] btrfs       rw,relatime,ssd,space_cache=v2,subvolid=257,subvol=/@.snapshots
>   ├─/mnt/archinstall/home                 /dev/loop0p2[/@home]       btrfs       rw,relatime,ssd,space_cache=v2,subvolid=258,subvol=/@home
>   ├─/mnt/archinstall/var/cache/pacman/pkg /dev/loop0p2[/@pkg]        btrfs       rw,relatime,ssd,space_cache=v2,subvolid=260,subvol=/@pkg
>   └─/mnt/archinstall/var/log              /dev/loop0p2[/@log]        btrfs       rw,relatime,ssd,space_cache=v2,subvolid=259,subvol=/@log
> ```
> 
> Not sure this is enough, I'm missing the `@home` when doing `ls -l /mnt/archinstall`

I think the output is totally Ok. You have the @home subvolume correctly mounted

Torxed:
Tested with and without encryption using `systemd-boot`.
I have not tested with grub yet, but so far un-encrypted works then but encryption fails on a BlockDevice lookup.

One thing that has me concerned is that every subvolume gets `Adding additional key-file ...`.
Because it's now treated as one separate "partition". This should be corrected as we probably will end up with too many keys (?) at some point.

I'm not far away at least.

wllacer:
> Tested with and without encryption using `systemd-boot`. I have not tested with grub yet, but so far un-encrypted works then but encryption fails on a BlockDevice lookup.
> 
> One thing that has me concerned is that every subvolume gets `Adding additional key-file ...`. Because it's now treated as one separate "partition". This should be corrected as we probably will end up with too many keys (?) at some point.
> 
> I'm not far away at least.
The btrfs code in 2.3.1 DOES NOT support encrypted partitions. It was clearly documented, and i thought i had a explicit branching in the code. If you want it, you need to put the whole  #838.

btw. i'm trying to test v.2.3.1.dev but I've run in an exception
```
 File "/home/werner/projects/archinstall/archinstall/lib/disk/partition.py", line 220, in partprobe
    if self.block_device and SysCommand(f'partprobe {self.block_device.device}').exit_code == 0:
AttributeError: 'float' object has no attribute 'device'
```
Do you remember having anything like that ?

Torxed:
I got it supporting encryption now. So far it encrypts the root partition `/dev/sda2` and then unlocks it, creates subvolumes inside it, unmounts it and with the unlocked volume mounts all the subolumes and continues with the installation.

I do not get `AttributeError: 'float' object has no attribute 'device'`. I don't know how `self.block_device` would be a float? humm

wllacer:
A correction, We managed to support encryption but not the generation of key-files for btrfs. It was the last thing that came out in #787

The reason why we didn't support encryption, was that encryption and key handling has to be done, in btrfs as in the rest of the filesystems, at the physical partition level.
To achieve this i had to totally rewrite *mount_ordered_partitions*, handling the encryption prior to subvol definition and mounting, and creating the keys after everything else was done

Torxed:
I solved that by checking if the keyfile already exists on the desired destination (which is based on loopdev name).
Still running a few last tests. `keyfile` is only really supposed to be used when there's multiple partitions that are individually encrypted.

Something that we don't really do on `btrfs` with subvolumes.
As they are all virtually the same logical `encrypted(device)`.

Torxed:
Old function: https://github.com/archlinux/archinstall/blob/becb4239599c0fbb59315dc3beaacd8dc826d0c2/archinstall/lib/user_interaction.py#L195-L204

This sets all partitions that are not `/` to generate a keyfile.
This has to be changed before this PR can go in.

wllacer:
> 



> I solved that by checking if the keyfile already exists on the desired destination (which is based on loopdev name).

Good idea
> Still running a few last tests. `keyfile` is only really supposed to be used when there's multiple partitions that are individually encrypted. I'm still not sure why it's even creating a keyfile for `/dev/sda2` as that's the encrypted volume. But it appears `generate-encryption-key-file` is set somewhere.

I remember the same problem, I solved it checking if the partitiion didn't contain the root / directory
> 
> Something that we don't really do on `btrfs` with subvolumes. As they are all virtually the same logical `encrypted(device)`.
btrfs handles partitions as encryption unit, not subvolumes

Torxed:
There we go. Some device-lookups later, it seems like we have a winner chicken dinner.
I'm using two methods of resolving `partition -> blockdevice-parent` lookups:
```python
result = min(SysCommand(f'lsblk -no pkname {partition.real_device}').decode().rstrip('\r\n').split('\r\n'), key=len)
block_device = BlockDevice(f"/dev/{result}")
```
And also:
```python
block_device_name = pathlib.Path(partition.real_device).stem
block_device_class_link = pathlib.Path(f"/sys/class/block/{block_device_name}")
if not block_device_class_link.is_symlink():
	raise ValueError(f"Could not locate blockdevice for partition: {block_device_class_link}")
block_device_class_path = block_device_class_link.readlink()
BlockDevice(f"/dev/{block_device_class_path.parent.stem}")
```
The later uses path traversal, I don't think a partition can be lower down than `/block/BDev/Partition`.
So that should probably be more safe than using `lsblk` which appears to switch around the order of things depending on what kind of partition we're working with.

This now successfully installs.

 - [x] systemd-boot with subvolumes
 - [x] systemd-boot + encryption with subvolumes
 - [ ] GRUB with subvolumes
 - [ ] GRUB + Encryption with subvolumes

Torxed:
Side note, I get this from time to time. Not sure why yet.
The command executes correctly but the shutdown on temporary boot fails:

screenshot

Torxed:
GRUB with encryption installs, but I can't get it to boot to the boot loader, it skips the HDD and goes straight to ISO.

Torxed:
Apparemtly `systemd-run --machine={self.container_name} --pty /bin/bash -c "shutdown now"` always exits with exit code 256 when it's doen good. (wraparound to 0?).

Instead, we check if either shutdown or the session exit code is 0, which should mean all is well.
`Boot()` now works as intended.

wllacer:
> GRUB with encryption installs, but I can't get it to boot to the boot loader, it skips the HDD and goes straight to ISO.
 cou
what kind of target disk are you using? I wasn't able to boot into Grub with a loopback device, IIRC not even with ext4. In fact the only way i could make it booting was within a virtualbox session, with an vdi target

Torxed:
> > GRUB with encryption installs, but I can't get it to boot to the boot loader, it skips the HDD and goes straight to ISO.
> 
> what kind of target disk are you using? I wasn't able to boot into Grub with a loopback device, IIRC not even with ext4. In fact the only way i could make it booting was within a virtualbox session, with an vdi target

I'm using [archtest](https://github.com/Torxed/Scripts/blob/master/python/archtest.py)` --branch torxed-fix-btrfs-mounts --bios --new-drives`. So it's KVM, two normal block devices. Skipping `--bios` when testing UEFI.

Torxed:
screenshot

Grub installs on UEFI, but isn't bootable. I'm wondering if it's because `ovmf` and grub some how.

Torxed:
Think I found the cause, these two lines doesn't do what it's supposed to: https://github.com/archlinux/archinstall/blob/0761f528aca836ccd1ffb98a5ac13590fd3e9a1d/archinstall/lib/installer.py#L731-L732

It's called *(verified through `cmd_history.txt`)*, but it doesn't replace the file content as it should. Not sure how this ever worked.

Torxed:
Everything appears to work correctly, commands are executed as they should.
But for whatever reason, it doesn't work yet.

screenshot

screenshot1

screenshot2

screenshot3

screenshot4

Torxed:
`GRUB` on `EFI` **without** subvolumes is also broken. Even in `v2.3.1-dev`.
I'm tempted to merge this as it's probably better to try to solve `GRUB` + `UEFI` issues in a separate PR. This PR will not worsen, only make other aspects work better.

Torxed:
Works now with the latest change from #793 :D

Torxed:
screenshot

If everyone is fine with this PR, I'm merging it. Pinging a few that I know has experience with these things: @wllacer @dylanmtaylor

wllacer:
I like what you did on the naming of the encrypted devices ... as for the rest, it feels a lot better
With a 2.3.1-dev from yesterday I've tested Grub+btrfs+ Uefi under VirtualBox. I could boot into Archlinux, almost ... It hung while loading the initial ram disk

Torxed:
> I like what you did on the naming of the encrypted devices ... as for the rest, it feels a lot better With a 2.3.1-dev from yesterday I've tested Grub+btrfs+ Uefi under VirtualBox. I could boot into Archlinux, almost ... It hung while loading the initial ram disk

What specifically do you not like? I can change things around and modify it.
But I would need to know what doesn't feel great.

wllacer:
Just the opposite. I find the final result very good.
Sorry for Haven't been able to follow the grub issue further. Today isn't
the BEST day

Torxed:
> Just the opposite. I find the final result very good. Sorry for Haven't been able to follow the grub issue further. Today isn't the BEST day

Ah apologies, my English skills sometimes messes with me hehe. Read your message as if there was something wrong with `v2.3.1-dev` heh. My bad, and I hope your day gets better! :D

PR #914, opened 2022-01-27
Move from the basic pipewire media session manager to wireplumber

### LINKED ISSUE #912: desktop profiles, replace pipewire-media-session with wireplumber, wireplumber seembs to be the newest replacement media-session replaceing pipewire-media-session
hi,
would it be possible to replace pipewire-media-session with wireplumber?
or, to make things flexable,
add a question in asking,
"would you like to use wireplumber or pipewire-media-session?
p.s
i'm using wireplumber currently.
wireplumber will remove pipewire-media-session and the systemd units.
a total replacement.
thank you for considdering my request!
Majid Hussain

dylanmtaylor:
It looks like wireplumber is now the preferred session manager for desktop audio, and Fedora in its latest release also made this same change. https://fedoraproject.org/wiki/Changes/WirePlumber

pipewire-media-session is provided by the wireplumber package. https://archlinux.org/packages/extra/x86_64/wireplumber/

https://wiki.archlinux.org/title/PipeWire#WirePlumber

It looks like starting the wireplumber user service should be handled by the package.

Closes #912

### DIFF
--- a/profiles/applications/pipewire.py
+++ b/profiles/applications/pipewire.py
@@ -3,7 +3,7 @@ import logging
 
 # Define the package list in order for lib to source
 # which packages will be installed by this profile
-__packages__ = ["pipewire", "pipewire-alsa", "pipewire-jack", "pipewire-media-session", "pipewire-pulse", "gst-plugin-pipewire", "libpulse"]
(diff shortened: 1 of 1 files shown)


dylanmtaylor:
I have tested this change on the KDE desktop environment and my audio is still functional afterwards.

Torxed:
Quick great work as usual!

PR #915, opened 2022-01-27
Fix trailing comma in schema.json

Sebas-h:
(Tried to [generate docs](https://github.com/coveooss/json-schema-for-humans) and noticed the comma preventing that)

🚨 PR Guidelines:

# Describe your PR

Remove invalid trailing comma in json file.


Torxed:
Good catch, old habits of Python :)

Issue #911, opened 2022-01-26
guided install fails with error for key-checks


triplem:
When installing the gnome package with the latest ArchIso (2022.01.01) the installation fails with an error in the key-check for the package libspeechd.
Obviously the workaround is to install a minimal system without gnome and do a pacman -S archlinux-keyring. 

Could this be implemented in the guided installer itself?

Torxed:
Duplicate of https://github.com/archlinux/archinstall/issues/907.
Discussion about implement `pacman -S archlinux-keyring` is ongoing there.

I would just like to verify any side effects of running `pacman -S archlinux-keyring` blindly before committing to it.

Issue #907, opened 2022-01-25
Could not strap in package:256


EzPecks:
Hi everyone! I am rookie and i want to install Archlinux on my secondary PC. (AMD Ryzen 5 1600, PowerColor 6600XT, 16gb ram 2666mhz, 1TB WB green) Im used the "command" archinstall and followed all steps correctly. It start install, but in certain moment, it's stop, a lot of words in red and the last things its says is "Could not strap in package 256". I searched in a all web and nothing. Please if anyone knows how to solve this, can explain like im a fool.

Thanks you!

dylanmtaylor:
Can you take a picture of the error and describe the configuration you chose?

Torxed:
Usually `Could not strap in packages: 256` means there was a network related issue.
Probably a mirror that timed out or something similar, re-running the installer should work.

timmypidashev:
@EzPecks run this command before you run archinstall: ```bash
pacman -Sy archlinux-keyring
```
I just had this issue as well, the keyring just needs to be synced because of some picky packages :) Also if a dev is reading this it would be nice if this is automatically included in the script. I can open a pr or something maybe.

timmypidashev:
> Usually `Could not strap in packages: 256` means there was a network related issue. Probably a mirror that timed out or something similar, re-running the installer should work.

@Torxed usually but the op did not specify the full issue. The pacstrap did not happen as a result of package integrity failure

Torxed:
> > Usually `Could not strap in packages: 256` means there was a network related issue. Probably a mirror that timed out or something similar, re-running the installer should work.
> 
> @Torxed usually but the op did not specify the full issue. The pacstrap did not happen as a result of package integrity failure

I don't know if pacstrapping in the keyring blindly is considered ok? If it is, we should definitely do it to avoid this issue. PR would be welcome :)

timmypidashev:
Ok so I've spent a fair bit of of time researching this topic, and while it isnt that hard to just hard-code `pacman -Sy archlinux-keyring` in __init__ or elsewhere, this script aims to be more than that, offering customizable installations and acts as a library. I am not experienced enough to do this yet, although I do marvel at how efficiently designed this python package is(TY to the devs). Anyways I will back off this pr, although I do think it would be nice to have archinstall update the keyring for inexperienced users trying to get into arch(me as well).

Torxed:
[...]

PR #917, opened 2022-01-28
Synchronize `archlinux-keyring` when running guided.py

timmypidashev:
### Error
To reproduce error, use guided install without first synchronizing `archlinux-keyring`

### Manual Fix
Simply synchronize the **keyring** before running `archinstall`:
```bash
pacman -Sy archlinux-keyring
```
### Installer fix:
Simply run the command in `guided.py`
```python
archinstall.log(f"Updating achlinux-keyring.")
os.system("pacman -Sy archlinux-keyring")
``` 
**NOTE**: Untested code, may not work as intended!

### Reason for PR:
New users and even experienced users won't usually think about synchronizing the keyring, which almost always ends up failing the installer when running in **guided** mode. This wastes time for experienced users, forcing them to rerun the installation, and scares away the new users, who then open issues such as: #907, while the experienced make notice about it in #911. I am by no means an experienced arch linux user, and have made an untested _blind_ hotfix. Edits are enabled should my code need changes :) Hope this helps out!

### DIFF
--- a/archinstall/lib/networking.py
+++ b/archinstall/lib/networking.py
@@ -40,6 +40,16 @@ def check_mirror_reachable() -> bool:
 
 	return False
 
+def update_keyring() -> bool:
+	log("Updating archlinux-keyring ...", level=logging.INFO)
+	if SysCommand("pacman -Sy archlinux-keyring").exit_code == 0:
+		return True
+	
+	elif os.geteuid() != 0:
+		log("update_keyring() uses 'pacman -Sy archlinux-keyring' which requires root.", level=logging.ERROR, fg="red")
+
+	return False
+
 
 def enrich_iface_types(interfaces: Union[Dict[str, Any], List[str]]) -> Dict[str, str]:
 	result = {}
--- a/docs/archinstall/general.rst
+++ b/docs/archinstall/general.rst
@@ -74,6 +74,8 @@ Networking
 
 .. autofunction:: archinstall.check_mirror_reachable
 
+.. autofunction:: archinstall.update_keyring
+
 .. autofunction:: archinstall.enrich_iface_types
 
 .. autofunction:: archinstall.get_interface_from_mac
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -11,10 +11,6 @@ if archinstall.arguments.get('help'):
 if os.getuid() != 0:
 	print("Archinstall requires root privileges to run. See --help for more.")
 	exit(1)
-	
-# update the arch linux keyring to ensure package integrity
-archinstall.log(f"Updating achlinux-keyring.")
-os.system("pacman -Sy archlinux-keyring")
 
 # Log various information about hardware before starting the installation. This might assist in troubleshooting
 archinstall.log(f"Hardware model detected: {archinstall.sys_vendor()} {archinstall.product_name()}; UEFI mode: {archinstall.has_uefi()}", level=logging.DEBUG)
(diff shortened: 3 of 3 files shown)


Torxed:
Thank you, looks pretty good!
I have a small favour to ask, could you move it after: https://github.com/archlinux/archinstall/blob/8a39d4dc98b7a4bd8c5efc14957b69d04a1deba5/examples/guided.py#L311

So that we first check for mirror reachability, and then do the pacman call? :) Perhaps on line 315 would be perfect.
Maybe we also do: `if os.system('pacman ...') != 0: log("something about not being able to update keyring"); exit(1)` so it doesn't silently pass? : )

timmypidashev:
Alright will do :)

timmypidashev:
Done!

timmypidashev:
Quick side note: `networking.py` might not be the best place to put the **update_keyring** function. Not sure if that's something that is needed, but when running the build iso from this patch prior to 361a6f1, df1d5d4, e695052, the system call to update **archinstall-keyring** brought up a [Y, n] prompt from pacman, hence the latter commits.

Torxed:
Looks good, I don't mind it living in `networking` for now, it's either there or [packages](https://github.com/archlinux/archinstall/blob/e6950526975ec2b8cd46e5072c035960662fe005/archinstall/lib/packages.py). But either way tbh. I'm merging this now : ) Great work and thank you for your contribution!

AviVahl:
Heya @Torxed @timmypidashev  is it possible the pacman command is missing the `--noconfirm` flag?

I've noticed `archinstall`'s master freezes on `Updating archlinux-keyring...` stage. running the command manually asked me to confirm, and once I did it finished instantly.

timmypidashev:
Yes it is. If that's the reason I'll make a quick fix...

timmypidashev:
Wow yea that was a pretty big oversight on my part, my apologies

timmypidashev:
#922 should fix this, again my deepest apologies I can't believe I missed that.

Torxed:
Quick thinking from you both, merged the PR! :) Should be solved now @AviVahl !

Torxed:
We bumped into a problem strapping in packages with this fix:
screenshot

I'm adding a check to see if the installed version is already the latest, that way we'll at least avoid this specific issue.
This fix can be removed once this is fulfilled: https://gitlab.archlinux.org/archlinux/archlinux-keyring/-/issues/4 *(a hidden project, but linking anyway for reference)*

To quote the release engineering team:
> I know you can't see this ticket, but https://gitlab.archlinux.org/archlinux/archlinux-keyring/-/issues/4 is about getting a daily pipeline running that checks for soon expiring keys/subkeys
> it's unfortunate that in the past few months we couldn't even guarantee archlinux-keyring to be "good to use" for a timespan of a month...
> I really hope this improves. I want to arrive at a guarantee of at least 6 months for this

Torxed:
The version-detection PR to avoid pulling it "in vain" is here: #933 (WIP)

timmypidashev:
Interesting, that is a better way of handling it altogether.

Torxed:
Thank you. It requires a lot of additional functionality that wasn't there, so it was a good time for me to throw some more lib-stuff in there ^^

timmypidashev:
Yea, as a followup(though not directly pertaining to the issue in this repo) the archlinux-keyring is really buggy, with pacman giving:
```bash
:: File /var/cache/pacman/pkg/archlinux-keyring-20220125-1-any.pkg.tar.zst is corrupted (invalid or corrupted package (PGP signature)).
```
Not sure if this is my installation medium being faulty(I have used the latest image fairly often in the last several days), going to reimage the drive and see what's happening, but it seems to be a faulty pgp signature from that secret repo mentioned in this issue: https://gitlab.archlinux.org/archlinux/archlinux-keyring/

Torxed:
That happens usually when you try to update the keyring when it’s already the latest version. Or that’s my theory at least :)

PR #924, opened 2022-01-30
Added a qtile profile

edl2:
Working on adding a qtile profile, should nitrogen be included and nano vs vim and dmenu vs rofi should be discussed.

### DIFF
--- a/profiles/qtile.py
+++ b/profiles/qtile.py
@@ -11,11 +11,7 @@ __packages__ = [
 	'alacritty',
 	'lightdm-gtk-greeter',
 	'lightdm',
-	'dmenu',
-	'pcmanfm',
-	'vim',
-	'nano',
-	'nitrogen',
+	'dmenu'
 	
 	
 ]
--- a/profiles/qtile.py
+++ b/profiles/qtile.py
@@ -22,8 +22,8 @@ def _prep_function(*args, **kwargs):
 	for more input before any other installer steps start.
 	"""
 
-if __name__ == 'qtile':
 
+if __name__ == 'qtile':
 	# Install dependency profiles
 	archinstall.storage['installation_session'].install_profile('xorg')
 
--- a/profiles/qtile.py
+++ b/profiles/qtile.py
@@ -1,0 +1,55 @@
+# A desktop environment using "qtile" window manager with common packages.
+
+import archinstall
+
+is_top_level_profile = False
+
+# New way of defining packages for a profile, which is iterable and can be used out side
+# of the profile to get a list of "what packages will be installed".
+__packages__ = [
+	'qtile',
+	'alacritty',
+	'lightdm-gtk-greeter',
+	'lightdm',
+	'dmenu',
+	'pcmanfm',
+	'vim',
+	'nano',
+	'nitrogen',
+	
+	
+]
+
+
+def _prep_function(*args, **kwargs):
+	"""
+	Magic function called by the importing installer
+	before continuing any further. It also avoids executing any
+	other code in this stage. So it's a safe way to ask the user
+	for more input before any other installer steps start.
+	"""
+
+
+
+	# qtile requires a functioning Xorg installation.
+	profile = archinstall.Profile(None, 'xorg')
+	with profile.load_instructions(namespace='xorg.py') as imported:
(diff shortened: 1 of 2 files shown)


Torxed:
Looks sensible to me. Looking forward to testing a python based WM :)

dylanmtaylor:
> Looks sensible to me. Looking forward to testing a python based WM :)

I'm not really sure we want too many more desktop environment options - especially if it's somewhat opinionated by us, for example, by picking things like vim, nano, and nitrogen. We should try to ship upstream without packages like these. Also, vim/nano is redundant with base packages, IMO. They don't really belong in the desktop profile.

profiles/qtile.py
```diff
@@ -0,0 +1,55 @@
...
+# New way of defining packages for a profile, which is iterable and can be used out side
+# of the profile to get a list of "what packages will be installed".
+__packages__ = [
+	'qtile',
+	'alacritty',
+	'lightdm-gtk-greeter',
+	'lightdm',
+	'dmenu',
+	'pcmanfm',
+	'vim',
```
dylanmtaylor, line 16: remove vim and nano, these aren't DE specific and don't belong in the desktop profile.

profiles/qtile.py
```diff
@@ -0,0 +1,55 @@
...
+	'nano',
+	'nitrogen',
+
```
dylanmtaylor, line 19: delete blank spaces here for linter, and remove trailing comman from nitrogen. Also, is nitrogen really needed?

profiles/qtile.py
```diff
@@ -0,0 +1,55 @@
...
+			return imported._prep_function()
+		else:
+			print('Deprecated (??): xorg profile has no _prep_function() anymore')
+
+
+if __name__ == 'qtile':
+
+
+	# Install dependency profiles
+	archinstall.storage['installation_session'].install_profile('xorg')
```
dylanmtaylor, line 47: qtile looks like it can run with Wayland, maybe this isn't needed?

dylanmtaylor:
Please correct the flake8 errors - only a few of them, so shouldn't take too long. https://github.com/archlinux/archinstall/runs/4999156240?check_suite_focus=true

profiles/qtile.py
edl2: For linter and the flake errors do I remove all blank spaces ?

dylanmtaylor: Just the blank lines where it's complaining.

dylanmtaylor:
Hmm, on second thought, we might want to do what we do in GNOME and install Xorg here, for now: https://github.com/archlinux/archinstall/blob/master/profiles/gnome.py#L24

edl2:
keep in mind that reverting 34e89b1 would also revert some spacing error fixes

tuhanayim:
You should add this profile to `profiles/desktop.py`.
https://github.com/archlinux/archinstall/blob/master/profiles/desktop.py#L23

dylanmtaylor:
> keep in mind that reverting 34e89b1 would also revert some spacing error fixes

Ah. In that case, maybe just copy the block out of the history. And the other commentator is right, it needs to be in desktop.py for it to show in the list.

Torxed:
> Hmm, on second thought, we might want to do what we do in GNOME and install Xorg here, for now: https://github.com/archlinux/archinstall/blob/master/profiles/gnome.py#L24
> 
> This would just be a revert of [34e87b1](https://github.com/archlinux/archinstall/commit/34e87b13accfba16cd8a145e513f225de7fd86fc)
> 
> @Torxed , thoughts?

I agree on this one.
I revert the commit-hash in question, since this is the first time committing to a project - in case git is new too here's how I did it:
```
cd archinstall
git revert 34e87b13accfba16cd8a145e513f225de7fd86fc
# modify/fix the merge issue it complained about with sublime
git add profiles/qtile.py
git revert --continue
# Used the commit message as-is
git push
```

Torxed:
> You should add this profile to `profiles/desktop.py`. https://github.com/archlinux/archinstall/blob/master/profiles/desktop.py#L23

Added it to the desktop selection :) Nice to point out where to add it to new contributors, helps a lot!

I think this PR is ready to go in.

PR #929, opened 2022-02-01
profiles/kde: replace `kate` with `kwrite`

tuhanayim:
# Describe Your PR
[The merge request on invent.kde.org that replaces Kate with Kwrite has been merged.](https://invent.kde.org/plasma/plasma-desktop/-/merge_requests/704)
# Quote From Merge Request
```
Kate is a programmer's text editor that may not be the best choice for a
general-purpose note taking or text editing tool. Furthermore, its name
and icon do not make its purpose obvious, and this is important for the
Favorites view, where any given menu representation is not guaranteed to
show the description (and in fact most do not) so being recognizable by
icon or name is important.
```

### DIFF
--- a/profiles/kde.py
+++ b/profiles/kde.py
@@ -7,7 +7,7 @@ is_top_level_profile = False
 __packages__ = [
 	"plasma-meta",
 	"konsole",
-	"kate",
+	"kwrite",
 	"dolphin",
 	"ark",
 	"sddm",


dylanmtaylor:
Generally, we should strive to ship as close to what the upstream DE defaults to as possible. Even though I use Kate, I'm fine with this. +1

Torxed:
Great work and thank you for taking the time to link upstream information.
Helps in the decision making process!

I agree with Dylan, and it's now merged! :)

PR #933, opened 2022-02-02
SysCommand() to remove ANSII VT100 Esc codes & archlinux-keyring fix

Torxed:
I also moved `package.py` into its own folder `lib/packages/*.py`, as that's something I want to expand on a lot, so package related stuff should go in there.
* I created a `installed_package()` function which gets information about the locally installed package.
* I changed so that `find_packages()` and `find_package()` returns a data-model instead for the package information. This should unify and make sure we detect issues down the line.
* Added support for comparing versions against each other, between upstream package searches and local package data models
* Fixed issue with `SysCommand` and `SysCommandWorker` where it would not remove ANSII VT100 escape codes, which meant that they made their way into the output result. By default it now cleans up any known escape codes.
* With the above support, we now only run `pacman -S archlinux-keyring` in **guided** when necessary. This should avoid the issue mentioned in #917 

This can be tested with:
```python
import archinstall

upstream = archinstall.find_package('archlinux-keyring')
local = archinstall.installed_package('archlinux-keyring')

print(f'Upstream: {upstream.pkg_version}')
print(f'Downstream: {local.pkg_version}')
print(f'Local outdated: {local.pkg_version < upstream.pkg_version}')
```


Torxed:
Just realized while testing that pydantic is not a builtin library >_<
I'll have to revert this one until further notice.

Issue #936, opened 2022-02-02
Profile selection aborts in master


wllacer:
When using guided.py the selection of a profile fails with:
```
 File "/home/werner/projects/archinstall/archinstall/lib/menu/selection_menu.py", line 373, in _select_profile
    profile = select_profile()
  File "/home/werner/projects/archinstall/archinstall/lib/user_interaction.py", line 780, in select_profile
    top_level_profiles = sorted(list(list_profiles(filter_top_level_profiles=True)))
  File "/home/werner/projects/archinstall/archinstall/lib/profiles.py", line 107, in list_profiles
    if Profile(None, profile).is_top_level_profile() is False:
  File "/home/werner/projects/archinstall/archinstall/lib/profiles.py", line 254, in is_top_level_profile
    with self.load_instructions(namespace=f"{self.namespace}.py") as imported:
  File "/home/werner/projects/archinstall/archinstall/lib/profiles.py", line 131, in __enter__
    self.execute()
  File "/home/werner/projects/archinstall/archinstall/lib/profiles.py", line 194, in execute
    self.spec.loader.exec_module(sys.modules[self.namespace])
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/home/werner/projects/archinstall/archinstall/profiles/xorg.py", line 15, in <module>
    *archinstall.lib.hardware.__packages__,
AttributeError: module 'archinstall' has no attribute 'lib'

```

Culprit is
https://github.com/archlinux/archinstall/blob/389feef035cfbb1bd2c4f8be070fa085c088f151/profiles/xorg.py#L10-L16
the reference in line 15 fails to load.

PR on its way

wllacer:
While testing the solution this one happened
```
Press Enter to continue.
 ! Formatting [BlockDevice(/home/werner/vms/testimage.img, size=8.0GB, free_space=5225kB+5243kB+1032kB, bus_type=None)] in 5....4....3....2....1....
Creating a new partition label on /dev/loop0
Traceback (most recent call last):
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/werner/projects/archinstall/archinstall/__main__.py", line 15, in <module>
    archinstall.run_as_a_module()
[...]

PR #916, opened 2022-01-27
Flexible menu 2

wllacer:
This is an replacement of PR #866 

This is **EXPERIMENTAL** code. It is more a working sample of how we could create a general selection menu framework, than a full blown solution; and can be a blueprint for exploration. 

What we have done is to extend the, otherwise excellent, code of @svartkanin for defining a *global selection menu*, with the 
intent of making it a reusable generic tool. Mostly, we have extracted/isolated all code which is directly dependent of Archinstall and the guided.py script, and done a couple of juditious (IOHO) extensions, so the code becomes as generic as possible

<h2> What we have done </h2>

* We have created a base class **GeneralMenu** which holds only the routines needed to handle an arbitrary menu backed by a data area (dictionary) for which, a menu entry is a first level item in the dict. Each menu entry is managed by a **Selector** object. The class is defined as a *context manager* (has \__enter__ and \__exit__ methods).

* We defined following signature for GeneralMenu

```python
	def __init__(self,
			data_store :dict = None,)
```
* *  **data_store** is the area (dictionary) where the selections will be stored, and transferred from/to surrounding code.

* * We have defined following public methods to the class
* * * enable
* * * run
* * * exec_option
* * * option
* * * synch
* * * list_enabled_options
* * * set_option
* * * set_mandatory
* * * mandatory_overview
* * * pre_callback
* * * post_callback
* * * exit_callback

* * This methods are abstract and have to be implemented for every realization of GeneralMenu
* * * _setup_selection_menu_options. Define which menu entries are to be displayed
* * * pre_callback. Code to be executed BEFORE any menu action
* * * post_callback Code to be executed AFTER any menu action
* * * exit_callback Code to be executed at the END of the menu processing

* * We expect some methods to be overloaded in child classes. Those are expected to invoke their parent implentation at one point or another
* * * \__init__  Basically, to prepare a data area to the flat dictionary standard needed for the menu. A call to super().\__init__ is necessary there, and it should be the last statement at the method (otherwise option definition could not work correctly)
* * \__enter__

* * This methods could be overloaded, but the callback methods in the  are thought for personalizing its behaviour under normal circumstances
* * * \__exit__ (use *exit_callback* for processing. Overload for error management)
* * * run  (use *pre_/post_callback*)

* We have expanded the signature of Selector:

```python
		exec_func :Callable = None,
		preview_func :Callable = None,
		mandatory :bool = False

```

[...]

PR #937, opened 2022-02-02
Issue936

wllacer:
Solves issue #936. A reference to a library object was missing / obsolete


Torxed:
I don't mind merging this, but I fail to see how this is different and why this works.
Because it works locally for me:

screenshot

Torxed:
This change was merged in #916

PR #948, opened 2022-02-06
Solving some issues after merging flexible_menu

wllacer:
I've solved a number of issues that the merging of flexible_menu has had.
* After the merge some common methods were specialized at GlobalMenu which used old versions of the code. That made *guide.py* still work, but neither *only_hd.py* nor *swiss.py* This has been cleaned up. All three work now in my tests
* A property (`text`) was taken off from Selector (with reason as it was dynamic and had to be updated, in several places). I've restored it, using a @property method, so the dynamicity is preserved and has no ripple effect in the code.
* I've restored the functionality for the  argument `!encryption-password` at the menu.


wllacer:
I promised some notes about *swiss.py*.
After I managed to convert to the new menu *only_hd* (not in the actual way) I needed to explore some alternatives, and a take for a menu which wasn't the standard menu or its variants. Thus *swiss* was born. ` 
* The first part (the menu invoked with `--advanced`) was a test bed for a menu for other purposes. And collect a series of parameters which would be interesting to set up before the execution of the installer procedure, and are/were not in archinstall.arguments (this could change)
* The second part is an attempt to put the different execution path of the install script under a single code base. Mantaining `only_hd` updated is a PITA [¹] and I suspect some of the other scripts are already bit-rotting. Now I have working three diferent paths with the same code base:
*  * **full**. Exactly *guided.py* (and the default)
* * **only_hd**.  
* * **only_os** this is new -perhaps incomplete-, and does the installation without the partition/hard disk managing.
* an option `mode=lineal`, now only full, to simulate the installation style without a central menu (as has been up to 2.3.1), but with all of its code used beneath. Originally, it was only an experiment, but i use it when I need a somehow faster interface. A user asked about accesibility of the new interface. i have some doubts about the menu as friendly as this lineal way.

In order to implement all of this I have done some imaginative usage of the menu framework, and can be execelent (if a little lengthly) example of how to use it. With a bit of more care (and less personal touches) it could also be the base -or at least an idea- of a v2 of the default installation script interface. But that's not up to me.

[1] And I get now the burden for two. Not very logical myself. ;-)

PR #953, opened 2022-02-09
Converting --key-file from /tmp/x.pw to /dev/stdin.

Torxed:
This should address the concerns of running archinstall on a non-iso setup where passwords would be stored on /tmp/.
I apologies for the delay in fixing this, but running on the ISO has been the top priority and there were some other bugs that needed squashing.

# Breaking changes
 * `archinstall.luks2()` old argument `key_file` is no longer an output for passwords, but rather an input for passwords. Meaning `.encrypt()` and `.unlock()` will source the `key_file` for the password if it exists *(and `password=` argument wasn't given)*. This way we can either pipe a password or a file to the class and it's methods. This will also prepare us for using a hardware token in the future.


Torxed:
It's working locally, but not in a VM. Can't explain why yet.
Something to do with how input is fed to the forked process `cryptsetup` via `SysCommandWorker()` and when I try to write to the `stdin` of the process.

I really can't explain this, both test-code (and guided) are identical.
But on the VM, the input that's supposed to go to the child process comes back into the "output" of the forked process.
But on the local machine, the input gets masked (as it's a password input prompt) and the process continues:

archinstall

And if I run the cryptsetup command manually, the input prompt doesn't echo back the password (as expected):
screenshot_archinstall

Torxed:
Think I've located the culprit!

svartkanin:
@Torxed I reckon this is fairly outdated, are you okay if we close it for now?

Torxed:
Totally fine! :D I can create a new one it if needed.
It was meant to partially solve the #1111

PR #958, opened 2022-02-10
Check if pacman is available

### LINKED ISSUE #957: Error when pacman is already running
When pacman is already doing something this error occures 
```
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/archinstall/archinstall/__main__.py", line 15, in <module>
    archinstall.run_as_a_module()
  File "/archinstall/./archinstall/__init__.py", line 267, in run_as_a_module
    script.execute()
  File "/archinstall/archinstall/lib/profiles.py", line 194, in execute
    self.spec.loader.exec_module(sys.modules[self.namespace])
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/archinstall/archinstall/examples/guided.py", line 238, in <module>
    if not (archinstall.check_mirror_reachable() or archinstall.arguments.get('skip-mirror-check', False)):
  File "/archinstall/archinstall/lib/networking.py", line 35, in check_mirror_reachable
    if SysCommand("pacman -Sy").exit_code == 0:
  File "/archinstall/archinstall/lib/general.py", line 409, in __init__
    self.create_session()
  File "/archinstall/archinstall/lib/general.py", line 460, in create_session
    with SysCommandWorker(self.cmd, callbacks=self._callbacks, peak_output=self.peak_output, environment_vars=self.environment_vars, remove_vt100_escape_codes_from_lines=self.remove_vt100_escape_codes_from_lines) as session:
  File "/archinstall/archinstall/lib/general.py", line 268, in __exit__
    raise SysCallError(f"{self.cmd} exited with abnormal exit code [{self.exit_code}]: {self._trace_log[-500:]}", self.exit_code)
archinstall.lib.exceptions.SysCallError: ['/usr/bin/pacman', '-Sy'] exited with abnormal exit code [None]: b'\x1b[?25l\x1b[1;34m::\x1b[0;1m Synchronizing package databases...\r\n\x1b[0m core\r\n\x1b[1F core is up to date\x1b[K\x1b[1E extra\r\n\x1b[1F extra is up to date\x1b[K\x1b[1E community\r\n\x1b[1F community is up to date\x1b[K\x1b[1E sublime-text\r\n\x1b[1F sublime-text is up to date\x1b[K'
```

I suppose it's a rather unlikely scenario on an ISO image but maybe it should be handled nevertheless

svartkanin:
Fixes #957 
Introduces a check if pacman is already running on the system to avoid a crash later on


Torxed:
Should we maybe put the check in the pacman function? I’m thinking if it’s not just guided but minimal and people using the library just to strap packages?

svartkanin:
Ahh there's a pacman function didn't see that one, I'll find that one and put it there

svartkanin:
@Torxed I had a look and the only function I could find is https://github.com/archlinux/archinstall/blob/master/archinstall/lib/installer.py#L291

However, adding it here won't do the trick because the error is essentially happening in the `guided.py` when calling `check_mirror_reachable` which is called before pacstrap. 
Shall we put this into the `__init__.py` maybe?

Torxed:
`__init__` is probably not the right place. I probably need to understand how Pacman itself isn’t available, that should almost Crete never happen on any arch system?

svartkanin:
It happened for me during development quite a few times now, mainly because I was running updates in the background. If pacman is already in use, lets say using pamac then archinstall isn't able to run pacman

Torxed:
Sorry, I assumed it was because the `pacman` command was missing :)
In that case we should look for any manual `SysCommand("pacman ...")` calls and replace them with a `def pacman()` function or something similar. Adding the fix to guided might work, `__init__.py` probably too.
But the cleanest would probably be to centralize those few calls, and put the check there.
As I suspect we'll run in to this issue elsewhere otherwise. Just a hunch tho, if you disagree I'll merge this and we can at least not worry about it for now :)

svartkanin:
Ahh okay, my description was probably not very clear. I'll have a look to see where this could fit in vest

svartkanin:
@Torxed I've created a new function for any calls with pacman, all parts of the code that were using pacman will now call this function which will perform a check if pacman is available or not before running it

Torxed:
Looks perfect! :) I'll fix the merge conflicts in a bit and merge this.
I might change the `exit(1)` into a loop that logs and then waits for the lock to disappear :)

svartkanin:
Ohh a loop that's a great idea I haven't thought of that actually :)

svartkanin:
Looks great, thanks for adding this @Torxed

svartkanin:
@Torxed I fixed the merge conflicts and tested it, it works :+1:

Issue #951, opened 2022-02-07
Installer loops on checking package validity when package is spelled incorrectly


mhussaincov93:
stepps to repro:
1.
when you go to add extra packages that you would like to be installed, add a spelling error like
pccmanfm
what should happen:
you should be returned to the box where you typed in your list of extra packages with the box being emptied so that you may try again.
before being returned to the prya box, you should be told that x package was not found.
in this case pccmanfm.
what happens:
archinstall enteres a loop of checking if the packages that you've entered are present.
trying this from latest iso.
Majid

svartkanin:
I've actually encountered a different error as well when adding additional packages and just entering rubbish

```
Write additional packages to install (space separated, leave blank to skip): asdf
Verifying that additional packages exist (this might take a few seconds)
Traceback (most recent call last):
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/archinstall/archinstall/__main__.py", line 15, in <module>
    archinstall.run_as_a_module()
  File "/archinstall/./archinstall/__init__.py", line 267, in run_as_a_module
    script.execute()
  File "/archinstall/archinstall/lib/profiles.py", line 194, in execute
    self.spec.loader.exec_module(sys.modules[self.namespace])
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/archinstall/archinstall/examples/guided.py", line 256, in <module>
    ask_user_questions()
  File "/archinstall/archinstall/examples/guided.py", line 90, in ask_user_questions
    global_menu.run()
  File "/archinstall/archinstall/lib/menu/selection_menu.py", line 261, in run
    if not self._process_selection(selection):
  File "/archinstall/archinstall/lib/menu/selection_menu.py", line 278, in _process_selection
    return self.exec_option(selector_name,selector)
  File "/archinstall/archinstall/lib/menu/selection_menu.py", line 297, in exec_option
    result = selector.func()
  File "/archinstall/archinstall/lib/menu/selection_menu.py", line 486, in <lambda>
    lambda: ask_additional_packages_to_install(storage['arguments'].get('packages', None)),
  File "/archinstall/archinstall/lib/user_interaction.py", line 408, in ask_additional_packages_to_install
[...]

PR #959, opened 2022-02-10
Fix errors on selection of additional packages

svartkanin:
This fixes both issues mentioned in #951.
 
In addition to the fixes it also introduces and new text input class that allows to pre-define text on an `input(...)` prompt. When valid and invalid packages are entered in the prompt, after the validation was run it will pre-fill the prompt with the valid packages so that they don't have to be retyped again and only remove the invalid ones.

### DIFF
--- a/archinstall/lib/packages/packages.py
+++ b/archinstall/lib/packages/packages.py
@@ -1,7 +1,9 @@
 import ssl
 import urllib.request
 import json
-from typing import Dict, Any, Tuple, Iterator
+from typing import Dict, Any, Tuple, Iterator, List
+
+import archinstall
 from ..general import SysCommand
 from ..models.dataclasses import PackageSearch, PackageSearchResult, LocalPackage
 from ..exceptions import PackageError, SysCallError
@@ -11,7 +13,7 @@ BASE_URL_PKG_SEARCH = 'https://archlinux.org/packages/search/json/?name={package
 BASE_GROUP_URL = 'https://archlinux.org/groups/search/json/?name={group}'
 
 
-def group_search(name :str) -> Iterator[PackageSearchResult]:
+def group_search(name :str) -> List[PackageSearchResult]:
 	# TODO UPSTREAM: Implement /json/ for the groups search
 	ssl_context = ssl.create_default_context()
 	ssl_context.check_hostname = False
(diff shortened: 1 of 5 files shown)


Torxed:
Regarding the group error, I’m waiting for this to go in: https://github.com/archlinux/archweb/pull/403

after which it will be easier to search for groups specifically.

And once https://gitlab.archlinux.org/archlinux/arch-repo-management becomes a thing we can actually parse JSON files directly instead of doing individual HTTP requests.

wllacer:
@svartkanin The TextInput class is a godsend. It avoids losing (or retyping) a lot of info. The ice in the cake would be a Menu with preselected values (not default ones), especially with multi select (wish, wish) ;-)

svartkanin:
@Torxed I left the group error as is, meaning that it'll still be thrown on anything but a 404 response. So that behavior didn't change I hope.

For the others, there was a bug in the validate function when accessible the returned findings object. 
Besides that I changed the return type of the validate which then also removes throwing errors in there.

Torxed:
> If this is a new piece that's not exposed yet (not sure if it was interested as an api) would you mind if I wrap it into a class? One of the reasons is that it gets easier to maintain since it's clearer which functions are public and which private. But it will also reduce the number of imports necessary in other files, especially looking at how the selection menu looks with all the function imports :)

I implemented/changed the `group_search()` *(previously called `find_group`)* function to use the new API endpoint that got merged about half an hour ago. And is already deployed upstream: https://archlinux.org/groups/search/json/?name=base-devel

Really quick work from Jelly! Much appreciated.

The group search returns all the packages in that group, instead of a "hit" if the group was found.
So that meant I had to change the `group_search` to generate an iterator *(or return the entire result list, but iterator felt nice)*.
But that meant `find_package` can now return more than one package *(which is great because I'm actually thinking of implementing one more feature to `archweb` that take `names=` instead of `name=` to do a bulk-search)*.

This meant that `find_packages` also had to deal with the new iterator. Which it does.
I'll test your changes + the ones I pushed to your branch and see if it all works out.. and if it does I'll merge it in the morning :)

archinstall/lib/packages/packages.py
```diff
@@ -1,7 +1,9 @@
 import ssl
 import urllib.request
 import json
-from typing import Dict, Any, Tuple, Iterator
+from typing import Dict, Any, Tuple, Iterator, List
+
+import archinstall
```
Torxed: Your IDE is acting up again ^^

svartkanin: Yes it is :(

svartkanin: Yeah no I get that, however, I can't see this import in my local branch nor can I see it in the `Files changes` tab either. Might this be from an older commit?

Torxed: Humm, maybe. I can see it in the change tab either.. GitHub playing tricks on my mind hehe. I’ll take a look tomorrow with less tired eyes and have a merge-spree ^^

Torxed:
Appears to install just fine :) And it handles packages spelled incorrectly. Looks good! : )

PR #967, opened 2022-02-12
Add a flag to install testing repositories

### LINKED ISSUE #686: ArchInstall should expose a flag to install a system with testing repo enabled
For trying things out on non-production machines with the testing repository, it'd be neat to be able to do a full system install with the unstable packages available in testing in a VM by doing something like `archinstall --testing`. This would simply make sure that this repository is not commented out in /etc/pacman.conf:

```
[testing]
Include = /etc/pacman.d/mirrorlist
```

Probably a low-priority feature request, but a nice to have.

dylanmtaylor:
This PR closes #686. Essentially, this will give users the option to install and play around with packages not in the main repositories yet, but in the testing repos. We'll want some sort of warning that this shouldn't be used for production systems.

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -304,7 +304,7 @@ class Installer:
 		with open("/etc/pacman.conf", "w") as pacman_conf:
 			for line in lines:
 				if pattern.match(line):
-					# If this is the 
+					# If this is the [] block containing 'testing', uncomment it and set the matched tracking boolean.
 					pacman_conf.write(line.lstrip('#'))
 					matched = True
 				elif matched:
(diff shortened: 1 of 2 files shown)


dylanmtaylor:
I tested the code in the method using a docker container since I couldn't get it to run on the live media. I expect that it works, but I haven't been able to run through a full install with it.

Torxed:
A bit dicey to assume `Server = `is on the line after `# [testing]` but as long as upstream doesn't introduce any surprises, it looks good enough to me :)

dylanmtaylor:
> A bit dicey to assume `Server = `is on the line after `# [testing]` but as long as upstream doesn't introduce any surprises, it looks good enough to me :)

Yeah, perhaps we should try to find some more robust way to handle that. It works with the current file format though.

Torxed:
> > A bit dicey to assume `Server = `is on the line after `# [testing]` but as long as upstream doesn't introduce any surprises, it looks good enough to me :)
> 
> Yeah, perhaps we should try to find some more robust way to handle that. It works with the current file format though.

I started on a `systemd.py` class holder, which essentially is a TOML/INI reader.
I could expand on that and migrate this code to using that in the future :)

Issue #968, opened 2022-02-12
Latest test ISOs aren't working due to missing locales directory


dylanmtaylor:
![VirtualBox_Arch_12_02_2022_11_42_39](https://user-images.githubusercontent.com/277927/153720036-4217ab12-84a5-467c-9f12-7123fb665b7e.png)

Torxed:
Yea this is due to #832 to some extent

PR #969, opened 2022-02-12
Modify setuptools configuration to fix development ISOs

dylanmtaylor:
This should close issue #968

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,3 @@
 import setuptools  # type: ignore
 
-setuptools.setup(include_package_data=True)
+setuptools.setup(package_data={'archinstall': ['locales/*']}, include_package_data=True)
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,3 @@
 import setuptools  # type: ignore
 
-setuptools.setup(package_data={'archinstall': ['locales/*']}, include_package_data=True)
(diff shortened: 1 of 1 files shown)


dylanmtaylor:
This *seems* to work. Instead of erroring out, I'm now able to launch archinstall from the development ISO.
VirtualBox_Arch_12_02_2022_12_35_28
VirtualBox_Arch_12_02_2022_12_34_52

dylanmtaylor:
It's a major improvement at least, considering it starts now. :) Still some crashes, like when selecting a DE (GNOME in this case):
VirtualBox_Arch_12_02_2022_12_38_00

Also crashes when trying to select a language.
VirtualBox_Arch_12_02_2022_12_39_26

dylanmtaylor:
Okay, so it's not copying recursively... eww. Going to need to come up with a solution here.

Torxed:
Can we get this to imclude the profiles/ as well?

dylanmtaylor:
> Can we get this to include the profiles/ as well?

They appear to be already there. This is using setuptools instead of flit so doesn't address that issue.
VirtualBox_Arch_12_02_2022_12_55_15

Torxed:
Ah! You are correct!

PR #971, opened 2022-02-12
Various Build System Improvements

### LINKED ISSUE #832: Not pulling profiles when installing via pip?
Hi,

not sure if it's my config or something but it seems that when using pip (for the pypi package and the raw repo) profiles are not packaged and when running archinstall I am not prompted with the choices (Desktop etc.)

Is this expected?

dylanmtaylor:
This closes #832.


dylanmtaylor:
Looks like it is not following symbolic links. Needs a little bit more work

dylanmtaylor:
By switching up the symlink so the actual files are in archinstall and the examples and profiles on the outer layer are the symlinks, it now seems to work.

dylanmtaylor:
Hmm, symlinks are not quite working, but it's building now...
image

FFY00:
What is the motivation behind this change?

dylanmtaylor:
> What is the motivation behind this change?

Based on the discussion here https://github.com/pypa/flit/issues/522, it feels like it would make sense for us to pivot away from flit, so I am working on a PR to propose a change to more standard tooling. In addition, I am trying to address concerns that certain files which use symbolic links do not seem to be getting included in the final distribution due to 'not being a regular file'.

FFY00:
Moving towards standard tooling does not require you to move away from flit, you can still use pypa/build with flit as a backend. If flit does not include symlinks in the distributions, but setuptools does, it could make sense to move, but it seems that even after that you are resorting to copying the files instead of keeping the symlinks, so I would guess it doesn't...

dylanmtaylor:
> Moving towards standard tooling does not require you to move away from flit, you can still use pypa/build with flit as a backend. If flit does not include symlinks in the distributions, but setuptools does, it could make sense to move, but it seems that even after that you are resorting to copying the files instead of keeping the symlinks, so I would guess it doesn't...

Hmm, you might be right. At the moment, this is something I'm playing around with and not something we've committed to. I'll explore that path of using pypa/build with flit. That might be the best solution.

dylanmtaylor:
I don't know if flit dropped symlink support or what - it looks like it used to work, but now it's not working anymore. I don't see the links in the final tarball.

dylanmtaylor:
The files built by this now seem to be correctly following the symlinks. [d909f80](https://github.com/archlinux/archinstall/pull/971/commits/d909f8007c3650696c89f1fe00666797e521b93c) seemed to be key to getting this working.

Torxed:
Yea it used to follow symlinks, I have an issue reported upstream about it but haven’t gotten a response yet

PR #973, opened 2022-02-12
Leverage same build system used by PyPi in the ArchInstall ISO

dylanmtaylor:
This should make testing ISOs more consistent with what we release.


dylanmtaylor:
This seems a lot faster, too :)
VirtualBox_Arch_12_02_2022_17_06_28
VirtualBox_Arch_12_02_2022_17_06_48

PR #975, opened 2022-02-12
Add multilib flag and enable multilib-testing if testing flag is also passed

dylanmtaylor:
This also fixes issues with the pacman configuration not being copied to the target system after installation


dylanmtaylor:
The -C /etc/pacman.conf fixes issues with pacstrap not using the testing repos we enabled

dylanmtaylor:
Testing working. Screenshot is from chroot with --multilb and --testing enabled
VirtualBox_Arch_12_02_2022_18_58_45

mhussaincov93:
hi,

would it be possible for archinstall to detect if we are running on a 64
bit machine, and if yes, enable the multilib repos?

or a question being asked if the end user would like the repos enabled?

PR #976, opened 2022-02-12
Fix #966 - DeferredTranslation errors

### LINKED ISSUE #966: Missing method in deferred translation
Testing the menu, I selected profile > desktop > budgie and got (really for xorg profile)
```
  File "/home/werner/projects/archinstall/archinstall/lib/menu/selection_menu.py", line 617, in _select_profile
    with profile.load_instructions(namespace=namespace) as imported:
  File "/home/werner/projects/archinstall/archinstall/lib/profiles.py", line 137, in __exit__
    raise args[1]
  File "/home/werner/projects/archinstall/archinstall/lib/menu/selection_menu.py", line 618, in _select_profile
    if not imported._prep_function():
  File "/home/werner/projects/archinstall/archinstall/profiles/desktop.py", line 58, in _prep_function
    with profile.load_instructions(namespace=f"{desktop}.py") as imported:
  File "/home/werner/projects/archinstall/archinstall/lib/profiles.py", line 137, in __exit__
    raise args[1]
  File "/home/werner/projects/archinstall/archinstall/profiles/desktop.py", line 60, in _prep_function
    return imported._prep_function()
  File "/home/werner/projects/archinstall/archinstall/profiles/budgie.py", line 26, in _prep_function
    with profile.load_instructions(namespace='xorg.py') as imported:
  File "/home/werner/projects/archinstall/archinstall/lib/profiles.py", line 137, in __exit__
    raise args[1]
  File "/home/werner/projects/archinstall/archinstall/profiles/budgie.py", line 28, in _prep_function
    return imported._prep_function()
  File "/home/werner/projects/archinstall/archinstall/profiles/xorg.py", line 28, in _prep_function
    archinstall.storage["gfx_driver_packages"] = archinstall.select_driver()
  File "/home/werner/projects/archinstall/archinstall/lib/user_interaction.py", line 925, in select_driver
    title += _('For the best compatibility with your Intel hardware, you may want to use either the all open-source or Intel options.\n')
TypeError: can only concatenate str (not "DeferredTranslation") to str

```
@svartkanin Seems the DeferredTranslation would profit from, at least an \__add\__ method ("string" concatenation in this case)
Just for reference [magic methods](https://docs.python.org/3/reference/datamodel.html). Thought I think to add them when they are needed is more sensible

svartkanin:
This fixes issues mentioned in #966 related to the DeferredTranslation class

### DIFF
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -203,6 +203,11 @@ def load_config():
 		storage['gfx_driver_packages'] = AVAILABLE_GFX_DRIVERS.get(arguments.get('gfx_driver', None), None)
 	if arguments.get('servers', None) is not None:
 		storage['_selected_servers'] = arguments.get('servers', None)
+	if nic_config := arguments.get('nic', {}):
+		if nic_config.get('nic', '') == 'Copy ISO network configuration to installation':
+			arguments['nic'] = {'type': 'iso_config'}
+		elif 'NetworkManager' in nic_config:
+			arguments['nic'] = {'type': 'network_manager', 'NetworkManager': True}
 
 
 def post_process_arguments(arguments):
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -208,7 +208,8 @@ def load_config():
 			arguments['nic'] = {'type': 'iso_config'}
 		elif 'NetworkManager' in nic_config:
 			arguments['nic'] = {'type': 'network_manager', 'NetworkManager': True}
-
+		else:
+			arguments['nic'] = {k if k != 'nic' else 'type': v for k, v in nic_config.items()}
 
 def post_process_arguments(arguments):
 	storage['arguments'] = arguments
--- a/archinstall/lib/menu/menu.py
+++ b/archinstall/lib/menu/menu.py
@@ -1,3 +1,5 @@
+from typing import Dict, List, Union, Any
+
 from archinstall.lib.menu.simple_menu import TerminalMenu
 from ..exceptions import RequirementError
 from ..output import log
@@ -7,7 +9,17 @@ import sys
 import logging
 
 class Menu(TerminalMenu):
-	def __init__(self, title, p_options, skip=True, multi=False, default_option=None, sort=True, preset_values=None, cursor_index=None):
+	def __init__(
+		self,
+		title :str,
+		p_options :Union[List[str], Dict[str, Any]],
+		skip :bool = True,
+		multi :bool = False,
+		default_option :str = None,
+		sort :bool = True,
+		preset_values :Union[str, List[str]] = None,
+		cursor_index :int = None
+	):
 		"""
 		Creates a new menu
 
--- a/archinstall/lib/menu/menu.py
+++ b/archinstall/lib/menu/menu.py
@@ -60,6 +60,11 @@ class Menu(TerminalMenu):
 			log(f"invalid parameter at Menu() call was at <{sys._getframe(1).f_code.co_name}>",level=logging.WARNING)
 			raise RequirementError('Menu.__init__() requires at least one option to proceed.')
 
+		if any([o for o in options if not isinstance(o, str)]):
+			log(" * Menu options must be of type string * ", fg='red')
+			log(f"invalid parameter at Menu() call was at <{sys._getframe(1).f_code.co_name}>",level=logging.WARNING)
+			raise RequirementError('Menu.__init__() requires the options to be of type string')
+
 		if sort:
 			options = sorted(options)
 
--- a/archinstall/lib/translation.py
+++ b/archinstall/lib/translation.py
@@ -8,6 +8,7 @@ from pathlib import Path
 from typing import List, Dict
 from .exceptions import TranslationError
 
+
 class Languages:
 	def __init__(self):
 		self._mappings = self._get_language_mappings()
@@ -45,6 +46,13 @@ class DeferredTranslation:
 
 	def __gt__(self, other) -> bool:
 		return self.message > other
+
+	def __add__(self, other) -> DeferredTranslation:
+		if isinstance(other, str):
+			other = DeferredTranslation(other)
+
+		concat = self.message + other.message
+		return DeferredTranslation(concat)
 
 	def format(self, *args) -> str:
 		return self.message.format(*args)
--- a/archinstall/lib/user_interaction.py
+++ b/archinstall/lib/user_interaction.py
@@ -30,7 +30,7 @@ from .mirrors import list_mirrors
 
 # TODO: Some inconsistencies between the selection processes.
 #       Some return the keys from the options, some the values?
-from .translation import Translation
+from .translation import Translation, DeferredTranslation
 from .disk.validators import fs_types
 from .packages.packages import validate_package_list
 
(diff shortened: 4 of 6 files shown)


wllacer (commented):
It's a pity the limitation to only explicit string at Menu, but I don't know what else we could do.
The problem with ` ask_to_configure_network`  is more complex than moving it to strings. The strings -in english- are significant elsewhere (at guided.py) and determine which workflow the installation will follow. If the installation process were just interactive it wouldn't mind, but when config files are in play, and the chance that they can be generated with one language and processed with others, ... well is a nightmare. Just compare your changes with version i was developing in parallel (it has other features as it allows a previous value, so i've supressed some lines for clarity). I lifted a bit from your code, btw. cleaner than mine 
```python
def ask_to_configure_network(preset :Dict[str, Any] = None) -> Dict[str, Any]:
	# Optionally configure one network interface.
	# while 1:
	# {MAC: Ifname}
	iso_config = str(_('Copy ISO network configuration to installation'))
	iso_config_nt = 'Copy ISO network configuration to installation' # do not translate
	network_manager = str(_('Use NetworkManager (necessary to configure internet graphically in GNOME and KDE)'))
	network_manager_nt = 'Use NetworkManager (necessary to configure internet graphically in GNOME and KDE)' # do not translate

	interfaces = {
		'ISO-CONFIG': iso_config,
		'NetworkManager': network_manager,
		**list_interfaces()
	}
	# for this routine it's easier to set the cursor position rather than a preset value
...
	nic = Menu(_('Select one network interface to configure'), interfaces.values(),cursor_index=cursor_idx).run()

	if nic and nic != iso_config:
		if nic == network_manager:
			nic = network_manager_nt
			return {'nic': nic, 'NetworkManager': True}

...
    elif nic:
		nic = iso_config_nt
		return nic

	return {}

```
As you see, Both versions of the string are needed, one for display and one for internal workflow control. Don't worry about that
I will upload now my version with the complete infraestructure for preset values at GlobalMenu as the first sample of how to use it (albeit partially, i waiting for your TextInput) and  some other minor corrections,

svartkanin:
@wllacer I see what you mean, I missed that the actual selection is returned, I thought it was only used to determine what the actual setting to be stored should be. 
Do you already have a PR/branch up that includes this change or can I push an update into this PR?

Torxed:
Appears that #978 was created yes, but that this one should go in first.
Before I merge, just so I understand the solution to the issue.

Would this work as well/instead or what would happen:
```python
def __add__(self, other) -> str:
		if isinstance(other, DeferredTranslation):
			other = DeferredTranslation(other).message

		return self.message + str(other)
```
Or would that break translation later in the code when the object is used for some reason?

svartkanin:
I think that might break the translation, but also not solve the problem I think.

The issue is that we're using a translatable text as a reference point. So if we store lets say an english translation as a selected value, but then change the language this value is no longer valid (because it didn't get updated). It's a bit of a mess because we're using the translatable messages as selectable values directly. What @wllacer proposed is actually good, we're creating a map between the translations only used for display and the value then stored.

A nicer way could be to use short keys instead of the English text as a value.

svartkanin:
@wllacer @Torxed I've update the network configuration values, the text displayed is only used for that purpose and will shortly after be mapped to key words. The return of the function was updated to 
```
{'type': NEW_TYPE, ...}
```

I've handled (hopefully) all cases where this setting is accessed in the code and I hope not to have broken anything :)

wllacer:
> @wllacer @Torxed I've update the network configuration values, the text displayed is only used for that purpose and will shortly after be mapped to key words. The return of the function was updated to
> 
> ```
> {'type': NEW_TYPE, ...}
> ```
> 
> I've handled (hopefully) all cases where this setting is accessed in the code and I hope not to have broken anything :)

Good idea. There is one piece missing. Backward compatibility ... This should go into \__init\__.py, inside the load_config code
IIRC the algorithm should be
If   isinstance(['nic'],str) --> is an ISO configuration
If   "NetworkManager" in ['nic¡] ---> is Network Manager
Else it is an actual device configuration
(Sorry, today i won't be able to handle it miself :-(

svartkanin:
@wllacer I think I'm not quite following, in the `load_config` in the __init__.py there's no arguments for the nic explicitly set so it can't be specified by command line. But I guess it can be specified within the `--config` file? 

In case someone specified the old format in the config 
```{'nic': 'Copy ISO network configuration to installation'}```
that will now not work anymore, since the newly expected value is 
```{'nic': 'iso_config'}```

I might be wrong here but I think the application could still break without making this change. As an example, if someone uses an old script but now uses a different language than English this value from the config will not work anymore since it must contain the actual translated value of the language. 

Maybe I'm missing here something :)

**Comment by wllacer** [comment]
Yes it's about the.config file.
Guided.py checks for the english string IIRC (although just the type of the
argument could work.in this case)
At __init__ we can adapt the old format to the new, so we can maintain
compatibility. As long as the new format does not use translatable strings.
We can not internationalize a string which goes into the config file,
because we can produce and use the file under different locales.
IIRC is the only case we have.to deal with (at least at the first menú
level, and could not test the profiles)

**Comment by svartkanin** [comment]
I've updated guided.py to check for the type so that should work fine regarding not using nationalization in the configs. 
For backwards comparability in the provided config files, we could have a mapper inside the __init__.py for something like

```
mapper = {'Copy ISO network configuration to installation': 'iso_config'}
config_data = json.loads(...)
if config_data.get('nic', {}):
    nic = config_data.get('nic', {}).get('nic', '')
    config_data['nic'] = {'type': mapper[nic]}
```

What do you think?

**Comment by Torxed** [comment]
I’ll chime in and say we probably should change the network config options all together. It should be a deterministic key:value or just value. Not both. The only reason it is like this is due to manual IP config. But perhaps we move that to a separate argument or storage structure. That way we can have “nic” be one of three things, and if set to “manual” it can look in a different key for the info it needs. For instance

**Comment by Torxed** [comment]
Separate PR for that tho I think

**Comment by svartkanin** [comment]
Just to understand, would we then do this as a breaking change for the nic and/or possibly handle existing configs with something like I showed in the previous comment?

When selecting manual, would the ip address info be a sub structure of the nic or independent? That would probably required some default value then...

**Comment by Torxed** [comment]
I think it should be independent. I can draw up a PR a bit later today just to show the concept :)

**Comment by svartkanin** [comment]
In the meantime how shall we proceed with this one here :)

**Comment by wllacer** [comment]
I'd  prefer it staying only under one header, unless there is a user need of configure both the automatic setups AND explictily configuring the nics.
Perhaps an structure like (in pseudo BNF)
```
nic := type | (type, ifname, (netparms ))*
ifname := IFNAME | *    #* a filler character
type := ISO | NW | manual | IWD ... whatever mode of network management we support
neparms := ipaddr, (gateway), (dns), ...| dhcp  ... whatever feels interesting
```
if we only specify *type* all the interfaces will be configured by the mechanism.
If we specify a list of triplets we configure each interface in another way (f.i. the ethernet port manual and the wlan via NM. and someday configure VPN ports)

IN the meantime, we could live for a couple of days without networking (ughh ) or go to the simplest solution (the untranslated string as parameter)

@svartkanin Daniel, the process to convert between versions as \__init\__.py is more simple than that. As an example of what we do today see:
https://github.com/archlinux/archinstall/blob/d9118a33b3a25839fde8626cde313115ddb79376/archinstall/__init__.py#L192-L197
Once we read the config file we simple change, add or delete the corresponding `arguments` entry

svartkanin:
I would also lean more towards a nested object since it will clearly state the connectivity between the nic config and all it's sub config elements. Unless, @Torxed you have something else in mind to disconnect the IP and other settings depending on the nic type and make them act and serve as an indipendent configuration

Torxed:
Go with your ideas, they’ll hold up!

svartkanin:
@Torxed @wllacer If you're happy with the state of this PR could we merge it in? It contains fixes that are quite helpful for others, especially the concatenation fix for DeferredTranslation. 

I'm happy to rework the nic configuration part in a next PR

Torxed:
Say no more :) just waiting for all discussion to finish hehe. Merged!

PR #977, opened 2022-02-13
Switch from flags to an 'additional repository' menu

dylanmtaylor:
This is a new rework, in addition to the fixes in #975.  The idea is that you can optionally select 'multilib' and 'testing' and it'll set the booleans to use the existing logic from the multilib and testing PRs. These will also be added to the schema and work with the configuration file.

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -353,7 +353,7 @@ class Installer:
 		self.log(f'Installing packages: {packages}', level=logging.INFO)
 
 		if (sync_mirrors := SysCommand('/usr/bin/pacman -Syy')).exit_code == 0:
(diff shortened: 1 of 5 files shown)


Torxed:
This feels like a core part of Arch Linux, so I don't see why we can't add one more menu item to the table.
I've merged the other PR and I can fix the merge conflicts in guided in a bit :)

dylanmtaylor:
> This feels like a core part of Arch Linux, so I don't see why we can't add one more menu item to the table. I've merged the other PR and I can fix the merge conflicts in guided in a bit :)

Awesome! I just fixed the merge conflict. I'd actually like to have a plugin that puts 'aur' in this menu, unofficially ;)

Torxed:
That would be legendary. Add a plug-in hook during setup of the menu. Which can load additional menu entries?

PR #978, opened 2022-02-13
enhacements to the menu infraestructure

wllacer:
Activate load of preset values in GeneralMenu
    Changed all select_functions definitions to the need of passing the preset value
 Partial management of preset values added to `ask_to_configure_network`, and corrected some problems re internationalization. See issue #966 and PR #976 (there is some duplicated code there. 976 should come first)


wllacer:
@torxed. Put this on hold until #976 is upstreamed and i'll have the time to merge the code for preset values into `ask...network`. A little delay is necessary

archinstall/lib/user_interaction.py
```diff
@@ -424,24 +424,38 @@ def read_packages(already_defined: list = []) -> list:
...
-def ask_to_configure_network() -> Dict[str, Any]:
+def ask_to_configure_network(preset :Dict[str, Any] = None) -> Dict[str, Any]:
 	# Optionally configure one network interface.
 	# while 1:
 	# {MAC: Ifname}
-
-	iso_config = _('Copy ISO network configuration to installation')
-	network_manager = _('Use NetworkManager (necessary to configure internet graphically in GNOME and KDE)')
+	iso_config = str(_('Copy ISO network configuration to installation'))
+	iso_config_nt = 'Copy ISO network configuration to installation' # do not translate
```
dylanmtaylor, line 432: This could be rewritten to be a little bit less repetitive.

iso_config = str(_(iso_config_nt))
network_manager = str(_(network_manager_nt))

wllacer:
Sure, but i am in doubt if the _nt strings will be catched that way by
internacionalización procedure.
And i was yesterday in a hurry to put It out, so i opted for the most
conservative solution.
Will check It as @svartkanin has a solution for the bug, so i have more
time for.the enhancements

wllacer:
@torxed The changes from PR #976 are integrated. So you can do what you want with it

As a bonus point i've put the use of preset values to most of the functions of the menu.
Still missing are
* harddrives
* disk_layout
* encryption_password
* superusers
* users
* profiles
Which need some special care. Probably disk_layout will be impossible to port as it is nowadays (but will be hopefully in the future)
Archinstall Language, root password and the --advanced options at swiss.py won't be using preset values

archinstall/lib/user_interaction.py
wllacer: I've adopted @svartkanin 's version, so this is solved

wllacer:
Wait a bit. I've found a small bug in the conversion of the nic formats

svartkanin:
Thanks for solving the nic bug

Torxed:
let me know if you want to complete the list above or if you want this merged :)

PR #979, opened 2022-02-13
Use archlinux as the default hostname

dylanmtaylor:


dylanmtaylor:
VirtualBox_Arch_12_02_2022_22_00_26
This is how the default looks with this PR applied

PR #980, opened 2022-02-13
Set default bootloader based on presence of UEFI

### LINKED ISSUE #970: For 2.4.0, if it's a BIOS system and only GRUB is available, default the selection to grub-install

dylanmtaylor:
Closes #970


dylanmtaylor:
This is tested and seems to be working :)

Torxed:
Nice one! :D

mhussaincov93:
hey,
if this is now a yes,
then does that mean that if detected that we are running on a bios based system as an exammple, we could doo away with the /boot seperate partition?
since bios systems don't require a seperate partition?
personal interest, I am on a bios based system.
Majid

PR #981, opened 2022-02-13
Set the default timezone to UTC in the menu

dylanmtaylor:
It's already the default, this just keeps the user from needing to make the selection unless they desire another timezone.


Torxed:
Awesome. I was on my phone yesterday so couldn't remember which ones needed fixing :) Great work!

PR #983, opened 2022-02-14
Add password strength check

mmdbalkhi:
# Add password strength check

Hello. I wrote a new feature for archinstall. This feature tells the user when the password is weak that the password is weak and it is better to write the password again. 

The reason for putting TODO in the code was also because of the interesting idea of @Torxed, for which I am very grateful ([in the discord](https://discord.com/channels/726808923662712894/726810321125376092/942832664396832899)).


dylanmtaylor:
We need to make sure that this does not cause issues with the configuration file. Also, I'm not sure if it should be strongly enforced. If I want to have my username be 'd' and my password also be 'd', nothing should stop me.

mmdbalkhi:
> We need to make sure that this does not cause issues with the configuration file. Also, I'm not sure if it should be strongly enforced. If I want to have my username be 'd' and my password also be 'd', nothing should stop me.

that's right. The function asks if you want to continue with this weak password

archinstall/lib/user_interaction.py
```diff
@@ -98,22 +98,51 @@ def sig_handler(sig :int, frame :Any) -> None:
...
+		and len(passwd) >= 8
+	):
+
+		log(
+            " * Your password too weak *\n",
+            "Strong passwords must contain lowercase, uppercase,",
+            "numbers and longer than 8 letters.",
+            fg="yellow",
+        )
+		time.sleep(2) # TODO: This is a hack, we should have a better way to do this.
```
dylanmtaylor, line 116: Why is there a sleep here?

mmdbalkhi: Unfortunately, the confirmation comes on the log, so I put it.

dylanmtaylor: Can we do this instead? If a user enters a weak password, they get a prompt saying that a weak password was provided and asking if they want to continue using the same selection system we have for the menus.

mmdbalkhi: > Can we do this instead? If a user enters a weak password, they get a prompt saying that a weak password was provided and asking if they want to continue using the same selection system we have for the menus.

First I wanted to do this using 'input'. But I thought it might not be interesting. If you have a suggestion, I will gladly implement it.

Torxed: We have a TextField I believe, that supports translation. We should use that :) @svartkanin knows most about this

svartkanin: You can remove the log and the sleep and move the current log text directly into the confirmation prompt that way it will be displayed on top.

Torxed: Merged :)

mmdbalkhi: So in this case we only use `select` class?

archinstall/lib/user_interaction.py
```diff
@@ -98,22 +98,45 @@ def sig_handler(sig :int, frame :Any) -> None:
...
 		if passwd != passwd_verification:
 			log(' * Passwords did not match * ', fg='red')
 			continue
 
-		if len(passwd.strip()) <= 0:
-			break
-
 		return passwd
-	return None
+	return
```
Torxed, line 137: linters and guidelines tells us to explicitly return `None` even tho they in the end does exaxtly the same thing IIRC

archinstall/lib/user_interaction.py
```diff
@@ -98,22 +98,43 @@ def sig_handler(sig :int, frame :Any) -> None:
 
 	return True
 
+def is_strong_password(passwd :str) -> bool:
+
+	if not (
+		any(char.isdigit() for char in passwd)
```
mmdbalkhi, line 104: I use `black` and it suggested that I break the line. I will solve the same now.

Torxed:
Just to update with some of the information from Discord.
* We don't want the `is_password_strong` function to be intrusive, meaning it should only report True or False if the password is strong or not. And any other menu or notification should be done elsewhere
* We want the function to be simple but yet support the basic patterns
* No external libraries should be needed but inspiration can be gathered

Example passwords to test:
 * iamastrongpasswordbecauseiamlongeventhoughiamjustlowercase
 * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
 * hdtvsfeosnrsrdv
 * fooBARb@zzThisIsAsecretPassword&probablyV€rySAFE

mmdbalkhi:
Hi @Torxed 

## The result of checking the passwords (only the power of the passwords without any menu, etc.)

```python
In [1]: def check_password_strong(passwd: str) -> bool:
   ...: 
   ...:     symbol_count = 0
   ...:     if any(character.isdigit() for character in passwd):
   ...:         symbol_count += 10
   ...:     if any(character.isupper() for character in passwd):
   ...:         symbol_count += 26
   ...:     if any(character.islower() for character in passwd):
   ...:         symbol_count += 26
   ...:     if any(not character.isalnum() for character in passwd):
   ...:         symbol_count += 40
   ...: 
   ...:     if symbol_count ** len(passwd) < 10e20:
   ...:         return False
   ...: 
   ...:     return True

In [2]: check_password_strong("fooBARb@zzThisIsAsecretPassword&probablyV€rySAFE")
Out[2]: True

In [3]: check_password_strong("123")
Out[3]: False

In [4]: check_password_strong("hdtvsfeosnrsrdv")
Out[4]: True

In [5]: check_password_strong("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
Out[5]: True

In [6]: check_password_strong("iamastrongpasswordbecauseiamlongeventhoughiamjustlowercase")
Out[6]: True
```

If there is a problem, tell me so I can solve it.

I do not write commit because the number of useless commitments increases :)

PR #987, opened 2022-02-16
profiles/sway: replace alacritty with foot

panchoh:
In Sway 1.7, the terminal emulator in the default config file has been changed to foot.

Without this change, on first boot there is no way to launch a terminal inside sway, since `dmenu` seems to be botched. 

See https://github.com/swaywm/sway/releases/tag/1.7

Cheers, fellas.

### DIFF
--- a/profiles/sway.py
+++ b/profiles/sway.py
@@ -13,7 +13,7 @@ __packages__ = [
 	"grim",
 	"slurp",
 	"pavucontrol",
-	"alacritty",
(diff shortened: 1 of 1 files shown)


Torxed:
thanks for not only bringing it to our attention but also fixing it :)

panchoh:
> thanks for not only bringing it to our attention but also fixing it :)

Glad to be of service!

Issue #986, opened 2022-02-15
additional user selection not visible in main menu


Byter09:
Hello,

I wanted to try out the latest master on my PC because I ran into an UTF-8 related error because apparently the manufacturer put `a1 b1` bytes into the model name. Hah.

![IMG_20220215_184812](https://user-images.githubusercontent.com/16060697/154125113-bda5f2b3-bec7-456e-8373-5868140d3562.jpg)

So I followed the steps in the readme, checked out master installed, ran, and was pleasantly surprised by a nice menu.

Everything was fine (even the harddrive selection didn't stutter because of the weird model name), until I reached the part about additional users.

I entered my preferred name, typed the password in twice, and was sent back to main menu, but it still showed an empty array like this:

![IMG_20220215_182752](https://user-images.githubusercontent.com/16060697/154125174-9dc60817-f223-40e5-8ce1-f070755fde37.jpg)

I tried again, same result. I didn't want to continue at that point because I knew I'm on master and something may be broken. I saw issues and pull requests regarding this main menu persistence, but couldn't find anything about this specifically. I hope it's not a duplicate!

My question is, is this a bug? Does it still correctly save my input internally and I just can't see it, but could safely resume?

Thanks for the installer. Makes Arch installation less of a scary beast to me haha.

Torxed:
ah, yea because when you set a root password it resets the user list. well get that sorted out :)

Torxed:
It's our code - it resets the user and superuser options I believe.
This is the code that's the issue (for reference): https://github.com/archlinux/archinstall/blob/d9118a33b3a25839fde8626cde313115ddb79376/archinstall/lib/menu/selection_menu.py#L575-L584

And it's called every time the menu system is shown, so not only during startup: https://github.com/archlinux/archinstall/blob/d9118a33b3a25839fde8626cde313115ddb79376/archinstall/lib/menu/selection_menu.py#L469

So every time the main menu is drawn, it resets the user and superuser entries :)
We'll fix that before launch.

svartkanin:
Actually I think @Torxed that @Byter09 had a point there, I don't think we reset the user/superuser if a root password is entered first. I'll try to reproduce this somehow...

svartkanin:
The reason for the above problem is, that a user was specified and then set as a superuser, but the display in the menu only displays regular users. So if the installation was actually completed the everything would have been created correctly. 

[...]

PR #993, opened 2022-02-18
Rework the user/superuser configuration

svartkanin:
This addresses #986.

This is reworking the user/superuser configuration flow.
Previously the root password and superusers where mutually exclusive (kinda), one could only select a root password or a superuser. However, there was a sneaky way to get around this by simply setting a root password and then creating a user and promoting them to superuser. 

With these changes:
- Allow both a root password and a superuser to be configured
- Allow to configure multiple superusers at the same time via the superuser menu entry
- Downgrade the user select to be only a user select


wllacer:
There is a use case you should check. When the same userid appears in both lists.  Now it is very difficult to check (or at least ugly coding) and -honestly- i have no idea what the result would be.
I think long term we should only have one list (dict) with a sudoer attribute to distinguish each case. I'm doing some work around it in my attempt to keep previous values, and -although i'm now stuck because some UI design issues, so it is not for the short term- using it internally. The format conversion is a bit ugly, but gives a lot of flexibility of use

Torxed:
> I think long term we should only have one list (dict) with a sudoer attribute to distinguish eac

I think the result would be that it will simply try to create the user twice, which should be accepted but errors ignored.
And in the second/first attempt of creating it as a super-user, it will simply create a `/etc/sudoers.d/entry` making either the new or existing user a sudo user :)

svartkanin:
> I see you have done away with the forced parameter for superusers. It was there to "force" the existence of having at least one superuser if no root password was specified. I miss this control in your code.
It is treated at _missing_config, and thus in the exec_func of install. I have the feeling that I missed something in such a case (not two options can be none). Please have a look at it and give me your opinion

The thing I have removed was that the superuser entry does no longer depend on the root password config by removing the `dependencies_not` which served as a XOR for root password <-> superuser. 
The missing configuration check remained as is and I don't change anything there (except for removing the timezone check because we have a default value there now). 
Please let me know if you think any logic got messed up with the changes

svartkanin:
Are there any concerns with this left?

Torxed:
Dont think so :) will merge during the day

Torxed:
Maybe the `x: self...` should have been there >_O

PR #996, opened 2022-02-19
Fix (some) mypy things

svartkanin:
An incredibly unspectacular PR to fix some of the mypy issues, I'll try to add more soon

### DIFF
--- a/archinstall/lib/disk/filesystem.py
+++ b/archinstall/lib/disk/filesystem.py
@@ -7,6 +7,7 @@ from typing import Optional, Dict, Any, TYPE_CHECKING
 # https://stackoverflow.com/a/39757388/929999
 if TYPE_CHECKING:
 	from .blockdevice import BlockDevice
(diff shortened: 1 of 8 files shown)


Torxed:
We should cross reference this with #846 : ) I guess I can merge this one after the flake8 issues are fixed and then merge those into my WIP PR.

svartkanin:
Ahh cool didn't know there was an open issue for this

svartkanin:
I see there's a mypy type checking running in the pipeline, I suspect that one is not actually stopping any mypy issues?

Torxed:
the pipeline has `|| exit 0`: https://github.com/archlinux/archinstall/blob/4b3b21ed756463914d618abe11345c9839217473/.github/workflows/mypy.yaml#L16

which tells github that it "always exits nicely" until we got all the mypy issues sorted. its not a huge priority but would be nice :)

Issue #991, opened 2022-02-18
Rework the network configuration


svartkanin:
As discussed with @Torxed and @wllacer we'll update the network configuration to a single type and include type specific configuration as a subelement of the data structure. I'll have a look at this soon and will keep the backwards compability (which we put in with the prev PR already) in place, unless we want to only support the new configuration?

Torxed:
Perfect, thanks!

svartkanin:
What's the best way to document them?

Torxed:
> What's the best way to document them?

I will put it in the release notes, currently the format for v2.4.0 is:

---------

# Breaking Changes

 * `archinstall.SysCommand*` now properly raises exceptions when exit codes are bad. This used to go by unnoticed in some cases, which caused silent errors. The flip-side to this is that it may now be overly aggressive and cause breakage at certain points. In which case the call itself should be wrapped around a catch block. As an example:
   ```python
   if (result := archinstall.SysCommand("ls --breaking-arguments")).exit_code != 0
   ```
   Should now be:
   ```python
   try:
       result = archinstall.SysCommand("ls --breaking-arguments")
   except SysCallError as error:
       if error.exit_code != 0:
   ```
   This avoids the risk of calls silently passing errors.
 * `archinstall.Filesystem().load_layout()` now looks for the keyword `"wipe": True` and not `"format": True`. This is to avoid confusion as to what the menu alternatives does, for instance *"mark partition for wiping"* as well as avoid mixing up the format options like `"filesystem"` and other options during formatting.

# New Features

 * A completely new menu system *(still accessibility friendly)* using [simple-term-menu](https://github.com/IngoMeyer441/simple-term-menu) as the base *(bundled with this source to avoid external dependencies. A big thank you to Ingo Meyer and his employer at the time)*:
   ![screenshot](https://user-images.githubusercontent.com/861439/152638047-abe56543-b109-4e7e-96e4-3e973c56c9fa.jpg)
 * Localization/Translation options for all archinstall output *(Currently `us`, `es` and `se` is available)*. To contribute please read the [translation contribution README](https://github.com/archlinux/archinstall/blob/f6d133804b7fdfab5a00d95a1985c3e220935ac1/archinstall/locales/README.md)

# Minor changes and bug fixes

 * `archinstall.define_arguments()` has been added to separate the logic of setting up `ArgumentParser` arguments.
 * Fixed all arguments so that they support both `--x="something"` and `--x "something"`.
[...]

PR #1001, opened 2022-02-22
Rework network config

svartkanin:
Addresses #991

This is a rework of the network configuration. The new input structure, mainly relevant for the configuration files is 
```
"nic": {
    "type": <ISO|NM|MANUAL>,
    "iface": <str>,  # interface name
    "ip": <ip>,
    "dhcp": <boolean>,
    "gateway": <ip>,  #optional
    "dns": [<str>]  #optional
}
```

I've created a new container `NetworkConfiguration` to store the configuration data. This allows to group some functionality together and give guided.py and swiss.py a bit more air to breath. 

I've updated the documentation as well (hopefully I didn't miss anything). 

# Breaking changes
Existing custom configuration files will not work anymore with the new format and if any `nic` attribute is defined it will be ignored unless there are the new fields specified. 
The new configuration format for `nic` is expected to be in the following format 
```
"nic": {
    "type": <ISO|NM|MANUAL>, 
    "iface": <str>, 
    "dhcp": <boolean>,
    "ip": <ip>,
    "gateway": <ip>,  #optional
    "dns": [<str>]  #optional
}
```

- type: ISO -> Copy the current iso configuration, NM -> Setup and configure NetworkManager, MANUAL -> Use custom config
- iface: Interface to configure
- dhcp: If set to false  `auto DHCP` will be used otherwise `static IP`
- ip: Ip address, is mandatory if DHCP is false
- gateway: Optional gateway 
- dns: Optional list of DNS servers

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -6,5 +6,5 @@ max-complexity = 40
 max-line-length = 236
 show-source = True
 statistics = True
-per-file-ignores = __init__.py:F401,F403,F405 simple_menu.py:C901,W503 guided.py:C901
+per-file-ignores = __init__.py:F401,F403,F405 simple_menu.py:C901,W503 guided.py:C901 network_configuration.py:F821
 builtins = _
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -123,7 +123,7 @@ def parse_unspecified_argument_list(unknowns :list, multiple :bool = False, erro
 					print(f" We ignore the entry {element} as it isn't related to any argument")
 	return config
 
-def get_arguments():
+def get_arguments() -> Dict[str, Any]:
 	""" The handling of parameters from the command line
 	Is done on following steps:
 	0) we create a dict to store the arguments and their values
@@ -203,6 +203,8 @@ def load_config():
 		storage['gfx_driver_packages'] = AVAILABLE_GFX_DRIVERS.get(arguments.get('gfx_driver', None), None)
 	if arguments.get('servers', None) is not None:
 		storage['_selected_servers'] = arguments.get('servers', None)
+	if arguments.get('nic', None) is not None:
+		arguments['nic'] = NetworkConfiguration.parse_arguments(arguments.get('nic'))
 
 
 def post_process_arguments(arguments):
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -203,13 +203,7 @@ def load_config():
 		storage['gfx_driver_packages'] = AVAILABLE_GFX_DRIVERS.get(arguments.get('gfx_driver', None), None)
 	if arguments.get('servers', None) is not None:
 		storage['_selected_servers'] = arguments.get('servers', None)
-	if nic_config := arguments.get('nic', {}):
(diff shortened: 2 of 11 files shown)


Torxed:
Great PR! And Thank you for summarizing the breaking changes.
I'll have a merge spree this weekend :) So hang tight!

wllacer:
Maybe a bit too late, but couple of comments. I like the way you have solved it, but, could it be extended to a list/dict with several entries?. If a machine has ethernet and wifi, now you define explicitly either of them or are both left to NM to manage. There is no way to explicitly defined both (dual ethernet configurations are not that popular as they were, but would fall in the same category).

You seem to have done away with the conversion from the old to the new format at `load_config`. Am I right? Any technical reason behind? It's a matter of schools, but intentional compatibility breaks make me very nervous (especially if they seem relatively easy to brigde)

svartkanin:
> Maybe a bit too late, but couple of comments. I like the way you have solved it, but, could it be extended to a list/dict with several entries?. If a machine has ethernet and wifi, now you define explicitly either of them or are both left to NM to manage. There is no way to explicitly defined both (dual ethernet configurations are not that popular as they were, but would fall in the same category).

That's a great point actually. I was mainly trying to keep the existing functionality but thinking about it I think it makes a lot of sense to be able to configure multiple ifaces.
With this there would be some significant changes though, we will have to rework how the network menu works to allow multiple configurations.

@Torxed @wllacer My proposal in that case is to allow 
"Copy..." as rhe single option and then as an alternative to configure multiple interfaces manually. Meaning that qe don't allow copy the current config and configuring an interface.
If it's okay I'd prefer to do that in a different PR though 

@wllacer My reasoning for removing the backwards compatibility was that it's not a major break, meaning that it won't break resulting in an error but simply take it as "no config provided". It would mean that we don't have to keep maintaining it. If you think it's worth it I'm happy to keep it in there :)
As an alternative, what about printing a message to inform that they're using an old incompatible network config and should switch to the new one?

svartkanin:
> A thought, is it worth creating a config lifting tool? --convert-conf for instance?

Would that be to convert older to non backwards compatible configs?

Torxed:
I'm happy with this as it is, and we can open separate PR's for the rest.

I was thinking the tool could read the old config format and convert it - making breaking changes smoother. Probably a lot of effort for this tho.

svartkanin:
Sounds good, I'll pick up the extension as this is merged.

I think having a whole tool just for this change is a bit too much :) 

Shall we put in the backwards compatibility again or we're happy with the breaking changes?

Torxed:
> Sounds good, I'll pick up the extension as this is merged.
> 
> I think having a whole tool just for this change is a bit too much :)
> 
> Shall we put in the backwards compatibility again or we're happy with the breaking changes?

Happy with it. We already have a lot of breaking changes, I'll announce it and see if I can come up with a technical solution to pipe old configs to the new layout transparently or something :)

Torxed:
There were some parallel changes to `user_interaction.py`. I think I got it clean in, but poke me if something's wrong :)

Issue #992, opened 2022-02-18
AttributeError: 'NoneType' object has no attribute 'get'


tttffff:
When passing in `--disk_layout` with my admittedly non-standard layout, I am getting odd behavior. I can not replicate this when using the standard layout, but maybe it is a symptom of another bug. This is using the latest [commit on master](https://github.com/archlinux/archinstall/commit/62a6aec197b0e83ab91fdec1ab15427ecf749dac).

- If `"encryption": true` is set and nothing is entered for `Set encryption password` on the interactive screen, I am prompted to enter the encryption password after clicking `Install` and the installation is a success (this is expected behavior, and how I am installing Arch at the minute).
- If `"encryption": true` is set and something is entered into `Set encryption password` on the interactive screen, the below error message is given (not-expected behavior).
- If `"encrpytion": false` is set and nothing is entered for `Set encryption password` on the interactive screen, installation will complete without encryption of the drive (expected behavior).
- If `"encryption": false` is set and something is entered into `Set encryption password` on the interactive screen, the below error message is given (not-expected behavior).
- Not setting `"encryption"` in the disk_layout acts the same as setting it to false.

The error message:
```
 ! Formatting [BlockDevice(/dev/sda, size=32.0GB, free_space=3145kB+3146kB, bus_type=sata)] in 5....4....3....2....1....
Creating a new partition label on /dev/sda
Adding partition....
Adding partition to BlockDevice(/dev/sda, size=32.0GB, free_space=34.4GB, bus_type=sata), 3MiB->2048MiB
Formatting /dev/sda1 -> vfat
Marking partition Partition(path=/dev/sda1, size=2.0, PARTUUID=40e9df38-01, fs=vfat) as bootable.
Setting boot on on (parted) partition index 1
Adding partition....
Adding partition to BlockDevice(/dev/sda, size=32.0GB, free_space=3145kB+32.2GB, bus_type=sata), 2051MiB->100%
Encrypting Partition(path=/dev/sda2, size=30.0, PARTUUID=40e9df38-02, parent=/dev/sda, fs=crypto_LUKS) (This might take a while)
Formatting /dev/mapper/ainstsda2 -> btrfs
'NoneType' object has no attribute 'get'
Traceback (most recent call last):
  File "/root/kickassarchinstall/archinstall/archinstall/examples/guided.py", line 129, in perform_installation
    installation.mount_ordered_layout(archinstall.arguments['disk_layouts'])
[...]

PR #1003, opened 2022-02-22
Solves Issue992

wllacer:
A oneliner (i dread them) to solve the issue behind #992. 
Really it only happens if there is also a `generate-encryption-key-file` entry  in the disk_layout structure. I thought we had gotten rid of it, if it was a single disk installation. The error happened when checking if it is needed, so now it is indiferent


wllacer:
OOps i ddn't add the commit. And now it doesn't let me push it. Closing and creating other

PR #1004, opened 2022-02-22
Issue992

wllacer:
A oneliner (i dread them) to solve the issue behind https://github.com/archlinux/archinstall/issues/992.
Really it only happens if there is also a generate-encryption-key-file entry in the disk_layout structure. I thought we had gotten rid of it, if it was a single disk installation. The error happened when checking if it is needed, so now it is indiferent.

Help. Why do I push that many old commits every time I upload something ?


Torxed:
The reason is that branch issue992_2 was forked from https://github.com/wllacer/archinstall/commit/bb98594054939e469d07f704e70f597a8b379657 and master has essentially not been synced in since then :) You have to options:
```
$ git checkout issue992_2
$ git pull origin master
```
And it will sync your master into your branch, eliminating *some* of the commits because git realizes they are already included.
Or, to get a real truth:
```
$ git add remote upstream https://github.com/archlinux/archinstall.git
$ git checkout issue992_2
$ git pull upstream master
```
And this will sync upstream master into your branch, so anything that exists upstream will not be visible in the PR :)
I did the later now. It won't remove them after the fact here on github, but it's only really going to merge your changes in the background :)

PR #1005, opened 2022-02-25
add new widget ListManager

wllacer:
This adds a new "widget" i think can be useful to others. I decided to upstream it standalone, so anyone can use it, if it's useful for their needs, as I have a lot of work to do before any of the application specific enhancements I will use it, are ready to upstream (or have to change things, as , as usual @svartkanin has been faster than me ;-)
  
## Purpose
ListManager is a widget based on `menu` which allows the handling of repetitive operations in a list.
Imagine you have a list and want to add/copy/edit/delete their elements. With this widget you will be shown the list
```
Vamos alla

Use ESC to skip


> uno : 1
  dos : 2
  tres : 3
  cuatro : 4
  ==>
  Add
  Confirm and exit
  Cancel
(Press "/" to search)
```
Once you select one of the elements of the list, you will be promted with the action to be done to the selected element
```

  uno : 1
  dos : 2
> tres : 3
  cuatro : 4
  ==>
  Confirm and exit
  Cancel
(Press "/" to search)

Select an action for < {'tres': 3} >


  Add
  Copy
  Edit
  Delete
> Cancel
```
You execute the action for this element (which might or not involve user interaction) and return to the list main page
till you call one of the options `confirm and exit` which returns the modified list or `cancel` which returns the original list unchanged.
If the list is empty one of the actions can be defined as default (usually Add). We call it **null_action**

The default implementation can handle simple lists and a key:value dictionary. The default actions are the shown above.
A sample of basic usage is included at the end of the source.

More sophisticaded uses can be achieved by
* changing the action list and the null_action during intialization
```
	opciones = ListManager('Vamos alla',opciones,[str(_('Add')),str(_('Delete'))],_('Add')).run()
```
* And using following methods to overwrite/define user actions and other details:
* * `reformat`. To change the appearance of the list elements
* * `action_list`. To modify the content of the action list once an element is defined. F.i. to avoid Delete to appear for certain elements, or to add/modify action based in the value of the element.
* * `exec_action` which contains the actual code to be executed when an action is selected

The contents in the base class of this methods serve for a very basic usage, and are to be taken as samples. Thus the best use of this class would be to subclass in your code

```
	class ObjectList(archinstall.ListManager):
			self.ObjectAction = [... list of actions ...]
			self.ObjectNullAction = one ObjectAction
		def __init__(prompt,list):
			super().__init__(prompt,list,ObjectActions,ObjectNullAction)
		def reformat(self):
			... beautfy the output of the list
		def action_list(self):
			... if you need some changes to the action list based on self.target
		def exec_action(self):
			if self.action == self.ObjectAction[0]:
				performFirstAction(self.target, ...)

	...
	resultList = ObjectList(prompt,originallist).run()
```
## things that can be improved or changed within the basic framework
- [ ] Visual enhancements, like separators for the submenu. There are limitations, though, of what we can achieve.
- [ ] Use separated screens for the submenu actions. I prefer it being on the same page (showing context) but others might think otherwise
- [ ] Paging for long lists. I don't believe there is much need for it at Archinstall, but noone knows
- [ ] Ensure that accesibility is not impaired excesively

### DIFF
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -34,6 +34,8 @@ from .lib.storage import *
 from .lib.systemd import *
 from .lib.user_interaction import *
 from .lib.menu import Menu
+from .lib.menu.list_manager import ListManager
+from .lib.menu.text_input import TextInput
 from .lib.menu.selection_menu import (
 	GlobalMenu,
 	Selector,
--- a/archinstall/lib/menu/list_manager.py
+++ b/archinstall/lib/menu/list_manager.py
@@ -10,33 +10,33 @@ Use ESC to skip
 
 
 > uno : 1
-  dos : 2
-  tres : 3
-  cuatro : 4
-  ==>
-  Confirm and exit
-  Cancel
+dos : 2
+tres : 3
+cuatro : 4
+==>
+Confirm and exit
+Cancel
 (Press "/" to search)
 ```
 Once you select one of the elements of the list, you will be promted with the action to be done to the selected element
 ```
 
-  uno : 1
-  dos : 2
+uno : 1
+dos : 2
 > tres : 3
-  cuatro : 4
-  ==>
-  Confirm and exit
-  Cancel
+cuatro : 4
+==>
+Confirm and exit
+Cancel
 (Press "/" to search)
 
 Select an action for < {'tres': 3} >
 
 
-  Add
-  Copy
-  Edit
-  Delete
+Add
+Copy
+Edit
+Delete
 > Cancel
 ```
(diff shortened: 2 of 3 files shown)


wllacer:
Small additions after some experience with the widget
Null_action appears now in the main list (to simplify additions to the list). Examples above are updated to show the result
Formatted data are now at the from to the actions submenu

Torxed:
I'm genuinely impressed with the cool new stuff you guys are doing with the menu end of things.
This will come to be very useful in several steps, like removing a user after it was added :)

PR #1008, opened 2022-02-27
User Management via lists

wllacer:
Management of users and superusers in the menu is done via list managers

Includes/relies  on @svartkanin 's PR #993  and my PR #1005. An offshot of my own PR #978, but not using the mechanism ins
Selecting one of  the user options will result in, when an user is selected 
user_list2
Depending which type we are editing adding / modifing users is only possible if they are of the same type. Those of the other can only be converted to/from superuser status to ordinary status.
Code is -almost- ready for common user management but it needs changes at Installer level.

I'd appreciate suggestions regarding the user interface and handling.

### DIFF
--- a/archinstall/lib/user_interaction.py
+++ b/archinstall/lib/user_interaction.py
@@ -9,13 +9,16 @@ import signal
 import sys
 import time
 from collections.abc import Iterable
+from copy import copy
 from typing import List, Any, Optional, Dict, Union, TYPE_CHECKING
 
 # https://stackoverflow.com/a/39757388/929999
 from .menu.text_input import TextInput
+from .models.network_configuration import NetworkConfiguration, NicType
 
 if TYPE_CHECKING:
 	from .disk.partition import Partition
+	_: Any
 
 from .disk import BlockDevice, suggest_single_disk_layout, suggest_multi_disk_layout, valid_parted_position, all_blockdevices
 from .exceptions import RequirementError, DiskError
@@ -34,6 +37,10 @@ from .mirrors import list_mirrors
 from .translation import Translation, DeferredTranslation
 from .disk.validators import fs_types
 from .packages.packages import validate_package_list
+
+
+# used for signal handler
+SIG_TRIGGER = None
 
 
 # TODO: These can be removed after the move to simple_menu.py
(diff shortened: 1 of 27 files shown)


svartkanin:
@wllacer this is really neat! Some things I've noticed, when adding superusers, they will appear as regular users in the main menu. Is that intentional? Or maybe it doesn't include the my PR yet that is supposed to address that? 

When entering the user creation menu the first time, it will do the single input, would it be possible to skip that one and immediately go to this new list widget instead? I suppose it would then look someting like 

```  
  ==>                                                                                                                                                          
  Add an user                                                                                                                                                  
  Confirm and exit                                                                                                                                             
  Cancel
```

I understand that `==>` is merely used as a delimiter, I wanted to look into the simple_menu.py file to see how easy it'd be to modify that to allow for a visual separater instead, which might be also useful for the main menu to have.
```
  Set automatic time sync (NTP)       SET: True                                                                                                                

  Install (2 config(s) missing)                                                                                                                                
  Abort   
```

svartkanin:
Ohh this might also be very useful for the partitioning actually!

archinstall/lib/user_interaction.py
```diff
@@ -1095,3 +1059,122 @@ def generic_multi_select(p_options :Union[list,dict],
 						sort=sort,
 						multi=True,
 						default=default)
+
+
+class UserList(ListManager):
```
svartkanin, line 1064: I think it'd be good to extract this into it's own file, since user_interactions.py is already massive :)

wllacer: Totally Agree. It's just where to put it ...  Perhaps the best solution would be to convert the current file into a directory. It is one of those ideas waiting for time

wllacer:
> @wllacer this is really neat! Some things I've noticed, when adding superusers, they will appear as regular users in the main menu. Is that intentional? Or maybe it doesn't include the my PR yet that is supposed to address that?

I used your PR as starting point, but for the superuser screen, all users are displayed, but the only action you can do with regular users is to promote them to superusers. And the opposite case for regular users. I know it may be look strange for the first time, but during testing i felt it was the most confortable way to manage users/superusers, specially if i created one user with the wrong attributes; barring having an unified menu entry. The program takes care to put them in the correct place afterwards

> 
> When entering the user creation menu the first time, it will do the single input, would it be possible to skip that one and immediately go to this new list widget instead? I suppose it would then look someting like
> 
> ```
>   ==>                                                                                                                                                          
>   Add an user                                                                                                                                                  
>   Confirm and exit                                                                                                                                             
>   Cancel
> ```
It can be skipped (is a parameter during the List initalization) but the Add action wouldn't appear. I didn't thought that you could not want an action when the list is empty and a diferent action from the main menu. I will give it some tought. I believe it will be an easy and interesting addition
 
> I understand that `==>` is merely used as a delimiter, I wanted to look into the simple_menu.py file to see how easy it'd be to modify that to allow for a visual separater instead, which might be also useful for the main menu to have.

> 
> ```
>   Set automatic time sync (NTP)       SET: True                                                                                                                
> 
>   Install (2 config(s) missing)                                                                                                                                
>   Abort   
> ```

wllacer:
Me too, but if It is posible , i Haven't been able to find It at
simple_menu.py, and the separator and it's handling is done by me at
ListManager.py
I have a couple of half backed ideas to have a Sort of general framework
for separators and exit functions at this Menu derived classes. But i need
more refinement (more use cases)

wllacer:
I developed It with partitioning in  mind  but User Management is a lot
easier target for a first implementation.
Some preliminary work is already done, though

wllacer:
> 
> When entering the user creation menu the first time, it will do the single input, would it be possible to skip that one and immediately go to this new list widget instead? I suppose it would then look someting like
> 
> ```
>   ==>                                                                                                                                                          
>   Add an user                                                                                                                                                  
>   Confirm and exit                                                                                                                                             
>   Cancel
> ```
I've made a change to the base ListManager (PR #1005 and here) that allows this behaviour. In fact it will allow to have an action to start when the list is empty and one as general action (like `Add an user` above). In this particular case ( User Management) , I think, forcing to add an user if the list is empty is the most sensible path.
OTOH, it only happens the first time you enter one of the user/superuser entries. When accessing the other list, you will already have some users defined. It's just the point I'm not that happy with the result. If I would start with the empty list menu, would be not that weird.

archinstall/lib/user_interaction.py
wllacer: I've opened a discussion #1010 about the future of user_interaction.py. Let's hope we get a good solution on the long term for it

wllacer:
@torxed. No the problem was previous to merging this PR. Ought to be solved via another PR . This had some dumb merge conflicts. I hope I have solved them also

Issue #1002, opened 2022-02-22
Save current configuration


svartkanin:
I'd like to add a new menu option to save the current configuration to file. 
I imaged it as a simple
```
Install (3 config(s) missing)                                                                                                                                
Save configuration
Abort             
```

and then specifying a filename when clicking on it

Torxed:
Sounds good. its a bit more clear than `--dry-run` and since we now store it neatly in a next-gen (buzzzz) menu system, it makes perfect sense :)

wllacer:
Great idea. You can use and extend the `output_configs` routine at *configurations.py*

wllacer:
@svartkanin  How do you limit the number of options shown in the menu? 
Btw. we would need some kind of visual clue that more options are available.

wllacer:
No. The last one is profiles.
I have to move with the cursor down to get at the end.
and i have a lot of free space in my terminal ...

Torxed:
@svartkanin I think wllacer is referring to this:

![archinstall](https://user-images.githubusercontent.com/861439/156076634-9dd18d13-4d57-4d84-a1f8-a7c2085b4def.gif)

svartkanin:
Ohh wow I did not see that behavior on mine :O
My first assumption is that the preview takes too much space, that's an easy fix

OJFord:
Or/also have `--dry-run` create a file. A naive installer otherwise reaches the end and then gets only last part of the JSON document on screen, the rest disappeared with no scrollback.

Log/interactive messages also go to stdout (seems wrong itself), so you also can't `archinstall --dry-run > config.json`. If you `| tee config.json` then of course it's not a JSON file without first removing all the prompts and responses from the beginning, and it runs in a unpaged/non-interactive/curses-less way.

Torxed:
> Or/also have `--dry-run` create a file. A naive installer otherwise reaches the end and then gets only last part of the JSON document on screen, the rest disappeared with no scrollback.
> 
> Log/interactive messages also go to stdout (seems wrong itself), so you also can't `archinstall --dry-run > config.json`. If you `| tee config.json` then of course it's not a JSON file without first removing all the prompts and responses from the beginning, and it runs in a unpaged/non-interactive/curses-less way.

`--dry-run` saves the configuration. if it doesnt its a bug :) No need to pipe anything, files are created under /var/log/archinstall

OJFord:
[...]

PR #1009, opened 2022-02-28
Menu option save config

svartkanin:
Addressing #1002

Adding a new menu option to save the current configuration to file. It allows for saving individual configurations or all of them.

save_option

In addition, I've also included a preview for missing configurations to start the installation process (in case people might not find it obvious what's missing)

install_output

### DIFF
--- a/archinstall/lib/menu/selection_menu.py
+++ b/archinstall/lib/menu/selection_menu.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 import sys
 import logging
 
-from typing import Callable, Any, List, Iterator
+from typing import Callable, Any, List, Iterator, Tuple, Optional
 
 from .menu import Menu
 from ..general import SysCommand, secret
@@ -12,7 +12,7 @@ from ..output import log
 from ..profiles import is_desktop_profile
 from ..disk import encrypted_partitions
 from ..locale_helpers import set_keyboard_language
-from ..user_interaction import get_password, ask_for_a_timezone
+from ..user_interaction import get_password, ask_for_a_timezone, save_config
 from ..user_interaction import ask_ntp
 from ..user_interaction import ask_for_swap
 from ..user_interaction import ask_for_bootloader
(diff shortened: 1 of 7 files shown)


wllacer:
Wow !!!
Have a look at the changes for menu.py at my PR #1005.  There i decided to stop adding more parameters to Menu and use **kwargs for new additions to be passed to TerminalMenu (simple_menu.py), so there is no need for you to handle at Menu level the preview* parameters

archinstall/lib/menu/selection_menu.py
```diff
@@ -544,31 +582,37 @@ def exit_callback(self):
...
-			missing += 1
-		if not check('timezone'):
-			missing += 1
+			missing += ['Audio']
 		if not check('!root-password') and not check('!superusers'):
-			missing += 1
+			missing += [str(_('Either root-password or at least 1 superuser must be specified'))]
 		if not check('harddrives'):
-			missing += 1
+			missing += ['Harddrives']
```
dylanmtaylor, line 612: Nit pick: This shouldn't be one word.

dylanmtaylor: Or, alternatively it could just be 'Drives'.

Torxed: Might need to be changed elsewhere too if we do. Visually it probably doesn't matter if we only change here. But the heck for `if .."harddrives"` has to be changed to `if .."hard drives"` as well where we rely on this keyword :)

Torxed:
Managed to peace together the parameter changes I think :)

svartkanin:
> Have a look at the changes for menu.py at my PR #1005. There i decided to stop adding more parameters to Menu and use **kwargs for new additions to be passed to TerminalMenu (simple_menu.py), so there is no need for you to handle at Menu level the preview* parameters

@wllacer yes I saw that but I had some concerns with that (i think I left a comment on your pr). I think kwargs hides information and it makes it difficult to search for usages of parameters throughout the code. I get that the parameters here are quite extensive, however, I think the benefits are still higher in keeping it explicit :)

Torxed:
Yea I stopped using `**kwargs` only because it's ambiguous.
It also makes it hard for code linting and stuff, as the list of supported objects that `kwargs` can take can grow indefinitely.

I like the method tho, makes the init look a lot cleaner ^^

PR #1015, opened 2022-02-28
Fixed Readme.md with Absolute Path

garvit-joshi:
While going through pip package some of the hyperlinks have relative path instead of the absolute path which when clicked goes to a non-relevant page.

image


Torxed:
Thank you! I'll bare this in mind going forward.

Issue #1013, opened 2022-02-28
[master] Cannot select/enter super-users (TypeError: GlobalMenu._setup_selection_menu_options)


Torxed:
![screenshot](https://user-images.githubusercontent.com/861439/156018652-83a31d43-07b4-4e3d-bd6d-09d59c0e99dd.jpg)

@wllacer / @svartkanin I think I bodged a PR. I think your help would solve this faster than I can.
If you're busy, that's fine - I'll try again tomorrow.

bebetoalves:
Maybe it's because of #993 ?

Torxed:
> Naw. It was just a parameter definition at a lambda, which went missing during the merges ... Apropos. the password weakness check (#983) almost drove me crazy the first time i saw it. Good feature but a PITA for tests ;-)

Speaking of which, we should make sure that doesn't trigger if `--advanced` or something is given. And especially `--silent`.

PR #1016, opened 2022-02-28
Solves issue 1013

wllacer:
Simple patch to solve issue #1013. Superuser management can be invoked now


Torxed:
Thank you! Didn't have time to investigate earlier :)

PR #1017, opened 2022-03-01
A couple of corrections for the user interface

wllacer:
The call to the `save_config` entry missed an argument.
I've made the `preview_size` an argument at GeneralMenu, so it can be changed for each specific issue (and set the default to .2 so the GlobalMenu fits in my screen ;-)

Warning. As of now `only_hd.py` does not work. Will send a separate correction for it when ready, as i have to make it and `swiss.py` up to date


Torxed:
Quick work! Thank you! :)

PR #1019, opened 2022-03-02
Deflate user interactions

svartkanin:
An attempt to split up the `user_interactions.py` file into it's own module with separate files. I tried to categorize them together as much as possible, but nothing is set in stone so happy for any suggestions :)


wllacer:
@svartkanin 
Hi Daniel. I have a request for you. Could you include in the new directory the `GlobalMenu` object ? I am running in some problems ( circular dependencies re `user_interaction`) having it and `GeneralMenu` in the same source file, once the use of the later is more widespread. It's a bit silly that I create some new files somewhere else, and not using your proposal, and putting `GlobalMenu` in the new `user_interaction` directory is its natural place. 
I had also problems with  `select_archinstall_language` (you have in the `general_conf.py` file). I did move it inside the  now cleaned `selection_menu.py`  ( with Selector and GeneralMenu), where it is used. But this might have some other solutions (still have to test it with the full new structure)

svartkanin:
@wllacer yes I think it's a good idea to move the GeneralMenu into the directory.
Regarding language selection, i think it is similar to the existing user interaction APIs and should therefore probably live alongside those?

@Torxed any thoughts on this PR?

wllacer:
> @wllacer yes I think it's a good idea to move the GeneralMenu into the directory. Regarding language selection, i think it is similar to the existing user interaction APIs and should therefore probably live alongside those?
> 
> @Torxed any thoughts on this PR?

NO.  From the contents of `lib/menu/selection_menu.py` the part that should/need to go into `user_interaction` is `GlobalMenu` (the application main menu, the one you wrote). The GeneralMenu (the superclass) as an abstract object is better kept (with Selector) under the `menu` directory (which becomes in fact now a  `widgets` directory, but i don't think the name change is worth).
At least it is how i managed to avoid the problems i was experiencing

wllacer:
@svartkanin Daniel, I did it myself, but at a separate PR #1032, where it was totally necesary, using yours as a base.
The problem with select_archinstall_language has been "solved" copying it at `selection_menu.py` as an inline function. Not the smartest move, but allows to proceed further

svartkanin:
> NO. From the contents of lib/menu/selection_menu.py the part that should/need to go into user_interaction is GlobalMenu (the application main menu, the one you wrote). The GeneralMenu (the superclass) as an abstract object is better kept (with Selector) under the menu directory (which becomes in fact now a widgets directory, but i don't think the name change is worth).
At least it is how i managed to avoid the problems i was experiencing

Yeah sorry that's the one, was typing on the phone and too lazy to look up the correct name ;)

Torxed:
> @Torxed any thoughts on this PR?

I think that's a lot of (good) work! And I think this should go in.
Essentially all menu entries and stuff has to be properly tested.

There's also a lot of branches depending on this one, so I'm going to go ahead and merge it before people start touching these files too much :P

Issue #985, opened 2022-02-15
Add sudoers.d file instead of appending to sudoers


KenJyn76:
A cleaner way to add a user as a superuser would be to create a file in /etc/sudoers.d, and add the %user ALL=(ALL) ALL syntax in that file. I'd make a PR, but I'm terrible with git :)

Torxed:
i like this idea a lot. it would also mean that plugins and stuff can place for instance temporary build users there while building packages in plugins :D

eugeniopacceli:
An attempt was made :p

Torxed:
Functionality was added in #1025

PR #1025, opened 2022-03-07
Adding sudoers.d file instead of appending to sudoers

eugeniopacceli:
# Describe your PR

Addresses issue #985, creating one file in sudoers.d per user added as a sudoer user during the installation process.

This implementation:
* Makes sure sudoers.d exists and is referenced through `@includedir` in `/etc/sudoer`
* Creates one rule file per sudoer user, in the format: `00_user1`, `01_user2`, `02_user3`, and so on...
* Makes sure the created filenames do not contain [illegal characters](https://stackoverflow.com/a/35352640/2684616)
* Makes sure the recommended permissions are applied, as specified in [Arch Wiki - sudo](https://wiki.archlinux.org/title/sudo#Sudoers_default_file_permissions)

# Summed up contribution

https://github.com/archlinux/archinstall/pull/1025/files

# Testing

* Create multiple users that can sudo through the interactive CLI installer `python -m archinstall` modified with this contribution.
* Reboot (or chroot)
* As superuser, `cd /etc/sudoers.d` and then `ls -la`. You should also log in as one of the created sudoer users and attempt to sudo.

I tested it in `virtualbox` using a fresh environment and following the instructions at `README.MD#Testing`

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -956,6 +956,7 @@ class Installer:
 
 	def enable_sudo(self, entity: str, group :bool = False) -> bool:
 		self.log(f'Enabling sudo permissions for {entity}.', level=logging.INFO)
+		
 		sudoers_dir = f"{self.target}/etc/sudoers.d"
 		
 		# Creates directory if not exists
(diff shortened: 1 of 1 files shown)


Torxed:
Looks clean and thoughtful and useful comments.
Thank you for this PR! : )

PR #1028, opened 2022-03-09
Additions to Listmgr et alia

wllacer:
Several minor changes:
* The real marker that a btrfs subvolume is in place is the bracket notation for *bind mounts* not the at sign
* Ability for ListManager to have a list of default actions (action without a defined element) instead of a single element
* Ability for Menu and subclasses to define a header for the list, like in this screen
header
In case you wonder this is a real menu, not a mockup (Real Soon Now in your repositories ;-))

### DIFF
--- a/archinstall/lib/disk/btrfs.py
+++ b/archinstall/lib/disk/btrfs.py
@@ -25,7 +25,7 @@ class BtrfsSubvolume:
 	root :bool = False
 
 def get_subvolumes_from_findmnt(struct :Dict[str, Any], index=0) -> Iterator[BtrfsSubvolume]:
-	if '@' in struct['source']:
+	if '[' in struct['source']:
 		subvolume = re.findall(r'\[.*?\]', struct['source'])[0][1:-1]
 		struct['source'] = struct['source'].replace(f"[{subvolume}]", "")
 		yield BtrfsSubvolume(
--- a/archinstall/lib/menu/list_manager.py
+++ b/archinstall/lib/menu/list_manager.py
@@ -90,9 +90,10 @@ from .menu import Menu
 from ..general import RequirementError
 from os import system
 from copy import copy
+from typing import Union
 
 class ListManager:
-	def __init__(self,prompt :str, base_list :list ,base_actions :list = None,null_action :str = None, default_action :str = None):
+	def __init__(self,prompt :str, base_list :Union[list,dict] ,base_actions :list = None,null_action :str = None, default_action :Union[str,list] = None):
 		"""
 		param :prompt  Text which will appear at the header
 		type param: string | DeferredTranslation
@@ -108,7 +109,7 @@ class ListManager:
 
 		param: default_action action which will be presented at the bottom of the list. Shouldn't need a target. If not present, null_action is set there.
 		Both Null and Default actions can be defined outside the base_actions list, as long as they are launched in exec_action
-		type param: string
+		type param: string or list
 		"""
 
 		if not null_action and len(base_list) == 0:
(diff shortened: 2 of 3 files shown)


Torxed:
Im not sure we can fit that much info onto a 90 wide monitor, or whatever the default width of for instance virtualbox is? 😅

wllacer:
## programming notes:
To create a header just define a string or a list and pass it as a parameter `header=` to Menu or ListManager. Can be also done inside the  object using the variable `self.header` as in this sample
```python
class DevList(archinstall.ListManager):
	def __init__(self,prompt,list):
		self.ObjectActions = [
			'Add disk to installation set',          # 0
			'Add partition',                         # 1
			'Clear disk (delete disk contents)',     # 2
			'Clear Partition & edit attributes',     # 3
			'Edit partition attributes',             # 4
			'Exclude disk from installation set',    # 5
			'Exclude partition from installation set', # 6
			'Delete partition'                       # 7
		]
		self.ObjectNullAction = None
		self.ObjectDefaultAction = 'Reset'
		self.partitions_to_delete = {}
		super().__init__(prompt,list,self.ObjectActions,self.ObjectNullAction,self.ObjectDefaultAction)
		bar = '|'
		self.header =((f"  {'identifier':^16}"
					f"{bar}{'wipe':^5}"
					f"{bar}{'boot':^5}"
					f"{bar}{'encrypted':^7.7}"
					f"{bar}{'start':^12}"
					f"{bar}{'size (sectors def.)':^12.12} "
					f"{bar}{'filesystem':^12}"
					f"{bar}{'mount at':^19}"
					f"{bar}{'currently mounted':^19}"
					f"{bar}{'uuid':^24}"),
                    f"{'-' * 18}{bar}{'-'*5}{bar}{'-'*5}{bar}{'-'*7}{bar}{'-'*12}{bar}{'-'*13}{bar}{'-'*12}{bar}{'-'*19}{bar}{'-'*19}{bar}{'-'*24}")

``` 
OTOH I promised @svartkanin  to include the possibility to sort the elements on the list. I discovered that there is no special need to have a general method for that as it can be done in the `reformat` method, as in the sample below (with a dictionary). In this sample i manage to have "hidden" elements of the list.
```python
	def reformat(self):
		blank = ''
		bar = r'\|'      #  if you don't escape it, simple_menu will behave in unexpected form
		self.reorder_data()
                # we exclude some elements
		tmp_list = list(filter(lambda x:not self.data[x].get('hide',False),self.data))
		return list(map(lambda x:pretty_disk(x,self.data[x]) if self.data[x]['class'] == 'disk' else pretty_part(x,self.data[x]),tmp_list))
        # in this case, i have three diferent kinds of objects in the same list, so the ordering is a bit more complex than a simple sorted. 
t       # the trick is to fill self.data
	def reorder_data(self):
		new_struct = {}
		tmp_disks = [disk for disk in sorted(list(self.data.keys())) if self.data[disk].get('class') == 'disk']
		for disk in tmp_disks:
			new_struct.update({disk:self.data[disk]})
			tmp_parts = [value for part,value in sorted(self.data.items()) if value.get('parent') == disk and value['class'] == 'partition']
			new_parts = create_gaps(tmp_parts,disk,GLOBAL_BLOCK_MAP[disk]['size'])
			new_struct[disk]['partitions'] = new_parts
		self.data = from_general_dict_to_display(new_struct)


```

wllacer:
> Im not sure we can fit that much info onto a 90 wide monitor, or whatever the default width of for instance virtualbox is? sweat_smile

Yupss ... But let this be a problem when the rest of it is ready ... I wish all problems were like this ;-)

Torxed:
Looks good, just a small note that we should probably keep even test code and comments in English even if it's just for graphical testing: https://github.com/archlinux/archinstall/blob/34ab21df1452ded447a6ff12eae874b8bac3b4c1/archinstall/lib/menu/list_manager.py#L271-L278

Merging! : )

PR #1032, opened 2022-03-13
Subvolume User Interface

wllacer:
At long last I have an interface for managing and defining subvolumes. It is an entry in current partition management menu
subvolume_1

Which invokes a list of the subvolumes and allows to manage them (of course only btrfs partitions)
subvolume_2

PR #1019 is a prerequiste for this. Otherwise a number of circular references between the list used at user_interaction, and the fact that GlobalMenu (which uses them) and GeneralMenu were on the same file. Instead of creating some isolated files I accepted @svartkanin solution and worked from there .

As bonus points, it streamlined some handling at ListManager (for instance, now allowing empty lists without a direct action), and allowed a filtering with free conditions at `select_partition`. I also updated `only_hd.py` and `swiss.py` which were a bit out of step


wllacer:
@torxed  
A **warning**, before this PR  (and #1019) is upstreamed every change since their creation at `user_interaction.py`  (for 1019) and `selection_menu` (for this one) has to be manually replicated. git is not smart enough when you split files :-(.
I will grab a copy of both  files at today's master, to have something to compare.

Torxed:
I think I managed to get everything in correctly, but it was a hell of a lot of changes.. So I wouldn't be surprised if something went haywire. But hopefully we're in a good state!

timakro:
@wllacer How about adding `noatime` to the menu? It's an option I always use.

wllacer:
> @wllacer How about adding `noatime` to the menu? It's an option I always use.

@timakro I do it also, so it is something of an oversight of my part.
When i developed this feature I found that many mount options for btrfs are whole file system related and not subvolume related (f.i. a more comprehensive `compress` option). Assigning them is a feature still not merged at the UI level ( can be done by hand with an array entry at the disk_layouts file   `filesystem.mount_options`). When I merged the subvolume handling I only put the two options which have an special behaviour, and totally forgot about `noatime`. 
There is still a *which mount options are fs related and which subvolume related on btrfs* entry in my TO_DO list. 
Sadly, i will be off line for some weeks. Feel free, in the meantime to add whichever option you think worth at 
https://github.com/archlinux/archinstall/blob/0604ed45c3e270ef39b6c80af3787d4ee025c5a1/archinstall/lib/user_interaction/subvolume_config.py#L126-L129

Issue #1035, opened 2022-03-15
Issues at menu with Network Handling


wllacer:
1) Configuration files of older versions (i.e. 2.3.2, at least) fail on being loaded at `guided`. It's either shows empty values or fails if it is an ISO definition 
2) Create a network configuration, return to menu and then return to edit the configuration. Process fails with 
```
 File "/home/werner/projects/archinstall/archinstall/lib/menu/selection_menu.py", line 341, in exec_option
    result = selector.func(presel_val)
  File "/home/werner/projects/archinstall/archinstall/lib/user_interaction.py", line 458, in ask_to_configure_network
    if preset['type'] == 'iso_config':
TypeError: 'NetworkConfiguration' object is not subscriptable
```

Torxed:
Closed with #1036

PR #1036, opened 2022-03-15
Issues with Network Management and user management in menu

wllacer:
Solves issue #1035
-> Compatibility with current stable version (v2.3) of the config file.
-> Reediting the network interface entry crashed
And solves an issue with default values (lack of) in the (super)user management lists at the menu, not prevously documented.

@svartkanin 
The solution is coded at the `NetworkConfiguration` level. 
For the first issue, I put  compatibility code at `parse_arguments`. Only remaining hook is that it is executed twice before entering the Menu. still have to pin where.
For the second issue i gave up trying to change `ask_to_configure_network`, where the problem lies (in many lines it still expect a dictionary and not a class), and went to add `__getitem__`  and `get` methods at NetworkConfiguration, to achieve a compatible interface. I went for the shortest line of resistence, i know.

**Updated**
One correction to the message error at `Installer.__exit__` (issue #1039) and

Another which performs an ordered exit when after not selecting a drive we decide not to continue (previous implementation, looped there, a good idea on paper -giving the user the possibility of selecting drives- but gave no possibility to escape. The new implementation is a bit cumbersone (return to the start of the process) but more flexible

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -149,7 +149,7 @@ class Installer:
 
 			# We avoid printing /mnt/<log path> because that might confuse people if they note it down
 			# and then reboot, and a identical log file will be found in the ISO medium anyway.
-			print(_("[!] A log file has been created here: {} {}").format(os.path.join(storage['LOG_PATH'], storage['LOG_FILE'])))
+			print(_("[!] A log file has been created here: {}").format(os.path.join(storage['LOG_PATH'], storage['LOG_FILE'])))
 			print(_("    Please submit this issue (and file) to https://github.com/archlinux/archinstall/issues"))
 			raise args[1]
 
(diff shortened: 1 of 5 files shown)


wllacer:
@torxed You're the boss, but i think this PR should be part of the RC1, because it is basically bug squashing. And the first two made the menu unusable at master

Torxed:
I can re-tag the release candidate, as it's not as set in stone as the prod releases should be. Merged and I'll retag in a second, just gotta go through the other PR's and see if something else belongs in there :)

wllacer:
@torxed.  Remember to include this at the RC1. I've done some light testing and this are the only crashes I could trigger at the Rc1 (easy, i know where to look at)

Torxed:
Tagging up a new release candidate today :)

Torxed:
@wllacer Tagged and ready with your latest changes + the split of user_interaction.
Great work on all of this, I'm really looking forward to v2.4.0! I think I will instruct giancarlo to keep it in `[testing]` for a while tho once we do release it. And we can use the power of the community to run via testing as well.

svartkanin:
@Torxed I have the multi network configuration almost ready do you think we can squeeze that in as well?

Torxed:
@svartkanin: Sure thing!

PR #1044, opened 2022-03-22
Clarify prompt text when providing superusers

psbleep:
# Describe your PR

When running `archinstall` I was confused by the prompt text when being asked to set up `sudo` users. I thought I was being asked if I wanted to define `sudo`-privileged users, and responded as if it was a `Y/n` prompt. The installer dutifully created a `sudo`-privileged user called `y`.

To me, clarifying that the prompt is asking for usernames would make it more obvious. There could be a better way of phrasing this, but I came up with:

```Define users with sudo privilege, by username: ```


Torxed:
Thank you for clarifying the prompt up, it's easy to miss these things as a developer when you run through the menus a few hundred times per week hehe.

PR #1049, opened 2022-03-29
Correct exit when no disk is selected and we don't wish to continue

wllacer:
Corresponds to commit 03f339591ab649cfd7a83bbd77aa5b6817676e5d which had been left out due to the recent file split.
If you don't select a harddrive and said no to use it, the previous implementation was to return to the disk selection. With the patch you exit the installation process. A bit more cumbersone if you made a mistake, but if you wanted to really exit, there is no alternative

### DIFF
--- a/archinstall/lib/menu/global_menu.py
+++ b/archinstall/lib/menu/global_menu.py
@@ -99,6 +99,7 @@ class GlobalMenu(GeneralMenu):
 			Selector(
 				_('Specify superuser account'),
 				lambda preset: self._create_superuser_account(),
+				default={},
 				exec_func=lambda n,v:self._users_resynch(),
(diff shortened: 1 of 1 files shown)


Torxed:
Ah, my bad. Did a manual guess and thought we didn't wanna hard exit but that there instead was some return magic.

Torxed:
Problem for the future hehe, works great for now :) So happy with how v2.4.0 has turned out.
I'll have to re-name it to `v2.4.1` because of slipup with the buildsystem and versioning.

twine read the version from `pyproject.toml` and not `__init__.py` so `v2.4.0` has already been published incorrectly to pypi.org.
I yanked the version, and before releasing `v2.4.0` I'll rename it to `v2.4.1` so it's back to being consistent.

Issue #1020, opened 2022-03-02
Multiple network configuration


svartkanin:
As discussed in https://github.com/archlinux/archinstall/pull/1001 we decided to upgrade the current network config from a single to a multiple configuration option for the interfaces. With the work from @wllacer 's `ListWidget` that should be quite nice! 

I'll get it kicked off asap!

wllacer:
Daniel. 
I have following changes to ListManager planned:
* Ability to define a header row or equivalent
* Ability to have multiple Default Actions (the ones below the list, to be executed for the list and not for one particular item
* Ability to define specific sort routines to the list 
None seems specially dificult, but they were low in my to-do list. If you think you could need them for the network management, i will be glad to implement them for you as soon as i am able

PR #1052, opened 2022-03-30
Support for multiple network interfaces

svartkanin:
This addresses #1020. 

This will support the configuration of multiple network interfaces.
image

There are some breaking changes to the current version in master. For the configuration file, the interfaces can now be defined as

```
{
...
"nic": {
    "type": "manual", 
    "config": [
        {<interface configurations},
        {<interface configurations}
    ]
}
...
}
```

### DIFF
--- a/.github/workflows/mypy.yaml
+++ b/.github/workflows/mypy.yaml
@@ -16,4 +16,4 @@ jobs:
 #            - name: run mypy
 #              run: mypy --strict --module archinstall || exit 0
             - name: run mypy
-              run: mypy --follow-imports=skip archinstall/lib/menu/selection_menu.py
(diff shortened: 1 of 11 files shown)


Torxed:
I like this one, I'll merge it later today.

wllacer:
@svartkanin . Yep. You're right, it ought to be self.data at the ListManager ... but it will break some code at PR #1053. Not that important (just a find/replace). Warn me when this is upstreamed, to make the changes

svartkanin:
@wllacer yes I did some refactoring there i hope that's okay. I got a bit confused which attributes are private and which ones public. I also updated some callbacks to actually pass the data into it instead of accessing it via the attributes.
 The reason is that for me as a newvie using this new class it's very difficult to determine which variable stores what and I had to go through the code to know what I should access. However, if you override a function that gets everything relevant as an argument I can simply look at those and decide what to use :)

svartkanin:
@Torxed Can we merge this if you're happy with the changes :)

Torxed:
I'm always happy with the changes :D Merged!

PR #1067, opened 2022-04-05
Autofilling harddrives

wllacer:
When using a `disk_layouts` argument, but no additional info, current implementation silently forgets the layout, unless we use an specific `--harddrives` argument. Good enough for me, but most user would find it difficult to use
With this change, we simply autofill the `harddrives` arguments from the contents of `disk_layouts`.
As an example if I call
`sudo python -m archinstall --disk_layout datos/btrfs_two_crypt.json`
without the change we get
before
and with  the change
after


Torxed:
I know you're on a well deserved rest.
But just wanted to let you know that I'm really grateful for this change! Works beautifully and even tho it's a really small change, it has huge implications! :) Thank you!

PR #1099, opened 2022-04-29
Fix #1096

### LINKED ISSUE #1096: Installer.zram_enabled is not set when swap is disabled
When trying to install with `"swap": False`, the installer fails prematurely when trying to install systemd-boot as it tries to access `self.zram_enabled`. This attribute is only set in `setup_swap()`, which does not get executed with swap disabled resulting in an exception.

logs: https://paste.dymstro.nl/?9fd3ee57a10a7725#5QguRT8KXp6jGQombKw7d2aDQmnjksrGC1TxsfFUFEkq

svartkanin:
Small fix for #1096


Torxed:
Thank you, quick work! :)

PR #1119, opened 2022-05-02
Fix #1106

### LINKED ISSUE #1106: Now way out of the desktop profile
Would be good to be able to see what desktops are available but choose none or have a chance to get back without choosing a desktop.

svartkanin:
- Fixes #1106
- Makes all profile selections cancellable with ESC
- Fixes some other bugs that were encountered during testing in the profile selections


Torxed:
You're quick! Thank you for this one!

PR #1120, opened 2022-05-02
Add network-manager-applet to NetworkManager profile

kpcyrd:
This adds the network-manager-applet package to the NetworkManager profile. :)


Torxed:
I would prefer if the package was not installed unless a desktop profile was installed.

dylanmtaylor:
> I would prefer if the package was not installed unless a desktop profile was installed.

Same, and I don't think all desktop environments using networkmanager would use this package.

kpcyrd:
I think we'd need some kind of solution for the i3 profile, because it's currently not possible to connect to a wifi through the gui if this package isn't present. :)

Torxed:
Just a friendly reminder:
> Your branch is ahead of 'origin/master' by 1662 commits.

I pulled in our master into your branch and it was a bit behind.
You can solve this by doing:
```
$ cd archinstall
$ git checkout master
$ git remote add upstream https://github.com/archlinux/archinstall.git
$ git pull upstream master
```
And then you can continue your normal workflow of doing `git checkout -b new-branch` and it should contain our latest changes.
Or just do `git pull upstream master` if you forget before creating the branch :)

I'm taking a look at this PR now.

Torxed:
Thank you for submitting this PR.
I tweaked it a tiny bit to look for the chosen profile through `storage["arguments"]`, as the configuration has been stripped down to only the related network configuration in the nic config model :)

PR #1122, opened 2022-05-03
Update Spanish translation

castillofrancodamian:
Signed-off-by: Franco Castillo <castillofrancodamian@gmail.com>


svartkanin:
Thanks @castillofrancodamian for the contribution! Would you mind also generating the `.mo` file otherwise the translation won't be applied

PR #1129, opened 2022-05-06
Added russian translation

Alexandr1322:


svartkanin:
It seems like we already had one up here #1014 :O

Torxed:
I'll close this one, as the original PR came first.
The reason behind it was we were releasing a new version and merging in language changes has a tendency to break even simple execution of the script. Now that we're in a phase of developing new features - it has been merged in :)

PR #1132, opened 2022-05-06
Fix minor issues

svartkanin:
- If not already selected, will not select an entry in a multi select anymore with enter  except when there's no other selection at all yet
- Skip over placeholders when automatically moving the cursor forward when exiting a submenu


Torxed:
https://i1.sndcdn.com/artworks-000203206908-h2j752-t500x500.jpg

PR #1140, opened 2022-05-08
Update base.po

roxfr:
Hello,
Here is the update for the French translation.
Regards,
Roxfr

### DIFF
--- a/archinstall/locales/fr/LC_MESSAGES/base.po
+++ b/archinstall/locales/fr/LC_MESSAGES/base.po
@@ -9,109 +9,179 @@ msgstr ""
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"X-Generator: Poedit 3.0\n"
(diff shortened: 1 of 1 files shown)


Torxed:
@svartkanin Would it make sense to move all strings into a "language source file", so we don't have to update the base.po files every time we move source code around? Or would that just complicate things a bit too much?

svartkanin:
@Torxed as I understand the gettext module, it will use the specified string in the text as a msg_id. So if that gets moved around in the code, then it shouldn't create a new entry in the .pot file for translation (I think), but I believe it will update that comment on top with the new location reference. I saw there's an option in the msgmerge command at least to turn off these comments.

However, if the string itself gets updated (fixing a typo for example) then there will inevitably be a new entry created because it is a new text essentially.

I'm not sure if a dedicated source file would help here, except for the generated comments, but we can try to turn those off.

Happy for alternative suggestions though :)

Torxed:
@svartkanin That's all i needed to hear to drop the idea hehe. Wouldnt be worth the very little if any gain it would give ^^ Good solid answer!

PR #1141, opened 2022-05-08
Update base.mo

roxfr:
Hello,
I would like to delete the .mo file in order to replace it with the new one.
Thank you.
Regards,
Roxfr


Torxed:
I don't think we need to delete it in order to update it.
You can simply push the new binary and it will replace it.

roxfr:
> I don't think we need to delete it in order to update it. You can simply push the new binary and it will replace it.

Hi,
I think I have successfully modified this query.
Thank you.
Roxfr

Torxed:
Good job and thank you!

PR #1145, opened 2022-05-08
Added portuguese language and generated files

CatKinKitKat:
Added portuguese language to locales and generated the files with the included .sh script

No problems detect on an VM using archinstall


svartkanin:
@CatKinKitKat thanks for the contribution!
There was another PR that got merged with some changes regarding the format of the base.pot file amd the subsequent.po files. Essentially removing the comments:)

Would you mind rerunning it on your end, I can also raise a PR against your branch if you like

PR #1152, opened 2022-05-09
Introduce ctrl+c and other bug fixes

### LINKED ISSUE #1130: Make Ctrl-C clear the menu option - and Esc to only back out of menu's
As pointed out in both [video format](https://www.youtube.com/watch?v=OtWWLN3wGNE) and in the [comments of the video](https://www.youtube.com/watch?v=OtWWLN3wGNE&lc=UgwWywhBjWoGfIvjard4AaABAg):

> ESC should just bring you back to previous screen. To reset, Ctrl+C

I think that makes sense so I'll post it here so we can track it.

svartkanin:
@Torxed this is big :) 

- Fix #1130:
    - Introduce the ctrl+c as a reset option for certain sub-menus (mainly those with multi-select)
    - I updated all occurrences of Menu(...) and I hope I didn't miss any
- Fix bug when exporting configuration files via the `Save configuration -> All` then the disk layout was not exported should now be fixed
- Fix profile selection
    - The profile selection was buggy (and may still be) but essentially when selecting profile server for example, exiting and then changing the selection to desktop, the previous server selection was not reset 
    - I tried to monkey patch it as good as I could and I tested the flows and :crossed_fingers: I didn't miss anything but this should be rewritten to something more maintainable
- Changed the audio selection of `none` to string `No audio server`
- I enabled the locale selections by default in the guided.py (let me know if it's more desirable to leave them hidden under --advanced)


Torxed:
Legendary! :D
I think you managed to squash almost everything mentioned in the reviews! Huge thanks!

svartkanin:
Most things I found annoying myself so it was a relief ;)

PR #1153, opened 2022-05-09
Set the unicode cyrillic font by default

### LINKED ISSUE #1138: Cyrillic problems (?)
I decided to test new translation and got this 
Снимок экрана от 2022-05-06 17-47-09

svartkanin:
Fix #1138 
This should not have any impact any other displayed text but will enable Cyrillic characters by default.

### DIFF
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -55,6 +55,10 @@ storage['__version__'] = __version__
 DeferredTranslation.install()
 
 
+def set_unicode_font():
+	SysCommand('setfont UniCyr_8x16')
+
+
 def define_arguments():
 	"""
 	Define which explicit arguments do we allow.
@@ -242,6 +246,9 @@ def post_process_arguments(arguments):
 
 	load_config()
 
+
+# to ensure that cyrillic characters work in the installer
+set_unicode_font()
 
 define_arguments()
 arguments = get_arguments()


Torxed:
Do we need to revert it if they change back to English?
Or are we fine with this being a one way transaction for now? :)

svartkanin:
@Torxed As I understand the font, it's universal, so it will include English as well as Cyrillic character sets. I don't see any downside in having this set generally
I tested it on the ISO and locally and there's no issues I could identify with English having it enabled :)

Torxed:
That's good news, merged! Good job as always!

Torxed:
c94c8cee-a623-4603-84c1-6b52182b7d48


This is running latest `releng` of archiso.
I think we're missing the font in the base packages?

If there's no way around it, I can ask if we can include a solution

svartkanin:
@Torxed apologies, I did test this on the iso in a vm and that worked fine :(

svartkanin:
@Torxed I gave this another shot and it does work for me, so I'm very confused now. 
- Downloaded the latest ISO
- Created a new VM with the ISO
- Cloned master
- Uncomment settting cyrillic unicode font
- Run with `python -m archinstall` 
- No error 

Am I missing here something?

svartkanin:
This one here seems to include it https://archlinux.org/packages/core/x86_64/kbd/files/
and it's present in the ISO 
```
$ pacman -Q kbd
kbd 2.4.0-2
```

Torxed:
I'll give it a try as well again, this was reported by another user. hmm..

PR #1171, opened 2022-05-12
Update validators.py

CYPT71:
Rewrites the nested conditions

🚨 PR Guidelines:

# New features *(v2.2.0)*

# Describe your PR

# Testing


Torxed:
I'm going to close this as well.
As it makes the code a bit less readable.

I get that they both do exactly the same thing, but nested `and` statements makes it hard to digest what is triggering what.

This function however should be re-written to be more optimized, perhaps a dataclass/model instead of a bunch of if/else :)

PR #1174, opened 2022-05-12
Add how to pass declarative configuration files as arguments to archinstall

i2:
🚨 PR Guidelines:

# Describe your PR

Just an update to README.md to firmly mention how --config, --disk-layout and --creds work. There are misleading tutorials and videos online that just use --config for all three types of config files and mentioning in the readme file will help minimize further confusions and errors.


Torxed:
That's a good idea. Even if technically using --config for all three works in some cases as they are merged internally :)

PR #1176, opened 2022-05-13
Adds nvidia-open package as an option.

dylanmtaylor:
Apologies for making the installation even more confusing for Nvidia users. This adds an option for the new open-source kernel module for Turing or later cards, https://github.com/NVIDIA/open-gpu-kernel-modules. nvidia-open is currently in the testing repository. Probably needs a discussion before merge as this is going to be SUPER confusing for Nvidia users with the latest developments. https://archlinux.org/packages/testing/x86_64/nvidia-open/

### DIFF
--- a/archinstall/lib/hardware.py
+++ b/archinstall/lib/hardware.py
@@ -51,6 +51,7 @@ AVAILABLE_GFX_DRIVERS = {
 		"intel-media-driver",
 		"vulkan-intel",
 	],
+	"Nvidia (new open-source kernel module)": ["nvidia-open"],
 	"Nvidia (open-source)": [
 		"mesa",
 		"xf86-video-nouveau",
--- a/archinstall/lib/hardware.py
+++ b/archinstall/lib/hardware.py
@@ -51,7 +51,7 @@ AVAILABLE_GFX_DRIVERS = {
 		"intel-media-driver",
 		"vulkan-intel",
 	],
-	"Nvidia (new open-source kernel module)": ["nvidia-open"],
+	"Nvidia (open kernel module for newer GPUs, Turing+)": ["nvidia-open"],
 	"Nvidia (open-source)": [
 		"mesa",
 		"xf86-video-nouveau",
--- a/archinstall/lib/hardware.py
+++ b/archinstall/lib/hardware.py
@@ -52,7 +52,7 @@ AVAILABLE_GFX_DRIVERS = {
 		"vulkan-intel",
 	],
 	"Nvidia (open kernel module for newer GPUs, Turing+)": ["nvidia-open"],
-	"Nvidia (open-source)": [
+	"Nvidia (open-source nouveau driver)": [
 		"mesa",
 		"xf86-video-nouveau",
 		"libva-mesa-driver"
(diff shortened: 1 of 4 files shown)


dylanmtaylor:
https://discord.com/channels/726808923662712894/726810321125376092/974489274697269259

Torxed:
@svartkanin I've been trying to set a value in the menu system if `nvidia-open` is selected. I simply want the `additional repositories` to show `["testing"]`, what am I doing wrong? I'm calling:
```python
storage['global_menu']._menu_options['additional-repositories'].set_current_selection(arguments.get('additional-repositories', []) + ['testing'])
```

svartkanin:
@Torxed If I may suggest to put this actually into the `gloabl_menu.py` instead https://github.com/archlinux/archinstall/blob/master/archinstall/lib/menu/global_menu.py#L327

something like

```
if archinstall.arguments('gfx_driver', None) == 'nvidia-open':
	if (package_info := find_package('nvidia-open')) and package_info[0].repo == 'testing':
		if 'testing' not in archinstall.arguments.get('additional-repositories', []):
			log(f"Enabling repository 'testing' due to nvidia-open being selected and it lives there", fg="orange", level=logging.WARNING)
			current = self._menu_options['additional-repositories'].get_selection()
			self._menu_options['additional-repositories'].set_current_selection(list(set(current + ['testing'])))
```
In case someone goes into the profile selection again and removes the driver or changes it, should this repo be removed again? That might be difficult here, since the user might have enabled it explicitly before having selected the driver. What's the downside of having this then always enabled?

Torxed:
Thank you! I'll move the code around.
The downside is that the user might get very unstable packages, that we really have to be clear on and warn about.

I was thinking of only showing the `nvidia-open` driver IF they manually select `testing` before selecting a profile.
But that would make it too anonymous I think and people would never find it. And sure that's just for nvidia this time, but I'm thinking this might happen in the future as well.

So I'll think about that one, it's a good question!

svartkanin:
We could not add the `testing` repository into the settings automatically, but in the installation flow when the point of enabling repositories is reached we could check explicitly for the gfx_driver selected there and enable the testing one if it wasn't chosen by the user

Torxed:
That might be one approach. I'm thinking enabling/disabling a repository for a specific step can get consequences tho.

svartkanin:
Good point!

dylanmtaylor:
> Thank you! I'll move the code around. The downside is that the user might get very unstable packages, that we really have to be clear on and warn about.
> 
> I was thinking of only showing the `nvidia-open` driver IF they manually select `testing` before selecting a profile. But that would make it too anonymous I think and people would never find it. And sure that's just for nvidia this time, but I'm thinking this might happen in the future as well.
> 
> So I'll think about that one, it's a good question!

I really feel like we shouldn't merge this until the package is in stable.

Torxed:
I'll revert my changes and agree that we'll merge this once it goes into stable :) Less code to maintain and cleaner overall.
Thank you both for the input!

dylanmtaylor:
> I'll revert my changes and agree that we'll merge this once it goes into stable :) Less code to maintain and cleaner overall. Thank you both for the input!

Hey @Torxed - looks like it just hit extra an hour ago. I hope you don't mind but I'm going to revert out the other changes on this PR and we can get this in cleanly :)

dylanmtaylor:
I reverted each commit individually so we'd preserve the history here, didn't want to do a reset to my commit and force push the branch :)

Torxed:
Great work, merged! That was quick getting it in the repos hehe

PR #1179, opened 2022-05-13
Internationalized portuguese and made it less formal

CatKinKitKat:
Internationalized portuguese and made it less formal

It works on my virtual machine


Alexmelman88:
You are changing files of other languages, this is wrong, pull request should only change your language.

CatKinKitKat:
Noticed now, I had some .bashrc aliases swapped and might have added the whole directory to the commit,  as for the other files, didn't touch them. Either vscodium did or your .sh script does more than it says.

Torxed:
I think someone added `_(...)` to new strings.
And the bash script found those new translations and added them to all the language files.
This is why it looks like this PR modifies the other languages, but in fact it only fixed them.

I'm ok merging this :)

Issue #114, opened 2021-03-12
PEP 561


advaithm:
currently archinstall is not PEP 561 complaint, if arch install were to become it would allow the end user to make their programs more "static", this would involve creating another package for types. This would allow users to use tools like `mypy` and `pytypes` on programs they create. 

useful info: https://www.python.org/dev/peps/pep-0561/

advaithm:
the easiest way I have found to make this happen is stubgen, which would generate the files required to be packaged by archinstall for a types package. however this only works if functions and class are type annotated. once I'm done with bios support I'll try and work on this

advaithm:
it should be possible to do this inline in they code using type anotations
for example
```python
def foo(bar:str) -> str:
    return bar
```

Torxed:
Exactly :D I like that approach hehe.
I've started on a few places:
 * https://github.com/Torxed/archinstall/blob/aceb0f3e98ea1009474f24cf02fdcf3e5923bcc7/archinstall/lib/output.py#L63

advaithm:
if you have mypy installed it'll throw a fit about return values which is pretty useful

Torxed:
I hope I'm doing IDE's justice now by doing:
 * https://github.com/Torxed/archinstall/blob/99dd9b1fb7856b864d816119f0742950941e97a8/archinstall/lib/profiles.py#L182

advaithm:
I think its `typing.Optional`

f-koehler:
I am really in favor of this. I use type annotations in most of my python code these days and find them really helpful.

Also running `mypy` via [`pre-commit`](https://pre-commit.com/)/GitHub actions is really useful in my experience. This can catch a lot of bugs and helps the development tools a lot. Also type hints can be added to the documentation automatically

advaithm:
@Torxed this shouldn't be closed

Torxed:
Yepp, this just gets automatically closed if a PR contains `fixes` and then a reference to the issue.
Thank github for this :P

Torxed:
Just as a reference, PR #846 addresses this

PR #1186, opened 2022-05-15
Fix mypy complaints in models.dataclasses.py

demipaki:
🚨 PR Guidelines:

# New features *(v2.2.0)*

# Describe your PR

I fixed all of the complaints from mypy in models.dataclasses.py. This was in attempt to help the library be more compliant with PEP 561 as discussed in issue #114.

You'll notice that in the comparison dunder functions, I replaced the type from `VersionDef` to `object`. This is because these methods can expect any type of object, which is why I raise `NotImplemeted` if it gets compared with any other type than itself. Inspired by this: https://stackoverflow.com/questions/54801832/mypy-eq-incompatible-with-supertype-object

# Testing

In another commit I wrote a few unit tests for the VersionDef class. That being said I noticed that there isn't a preexisting test suite yet (that's in another branch, presumably coming soon). So feel free to disregard that commit if you want.

### DIFF
--- a/archinstall/lib/models/dataclasses.py
+++ b/archinstall/lib/models/dataclasses.py
@@ -110,13 +110,13 @@ class PackageSearchResult:
 		if not isinstance(other, PackageSearchResult):
 			return NotImplemented
 
-		return VersionDef(self.pkgver) == VersionDef(other.pkgver)
+		return self.pkg_version == other.pkg_version
 
 	def __lt__(self, other :object) -> bool:
 		if not isinstance(other, PackageSearchResult):
 			return NotImplemented
 
-		return VersionDef(self.pkgver) < VersionDef(other.pkgver)
+		return self.pkg_version < other.pkg_version
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


archinstall/lib/models/dataclasses.py
```diff
@@ -14,38 +15,60 @@ def parse_version(self) -> List[str]:
...
 		return self.parse_version()[0]
 
-	@classmethod
 	def minor(self) -> str:
 		versions = self.parse_version()
 		if len(versions) >= 2:
 			return versions[1]
 
-	@classmethod
+		return ""
```
Torxed, line 26: I've only glanced over the code, and I honestly can't remember my own logic for this one. But I suspect we'd want to return `"0"` where version definitions are not found, as we'd want to convert this to something comparable?
I guess `""` is comparable as well, but something inside feels better comparing against numbers.

Torxed:
I'm not sure why, but this PR breaks `packages.py`.. They shouldn't have any correlation but yet it breaks?
screenshot

svartkanin:
@demipaki If this file is fully mypy compliant now, could you add it to the github action hook as well https://github.com/svartkanin/archinstall/blob/66765e1735e950a6d40c6d44c5615b065dc1d309/.github/workflows/mypy.yaml#L18-L18? 
This way it will remain clean since nobody can make modifications anymore that are not compliant :)

archinstall/lib/models/dataclasses.py
demipaki: Good point, I've changed the code to reflect your critique

demipaki:
I've just pushed commits to this branch by accident without doing the test, that broke packages.py, that @Torxed mentioned...

I'll request another review once I've done the test myself. Sorry, this is my first time contributing to a project on github :sweat_smile:

Torxed:
No worries hehe, and critique sounds so harsh :P More a gentle nudge ^^
I appreciate the PR!

PR #1187, opened 2022-05-15
This allows us to set LANG= in vconsole for keymap negotiation during crypt-unlock

### LINKED ISSUE #596: LUKS password fails with different locale
When performing the installation with a different locale than English, the LUKS password doesn't work. I assume once the installation finished and you boot the system it's again using the English locale to decrypt the LUKS partition?

Torxed:
This fixes #596 by explicitly setting `LANG=` in `/etc/vconsole.conf`.
Not sure why this is needed, but is apparently a known "weird" glitch.

And just to be clear, it's not always required. But it's hard to tell when it is.
And adding this shouldn't cause any other issues. The drawback is that we generate and run `mkinitcpio` twice.

I'm contemplating evaluating [sd-encrypt](https://wiki.archlinux.org/title/Dm-crypt/System_configuration#Using_sd-encrypt_hook) hook instead. But saving that for later.

### DIFF
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -32,7 +32,11 @@ from .lib.packages.packages import (
 from .lib.profiles import *
 from .lib.services import *
 from .lib.storage import *
-from .lib.systemd import *
+from .lib.systemd import (
+	localectl_status,
+	Boot,
+	Ini
+)
 from .lib.user_interaction import *
 from .lib.menu import Menu
 from .lib.menu.list_manager import ListManager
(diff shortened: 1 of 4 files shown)


Torxed:
I'm holding off on this one until I figure out why it doesn't work.
It implements the manual steps taken to get the issue fixed - but for some reason - it doesn't fix the issue (?).

h8d13:
@Torxed at least for grub LUKS password needs ckbcomp [see here.](https://github.com/h8d13/SYMAN-GRUB2/blob/master/grub_keymaps) or used to need now?

svartkanin:
I think this has been addressed with the linked pr, so I'll close it for now

PR #1191, opened 2022-05-16
Fixed base.pot

Alexmelman88:


Torxed:
Thank you! Some quirky message strings hehe

PR #1195, opened 2022-05-16
Fix lengthy translations in Main Menu (Important)

i2:
The motivation behind this cumbersome PR were:
1. In a YouTube video,  that archinstall was being reviewed, the person was complaining that menu items are very long. This is actually very true even more so for romantic languages. Entries go to next line which makes it not great to look at, especially on smaller screens!
2. Locating setting easier. There are a lot of entries that start with 'Select' or 'Specify'. These words are redundant. If I am looking, say for mirrors or network config, I would like the entry to start with Mirror and Netwok.

### DIFF
--- a/archinstall/lib/menu/global_menu.py
+++ b/archinstall/lib/menu/global_menu.py
@@ -146,7 +146,7 @@ class GlobalMenu(GeneralMenu):
 				default=[])
 		self._menu_options['additional-repositories'] = \
 			Selector(
-				_('Additional repositories'),
+				_('Optional repositories'),
 				select_additional_repositories,
 				default=[])
 		self._menu_options['nic'] = \
--- a/archinstall/lib/menu/global_menu.py
+++ b/archinstall/lib/menu/global_menu.py
@@ -42,33 +42,33 @@ class GlobalMenu(GeneralMenu):
 		# archinstall.Language will not use preset values
 		self._menu_options['archinstall-language'] = \
 			Selector(
-				_('Select Archinstall language'),
+				_('Archinstall language'),
 				lambda x: self._select_archinstall_language(x),
 				default='English')
 		self._menu_options['keyboard-layout'] = \
 			Selector(
-				_('Select keyboard layout'),
+				_('Keyboard layout'),
 				lambda preset: select_language(preset),
 				default='us')
 		self._menu_options['mirror-region'] = \
 			Selector(
-				_('Select mirror region'),
+				_('Mirror region'),
 				lambda preset: select_mirror_regions(preset),
 				display_func=lambda x: list(x.keys()) if x else '[]',
 				default={})
 		self._menu_options['sys-language'] = \
 			Selector(
-				_('Select locale language'),
+				_('Locale language'),
 				lambda preset: select_locale_lang(preset),
 				default='en_US')
 		self._menu_options['sys-encoding'] = \
 			Selector(
-				_('Select locale encoding'),
+				_('Locale encoding'),
 				lambda preset: select_locale_enc(preset),
 				default='UTF-8')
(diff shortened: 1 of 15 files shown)


archinstall/lib/menu/global_menu.py
```diff
@@ -87,82 +87,82 @@ def _setup_selection_menu_options(self):
...
 			Selector(
-				_('Additional packages to install'),
+				_('Additional packages'),
 				# lambda x: ask_additional_packages_to_install(storage['arguments'].get('packages', None)),
 				ask_additional_packages_to_install,
 				default=[])
 		self._menu_options['additional-repositories'] = \
 			Selector(
-				_('Additional repositories to enable'),
+				_('Additional repositories'),
```
Torxed, line 149: I would probably cange this to "Optional repositories" instead, while we're at it.

archinstall/lib/menu/global_menu.py
```diff
@@ -42,33 +42,33 @@ def _setup_selection_menu_options(self):
...
 		self._menu_options['sys-encoding'] = \
 			Selector(
-				_('Select locale encoding'),
+				_('Locale encoding'),
 				lambda preset: select_locale_enc(preset),
 				default='UTF-8')
 		self._menu_options['harddrives'] = \
 			Selector(
-				_('Select harddrives'),
+				_('Drives'),
```
Torxed, line 71: "Harddrives" maybe? it's what i would look for at least.

Torxed (changes requested):

archinstall/lib/menu/global_menu.py
i2: I thought about "Hard Drives" and "Disk Drives", but then again SSD is not a hdd ;) that's why I ended up with just Drives.
My personal preference would be: **Drive(s)** with parenthesis and all.

archinstall/lib/user_interaction/general_conf.py
```diff
@@ -79,7 +79,7 @@ def select_language(preset_value: str = None) -> str:
 	sorted_kb_lang = sorted(sorted(list(kb_lang)), key=len)
 
 	selected_lang = Menu(
-		_('Select Keyboard layout'),
+		_('Choose keyboard layout'),
```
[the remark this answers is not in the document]
i2: The reason I did this:
There are two sets of entries called "Select Keyboard layout" and "Select keyboard layout". (lower and upper case K) which was confusing. So I changed "Select Keyboard layout" to "Choose keyboard layout".

And the other change being  "Select keyboard layout" -> "Keyboard layout" is in line with mission for this PR (making things shorter, less verbose, easier to find).

archinstall/lib/menu/global_menu.py
Torxed: I might be an old soul, but "Drives" for me represent cars :P Is there anything else we can use? If not I'm ok with this.

i2: By "If not I'm ok with this." you mean the new term or the old one? Sorry for the confusion! :D

Torxed:
Great work!

i2:
The English menu works great, but for some reason, none of the other locales show correctly! Here's for instance French:
image

Torxed:
It's because some of the message ID's have moved/changed.
https://github.com/archlinux/archinstall/pull/1202 will correct this. However due to the changes someone needs to go through the pot file and update the incorrect ones.

screenshot

i2:
Cool! With #1202, all languages except French and Polish seem to be working now.

Torxed:
> Cool! With #1202, all languages except French and Polish seem to be working now.

Yea they need to go through a `poedit` again or something I think.

PR #1196, opened 2022-05-16
Added a HSM menu entry

### LINKED ISSUE #861: Support TPM for systemd-boot
Just as a reminder I'll put this conversation here: https://www.reddit.com/r/archlinux/comments/s21mab/comment/hscgy9o/?context=3

Torxed:
Also a safety check to make sure a FIDO device is connected.
Trying to verify against an actual FIDO2 compliant HSM before committing.

Following: https://wiki.archlinux.org/title/Universal_2nd_Factor#Data-at-rest_encryption_with_LUKS
Following: https://wiki.archlinux.org/title/Trusted_Platform_Module#systemd-cryptenroll
Following: https://wiki.archlinux.org/title/dm-crypt/System_configuration#Using_sd-encrypt_hook

This fixes #861


Torxed:
@svartkanin I can't find anything in this PR that would cause the issues mentioned in https://github.com/archlinux/archinstall/issues/1203.

Torxed:
This is ready for merger after I've figured out what the best way forward is for telling the initramfs that it's an option to unlock with HSM: https://wiki.archlinux.org/title/Trusted_Platform_Module#systemd-cryptenroll

dylanmtaylor:
Should this be an --advanced feature? Not sure how prevalent using something like a yubikey for disk encryption would be

Torxed:
It's at a state where it encrypts, enrolls the FIDO2 token, reboots and can successfully detect the partition as encrypted and tries to unlock it.

However, it uses `password` instead of `fido`/`tpm` to do so.
Not sure how to:
 * Get it to prefer tpm/hsm over password
 * Get it to use `PARTUUID` instead of `UUID` *(Maybe that doesn't matter too much, but it required some additional changes to the core of archisntall)*

Torxed:
Pretty sure we need the `/etc/crypttab.initramfs`
But with it, I get:
img

And without it, I get a password-prompt instead of fido.

svartkanin:
I find the --advanced features a tick to hidden :)
Would it be a good idea to possibly have them as a general menu entry behind an "Advanced" or "Experimental"? This way we could display some additional information as well

Torxed:
> I find the --advanced features a tick to hidden :) Would it be a good idea to possibly have them as a general menu entry behind an "Advanced" or "Experimental"? This way we could display some additional information as well

That could work, would clean up the main menu a bit too.
But let's hold off on it until this PR goes in, as it's driving me insane and keeping the GUI as-is would mean less hunting for me while trying to figure out the technical details around what's wrong with FIDO2.

I suspect it's a library missing that `mkinitcpio` isn't pulling in for the FIDO2 specifically.
But not sure how to debug that.

Torxed:
After installing the package [libfido2](https://archlinux.org/packages/extra/x86_64/libfido2/) before running `mkinitcpio -P` was the missing link. I came to the conclusion when I ran `strace -e trace=open /lib/systemd/systemd-cryptsetup attach 'luksdev' '/dev/sda2' 'none' 'luks,fido2-device=auto'` on the live ISO. That gave me a diff of a few libraries compared to what was installed on the system, but mainly `/usr/lib/libfido2.so.1` which was missing on the target system. `pkgfile /usr/lib/libfido2.so.1` gave me `libfido2` so I installed it.

I also had to remove `/etc/crypttab//etc/crypttab.initramfs` as it introduces a crash in the startup service.

I also had to remove `tpm2-device=auto` as it causes an override no matter the order these options come in to only use password during unlocking.

Torxed:
I think this is ready to go in now:
[IMAGE ALT TEXT HERE](https://www.youtube.com/watch?v=35L54syE8II)

PR #1206, opened 2022-05-17
Fix some consistency issues with 1195

dylanmtaylor:


Torxed:
Quick work! Thanks! :D

PR #1210, opened 2022-05-18
Enhance view

svartkanin:
Added menu preview for `Disks` and `Disk layout`:

image
image

I also realized that the mypy check in the github actions wasn't working properly since it passed `--follow-import=skip` instead of `silent`, which is fixed now too :)


svartkanin:
seems like mypy works now ;) but funny enough that some execution works on my local oO

Torxed:
> seems like mypy works now ;) but funny enough that some execution works on my local oO

I've witnessed the same weird behavior hehe.
It's a bit inconsistent at times..

Do you want to fix the mypy complaints before I merge now that it works? :sweat_smile:

svartkanin:
now even funnier, the mentioned file doesn't have a line `490`...

svartkanin:
okay maybe I should have downstreamed master first :facepalm:

Torxed:
Hehe, easy mistake ^^ Good job, preview looks great!

PR #1219, opened 2022-05-19
Clarify info message about profile installation

Didr:
Short and sweet :)
I was a bit confused about a message concerning "Installing **network profile** Profile{kde}".
I've replaced the word "network profile" with the word "archinstall profile"


Torxed:
We thank you for this contribution! 💪

Issue #1183, opened 2022-05-14
Renaming config files result in 2 configs missing warning


i2:
To type less when I am testing archinstall in a VM, I have renamed my config files to:
- cf.json (user config)
- dl.json (disk layout)
- cr.json (user credentials)

When I run `python3 examples/guided.py --config cf.json --disk-layout dl.json --creds cr.json`, in the menu, I get the following:
![image](https://user-images.githubusercontent.com/3228126/168438932-22cc8a57-1c9d-4069-889b-071310c7737b.png)

Now, if I click on "Save configuration" and hit Esc (even without saving any config files), that warning disappears!

Torxed:
I think it's due to a static setup step fo the dependency checks. They don't take into consideration of what's been given in the `--conf` fully. We'll look in to it :)

svartkanin:
@i2 I have difficulties reproducing this, could you share which configurations where reported as missing? Or did you by any chance notice which ones did not get imported from the config files?

i2:
@svartkanin Sorry I missed this! And thanks for the fix.

Issue #1182, opened 2022-05-14
Creating user and promoting him does not show in the script


erikdubois:
When you create a user I did not get the first time I have to promote him to be superuser.
It does not say I should do that anywhere.

I would assume a user can update and install his system by default.

That happened in this video at that moment in time.

https://youtu.be/boWn4KnOc24?t=281

I tested it again lateron and I noticed that we do not see that the user has been created IF we promote him.
So we are uncertain if a user has been created for that matter or if he has root privileges. 

https://youtu.be/OZAyOXb_eC4?t=65

Everything works like a charm once you know you have to take an extra step in the user.
![ArcoLinux-2022-05-14_09-01](https://user-images.githubusercontent.com/10594806/168414901-cc92c4bb-a86d-4a4b-a84f-b5d6500adf71.png)

Torxed:
I agree that this is a bit confusing.
The option is a bit hidden and a bit unclear.

We'll re-work this for the next minor release.
Cheers for reporting it!

erikdubois:
I love the way you have changed since last time I looked at it.
the / to search is awesome

Torxed:
Hehe thanks @erikdubois. All credit to the new menu system goes out to @svartkanin and @wllacer who has put in hours and hours making it look the way it is :) I just merged a few PR's hehe.

svartkanin:
Yeah we should rework this user selection to make it clearer

erikdubois:
complimenting you on a job well done
just installed 22.06.01 of ArchLinux
the user creation is solved in a very user friendly way
kudos - I will make a video shortly for anyone who needs it

Issue #1112, opened 2022-05-01
Generated user_*.json files do not appear to store all answers


jamincollins:
In attempting to use the `user_configuration.json` and `user_credentials.json` output from a previous installation/run for a subsequent installation/run.  I've found that they do not appear to store all answers.  Such as prompts that are to be left blank to skip.  Such as whether to set a root password or create additional users.  There may be other prompts that this applies to as well.

Torxed:
they are stored in user_credentials and subsequent questions can be ignored with `--silent`

jamincollins:
The questions where the prompt was left blank _are not_ recorded in **user_credentials**.  I've reviewed the file and when electing: not to set a root password, create only one user with sudo privileges, and no additional users.  The file contains only one entry and subsequent runs prompt for both the root password and additional users.  I guess passing `--silent` _could_ be seen as a workaround, but I would expect the previous answers (electing to leave blank is an answer) would have been recorded and as a result used to answer the same question on subsequent runs.

Issue #1109, opened 2022-05-01
Prompt for creating a user with sudo privileges is unclear as to whether it's a question or a request for a user name/login.


jamincollins:
Prompt for creating a user with sudo privileges is unclear as to whether it's a question (should it be done) or a request for a user name/login.  Turns out it's a request for a user name/login.

> Create a required super-user with sudo privileges: 

Perhaps changing this to something less question-like, such as:

> Please enter the desired login (will be granted sudo privileges):

Torxed:
Easy enough change, should probably adapt this to the new menu system overall and have a tickbox or something rather than my old menu output :)

SoongVilda:
I also found unclear user and superuser configuration. 
As for the me, what's make sense.

1. Input: Do you want to create user? Yes/no/exit
Output: Please specify username: "user"
2. Input: Should "user" have root privileges?
Output: Yes/no/exit 

It will lead you naturaly to configuration with user and possibility to use sudo.

PR #1220, opened 2022-05-20
Rework user management

svartkanin:
# Re-worked user management

## Fixes
- https://github.com/archlinux/archinstall/issues/1112
- https://github.com/archlinux/archinstall/issues/1183
- https://github.com/archlinux/archinstall/issues/1182
- https://github.com/archlinux/archinstall/issues/1109

### New file format for user credentials 
The format for the credentials config file has changed to allow specifying multiple users. This change is backwards compatible, so if old configurations get passed they are parseable and will internally be transformed into the new format. When saving configurations they will be saved in the new format. 

Old format
```
{
	"!root-password": "<root password>",
	"!users": {
		"username": {"!password": "<user password>"}
	},
	"!superusers": {
		"admin": {"!password": "<admin password>"}
	}
}
```

New format
```
{
	"!root-password": "<root password>",
	"!users": [
		{
			"username": "<USERNAME>",
			"!password": "<PASSWORD>",
			"sudo": false
		},
		{
			"username": "<SUDO_USERNAME>",
			"!password": "<PASSWORD>",
			"sudo": true
		}
	]
}
```

### User object
`!superuser` does not exist anymore, instead only the internal `!users` is handled whereas the entries are now a `User` class object 

```
@dataclass
class User:
	username: str
	password: str
	sudo: bool
```

This greatly simplified the code in some parts and makes it easier to avoid bugs :)

### Update to menu 

The menu has now only 1 entry to define users in which the sudo privileges can be specified. 

image

@Torxed I'll let you know when it's ready, probably tomorrow :)


dylanmtaylor:
Can you also update the docs and schema for this change?

svartkanin:
Yep will do that too :)

Torxed:
I don't know why i did'nt do this right away haha, way more clean! Thank you!

dylanmtaylor:
May I suggest we don't merge this until 2.5.0 since it's a schema-breaking change?

svartkanin:
It's not a breaking change since it's backwards compatible :) 

I think the current state is way more fragile since we only support a single user in a file. Which means - and has been pointed out in the issues - that users will lose their actual settings if they try to export them.

In addition I think the 1 different settings of "users" and "superusers" seemed to have stirred quite some confusion which this resolves as well

archinstall/lib/installer.py
```diff
@@ -1054,17 +1055,15 @@ def enable_sudo(self, entity: str, group :bool = False) -> bool:
 
 		return True
 
-	def user_create(self, user :str, password :Optional[str] = None, groups :Optional[str] = None, sudo :bool = False) -> None:
-		if groups is None:
-			groups = []
-
+	def user_create(self, users: List[User]) -> None:
```
Torxed, line 1058: The function should stay as it were, in the sense that it was a function to perform one action.
It made it's purpose a bit more clear. And I propose we create a secondary function that deals with the loops and stuff, which would be called `load_users_layout()` or something similar to `load_disk_layout()`.
Thoughts?

Torxed:
Yea it appears backwards compatible, I get that there's still some work left.
But I'd be happy to run tests on this on the backwards compatibility side and if they're successful I'd be happy to merge this in time for the release :)

archinstall/lib/installer.py
svartkanin: Yes that's a great point! I'll update that one :)

dylanmtaylor:
> Yea it appears backwards compatible, I get that there's still some work left. But I'd be happy to run tests on this on the backwards compatibility side and if they're successful I'd be happy to merge this in time for the release :)

I like this plan. I was under the impression that this was a breaking change

svartkanin:
@Torxed This should be ready for a review now :)

Torxed:
> @Torxed This should be ready for a review now :)

Testing commences now! :D

Torxed:
It loads the old format where I created a `!superuser` and a `!user`.
screenshot

It also installs with the old configuration:
screenshot

And no shocker i guess, but the permissions are set correctly to:
screenshot

Torxed:
Great work, especially making it backwards compatible. Merged and done! :)

svartkanin:
Fantastic, thanks for double checking everything :)

svartkanin:
I think the related issues where not closed automatically

Torxed:
You are correct heh, gotta love Github.

svartkanin:
I'm not sure what triggers it, maybe only mentioning it in the title 🤷‍♂️

Torxed:
It's if the word `fix` is in the same sentence as the issue link I think.
There are some other aliases for it, but it has to be in the same sentence i recon.
But I don't mind closing the manually too hehe, gives me a sense of purpose :angel:

erikdubois:
Could you add a text when we can start testing it?

svartkanin:
It is in master now so basically from there now :), @Torxed will this make it into the new iso?

Torxed:
> It is in master now so basically from there now :), @Torxed will this make it into the new iso?

It will :) I'm planning on running several tests throughout the day.
If they all succeed, I'm aiming to release the next version later this evening or very early tomorrow :)

PR #1221, opened 2022-05-20
Updated base.pot, added pt_BR locale

Alexmelman88:


dylanmtaylor:
Is this cain-dev's work? It'd be nice to give them authorship on the commits

Alexmelman88:
> Is this cain-dev's work? It'd be nice to give them authorship on the commits

Yes

dylanmtaylor:
One way of doing so would be to rebase this on their branch, so it keeps the commits they already made.

Alexmelman88:
Okay, I'm trying to figure out how I can do that :)

Torxed:
I think I got it right :) There was one collision and that was:
*"Se deseja um navegador web, como firefox ou chromium, pode especificá-lo "*
<<<<<<< HEAD
"na próximo prompt."
=======
"no próximo prompt."
>>>>>>> f20e0e97cc2649820eeb4d92415d8ccdbba27161 

After consulting with a authentic Brazilian, `no` appears to be the correct choice.
So I chose that one ^^ And I assume the locale should be `pt_BR` and not `pt-BR`. But I'm not sure if the translation lib works differently from the NIX style of locale syntaxes.

Issue #1227, opened 2022-05-21
Installation failure on master, latest commit.


dylanmtaylor:
Unlike the previous failure I reported, this one didn't have encryption enabled at all.

configuration:

![image](https://user-images.githubusercontent.com/277927/169657872-8be56723-37e3-44ad-9be8-432205f4ba2e.png)
![image](https://user-images.githubusercontent.com/277927/169657877-0e512373-9ad4-4e7d-a284-ac0eec92bb82.png)

cmd history:
![image](https://user-images.githubusercontent.com/277927/169657882-074ac181-14d2-4738-b9a7-f6ecf83f80c8.png)

install.log:

![image](https://user-images.githubusercontent.com/277927/169657902-4dffa8a6-4727-41ce-bd67-09379bfeff2c.png)

stack trace:

![image](https://user-images.githubusercontent.com/277927/169657905-72d99065-556b-4bc9-89e1-d0d684b47369.png)

Torxed:
So I've reproduced the issue, and boiled it down to `device_instance` being the culprit.
It's the only non-standard object in the `deepcopy(partition)` step.

I'm trying to figure out why this happens.
Because to me the code for this particular config and code path hasn't changed much at all..

But some underlying variables appears to have.
After solving this bug, by implementing my own `deepcopy` *(just a recursive dict copier)* we'll get this:
![bug](https://user-images.githubusercontent.com/861439/169684641-224b37f0-7a39-4b9e-96e0-e779faf053ca.png)

Which indicates that `.real_device` or something similar isn't stripping subvolume names correctly anymore.
I'll get this sorted but I need to figure out the root cause first.

Torxed:
The path some how gets mangled up by: https://github.com/archlinux/archinstall/blob/33ce7dbd8baf5402a20e850f503390b9c229e698/archinstall/lib/disk/btrfs.py#L225

I understand why, as `.path` inserts `.path` + `new_subvolume`, appending on the already given path.
I think I will have to re-write this logic to handle it elsewhere, as string management is probably not the best fit here.

This will take a day or so. But for now setting: https://github.com/archlinux/archinstall/blob/33ce7dbd8baf5402a20e850f503390b9c229e698/archinstall/lib/disk/partition.py#L492-L493

To:
```python
	def unmount(self) -> bool:
		worker = SysCommand(f"/usr/bin/umount {self.device_path}")
```

But that will cause:
[...]

PR #1234, opened 2022-05-22
Rework btrfs handling

Torxed:
I have created more sparation of duties by making `BtrfsSubvolume` responsible for its things.
And a `BTRFSPartition` that inherits `Partition()` but deals with all the quirks of BTRFS on a partition.

This fix: #1227


Torxed:
I almost have it installing again, I just need to make better handling of the JSON struct supplied by the user - as there are still a lot of for loops and if cases to deal with it bundled in "other code". I'd like to separate that out. But it's almost there :)

Torxed:
Un-encrypted layouts now work:
archinstall

Which means I think this is ready to go in.
The actual re-work of BTRFS I think is complete. But it could be cleaned up some more in regards to the conversion/handling of user input -> convert that to a `BTRFSSetup` object or something with better error handling, and have all the setup stuff there. But I think it's good enough for now.

It encrypts, it formats, sub-volumes are there and mounted in the correct order.

Encryption:
screenshot

PR #1243, opened 2022-05-24
Dmenu not needed

edl2:
Qtile comes with simple run prompt already, want to keep as vanilla as possible.


Torxed:
Thank you! I have yet to test this DE!

PR #1247, opened 2022-05-25
Remove unneeded packages from awesome profile?

edl2:
This profile currently installs a nemo - file manager, gpicview - image viewer, and maim - screenshot taker. all of these items do not typically come with awesome and aren't awesome specific and should be installed by the user. (also not required for the system to work). Especially since these are not even the most popular tools for each use.


dylanmtaylor:
These are leftover from when this project was in its infancy and Torxed was the only user of that profile for the most part. I agree these should be removed.

Torxed:
Goodbye old friends, you've been with me since the start in 2012! :headstone:
Jokes aside, thank you for cleaning up that profile :)

PR #1250, opened 2022-05-27
Fix dual-booting

Torxed:
# Fixes

 * Optimized partition lookups
 * Fixed re-use of partition UUID's
 * `BlockDevice().get_partition()` now supports looking up both `PARTUUID` and `UUID` for a partition under itself
 * Partitions listed in `--disk-layout` that doesn't have a PARTUUID/UUID should no longer cause an exception, but instead logs a warning and they will simply be ignored
 * `Filesystem().add_partition()` now handles `DiskError` raised by `partition.part_uuid`
 * Fixed issue on normal partitions where the device was not properly frozen in `lambda` calls, meaning two or more mount-points shared the same `device_instance`.
 * Lowered global `DISK_RETRY_ATTEMPTS` to 5, as the timeouts are linear *(`range(DISK_RETRY_ATTEMPTS) * DISK_TIMEOUTS`)*

### DIFF
--- a/archinstall/lib/disk/blockdevice.py
+++ b/archinstall/lib/disk/blockdevice.py
@@ -293,6 +293,7 @@ class BlockDevice:
 		count = 0
 		while count < 5:
 			for partition_uuid, partition in self.partitions.items():
+				print('Comparing:', [partition.part_uuid.lower(), uuid.lower()])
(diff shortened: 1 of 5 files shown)


Torxed:
Verified that it works: https://www.youtube.com/watch?v=RoTe4uAqIHA

PR #1262, opened 2022-05-27
Fix issue with multiples spaces in additional packages

dylanmtaylor:
Still need to test this.

### DIFF
--- a/archinstall/lib/user_interaction/general_conf.py
+++ b/archinstall/lib/user_interaction/general_conf.py
@@ -172,7 +172,7 @@ def ask_additional_packages_to_install(pre_set_packages: List[str] = []) -> List
 
 	def read_packages(already_defined: list = []) -> list:
 		display = ' '.join(already_defined)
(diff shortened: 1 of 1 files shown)


archinstall/lib/user_interaction/general_conf.py
```diff
@@ -172,8 +172,8 @@ def ask_additional_packages_to_install(pre_set_packages: List[str] = []) -> List
 
 	def read_packages(already_defined: list = []) -> list:
 		display = ' '.join(already_defined)
-		input_packages = TextInput(_('Write additional packages to install (space separated, leave blank to skip): '), display).run()
-		return input_packages.split(' ') if input_packages else []
+		input_packages = TextInput(_('Write additional packages to install (space separated, leave blank to skip): '), display).run().strip()
```
Torxed, line 175: You probably want `.replace("  ", " ").strip()`

dylanmtaylor: I think .split() without ' ' as an argument treats multiple spaces as one.

dylanmtaylor:
Tested and working.

PR #1263, opened 2022-05-28
Install font packs by default on desktop systems

kpcyrd:
Especially having emojis working by default is probably very helpful for beginners :cat::v:

### DIFF
--- a/profiles/desktop.py
+++ b/profiles/desktop.py
@@ -25,6 +25,9 @@ __packages__ = [
 	'wpa_supplicant',
 	'smartmontools',
 	'xdg-utils',
+	'noto-fonts',
+	'noto-fonts-cjk',
+	'noto-fonts-emoji',
 ]
 
 __supported__ = [


dylanmtaylor:
This was shot down previously as too opinionated, and we aim to ship a minimal installation by default with users adding what they want and configuring the system after installation.

svartkanin:
This can easily be added as additional packages in the menu or just as a post installation script. It's more of a system setup rather than a system installation property

dylanmtaylor:
I'm not strongly against this, it's valuable to have support for displaying text in various languages, and people enjoy having emojis on their system. I'd install these packages on my personal systems without hesitation.However, it does add to the install size, SIGNIFICANTLY. And some people want a minimal 'desktop' install on a system with a small drive like a VPS cloud system just to make administration easier for them (sample use case).

These packages add:

noto - 112.6 MB

noto cjk - 297.8 MB
noto emoji - 9.4 MB
noto extra - 364.9 MB

That's nearly 800 MB of space and is going to be considered unnecessary bloat for a lot of people. There will be complaints. The default needs to be slim for people with limited space and bandwidth, especially slower connections like DSL. This is why this makes more sense to install, post-install if desired, or to add to additional packages.

svartkanin:
That was kind of my point, most likely not expressed too well :)

This would go beyond a typical installation imho, like it is not a necessity to get the selected profile installed and running but a feature for certain users to enjoy after installation :)

Torxed:
I'm closing this, as it's a pretty simple "additional packages to install".
And if users forget it's one `pacman -S <fonts>` away. I get the convenience argument but I would for instance almost quadrupling the awesome desktop installation for no additional benefit since I don't use moden emoji's :)

PR #1273, opened 2022-05-28
Don't offer to encrypt /boot, exlude it from the set available to the user

### LINKED ISSUE #1269: [critical] master installation failure: 'MapperDev' object has no attribute parent
installation video: https://www.youtube.com/watch?v=P-ZpZne7Pks

Installation: BIOS system on KVM, wiping whole drive, encryption on. Problem: /boot is errantly encrypted and is a mapper device. This causes bootloader installation to fail.

image

### LINKED ISSUE #1270: [critical] master installation failure: /dev/block/252:0 is not located on a partitioned block device
installation video (sorry for not being cropped, I only recorded half my screen). https://www.youtube.com/watch?v=m2Sk7XLH0cg

Setup: UEFI on KVM, using entire drive (wipe) with btrfs and all options selected. Encryption on.

image

This also seems to have an encrypted /boot and shouldn't.

image

### LINKED ISSUE #1271: Select which partitions to encrypt should not offer to encrypt /boot until we support that
This is the real problem for issues #1269 and #1270. The installs for both failed with different messages because /boot was encrypted. This isn't supported but users won't know that. Let's not even allow it until we support it.

dylanmtaylor:
image
Closes #1271
Closes #1270 
Closes #1269

### DIFF
--- a/archinstall/lib/menu/global_menu.py
+++ b/archinstall/lib/menu/global_menu.py
@@ -207,7 +207,8 @@ class GlobalMenu(GeneralMenu):
 					if storage['arguments']['disk_layouts'][blockdevice].get('partitions'):
 						for partition_index in select_encrypted_partitions(
 								title=_('Select which partitions to encrypt:'),
-								partitions=[p for p in storage['arguments']['disk_layouts'][blockdevice]['partitions'] if p['mountpoint'] != '/boot']
+								partitions=storage['arguments']['disk_layouts'][blockdevice]['partitions'],
+								filter_ = (lambda p: p['mountpoint'] != '/boot')
 							):
 
 							partition = storage['arguments']['disk_layouts'][blockdevice]['partitions'][partition_index]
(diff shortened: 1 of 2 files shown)


dylanmtaylor:
So this actually needs more work. The index no longer matches, so it results in a failure still. If I mark /root for format, and it is the only item in the list, it encrypts /boot as it is first in the dict

image

dylanmtaylor:
Might need to patch this function too. https://github.com/dylanmtaylor/archinstall/blob/3328779457db0cbbad90906fee41dc5dee41cc2d/archinstall/lib/user_interaction/partitioning_conf.py#L366

dylanmtaylor:
Much better. Now the index is correct.

image

dylanmtaylor:
Although I bet I can exclude it from the list and also get the correct index...

dylanmtaylor:
Perfect. Ready to go in.

image

Torxed:
Great work, and happy to see that the filter function got put to use hehe.

PR #1277, opened 2022-05-29
Fix subvol selection

svartkanin:
Quick fix for the subvolume blowing up, but this is not the end since the Subvolume menu seems to be quite broken at the moment.


Torxed:
Thank you for the quick work! I'll be tagging the release in a few moments :)

PR #1280, opened 2022-05-29
Updated base.pot

Alexmelman88:


archinstall/locales/base.pot
```diff
@@ -652,7 +652,7 @@ msgid ""
 msgstr ""
 
 msgid ""
-"Choose which servers to install, if none then a minimal installation wil be "
+"Choose which servers to install, if none then a minimal installation will be "
```
dylanmtaylor, line 655: Can you make this correction in all languages? The translated string will be the same

Alexmelman88: Ok

PR #1289, opened 2022-05-30
Update blockdevice

svartkanin:
This should make the BlockDevice class more robust and less prone to issues. It introduces dataclasses to store the Blockdevice metadata and lets callers retrieve it more easily this way. 
This now also removes the previously introduced `cached_property` since it's redundant now and should also eliminate confusion what the heck that is :) 

I've also adjusted the menu preview to be aligned with the other previous to be in a table format. 

image

@Torxed I've tested this as good as I could but I might have missed some cases, hopefully there's enough time now till the next release to be certain :)


svartkanin:
@Torxed fixed the conflicts, should be good now :)

Torxed:
Thank you! Good work as always, merged and it's in! :)

PR #1290, opened 2022-05-30
Divide Chinese locales into 2 variants

toto6038:
# Describe your PR

Change Chinese language code to its two popular variants to avoid future renaming once translation of 2 variants are submitted.


dylanmtaylor:
We probably need to update this for pt-br as well

PR #1295, opened 2022-06-02
Fix swiss script bug

LittleboyHarry:
# BUG

    MyMenu.post_callback() missing 1 required positional argument: 'option'

图片


JoelLarson (approved):
`archinstall/lib/menu/selection_menu.py` has a post_callback definition of

```
def post_callback(self, selection_name: str = None, value: Any = None):
```

svartkanin:
@LittleboyHarry thanks for fixing that, i missed it in a previous one :)

PR #1300, opened 2022-06-03
Add `packagekit-qt5` for kde

### LINKED ISSUE #1041: Archinstall discover shop non-functional.
Hello,

I have installed Arch with archinstall twice now, selected the desktop option then KDE but I noticed that by default the "Discover" shop does not want to function I have to download the packagekit-qt5 package then it functions. Just wanted to let you know.

LittleboyHarry:
Fix #1041: Discover is missing `packagekit-qt5`

### DIFF
--- a/profiles/kde.py
+++ b/profiles/kde.py
@@ -13,6 +13,7 @@ __packages__ = [
 	"sddm",
 	"plasma-wayland-session",
 	"egl-wayland",
+	"packagekit-qt5",
 ]
 
 


dylanmtaylor:
How much space does this add to the installation? I suppose if we have the gnome software packagekit integration we should have parity for the kde profile.

dylanmtaylor:
Installed Size:	373.4 KB

And depends on packagekit... So maybe 2-3 MB more. +1 on this, enough people will want it that we ought to ship it by default IMO

JoelLarson:
This seems related: https://wiki.archlinux.org/title/KDE#Discover_does_not_show_any_applications

Torxed:
Thank you for fixing this profile! :) And thanks dylan for checking the space requirement and Joel for the wifi reference (that helps with quickly identifying recommended steps from the wiki)

Issue #1301, opened 2022-06-03
Needed "ESC" functionality for selecting desktop environment menu and line space below "Use ESC to skip" statement


0xShree:
I found it little bit difficulty to navigate in **_selecting desktop environment_** from the menu.
If the user chose the desktop profile and didn't want to choose the desktop environments from the list, there is no skip funct. to get back to previous menu. User needs click `ENTER` it will go to **_selecting driver menu_**, there can be skipped.
![1](https://user-images.githubusercontent.com/49250702/171786941-cc3aa2d3-221b-46ae-98ab-a7d413dd439e.png)
Also it requires a line space below **_Use ESC to skip_** statement, which is much visually distinguishable
![2](https://user-images.githubusercontent.com/49250702/171786758-e1c9a3ad-90c3-4702-8fca-f0fe264e5c3e.png)

svartkanin:
@Torxed i think we can close this as it was confirmed that it's fixed in master :)

PR #1302, opened 2022-06-03
Adding `Esc` fucntionality to desktop selection menu

0xShree:
Adding small offline functionality https://github.com/archlinux/archinstall/issues/1301
* Adding "ESC" functionality to desktop enviroment selection menu

* Adding line space below to "Use Esc to skip" statement

### DIFF
--- a/archinstall/lib/menu/menu.py
+++ b/archinstall/lib/menu/menu.py
@@ -158,7 +158,7 @@ class Menu(TerminalMenu):
 
 		action_info = ''
 		if skip:
-			action_info += str(_("Use ESC to skip"))
+			action_info += str(_("Use ESC to skip\n"))
 
 		if self._explode_on_interrupt:
 			if len(action_info) > 0:
--- a/profiles/desktop.py
+++ b/profiles/desktop.py
@@ -52,7 +52,7 @@ def _prep_function(*args, **kwargs) -> bool:
 	other code in this stage. So it's a safe way to ask the user
 	for more input before any other installer steps start.
 	"""
-	choice = Menu(str(_('Select your desired desktop environment')), __supported__, True).run()
+	choice = Menu(str(_('Select your desired desktop environment')), __supported__).run()
 
 	if choice.type_ != MenuSelectionType.Selection:
 		return False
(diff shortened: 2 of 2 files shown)


archinstall/lib/menu/menu.py
```diff
@@ -158,7 +158,7 @@ def __init__(
 
 		action_info = ''
 		if skip:
-			action_info += str(_("Use ESC to skip"))
+			action_info += str(_("Use ESC to skip\n"))
```
JoelLarson, line 161: Does adding `\n` to the transaction string cause the localization to break? It might need to go outside of the `_()`.

profiles/desktop.py
```diff
@@ -52,7 +52,7 @@ def _prep_function(*args, **kwargs) -> bool:
 	other code in this stage. So it's a safe way to ask the user
 	for more input before any other installer steps start.
 	"""
-	choice = Menu(str(_('Select your desired desktop environment')), __supported__).run()
+	choice = Menu(str(_('Select your desired desktop environment')), __supported__, skip=True).run()
```
JoelLarson, line 55: [Skip is True by default](https://github.com/archlinux/archinstall/blob/master/archinstall/lib/menu/menu.py#L48)

0xShree: Yeah **True** is redundant

archinstall/lib/menu/menu.py
0xShree: `\n` added to **Ctrl+C** functionality too

JoelLarson: https://github.com/archlinux/archinstall/blob/7943dd82365fd9fb5034a0f1c05de3ccabda468a/archinstall/locales/en/LC_MESSAGES/base.po#L369

I see that this localization exists in English, so this change shouldn't require a new localization line.

0xShree: Ok, then why there is no line space?

JoelLarson: Apologies for the confusion. I was trying to say that `_("Use ESC to skip\n")` will resolve to the LC_MESSAGES msgid entry. What you have here will work because the `base.po` line I linked exists. The code you have here looks correct.

0xShree:
VS code autocomplete really sucks 😵‍💫. It added some random library for skip, then I had to revert the commit

0xShree:
I got it know, the `ESC` functionality for desktop env menu already implemented in master branch. I was testing on released version

PR #1308, opened 2022-06-06
Fix #1304 - Make password validation less intrusive

### LINKED ISSUE #1304: Suggestion: babysitting password strength is not a necessary step?
Arch itself does not do it, unlike some other distros that by default force requirements on you.

Just maybe the script should skip judging the password and not ask for additional input of yes/no, if you are aware your password is not very secure.

Arent even actually secure passwords ridiculously long? Like 15 chars? So its more of a show of security unless it stops yelling at you after really strong password only.

svartkanin:
Fixes #1304 
This will make the password validation less intrusive and will only display an info message about the strength

image
image
image

### DIFF
--- a/archinstall/lib/models/password_strength.py
+++ b/archinstall/lib/models/password_strength.py
@@ -1,0 +1,85 @@
+from enum import Enum
+
+
+class PasswordStrength(Enum):
+	VERY_WEAK = 'very weak'
+	WEAK = 'weak'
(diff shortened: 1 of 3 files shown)


Torxed:
That looks really great. I like the use of `Enum` here too.
I should get around converting `GPT` and `MBR` lookups to that too hehe.

PR #1309, opened 2022-06-06
Handle cyrillic characters

### LINKED ISSUE #1138: Cyrillic problems (?)

svartkanin:
This should fix #1138 now in a safe way :) 

When choosing a cyrillic languages we'll set the `UniCyr_8x16` language, when a non cyrillic language is set we'll reset the font to a default font


Torxed:
Nice, I'll test this one out in a couple of days but it looks non breaking so I'm merging it :)

PR #1314, opened 2022-06-07
Fix disk layout display

svartkanin:
This should fix the disk layout display to show the btrfs subvolumes in a readable manner

image

### DIFF
--- a/archinstall/lib/user_interaction/partitioning_conf.py
+++ b/archinstall/lib/user_interaction/partitioning_conf.py
@@ -5,7 +5,7 @@ from typing import List, Any, Dict, Union, TYPE_CHECKING, Callable, Optional
 
 from ..menu import Menu
 from ..menu.menu import MenuSelectionType
-from ..output import log
+from ..output import log, FormattedOutput
 
 from ..disk.validators import fs_types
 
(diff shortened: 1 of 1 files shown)


Torxed:
I'm starting to wonder how much we can fit on the main menu hehe. Feels like a lot more than i thought ^^

dylanmtaylor:
> I'm starting to wonder how much we can fit on the main menu hehe. Feels like a lot more than i thought ^^

I imagine this is highly dependent on screen resolution

svartkanin:
Yeah it will possibly not fit in a vm window and in that case the scrolling is activated automatically. 
Not sure if that will cause confusion or not

Torxed:
I think that's fine, but we should add back the indexing in the menues so people can hit 0 on the numpad for instance to jump to 0. I saw that the lib we use support splitting on `4|text` or something like that for instance :)

svartkanin:
@Torxed I might have missed something here, where was this indexing being used?

Torxed:
> @Torxed I might have missed something here, where was this indexing being used?

People with accessability features used this for two things:
1. espeakup appears to work better with the default espeakup profile when there is a clear indexed order
2. As shortcuts through the menu system. They would have the options read out, press 5 to jump to menu item 5, thus making navigating a bit less troublesome for those users.

Both appear to be supported by the menu library already :)

PR #1315, opened 2022-06-07
Introduce '--mirrors' cmd line option which allows the user to add custom repositories/mirrors

### LINKED ISSUE #814: Using custom repos with installed system
I would like to use archinstall, but initially use and get packages from the unofficial ALHP repos (see https://git.harting.dev/ALHP/ALHP.GO & https://wiki.archlinux.org/title/unofficial_user_repositories#ALHP). Since these are custom, optimized builds of Arch Linux official packages, the custom repos need to be added to `/etc/pacman.conf` _before_ the declarations for [core], [extra], and [community], because the official repositories need to be a lower priority and only used as a fallback. I know there is a function, `add_custom_mirrors()`, that allows for _appending_ custom repos to `/etc/pacman.conf`, but that is only for the live ISO, I believe, and will therefore not be carried over to the installed system. Additionally, it only supports appending custom mirrors/repos, rather than supporting both prepending and appending.

There are many reasons advanced users may want to have unofficial/custom repositories as sources for their installation, for example if they want to have packages from them automatically installed, or in my case, to override the official package versions. In many cases, these custom repositories should be carried-over to the installed system in the created `/etc/pacman.conf` file (I think this is the best default behavior), though an option should allow for them to not be carried-over.

I am fully aware that _specific_ custom repositories are not supported or endorsed, but giving the option for users to add them should not be a problem, I hope. Otherwise, I would have to install Arch Linux using the official packages... just to have to re-install everything again after adding them myself, which really doesn't make any sense if my goal is to use custom package builds. This change would also allow specifying and using _official_ repositories such as testing, gnome-unstable, kde-unstable, etc. Any repositories (name, URL/Include (mirrorlist), SigLevel) could be added and optionally preserved on the installed system.

Tereius:
With the cmd line option _--mirrors_ user-defined repositories/mirrors can be made available during/after installation (might close #814).
Following format is expected:
```
archinstall --mirrors <REPO_NAME>,<INDEX>@<REPO_URL>
```
This will create a new repo section in /etc/pacman.conf right **at the beginning** of the guided installer.
The new section looks like this:
```
[<REPO_NAME>]
SigLevel = Optional TrustAll
Server = <REPO_URL>
```
The `<INDEX>`  determines the position in pacman.conf (an index of zero will place it right at the top).
The new `add_custom_repository()` function in mirrors.py is used for this.

Because the guided installer now changes the pacman.conf on the host system, I added a new Context Manager `PacmanConfBackup` in pacman.py that does a backup of the original pacman.conf file. If the Context Manager goes out of scope the backup is restored.

In `find_package()` in packages.py I had to replace `package_search()` with the new function `pacman_package_search()` which searches for packages via `pacman -Ss` instead of http search requests. This is necessary because custom repositories may contain packages not found in the official Arch repos.

In `installation.minimal_installation()` I added an argument to copy over the /etc/pacman.conf to the target system. The guided installer always copies the pacman.conf file over if the _--mirrors_ argument was provided.


svartkanin:
Custom mirrors are supported in the latest version, as this PR is fairly old I'll close it for now but if it adds new functionality feel free to open a new one

Issue #1312, opened 2022-06-07
No connection after an installation with manual configuration


castillofrancodamian:
In a new installation and manual network configuration, I have not been able to connect to the internet or ping to verify and I see that adding `Gateway=ip-dns` solves it.

svartkanin:
@castillofrancodamian thanks for reporting would you mind sharing which network configuration you did exactly? Was it using auto dhcp or manual one?

dylanmtaylor:
> @castillofrancodamian thanks for reporting would you mind sharing which network configuration you did exactly? Was it using auto dhcp or manual one?

The title says it was manual.

Torxed:
> > @castillofrancodamian thanks for reporting would you mind sharing which network configuration you did exactly? Was it using auto dhcp or manual one?
> 
> The title says it was manual.

It could be manual interface configuration but with DHCP or static IP?

Torxed:
It might be me miss-remembering the options from the top of my head.
IIRC the "manual configuration" used to be called that, meaning you can manually configure each individual NIC.
But Manual could also mean static IP manual configuration.. The irony/confusion doesn't get lost on me here and we should improve the language used I think.

With static IP, did you define a gateway at all during archinstall steps?
Easiest would be if you supplied the `/var/log/archinstall/install.log` and we can check what options you gave :)

svartkanin:
That helps a lot thanks! I can have a look at this one

Torxed:
`gateway: null` has me worried :) Might that be the issue?

svartkanin:
Most likely, if the dhcp is not set to auto I think the gateway is an optional field (or also with dhcp!?)

So not sure atm if this is a bug or just something that has to be remembered to be set by the user :)

castillofrancodamian:
I don't use DHCP, I use static IP. The installer asks me for an IP for the device, a gateway and a DNS server, but I see that the gateway is not saved.

svartkanin:
okay cool, so you did specify the gateway but it didn't save it, I'll look into that

castillofrancodamian:
That's how it is. I enter the data and the way it asks me.

castillofrancodamian:
Thank you!

PR #1320, opened 2022-06-09
Fancy user interface

svartkanin:
# Fixes
I happened to stumble upon this one https://github.com/archlinux/archinstall/issues/1312 so that is fixed now :) 

# Table selection feature 
In the sub-menus that perform management actions, existing entries are now displayed as tables and the rows are basically the selection options

Subvolume menu 
image

User management
image

Network interfaces
image

# Other bug fixes

- Network configuration parser seemed to have been broken for new configurations, that should be fixed now (I'm unaware that anyone raised an issue) 
- When pressing ESC in the management menus then the screen would render a new menu without deleting the old one
- I've update a single translation string manually in all files (not actual text translation but simply `< {} > ==> '{}'`


Torxed:
Looks great! And thank you for fixing that gateway thing!

PR #1327, opened 2022-06-12
Fix spacing "The password you are using seems to be weak,"

andrecaspar:
Fix spacing in the message "The password you are using seems to be weak,are you sure you want to use it?" for every locale.

[Current state.](https://www.youtube.com/watch?v=leQbSsu-7F4&t=365s)

### DIFF
--- a/archinstall/locales/base.pot
+++ b/archinstall/locales/base.pot
@@ -505,7 +505,7 @@ msgstr ""
 msgid "Not a valid directory: {}"
 msgstr ""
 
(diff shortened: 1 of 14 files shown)


dylanmtaylor:
Needs to be updated in the source code too for this to work

svartkanin:
I think this text isn't actually used anymore in a new version in master, but not 100% sure.

Torxed:
Correct, this particular string has been replaced.

Issue #1181, opened 2022-05-14
Could not get PARTUUID of partition using 'blkid -s PARTUUID -o value /dev/nvme0n1p2'


EdisonJwa:
I was trying to install archlinux with archinstall on an old laptop but seems there are some things wrong with `blkid`

<details>
<summary>install.log Details</summary>

```
Hardware model detected: LENOVO 80UE; UEFI mode: True
Processor model detected: Intel(R) Core(TM) i5-6260U CPU @ 1.80GHz
Memory statistics: 6887984 available out of 8044508 total installed
Could not detect virtual system: ['/usr/bin/systemd-detect-virt'] exited with abnormal exit code [256]: b'none\r\n'
System is not running in a VM: ['/usr/bin/systemd-detect-virt'] exited with abnormal exit code [256]: b'none\r\n'
Virtualization detected: None; is VM: None
Graphics devices detected: dict_keys(['Intel Corporation Iris Graphics 540 (rev 0a)'])
[...]

PR #1333, opened 2022-06-16
Cleanup partition

svartkanin:
This should cleanup the partition class a bit and make things for removing the native types soon.

### DIFF
--- a/archinstall/lib/disk/blockdevice.py
+++ b/archinstall/lib/disk/blockdevice.py
@@ -118,6 +118,8 @@ class BlockDevice:
 	def _load_partitions(self):
 		from .partition import Partition
 
+		self._partitions.clear()
+
 		lsblk_info = self._call_lsblk(self._path)
 		device = lsblk_info['blockdevices'][0]
 		self._partitions.clear()
@@ -230,8 +232,6 @@ class BlockDevice:
 
 	@property
 	def partitions(self) -> Dict[str, 'Partition']:
-		self._partprobe()
-		self._load_partitions()
 		return OrderedDict(sorted(self._partitions.items()))
 
 	@property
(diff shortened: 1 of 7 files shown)


svartkanin:
Since this PR was still open I pushed more updates to the partitioning into here. 

I have tested that archinstall can still 
- Create ext4 partitioning
- Create btrfs partitioning
- Create encrypted partitions

Since this is quite dangerous update I'd really like to have someone test this as well since it is the essence of the whole cli :)

svartkanin:
I'm going to claim (without proof) that these changes might solve: 
- https://github.com/archlinux/archinstall/issues/1181

Torxed:
Thank you, as always nice work! I'm starting to pull in all the PR's and master will be a bit unstable while I do this (for anyone reading the latest activity on the repo)

Torxed:
I think this PR might have introduced: https://github.com/archlinux/archinstall/issues/1411
I'm trying to figure out exactly why.

Because we still have `BlockDevice.partitions`.

Torxed:
It appears there's a setback here: https://github.com/archlinux/archinstall/blob/3842f1eb5644851ad3af84d84e704c0b33b2b204/archinstall/lib/installer.py#L1001-L1002

Where we no longer get the `BtrfsPartition()` objects, but instead `Partition()` which only has one mountpoint available, and it will be `/var/log/archinstall/var/log` instead of the different nuances that btrfs subvolumes can be mounted at, sharing the same partition.

Torxed:
Trying to think of the best approach forward here.
Because technically those paths are on that partition, so that's not a lie.
But that partition doesn't have a mountpoint, because the subvolumes do.

svartkanin:
@Torxed yeah sorry about the bugs, I can take a look at this today :)

svartkanin:
@Torxed I saw that you have already merged in 2 fixes, is there anything outstanding?

Torxed:
I think it's all good :D I mostly commented to track the relationship between the too.
It was really easy to miss this and I didn't pick up on it during merging so it's on me!

PR #1339, opened 2022-06-20
Invert mypy file checking and add pre-commit hook

svartkanin:
Some changes to how mypy is applied: 

- We're now excluding files to be checked by mypy instead of including them explicitly, which means that all new files will have to be compliant

Add an OPTIONAL pre-commit hook configuration file, this run some checks before actually committing which should indicate if github actions are going to fail :) 
The hooks can easily be installed with 
```
pip install pre-commit
pre-commit install
```


svartkanin:
Extracted the mypy configuration from the github action into it's own `mypy.ini` configuration so it can be called easily from github and the pre-commit hook

svartkanin:
@Torxed If you're okay with this one could we merge it in? Otherwise it will inevitably cause a lot of conflicts whenever something else gets merged :)

svartkanin:

Issue #1344, opened 2022-06-22
archinstall might error out if /etc/locale.gen already is set up


fabiscafe:
Hello!
I just found this bug. I build my own archiso with a fully localized GNOME in german. To make this happen my /etc/locale.gen is already set up and contains an uncommented
```
de_DE.UTF-8 UTF-8
```

This however leads to a problem where Archinstall lists this locale later on as `e_DE.UTF-8`
![grafik](https://user-images.githubusercontent.com/18744080/175011148-60ccf8f7-fc2d-4a39-84dd-2029371489c8.png)

This results in a failed installation attempt

```
['/usr/bin/arch-chroot', '/mnt/archinstall', 'locale-gen'] exited with abnormal exit code [1024]: b"Generating locales...\r\n  en_US.UTF-8... done\r\n  e_DE.UTF-8...[error] cannot open locale definition file `e_DE': No such file or directory\r\n"
```

You can reproduce this with every locale you comment out prior archinstall launch.
This might be because of [this line](https://github.com/archlinux/archinstall/blob/master/archinstall/lib/locale_helpers.py#L23=)
```
text = entry[1:].strip()
```
That might work better with something like 
```
text = entry.replace("#", "").strip()
``` 
instead

This is with archinstall 2.5.0-1

PR #1345, opened 2022-06-22
Explicitly remove the hash character at the line-start in /etc/locale.gen

fabiscafe:
Currently the helper remove the first character of each line, this can
lead to cases, where important characters are removed. For example if
the locale is already set up. (hash is already removed) in that case the
helper would remove the first character of the locale and lead to a
broken attempt to set the locale later on. This change should avoid that
and only remove the hash.

I have to say that I'm not a dev so this might be completely wrong. Nevertheless I have tested this change on the official archiso as well as my own one, where I found the bug.

Note: In my mind `.strip('# ')` should also work, but crashes archinstall for a reason I don't understand. Thats why I haven't done this here.

Fixes: https://github.com/archlinux/archinstall/issues/1344

### DIFF
--- a/archinstall/lib/locale_helpers.py
+++ b/archinstall/lib/locale_helpers.py
@@ -20,7 +20,7 @@ def list_locales() -> List[str]:
 		entries.reverse()
 
 		for entry in entries:
-			text = entry[1:].strip()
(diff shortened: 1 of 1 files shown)


Torxed:
Thank you for this one!

Issue #1352, opened 2022-06-27
parted error location outside of device


JeremiahCheatham:
Unabled to use archinstall on 2 different laptops. hp probook 440 8250u with 2 drivies Windows on other drive. lenovo IdeaPad ryzen 5700u Windows on same drive. I didn't get the exact error copied it it ends with parted claming location is outside of device. Same error on both devices i also went and tried branch below with the same error. I have drived about 20 times now trying to set / and other ideas but always ends in same error.

Other branch i tried.
https://github.com/archlinux/archinstall/issues/1305#issuecomment-1146698663

I'm installing into an existing unformated linux partition on GPT  partiton layout. Using existing EFI vfat on first partition on both machines. I was not able to delete and recreate the partition as it doesn't seem to understand the existing space. So instead i am marking it as / (or not) choosing to format it as btrfs creating @, @home, @.snapshots @log, @pkg choosing to wipe (or not) choosing compress for btrfs (or not) but always getting that same error. It seems to be trying to create a partition even though i am not asking for a partition to be created.

Here is the last time i tried. Though i did try may different ways on both laptops.

````
{
    "HSM": null,
    "__separator__": null,
    "additional-repositories": [],
    "archinstall-language": "English",
    "audio": "pipewire",
    "bootloader": "systemd-bootctl",
    "config_version": "2.5.0",
    "debug": false,
    "desktop-environment": "kde",
    "gfx_driver": "All open-source (default)",
    "harddrives": [
        "/dev/sdb"
    ],
    "hostname": "archlinux",
    "kernels": [
        "linux"
    ],
    "keyboard-layout": "uk",
    "mirror-region": {
        "United Kingdom": {
            "http://archlinux.uk.mirror.allworldit.com/archlinux/$repo/os/$arch": true,
            "http://lon.mirror.rackspace.com/archlinux/$repo/os/$arch": true,
            "http://mirror.bytemark.co.uk/archlinux/$repo/os/$arch": true,
            "http://mirror.cov.ukservers.com/archlinux/$repo/os/$arch": true,
            "http://mirror.netweaver.uk/archlinux/$repo/os/$arch": true,
            "http://mirrors.manchester.m247.com/arch-linux/$repo/os/$arch": true,
            "http://mirrors.melbourne.co.uk/archlinux/$repo/os/$arch": true,
            "http://www.mirrorservice.org/sites/ftp.archlinux.org/$repo/os/$arch": true,
[...]

Issue #1343, opened 2022-06-22
I can't install ISO from commit #1331


castillofrancodamian:
```
Esta es su configuración elegida:
{
    "HSM": null,
    "__separator__": null,
    "additional-repositories": [],
    "archinstall-language": "Spanish",
    "audio": null,
    "bootloader": "systemd-bootctl",
    "config_version": "2.5.0",
    "debug": false,
    "harddrives": [
        "/dev/sda"
    ],
    "hostname": "archlinux",
    "kernels": [
        "linux"
    ],
    "keyboard-layout": "es",
    "mirror-region": {
        "Brazil": {
            "http://archlinux.c3sl.ufpr.br/$repo/os/$arch": true,
            "http://archlinux.pop-es.rnp.br/$repo/os/$arch": true,
            "http://br.mirror.archlinux-br.org/$repo/os/$arch": true,
            "http://linorg.usp.br/archlinux/$repo/os/$arch": true,
            "http://mirror.ufscar.br/archlinux/$repo/os/$arch": true
        }
    },
    "mount_point": null,
    "nic": [
        {
            "dhcp": false,
            "dns": [
                "192.168.1.203"
            ],
            "gateway": "192.168.1.1",
            "iface": "enp0s3",
            "ip": "192.168.1.150/24",
            "type": "manual"
        }
    ],
    "no_pkg_lookups": false,
    "ntp": true,
    "offline": false,
    "packages": [
        "htop",
        "nano",
        "neofetch",
        "screenfetch",
        "openssh"
    ],
    "plugin": null,
    "profile": {
        "path": "/usr/lib/python3.10/site-packages/archinstall/profiles/minimal.py"
    },
    "save_config": null,
    "script": "guided",
    "silent": false,
    "swap": false,
    "sys-encoding": "UTF-8",
    "sys-language": "es_AR",
    "timezone": "America/Argentina/Buenos_Aires",
    "version": "2.5.0"
}
{
    "/dev/sda": {
        "partitions": [
            {
                "boot": true,
                "encrypted": false,
                "filesystem": {
                    "format": "fat32"
                },
                "mountpoint": "/boot",
                "size": "512MiB",
                "start": "1MiB",
                "type": "primary",
                "wipe": true
            },
            {
                "encrypted": false,
                "filesystem": {
                    "format": "ext4",
                    "mount_options": []
                },
                "mountpoint": "/",
                "size": "100%",
                "start": "513MiB",
                "type": "primary",
                "wipe": true
[...]

PR #1355, opened 2022-06-30
Solves issue 1343. Could not locate partition after creation

wllacer:
The partuuid was missing and substituted for something else.
Solves issue #1343 and probably #1352


svartkanin:
@wllacer I'll merge this into my open PR as well that refractors things, I think that one fixes some additional bugs as well

Torxed:
Merged! There's a lot of changing crisscrossing because I've been slow on merging, so if this merger contained any errors do let me know :)

PR #1356, opened 2022-07-01
Downstream merged simple menu changes

svartkanin:
Downstream of the changes merged into the simple-term-menu https://github.com/IngoMeyer441/simple-term-menu


wllacer:
Daniel  Any interesting feature for us? I saw you made some changes there

svartkanin:
Only the one I've added to the upstream ;) I had added the ability to press ctrl+c and actually capture the keyerror on our side so we can use that to reset menu entries.
But i had included that already in our version previously just to be able to use it quickly. This downstream is merely a sync.

Issue #1357, opened 2022-07-01
archinstall do not work 


elvenaz:
hello, im a archuser and this is the first time y get this problem

![errorarchinstall](https://user-images.githubusercontent.com/69324978/176896832-4811c688-5bb1-4f52-8672-8cd078402b00.jpeg)

elvenaz:
> you don't have internet.
> 
> run the command "iwctl station wlan0 connect network"

I feel so stupid,,,, Thank you

svartkanin:
@elvenaz would you mind closing the issue if it's resolved for you now :)

We may consider catching that error and pointing to a connection problem since this seems to confuse people recently

ZeroIntensity:
> We may consider catching that error and pointing to a connection problem since this seems to confuse people recently 

yeah i had this problem the first time i tried archinstall, would have been nice if instructions were given or it at least just told you what the problem was

svartkanin:
Should be an easy one to capture this, I'll look into it

Issue #1347, opened 2022-06-23
Arch installer error 


Guicap:
I can't seem to use it because it gives me this error idk why
![IMG20220623235809](https://user-images.githubusercontent.com/108102257/175429286-6b57ab0e-dbd5-4b93-af89-e9f51f7386ed.jpg)

svartkanin:
@Guicap it seems the command `pacman -Sy` is failing, could you try to run that manually on the system to see what it returns?

Guicap:
When i run it it gives me  this
![IMG20220624083510](https://user-images.githubusercontent.com/108102257/175486897-4ac8dc25-a471-473a-9858-54191155e58a.jpg)
this

svartkanin:
Okay this seems to be a connection issue, do you have internet on this machine you're trying to run it on?

Guicap:
> Okay this seems to be a connection issue, do you have internet on this machine you're trying to run it on?

I did turn on my WiFi and it still gives me the same error i also tried pinging google on my wifi just to test and it gave me a error aswell

svartkanin:
Is this a vm or a physical machine?

Archinstall requires internet to work, so if that is not configured it won't start unfortunately

Guicap:
it is physical and i have internet on i just dont know why it gives me a error

dylanmtaylor:
> it is physical and i have internet on i just dont know why it gives me a error

Is your system clock set correctly? Could be related to SSL certificate validation

Alexmelman88:
You don't have Wi-Fi set up. Enter the following commands:

`iwctl`
`device list`
`station wlan0 scan`
`station wlan0 get-networks`
`station wlan0 connect` _name your wi-fi_ 
enter the network password
`exit`

wlan0 - Wi-Fi device. Its designation you will see after the `device list` command.

After you run archinstall.

Guicap:
> Is your system clock set correctly? Could be related to SSL certificate validation
I actually never thought about that how do u do that sorry for the late response i was busy

iitzrohan:
you have to run dhcpcd command after connecting to wifi on the latest arch iso.
Just connect to wifi with iwctl then run 'dhcpcd' to assign an IP.
After that run archinstall.
I was having the same problem.

Guicap:
> you have to run dhcpcd command after connecting to wifi on the latest arch iso.
> Just connect to wifi with iwctl then run 'dhcpcd' to assign an IP.
> After that run archinstall.
[...]

PR #1361, opened 2022-07-03
Handle no internet connection gracefully

svartkanin:
When starting archinstall and there's not internet access it will fail with a cryptic error message such as reported in https://github.com/archlinux/archinstall/issues/1357 and https://github.com/archlinux/archinstall/issues/1347. 

# Changes
This PR validates if there's an internet connection available on the system and if not will terminate gracefully with a respective message 

```
Your system must have internet access to be able to run archinstall
```

### DIFF
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -254,7 +254,7 @@ def verify_internet_connection():
 		req = urllib.request.Request(url, headers={'User-Agent': 'Mozilla/5.0'})
 		urllib.request.urlopen(req, timeout=5)
 	except Exception as e:
-		err = str(_('To run archinstall your system must have internet access.'))
+		err = str(_('Your system must have internet access to be able to run archinstall'))
 		log(err, fg="red", level=logging.WARNING)
 		sys.exit(1)
 
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -1,4 +1,5 @@
 """Arch Linux installer - guided, templates etc."""
+import sys
 from argparse import ArgumentParser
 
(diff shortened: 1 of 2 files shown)


dylanmtaylor:
We want to be able to handle offline installations with local mirrors

archinstall/__init__.py
```diff
@@ -245,6 +246,20 @@ def post_process_arguments(arguments):
 	load_config()
 
 
+def verify_internet_connection():
+	from urllib.request import urlopen
+	try:
+		import urllib.request
+		url = 'https://1.1.1.1'
```
dylanmtaylor, line 253: Instead of cloudflare, we should be checking an activity to the actual mirrors

dylanmtaylor:
We already tried to do a check for internet connectivity in guided with the line

 if not (archinstall.check_mirror_reachable() or archinstall.arguments.get('skip-mirror-check', False)):

Perhaps updating the message that is output if this does not succeed would be a better solution.

svartkanin:
That's a good point. I'll wrap that one instead of introducing a new one :)

Torxed:
I'm back from vacation starting today, this looks good and I'll catch up and merge things throughout the week.
People have been busy while I was cutting the grass and trimming hedges apparently ^^

svartkanin:
@dylanmtaylor thanks for the feedback, I've updated the PR now to actually capture the mirror call error instead

JoelLarson (approved):

Torxed:
It took 28 days, but I finally merged it! Thanks for improving this area, it's one of the features I use the most ^^

PR #1373, opened 2022-07-11
restoring ability to use a predefined installation directory

wllacer:
Seems that somewhere along the line we lost the ability to let archinstall use a predefined `mountpoint` (be it the default `/mnt/archinstall` or another one, at least using the `guided.py` script.
It happens at least from 2.5.0 and probably earlier.  Beside other minor troubles, the missing validator at GlobalMenu wasn't able to let an empty harddrives list be considered correct in this use case
This PR resolves this problem, and lets the way open to handle this use case thru an invocation argument (not implemented by now)

### DIFF
--- a/archinstall/lib/menu/global_menu.py
+++ b/archinstall/lib/menu/global_menu.py
@@ -42,6 +42,7 @@ if TYPE_CHECKING:
 
 class GlobalMenu(GeneralMenu):
 	def __init__(self,data_store):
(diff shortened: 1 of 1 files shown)


JoelLarson (approved):

Torxed:
Nice catch! There's probably a few other things missing so I'll keep an eye out!

PR #1377, opened 2022-07-15
Enhacements to scripts for developers

wllacer:
two small enhacements to script execution.
1) Ability to be invoked directly without the need of using the `-m archinstall` . Also, in PyCharm and Kdevelop at least i managed to execute the scripts inside the IDE (a bit more tinkering is requiered 'cause the sudo)
This is achieved playing with the python path:
```python
if __name__ == '__main__':
	# to be able to execute simply as python examples/guided.py or (from inside examples python guided.py)
	# will work only with the copy at examples
	# this solution was taken from https://stackoverflow.com/questions/714063/importing-modules-from-parent-folder/33532002#33532002
	import sys
	current_path = os.path.abspath(getsourcefile(lambda: 0))
	current_dir = os.path.dirname(current_path)
	parent_dir = current_dir[:current_dir.rfind(os.path.sep)]
	sys.path.append(parent_dir)
```
2) Ability to get functions called extenally from other scripts:

Just separating code with has not to be executed if it is imported as a library inside this block[¹]
```python
nomen = getsourcefile(lambda: 0)
script_name = nomen[nomen.rfind(os.path.sep) + 1:nomen.rfind('.')]
if __name__ in ('__main__',script_name):
    ....
```
In the other script the functions can be imported via [²]
```python
from archinstall.examples.guided import perform_filesystem_operations, perform_installation
```

[¹]the two previous lines are to get the name of the script from the system (could have done from archinstall.arguments[script], but is more generic).  That `getsourcefile` is from `inspect` -thus the weird syntax-
[²]As long as we keep the shadow copy between  archinstall/examples and examples

### DIFF
--- a/examples/guided.py
+++ b/examples/guided.py
@@ -1,6 +1,18 @@
 import logging
 import os
 import time
+from inspect import getsourcefile
+
+if __name__ == '__main__':
+	# to be able to execute simply as python examples/guided.py or (from inside examples python guided.py)
+	# will work only with the copy at examples
+	# this solution was taken from https://stackoverflow.com/questions/714063/importing-modules-from-parent-folder/33532002#33532002
+	import os.path, sys
+	from inspect import getsourcefile
+	current_path = os.path.abspath(getsourcefile(lambda: 0))
+	current_dir = os.path.dirname(current_path)
+	parent_dir = current_dir[:current_dir.rfind(os.path.sep)]
+	sys.path.append(parent_dir)
 
 import archinstall
 from archinstall import ConfigurationOutput, Menu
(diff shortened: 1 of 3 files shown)


Torxed:
There's one minor issue with this change.
I'll come back and see what it can be, but these behave differently:
```
$ sudo python examples/guided.py`
Testing connectivity to the Arch Linux mirrors ...
  File "/home/anton/github/archinstall-wllacer/examples/guided.py", line 288, in <module>
    if not archinstall.arguments['offline']:
KeyError: 'offline'
```
vs
```
$ cp examples/guided.py ./
$ sudo python guided.py
# Works
```

wllacer:
> There's one minor issue with this change. I'll come back and see what it can be, but these behave differently:
> 
> ```
> $ sudo python examples/guided.py`
> Testing connectivity to the Arch Linux mirrors ...
> Traceback (most recent call last):
>   File "/home/anton/github/archinstall-wllacer/examples/guided.py", line 288, in <module>
>     if not archinstall.arguments['offline']:
> KeyError: 'offline'
> ```
> 
> vs
> 
> ```
> $ cp examples/guided.py ./
> $ sudo python guided.py
> # Works
> ```

It works for me in both cases ( i've used the master after PR 1331). Pretty strange

Torxed:
It might be something local on my end, I'll give it some more tries and see what it's about :)

svartkanin:
This PR hasn't been updated in a while and it's quite outdated so I'll close this for now

Issue #1379, opened 2022-07-16
archinstall kde desktop option comes installed with kde discover but is missing dependencies


Kame03:
discover doesn't work out of the box if you try to launch it it will error saying no application backend found one simple way to fix it is by installing packagekit-qt5

I would like to request packagekit-qt5 to be installed with kde option selected.

Issue #1321, opened 2022-06-09
Please remove packagekit from GNOME profile


heftig:
Speaking as the maintainer of GNOME on Arch, please remove packagekit from the `gnome` profile.

It's split out from the rest of gnome-software for a reason - packagekit is utterly unsupported. I've never seen it work reliably and it breaks our assumptions about user attention during upgrades.

dylanmtaylor:
You would know better than anyone else if it's okay to ship this package. https://archlinux.org/packages/extra/x86_64/gnome-software-packagekit-plugin/ I see you maintain it. It's a shame that without this the GNOME Software app doesn't work at all though, and users get the impression that their system is broken.

heftig:
I'd rather our users not break their systems by using packagekit.

dylanmtaylor:
> I'd rather our users not break their systems by using packagekit.

I had no idea it was that bad on Arch Linux. I will make a pull request to remove it and do the same for KDE as well

dylanmtaylor:
By the way, is there an upstream bug with the packagekit developers tracking issues with Arch systems breaking or something?

Armadillidiid:
> Handling system packages via packagekit is just fundamentally incompatible with our high-maintenance rolling release distro, where any update might leave the system in an unbootable or otherwise unusable state if the user does not take care reading pacman's logs or merging pacnew files before rebooting.

@heftig Is this still true today? I'm working on a tool that involves PackageKit and I'm wondering whether to disable the ALPM backend.

heftig:
Yes, nothing fundamentally changed. Though note this isn't about ALPM itself, just how Arch Linux uses it. It is possible to make a distro using ALPM that avoids the pitfalls that make unattended upgrades dangerous.

PR #1381, opened 2022-07-16
Add Qt5 bindings for PackageKit for Discover to work properly

Kame03:
Add packagekit-qt5 dependency to fix discover from not working, because plasma-meta package comes with discover which won't work on arch Linux without packagekit-qt5 dependency

this pull request is related to this issue #1379

### DIFF
--- a/profiles/kde.py
+++ b/profiles/kde.py
@@ -12,7 +12,8 @@ __packages__ = [
 	"ark",
 	"sddm",
 	"plasma-wayland-session",
-	"egl-wayland"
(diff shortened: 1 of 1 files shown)


doubletimedan:
#1321 The same reason is for kde

dylanmtaylor:
We had this previously and it was removed in #1322.

Kame03:
I think that the solution to this would be making a Yes or No question during installation process asking users weather or not they want this to be included.

Issue #1386, opened 2022-07-20
Crashes if `additional-repositories` is omitted from config


wpaivine2:
When running the archinstall script from the command line, if `additional-repositories` is omitted from the config, this causes a crash.

Config files:

<details>
<summary>
user_configuration.json
</summary>

```
{
    "audio": "pulseaudio",
    "config_version": "2.5.0",
    "debug": false,
    "gfx_driver": "VMware / VirtualBox (open-source)",
    "harddrives": [
        "/dev/sda"
    ],
    "mount_point": null,
    "nic": {
        "dhcp": true,
        "dns": null,
        "gateway": null,
        "iface": null,
        "ip": null,
        "type": "iso"
    },
    "plugin": null,
    "profile": {
        "path": "/usr/lib/python3.10/site-packages/archinstall/profiles/xorg.py"
    },
    "script": "guided",
    "silent": true,
    "version": "2.5.0"
}
```

</details>

<details>
<summary>
user_disk_layout.json
</summary>

```
{
    "/dev/sda": {
        "partitions": [
            {
                "boot": true,
                "encrypted": false,
                "filesystem": {
                    "format": "fat32"
                },
                "mountpoint": "/boot",
                "size": "512MiB",
                "start": "1MiB",
                "type": "primary",
                "wipe": true
            },
            {
                "encrypted": false,
                "filesystem": {
                    "format": "ext4",
                    "mount_options": []
                },
                "mountpoint": "/",
                "size": "100%",
                "start": "513MiB",
                "type": "primary",
                "wipe": true
            }
        ],
        "wipe": true
    }
}
```

</details>

<details>
<summary>
user_credentials.json
</summary>

```
{
    "!users": [
        {
            "!password": "examplefakepassword",
            "sudo": true,
            "username": "exampleusername"
        }
    ]
}
```

</details>

Installation command:

```
archinstall --silent --config user_configuration.json --creds user_credentials.json --disk_layouts user_disk_layout.json
```

Error:

```
Traceback (most recent call last):
  File "/usr/bin/archinstall", line 8, in <module>
    sys.exit(run_as_a_module())
  File "/usr/lib/python3.10/site-packages/archinstall/__init__.py", line 281, in run_as_a_module
    script.execute()
  File "/usr/lib/python3.10/site-packages/archinstall/lib/profiles.py", line 195, in execute
    self.spec.loader.exec_module(sys.modules[self.namespace])
[...]

PR #1391, opened 2022-07-21
Fix check for additional repositories

codefiles:
- This fixes issue: #1386

## PR Description:

Fix crash that occurs when `additional-repositories` is omitted from a [configuration file](https://github.com/archlinux/archinstall/blob/master/examples/config-sample.json).

Fix related typos in logging for when multilib flag is not set.

## Tests and Checks
- [ ] I have tested the code!


Torxed:
Thanks for fixing this!

PR #1397, opened 2022-07-30
Option for Parallel Downloads

abhay-mohandas:
# Describe your PR

Adds the option to enable parallel downloads during installation.

[Note: This method directly edits the pacman.conf file to enable the parallel downloads]

### DIFF
--- a/archinstall/lib/menu/global_menu.py
+++ b/archinstall/lib/menu/global_menu.py
@@ -149,8 +149,8 @@ class GlobalMenu(GeneralMenu):
 			Selector(
 				_('Parallel Downloads'),
 				add_number_of_parrallel_downloads,
-				display_func = lambda x: x if x else '0',
-				default= None
+				display_func=lambda x: x if x else '0',
+				default=None
 			)
 
 		self._menu_options['kernels'] = \
--- a/archinstall/lib/menu/global_menu.py
+++ b/archinstall/lib/menu/global_menu.py
@@ -143,6 +143,15 @@ class GlobalMenu(GeneralMenu):
 				display_func=lambda x: x if x else 'None',
 				default=None
 			)
+
+		self._menu_options['parallel downloads'] = \
+			Selector(
+				_('Parallel Downloads')
+			)
+
+
+
+
 		self._menu_options['kernels'] = \
 			Selector(
 				_('Kernels'),
--- a/archinstall/lib/menu/global_menu.py
+++ b/archinstall/lib/menu/global_menu.py
@@ -149,6 +149,7 @@ class GlobalMenu(GeneralMenu):
 			Selector(
 				_('Parallel Downloads'),
 				add_number_of_parrallel_downloads,
+				display_func = lambda x: x if x else 'None',
 				default= 0
 			)
 
--- a/archinstall/lib/menu/global_menu.py
+++ b/archinstall/lib/menu/global_menu.py
@@ -32,6 +32,7 @@ from ..user_interaction import select_encrypted_partitions
 from ..user_interaction import select_harddrives
 from ..user_interaction import select_profile
 from ..user_interaction import select_additional_repositories
+from ..user_interaction import add_number_of_parrallel_downloads
 from ..models.users import User
 from ..user_interaction.partitioning_conf import current_partition_layout
 from ..output import FormattedOutput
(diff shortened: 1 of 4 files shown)


dylanmtaylor:
It might make more sense to hide this under the advanced flag so that users are not overwhelmed with too many options

abhay-mohandas:
> It might make more sense to hide this under the advanced flag so that users are not overwhelmed with too many options

Oh I didn't know archinstall had an advanced option. I was only aware of guided and minimal scripts and that's why I edited the guided script.
Also I personally feel like it is a reasonable option to be included in the guided script. It isn't a very confusing option and almost everyone prefers to enable parallel downloads to speed up their install process

svartkanin:
It has an advanced flag but I don't think it's a good way of handling such features. People who stumble upon archinstall will probably not look into the possible cmd flags it has, but rather expect to be guided via the "gui".

My suggestion would be to have a dedicated "Advanced" submenu option with such things listed, but It will then probably be random things under there. At least they could be found more explicitly

abhay-mohandas:
> It has an advanced flag but I don't think it's a good way of handling such features. People who stumble upon archinstall will probably not look into the possible cmd flags it has, but rather expect to be guided via the "gui".

Yeah even I didn't know about it. How can I enable advanced flag? I tried `archinstall --script` but it only had minimal and guided options. And I wasn't able to find anything online related to the advanced flag.

> My suggestion would be to have a dedicated "Advanced" submenu option with such things listed, but It will then probably be random things under there. At least they could be found more explicitly

That does make sense. Maybe even have a menu that lets the user to pick from guided, minimal and advanced option, with guided being the recommended option.

svartkanin:
I'll add this when the current open PRs start getting merged, it's a bit difficult atm because there's quite some conflicts having to be resolved

dylanmtaylor:
> I'll add this when the current open PRs start getting merged, it's a bit difficult atm because there's quite some conflicts having to be resolved

Some of the stuff under the advanced flag simply shouldn't be used in any case other than an experimental install. You can do things like install the system to an NTFS root partition with it, which is fun for playing around in the VM, but it's not something that 99% of users should ever even consider. Hence the advanced flag.

Torxed:
So I cleaned up this a bit, using `with open()` instead of `fh = open()`.
I also hid the option behind `--advanced`, and it's not because it's advanced in any way.. The functionality is pretty straight forward. But it would put quite a bit of extra strain on the mirrors, something I'm not sure upstream is to happy about.

The functionality is there, we can enable it.. and we'll have to come back and discuss how much of an impact this would have on overall mirrors if we made it "available" :)

abhay-mohandas:
> So I cleaned up this a bit, using `with open()` instead of `fh = open()`. I also hid the option behind `--advanced`, and it's not because it's advanced in any way.. The functionality is pretty straight forward. But it would put quite a bit of extra strain on the mirrors, something I'm not sure upstream is to happy about.
>
> The functionality is there, we can enable it.. and we'll have to come back and discuss how much of an impact this would have on overall mirrors if we made it "available" :)

It looks great now! Thank you for cleaning it up.
However there are a few things I did noticed about it.
1. Minimum value of parallel downloads needs to be changed to 1 instead of 0, otherwise pacman refuses to run at all.
2. New program was unable to enter `while input_number is None:` loop once a value is assigned. So the user can't change the value once it is assigned.

I am working on these issues and will upload them as soon as possible :D

abhay-mohandas:
> Great catch! It's fine to patch over the existing code if that helps, instead of reverting it : )

This is my first ever pull request so I'm not sure how to patch over the existing one... Instead i think i accidentally opened a new pull request... I guess...
I'm still figuring it out

Torxed:
The most efficient way is (in your repo, locally on your harddrive, do):
```bash
$ cd archinstal
$ git remote add upstream https://github.com/archlinux/archinstall.git
$ git pull upstream master
```
Then edit the code as you please locally, push the new changes as you would.
Then open a new PR with those changes in your repo :)

This will first sync upstream (our repo) into your repo.
Then you modify your stuff and submit a PR, that way you simply patch ontop of our current code base :)

abhay-mohandas:
Reopen Pull Request
> The most efficient way is (in your repo, locally on your harddrive, do):
> 
> ```shell
> $ cd archinstal
> $ git remote add upstream https://github.com/archlinux/archinstall.git
> $ git pull upstream master
> ```
> 
> Then edit the code as you please locally, push the new changes as you would. Then open a new PR with those changes in your repo :)
> 
> This will first sync upstream (our repo) into your repo. Then you modify your stuff and submit a PR, that way you simply patch ontop of our current code base :)
> 
> (Ideally you would want to create a branch, instead of working in your own `master`, but save those experiments to later I guess)

Thank you for the advice :D

Okay so I think that's what I did and i submitted a new pull request #1403 
I hope that it should work since it does show all my previous commits and the changes you made along with it.

PR #1403, opened 2022-08-02
Fixes for known issues in the Parallel Download option

abhay-mohandas:
- Known issue(s): 

> 1. Minimum value of parallel downloads needs to be changed to 1 instead of 0, otherwise pacman refuses to run at all.
> 2. New program was unable to enter `while input_number is None:` loop once a value is assigned. So the user can't change the value once it is assigned.

## PR Description:
Both the above issues has been resolved.
For fixing the second issue, the loop was changed back to `while True:` 

Additionally, a limit is set for the maximum number of parallel downloads

## Tests and Checks


Torxed:
I'm wondering if we shouldn't allow `0` after all, but if zero is given we leave the function commented out. That way the user can change their mind and turn it off. I get that `1` does the exact same thing as commenting it out, but linguistically I would expect `1` to turn on one parallel download, to my otherwise single download. And 0 would turn it off.

This is nitpicking at it's finest, but what do you think?

abhay-mohandas:
> I'm wondering if we shouldn't allow `0` after all, but if zero is given we leave the function commented out. That way the user can change their mind and turn it off. I get that `1` does the exact same thing as commenting it out, but linguistically I would expect `1` to turn on one parallel download, to my otherwise single download. And 0 would turn it off.
> 
> This is nitpicking at it's finest, but what do you think?
> 
> PR looks ok btw :)

Yeah I do agree with your take. So the new changes are:

-  0 disables this option by commenting the option. So only one download occur at a time, hence is the default.
-  Any other number between 1 to maximum value is treated as input+1. So atleast two downloads occur at a time.

I've also added few lines of explanation incase anyone is confused by the choices.

Torxed:
Nice touch, thank you for this contribution!

abhay-mohandas:
You're welcome : )

PR #1406, opened 2022-08-06
Add command line option to print the program version

strboul:
With that, it prints:

```
archinstall --version
# archinstall 2.5.0
```

GNU Coding Standards recommends:

> The standard `--version` option should direct the program to print information about its name, version, origin and legal status, all on standard output, and then exit successfully. Other options and arguments should be ignored once this is seen, and the program should not perform its normal function. 

https://www.gnu.org/prep/standards/standards.html#g_t_002d_002dversion

:star2:


Torxed:
Nice tough, thanks for putting it in there :)

Issue #1411, opened 2022-08-09
btrfs-progs missing during mkinitcpio generation


Torxed:
Strictly speaking this is an error that results in exit code 256 (1), but the image will still be generated.
But it causes archinstall to react on the exit code and aborts the installation.

Either we have to be more tolerant towards `mkinitcpio` and allow "bad" exit codes and check for `Image generation successful`, or we have to include `btrfs-progs` earlier in the steps.

![screenshot](https://user-images.githubusercontent.com/861439/183652260-462c5755-27e8-4f70-9f2d-adedf114a471.png)

Torxed:
This logic is broken: https://github.com/archlinux/archinstall/blob/94d611d22f178474d1f3a491cfe60e99bd3293b3/archinstall/lib/installer.py#L652-L656

Because at the time of: https://github.com/archlinux/archinstall/blob/94d611d22f178474d1f3a491cfe60e99bd3293b3/archinstall/lib/installer.py#L716

The following packages are installed: `base base-devel linux-firmware linux`
![screenshot](https://user-images.githubusercontent.com/861439/183656517-d53510ba-a46b-4cfb-ab3f-c64ed17db79d.png)

Which means this will "fail": https://github.com/archlinux/archinstall/blob/94d611d22f178474d1f3a491cfe60e99bd3293b3/archinstall/lib/installer.py#L748

Torxed:
Trying to think of the best approach forward here.
Because technically those paths are on that partition, so that's not a lie.
But that partition doesn't have a mountpoint, because the subvolumes do.

Torxed:
Easy tweak, I check if `installation.target` is inside the `Partition.mountpoints` instead of `.mountpoint` as we do recieve all the mountpoints from that function, including subvolumes.

Torxed:
All in all, it was an error in detecting `partitions` in use. Because of `pathlib.Path` being introduced, and `MapperDev` missing a `.mountpoints` function.

PR #1412, opened 2022-08-09
Converted a path to pathlib.Path

Torxed:
- This fix issue: #1411 

## PR Description:

In #1333 we cleaned up the partition handling a bit and introduced more strict types, among those were `pathlib.Path()` on partition mountpoints.

We forgot to add a conversion for `findmnt` paths when comparing against a list of partitions paths.
The error was here (kinda): https://github.com/archlinux/archinstall/blob/3842f1eb5644851ad3af84d84e704c0b33b2b204/archinstall/lib/disk/partition.py#L170

This caused an issue in comparison here: https://github.com/archlinux/archinstall/blob/3842f1eb5644851ad3af84d84e704c0b33b2b204/archinstall/lib/disk/helpers.py#L410-L411

## Tests and Checks

### DIFF
--- a/archinstall/lib/disk/helpers.py
+++ b/archinstall/lib/disk/helpers.py
@@ -403,6 +403,10 @@ def get_partitions_in_use(mountpoint :str) -> Dict[str, Any]:
 	log(f'Filtering available mounts {block_devices_mountpoints} to those under {mountpoint}', level=logging.DEBUG)
 
 	for mountpoint in list(get_all_targets(output['filesystems']).keys()):
+		# Since all_blockdevices() returns PosixPath objects, we need to convert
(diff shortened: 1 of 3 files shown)


Torxed:
This causes `Could not detect root (None) or boot (Partition(path=/dev/sda1...))...`.
Investigating.

Torxed:
Works for `btrfs` but not for `ext4` at this stage.

Torxed:
Fixed, by implementing a `.mountpoints` on `MapperDev` that behaves the same way as `Partition.mountpoints`. Using `pathlib.Path` throughout. There's probably some edge cases lurking here as changes to `Partition` logic always causes some durst to fly up.

PR #1423, opened 2022-08-14
Rework system locale

codefiles:
## PR Description:

- Update `list_locales()` to use `/usr/share/i18n/SUPPORTED`.
- Verify locales are supported by checking against `/usr/share/i18n/SUPPORTED`.
- Uncomment entries in `/etc/locale.gen` or append if an entry is not found to uncomment.
- Use `localedef` for generating locales instead of `locale-gen`.
- Do not write to `/etc/locale.gen` when it does not need to be modified.
- Do not generate locales when the given locales are already generated.
- Do not set the system locale when it is already set to the given locale.
- Strip encodings from options in the guided installer `Locale language` menu leaving only languages since encodings are selected in the `Locale encoding` menu.
- Add support for generating multiple locales (to add to the guided installer in the future).

## Tests and Checks
- [ ] I have tested the code!

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -12,7 +12,7 @@ from typing import Union, Dict, Any, List, Optional, Iterator, Mapping, TYPE_CHE
 from .disk import get_partitions_in_use, Partition
 from .general import SysCommand, generate_password
 from .hardware import has_uefi, is_vm, cpu_vendor
-from .locale_helpers import verify_keyboard_layout, verify_x11_keyboard_layout
+from .locale_helpers import Locale, LocaleUtils, verify_keyboard_layout, verify_x11_keyboard_layout
 from .disk.helpers import findmnt
 from .mirrors import use_mirrors
 from .plugins import plugins
@@ -440,32 +440,13 @@ class Installer:
 			fh.write(hostname + '\n')
 
 	def set_locale(self, locale :str, encoding :str = 'UTF-8', *args :str, **kwargs :str) -> bool:
-		if not len(locale):
-			return True
+		try:
+			locales = [Locale(locale, encoding)]
+		except ValueError as error:
+			self.log(f'Invalid locale: {error}', level=logging.ERROR, fg='red')
+			return False
 
-		modifier = ''
-
-		# This is a temporary patch to fix #1200
-		if '.' in locale:
-			locale, potential_encoding = locale.split('.', 1)
-
-			# Override encoding if encoding is set to the default parameter
-			# and the "found" encoding differs.
-			if encoding == 'UTF-8' and encoding != potential_encoding:
-				encoding = potential_encoding
-
-		# Make sure we extract the modifier, that way we can put it in if needed.
-		if '@' in locale:
-			locale, modifier = locale.split('@', 1)
-			modifier = f"@{modifier}"
-		# - End patch
-
-		with open(f'{self.target}/etc/locale.gen', 'a') as fh:
-			fh.write(f'{locale}.{encoding}{modifier} {encoding}\n')
-		with open(f'{self.target}/etc/locale.conf', 'w') as fh:
-			fh.write(f'LANG={locale}.{encoding}{modifier}\n')
-
-		return True if SysCommand(f'/usr/bin/arch-chroot {self.target} locale-gen').exit_code == 0 else False
+		return LocaleUtils(locales, self.target).run()
 
 	def set_timezone(self, zone :str, *args :str, **kwargs :str) -> bool:
 		if not zone:
--- a/archinstall/lib/locale_helpers.py
+++ b/archinstall/lib/locale_helpers.py
@@ -1,32 +1,394 @@
 import logging
+import re
+import pathlib
 from typing import Iterator, List, Callable
 
-from .exceptions import ServiceException
+from .exceptions import ServiceException, SysCallError
 from .general import SysCommand
 from .output import log
 from .storage import storage
 
-def list_keyboard_languages() -> Iterator[str]:
-	for line in SysCommand("localectl --no-pager list-keymaps", environment_vars={'SYSTEMD_COLORS': '0'}):
-		yield line.decode('UTF-8').strip()
+
+class Locale():
+	def __init__(self, name: str, encoding: str = 'UTF-8'):
+		"""
+		A locale composed from a name and encoding.
+
+		:param name: A name that represents a locale of the form language_territory[.encoding][@modifier]. An encoding within the name will override the encoding parameter.
+		:type name: str
+		:param encoding: The encoding of the locale; if omitted defaults to UTF-8.
+		:type encoding: str
+		"""
+		if not len(name):
+			raise ValueError('Locale name is an empty string')
+
+		if name.count('.') > 1:
+			raise ValueError(f"Locale name '{name}' contains more than one '.'")
+
+		if name.count('@') > 1:
+			raise ValueError(f"Locale name '{name}' contains more than one '@'")
+
+		self.name = name
+		self.encoding = encoding
+
+		# Extract the modifier if found.
+		if '@' in name:
+			name, potential_modifier = name.split('@')
+
+			# Correct the name if it has the encoding and modifier in the wrong order.
+			if '.' in potential_modifier:
+				potential_modifier, potential_encoding = potential_modifier.split('.')
+				name = f'{name}.{potential_encoding}'
+
+			self.modifier = potential_modifier
+		else:
+			self.modifier = None
+
+		if '.' in name:
+			self.language, potential_encoding = name.split('.')
+
+			# Override encoding if name contains an encoding that differs.
+			if encoding != potential_encoding:
+				self.encoding = potential_encoding
+		else:
+			self.language = name
+
+		if not len(self.encoding):
+			raise ValueError('Locale encoding is an empty string')
+
+		if not len(self.language):
+			raise ValueError('Locale language is an empty string')
+
+		self.str = f'{self.language}.{self.encoding}'
+
+		if self.modifier is not None:
+			self.str += '@' + self.modifier
+
+	def __str__(self) -> str:
+		return self.str
+
+	def __eq__(self, other) -> bool:
+		return self.str == other.str
+
+	def __lt__(self, other) -> bool:
+		return self.str < other.str
 
 
-def list_locales() -> List[str]:
-	with open('/etc/locale.gen', 'r') as fp:
-		locales = []
-		# before the list of locales begins there's an empty line with a '#' in front
-		# so we'll collect the localels from bottom up and halt when we're donw
-		entries = fp.readlines()
-		entries.reverse()
+class LocaleUtils():
+	def __init__(self, locales: List[Locale] = [], target: str = ''):
+		"""
+		Get locale information, generate locales, and set the system locale.
+		An instance can contain a list of locales and the target location.
 
+		:param locales: A list of locales, the first locale is intended as the system locale.
+		:type locales: List[Locale]
+		:param target: An installation mount point, if omitted default to the local system.
+		:type target: str
+		"""
+		self.locales = locales
+		self.target = target
+		self.locale_gen = pathlib.Path(f'{target}/etc/locale.gen')
+		self.locale_conf = pathlib.Path(f'{target}/etc/locale.conf')
+
+	def verify_locales(self) -> bool:
+		"""
+		Check if the locales match supported locales.
+		If a match is found update the name of the locale to the name of the matching entry if they differ.
+
+		:return: If matched return True else False.
+		:rtype: bool
+		"""
+		supported = []
+
+		for locale in list_locales(self.target):
+			supported.append(Locale(*locale.split()))
+
+		found_all = True
+
+		for locale in self.locales:
+			found = False
+			for entry in supported:
+				if locale == entry:
+					if locale.name != entry.name:
+						locale.name = entry.name
+					found = True
+					break
+
+			if not found:
+				found_all = False
+				log(f'Unsupported locale: {locale}', fg='red', level=logging.ERROR)
+
+		return found_all
+
+	def list_uncommented(self) -> List[str]:
+		"""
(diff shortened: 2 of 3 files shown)


Torxed:
This looks great! It's a breaking change as it will break all existing configurations (which is fine), so I'd like to propose that we postpone merging this in until next release, that way we can add a warning to the current functionalities that they're being deprecated.

Torxed:
This looks great! It's a breaking change as it will break all existing configurations (which is fine), so I'd like to propose that we postpone merging this in until next release, that way we can add a warning to the current functionalities that they're being deprecated.

Torxed:
This looks great! It's a breaking change as it will break all existing configurations (which is fine), so I'd like to propose that we postpone merging this (when you feel ready with the PR) until next release, that way we can add a warning to the current functionalities that they're being deprecated.

svartkanin:
In previous changes like this we did support backwards compatibility, for example superuser configuration. 
I think it shouldn't be too difficult to get that in here, essentially when parsing the config we need to check if there's an old version entry present and if so convert it to the new one.

Torxed:
> However, i am very much in favor of providing a deprecated warning (also introduced in users) to inform that this will go away. This way we'll clean up this code and don't have to maintain it.

Yea this would be optimal, since we have about 1-2 months between releases any way I think having a deprecation warning is more than enough. We could do the same for the superuser configuration tbh.

codefiles:
I've changed the scope of the pull request to focus on non-breaking changes. I will follow it up with a pull request for the unification of locale language and locale encoding into locale if that is a desired change.

Torxed:
`localedef` was a new one for me. Looks really good. I'll try to run a few tests in a day or so and merge this in :)

codefiles:
@Torxed considering what you mentioned here: https://github.com/archlinux/archinstall/issues/1200#issuecomment-1130564952, is there a reason locale and hostname are not parameters to `minimal_installation()`?
```diff
-def minimal_installation(self, testing=False, multilib=False) -> bool:
+def minimal_installation(self, testing :bool = False, multilib :bool = False, hostname :str = 'archlinux', locale :Tuple[str, str] = ('en_US', 'UTF-8')) -> bool:
```

Torxed:
@codefiles There's really no reason for it no.

*Long answer:*
`minimal_installation()` didn't even set those before, it only did the **bare** minimum to get something up and running. Setting a hostname wasn't considered a requirement for that. Equally so, the locale would default to something so there was no **need** to set one.

As the project grew, those are sane defaults in almost any minimal installation so the function evolved.
And since the values would get overridden by user choices later in `guided.py` we just set two sane default values in case the user skipped it. But making them into a parameter and sending it in and setting it there and then - makes a lot of sense.

codefiles:
@Torxed Thanks for the insight. I'm just getting familiar with the project. I wanted to make sure I wasn't overlooking anything because it seems like an easy fix that would be a non-breaking change.

I created a locale class in `locale_helpers.py`. Here is what it looks like when used on a local system:

```
>>> from archinstall.lib.locale_helpers import Locale
>>> Locale('en_US').set()
Setting locale: en_US.UTF-8
- Already generated
- Already set in /etc/locale.conf
- Already uncommented in /etc/locale.gen
Done
True
>>> Locale('en_US', 'ISO-8859-1').set()
Setting locale: en_US.ISO-8859-1
- Generating locale...
- Locale generated
- Updated /etc/locale.conf
- Updated /etc/locale.gen
Done
True
>>> Locale('sv_FI@euro').set()
Setting locale: sv_FI.UTF-8@euro
Entry 'sv_FI@euro UTF-8' not found in /etc/locale.gen
False
>>> Locale('sv_FI@euro', 'ISO-8859-15').set()
Setting locale: sv_FI.ISO-8859-15@euro
- Generating locale...
- Locale generated
- Updated /etc/locale.conf
- Updated /etc/locale.gen
Done
True
>>> Locale('en_US', 'UTF-8').set()
Setting locale: en_US.UTF-8
- Generating locale...
- Locale generated
- Updated /etc/locale.conf
- Updated /etc/locale.gen
Done
True
```

Torxed:
Great modification, very clean and separates the functionality well from the installer code.
I'll run a few tests in the afternoon (local time) and merge this later today if you feel ready with it :)

Torxed:
screenshot

Saving one of the odd results from my testing that stood out.
Not sure I entered a valid combination hehe. `locale-gen` accepted it, but it's not one of the valid pre-defined combinations from `/etc/locale.conf`.

codefiles:
The `locale-gen` validates a locale by checking that the _language_\__territory_ portion of the locale matches one found in the directory `/usr/share/i18n/locales`.

https://github.com/archlinux/svntogit-packages/blob/128eb3280665c0f227e7c514da1f7060efa5873f/trunk/locale-gen#L6

https://github.com/archlinux/svntogit-packages/blob/128eb3280665c0f227e7c514da1f7060efa5873f/trunk/locale-gen#L37

That validation step is being carried out in addition to validating against `/etc/locale.gen`. The idea behind this is that 

https://github.com/archlinux/svntogit-packages/blob/128eb3280665c0f227e7c514da1f7060efa5873f/trunk/PKGBUILD#L183-L186

https://sourceware.org/git/?p=glibc.git;a=blob;f=localedata/SUPPORTED;h=c8b63cc2fe2b4547f2fb1bff6193da68d70bd563;hb=HEAD

I'm not sure either of these would be a good idea but it allows it:

```
Generating locales...
  en_US.ISO-8859-15... done
  en_US.GB18030... done
Generation complete.
```

codefiles:
Not the right place for this but these Arch Linux glibc package changes might be improvements:
- Provide [localedata/SUPPORTED](https://sourceware.org/git/?p=glibc.git;a=blob;f=localedata/SUPPORTED;h=c8b63cc2fe2b4547f2fb1bff6193da68d70bd563;hb=HEAD) as `/usr/share/i18n/SUPPORTED` since `/etc/locale.gen` is intended to be edited.
- Remove `#  en_US.UTF-8 UTF-8` line from examples in [locale.gen.txt](https://github.com/archlinux/svntogit-packages/blob/128eb3280665c0f227e7c514da1f7060efa5873f/trunk/locale.gen.txt#L14) since that line, instead of the later `#en_US.UTF-8 UTF-8`, can become uncommented in `/etc/locale.gen` with the use of `localectl set-locale en_US.UTF-8` or remove all examples since they are superfluous.

codefiles:
Here is what this pull request looks like working with multiple locales:
```
>>> from archinstall.lib.locale_helpers import Locale, LocaleUtils
>>> locale_names = ['en_US.UTF-8', 'de_DE.UTF-8', 'de_DE.ISO-8859-15@euro']
>>> locales = []
>>> for locale_name in locale_names:
...     locales.append(Locale(locale_name))
...
>>> LocaleUtils(locales).run()
Uncommented entries in locale-gen configuration file
  de_DE.UTF-8 UTF-8
  de_DE@euro ISO-8859-15
  en_US.UTF-8 UTF-8
Generating locales...
  de_DE.UTF-8...
  de_DE.ISO-8859-15@euro...
  en_US.UTF-8...
Generation complete.
System locale set to en_US.UTF-8
True
```
This is ready to be reviewed and tested. More to come in future pull requests as mentioned here https://github.com/archlinux/archinstall/issues/1435#issuecomment-1236491516.

Torxed:
Great work on this, I'll pass along the message about glibc and the examples file.
I'll test this most likely next week, as this week is a bit busy and scheduled.

codefiles:
I appreciate you passing along the message about the Arch Linux glibc package. Would you please link to the comment when passing the information along? I have just edited that comment for additional clarity.

codefiles:
Another message regarding the Arch Linux glibc package:
- Debian [0][1] and Gentoo [2] process `localedata/SUPPORTED` to generate `/usr/share/i18n/SUPPORTD`. Use the following sed command on `localedata/SUPPORTED` to generate `/usr/share/i18n/SUPPORTED`.

```bash
sed -e '1,3d' -e 's|/| |g' -e 's| \\||g'
```

- Modify the line with the sed command for `/etc/locale.gen` to remove trailing whitespace,

```diff
-  sed -e '1,3d' -e 's|/| |g' -e 's|\\| |g' -e 's|^|#|g' \
+  sed -e '1,3d' -e 's|/| |g' -e 's| \\||g' -e 's|^|#|g' \
```

or use the following sed command on the generated `/usr/share/i18n/SUPPORTED` rather than `localedata/SUPPORTED`.

```
sed 's|^|#|g'
```

[0] https://salsa.debian.org/glibc-team/glibc/-/blob/sid/debian/rules.d/build.mk#L240-244
[1] https://salsa.debian.org/glibc-team/glibc/-/blob/sid/debian/generate-supported.mk
[2] https://gitweb.gentoo.org/repo/gentoo.git/tree/sys-libs/glibc/glibc-9999.ebuild#n1401

Torxed:
> `sed -e '1,3d' -e 's|/| |g' -e 's| \\||g'`

Isn't this a bit of a workaround for an upstream glibc issue tho? Or did I misunderstood why we remove trailing whitespaces?

codefiles:
> > `sed -e '1,3d' -e 's|/| |g' -e 's| \\||g'`
> 
> Isn't this a bit of a workaround for an upstream glibc issue tho? Or did I misunderstood why we remove trailing whitespaces?

You are misunderstanding why I'm requesting trailing whitespace to be removed. In the glibc PKGBUILD a sed command is being run on `localedata/SUPPORTED` to generate the `/etc/locale.gen` file. The command in the PKGBUILD is `sed -e '1,3d' -e 's|/| |g' -e 's|\\| |g' -e 's|^|#|g'` and the `-e 's|\\| |g'` portion replaces the backslash at the end of a line containing a locale such as `aa_DJ.UTF-8/UTF-8 \` with a space. Changing the `-e 's|\\| |g'` portion to `-e 's| \\||g'` removes the space at the end before the backslash and does not add a space.

The following demonstrates that it makes a difference:

```
$ cat localedata/SUPPORTED
# This file names the currently supported and somewhat tested locales.
# If you have any additions please file a glibc bug report.
SUPPORTED-LOCALES=\
aa_DJ.UTF-8/UTF-8 \
aa_DJ/ISO-8859-1 \
...
$ sed -e '1,3d' -e 's|/| |g' -e 's|\\| |g' localedata/SUPPORTED
aa_DJ.UTF-8 UTF-8
aa_DJ ISO-8859-1
aa_ER UTF-8
aa_ER@saaho UTF-8
aa_ET UTF-8
...
$ sed -e '1,3d' -e 's|/| |g' -e 's| \\||g' localedata/SUPPORTED
aa_DJ.UTF-8 UTF-8
aa_DJ ISO-8859-1
aa_ER UTF-8
aa_ER@saaho UTF-8
aa_ET UTF-8
...
$ diff <(!-2) <(!!)
diff -sq <(sed -e '1,3d' -e 's|/| |g' -e 's|\\| |g' localedata/SUPPORTED) <(sed -e '1,3d' -e 's|/| |g' -e 's| \\||g' localedata/SUPPORTED)
Files /dev/fd/63 and /dev/fd/62 differ
```

codefiles:
This may also clear things up:

```
$ curl -sO "https://salsa.debian.org/glibc-team/glibc/-/raw/sid/debian/generate-supported.mk"
$ make -f generate-supported.mk IN=localedata/SUPPORTED OUT=SUPPORTED.tmp > /dev/null
$ diff -sq SUPPORTED.tmp <(sed -e '1,3d' -e 's|/| |g' -e 's|\\| |g' localedata/SUPPORTED)
Files SUPPORTED.tmp and /dev/fd/63 differ
$ diff -sq SUPPORTED.tmp <(sed -e '1,3d' -e 's|/| |g' -e 's| \\||g' localedata/SUPPORTED)
Files SUPPORTED.tmp and /dev/fd/63 are identical
```

Torxed:
I see, I'll pass it along! :) Apologies for not getting it straight away, I should maybe have spent a few more minutes reading through your initial example. I was in a bit of a rush :)

codefiles:
It is ok, should be a lot clearer now. It is hard to find a middle ground between being terse and verbose when describing issues. Terse descriptions might lack clarity and verbose descriptions risk being ignored.

codefiles:
In regards to the Arch Linux glibc package, two of the bullets in my earlier comment, https://github.com/archlinux/archinstall/pull/1423#issuecomment-1238793882, have now been resolved by:
- https://github.com/archlinux/svntogit-packages/commit/d6f2e1f80af0f94e6b024552ad9a94842002356d
- https://github.com/archlinux/svntogit-packages/commit/b09829c6ca6b116f8d16db9f7de7fde72f47e758

The revised patch file [supported.diff](https://gist.github.com/codefiles/bfcb09c3e19f972685d2bd86bec3766b/8204768f1fe724bce630d7b1f43ddabff4880a4b) targets the second bullet from my earlier comment, https://github.com/archlinux/archinstall/pull/1423#issuecomment-1238793882, that is not yet addressed. If implemented it would:
- Eliminate the whitespace at the end of entries in `/etc/locale.gen`.
- Deduplicate the sed commands that are working on the glibc file `localedata/SUPPORTED` to create `/etc/locale.gen` and `/usr/share/i18n/SUPPORTED`.
- Reduce maintenance of the PKGBUILD should the glibc file `localedata/SUPPORTED` change in a way that required modification of the sed commands.

codefiles:
The latest commit requires the next release of the glibc package for https://github.com/archlinux/svntogit-packages/commit/b09829c6ca6b116f8d16db9f7de7fde72f47e758.

Torxed:
Perfect, I'll merge this as soon as https://archlinux.org/packages/core/x86_64/glibc/ is updated (the upstream commit is newer than latest package release)

codefiles:
The glibc package has been updated.

Be aware that if this is merged the current ISO (2022.10.01) will not meet the required version of glibc when using `master`.

Torxed (approved):
Changes looks promising, I understand I'm a bit late approving this. Will this interfere with the recent changes to the locale perhaps?

codefiles:
Let's not merge this yet. I feel that it is stale now so I am changing this to a draft till I review it again.

Issue #1175, opened 2022-05-12
Empty list of services results in a systemctl error


i2:
I like to have my user config file written in a way that can act as a template and placeholder for future edits. For instance, even though I don't need to enable a service right now. I have:

```
    "services": [
        ""
    ],
```

But this results in archinstall throwing an error that:
`arch-chroot /mnt/archinstall systemctl enable exited with abnormal exit code [256]: Too few arguments.`

Which makes sense, as my list was empty. It would be nice if this can be avoided.

svartkanin:
@Torxed we can close this as it should be resolved with the new profile manger

Issue #1154, opened 2022-05-09
add menu for selecting display manager


zlianon:
It would be convenient if the desktop profiles offered the ability to choose a display manager.
Currently that is locked in to the profiles, even even tho post-configuration can be done to change this quite easily, it would be beneficial to be able to change the option before the installation is complete.

Torxed:
Could you elaborate with a description? :)

zlianon:
e.g. i3 profile forces lightdm installation, it would be good to offer alternatives

Torxed:
I updated the description as to what I think you wish for.
I also made the argument for you why this would be a good thing.

This has historically not been in scope because it can be considered post-installation configuration.
And some people would surely argue that it's good enough to have something working and tweak it.

Which is why I'll leave this issue open for a while, to see what the general consensus is so we can discuss it :)

svartkanin:
@Torxed we can close this as it should be resolved with the new profile manger

Issue #999, opened 2022-02-21
server installation: leaving blank for a minimal selection loops


j-james:
At one point in archinstall after selecting the "server" profile, you're prompted to pick which servers to install with the message: `Choose which servers to install and enable (leave blank for a minimal installation)`.

Leaving this blank and pressing enter quickly flashes an error and loops you back to the same question again.

j-james:
Also, `ctrl+c` no longer works - I'm not sure if this is because something with archinstall crashed, or if archinstall just doesn't respect it at this point in the install.

Torxed:
I'll have a look :)

svartkanin:
@Torxed we can close this as it should be resolved with the new profile manger

Issue #740, opened 2021-11-19
[Feature Request] source code generation for custom profiles and scripts.


evan0greenup:
For user who want to install massive amount of computer with similar configuration and environment,  It would be awesome to provide a interactive interface to let user input essential information, and then generate relevant source code for script and profile.

For each computer instance, user just need to insert USB drive and then boot and then run the generated script. All the process will be done automatically.

And auto, for archinstall command, it would be nice to enable user to add custom profile, those custom profile will add to the candidate list when the archinstall let user to choose profile.

Torxed:
I'm not sure I understand this request. Archinstall supports the `--config`, `--disk-layouts` and `--creds` parameters which allow you to do a complete template install.

Archinstall also supports custom profile paths in the profile question. If added to `--config`:

```json
{
    "profile": "https://remote.lan/my_profile.py",
    ...
}
```
Archinstall will source that profile and you can do whatever you want in it :)

A separate project of mine archiffline even allows you to do all this offline from a USB stick :)

Is this what you needed?

bebetoalves:
> I'm not sure I understand this request. Archinstall supports the `--config`, `--disk-layouts` and `--creds` parameters which allow you to do a complete template install.
> 
> Archinstall also supports custom profile paths in the profile question. If added to `--config`:
> 
> ```json
> {
>     "profile": "https://remote.lan/my_profile.py",
>     ...
> }
> ```
> 
> Archinstall will source that profile and you can do whatever you want in it :)
> 
> A separate project of mine archiffline even allows you to do all this offline from a USB stick :)
> 
> Is this what you needed?

Maybe he's talking about this? #566

dylanmtaylor:
I am pretty sure we already have this. --config allows you to pass in pre-populated values to guided. ArchInstall is also a python library and you can write your own installer using its methods very easily (there are examples of this).

evan0greenup:
@Torxed @bebetoalves , is it possible to pass multiple profiles to archinstall?

Torxed:
> @Torxed @bebetoalves , is it possible to pass multiple profiles to archinstall?

Not in the sense you mean, `--profile=one.py,two.py` does not work *(yet at least)*. Because we don't iterate the profile parameter in any of our usages: https://github.com/archlinux/archinstall/blob/627b6b5207beffee2ed9507933cebe780c62a11d/examples/guided.py#L152-L156

[...]

PR #1434, opened 2022-08-27
Rework profile management

svartkanin:
# Reworking the current profile mangement

With the current version there are some major challenges. It is tightly coupled with `archinstall.storage` which means that any selection in the menu will modify the storage. This makes it very difficult to reset settings and/or override them without forgetting to cleanup the ones not set. 

These major changes aim to make the profile handling more flexible and dynamic and give users the ability to programmatically generate profiles. Unfortunately, this is a breaking change and the old version of profile definitions will not work anymore.

 @Torxed I've transferred over all existing profiles in the repository, but I tried my best to support an automatic transfer for profiles that have been defined by users. Unfortunately, that wasn't feasible without having to cover dozens of edge cases, so this will inevitably a breaking change :( 

## Related issues that should be fixed with this upgrade

* Fixes https://github.com/archlinux/archinstall/issues/234
* Fixes https://github.com/archlinux/archinstall/issues/395
* Fixes https://github.com/archlinux/archinstall/issues/601
* Fixes https://github.com/archlinux/archinstall/issues/606
* Fixes https://github.com/archlinux/archinstall/issues/611
* Fixes https://github.com/archlinux/archinstall/issues/612
* Fixes https://github.com/archlinux/archinstall/issues/740
* Fixes https://github.com/archlinux/archinstall/issues/999
* Fixes https://github.com/archlinux/archinstall/issues/1154
* Fixes https://github.com/archlinux/archinstall/issues/1175

### Profiles as classes

All profiles are now defined as a class. This allows that profiles can be inherited from each other, and therefore users can build 
profiles on top of other profiles. This will be essential when creating custom profiles :)

### Programmatically create profiles

User will now have the ability to programmatically create profiles and inject them, as an example 

```
class TestProfileV2(ProfileV2):
	def __init__(self):
		super().__init__('TestMe', ProfileType.CustomType)

       @property
		return ['my_fav_packages']

	def install(self, install_session: 'Installer'):
		install_session.add_additional_packages(self.packages())

handler = ProfileHandler()
handler.add_custom_profile(TestProfileV2())
```

This also has the major benefit that no separate files have to be create anymore for profiles. 

### Multiple profiles per file

Currently there's a single profile definition per file. One can now create profiles programmatically as shown above, or specify them 
as in the current way in a separate file.  
However, now one can also specify multiple profiles per file, since the actual class definitions will be recognized as the profile.

```
my_profiles.py

class ProfileOne(ProfileV2):
    ...

class ProfileTwo(ProfileV2):
    ...
```

### Ability to install multiple desktop profiles

There is now the option to install multiple desktop profiles, similar to multiple servers that already exists 

### Display the packages to be installed on the desktop profiles

image

### Select which greeter should be installed

Each desktop profile can define a default greeter. When selecting one or multiple desktop profiles, the user will get the option
to chose which greeter should be installed with these profiles. As the default there will be a recommendation of the 
default greeter (if only one or if all have the same default).

image

### Create custom profiles in the menu

In the profile menu, there's also the option to create custom profiles

image

image

They will also be exported as configuration settings. If they have been created and set to disabled, and another profile (e.g. Desktop) is set then they will still be exported, this case any definitions won't be lost.

image

### New option in the `Menu` for "Back"

In some Menus it makes sense to display a visual "Back" option. This option will behave the same way as ESC essentially does but it is more intuitive.

image

## Breaking changes

* Old profiles will not work anymore so they have to moved over to the new class format manually (all existing profiles in the repository have been moved over already) 

### Schema changes

The schema for the profile configuration has changed. Also the gfx driver configuration has been moved inside the profile configuration since it is never used as a standalone but only in combination with the respective profiles selected

```
        "profile": {
            "main":  {
                "description": "Main top level profile ",
                "type": "string",
                "enum": [
                    "desktop",
                    "minimal",
                    "server",
                    "xorg",
                ]
            },
            "details": {
                "description": "Specific profile to be installed based on the 'main' selection; these profiles are present in profiles_v2/, use the name of a profile to install it (case insensitive)",
                "type": "string",
                "enum": [
                    "awesome",
                    "bspwm",
                    "budgie",
                    "cinnamon",
                    "cutefish",
                    "deepin",
                    "desktop",
                    "enlightenment",
                    "gnome",
                    "i3-wm",
                    "i3-gasp",
                    "kde",
                    "lxqt",
                    "mate",
                    "sway",
                    "xfce4",
                    "qtile",
                    "cockpit",
                    "docker",
                    "httpd",
                    "lighttpd",
                    "mariadb",
                    "nginx",
                    "postgresql",
                    "sshd",
                    "tomcat"
                ]
            },
            "custom": {
                "description": "Specific profile definitions for custom setup profiles)",
                "type": "array",
                "items": {
                    "type": "object",
                    "properties": {
                        "packages": "string",
                        "services": "string",
                        "enabled": "boolean"
                    }
                }
            },
            "gfx_driver": {
                "description": "Graphics Drivers to install if a desktop profile is used, ignored otherwise.",
                "type": "string",
                "enum": [
                    "VMware / VirtualBox (open-source)",
                    "Intel (open-source)",
                    "AMD / ATI (open-source)",
                    "All open-source (default)",
                    "Nvidia (open kernel module for newer GPUs, Turing+)",
                    "Nvidia (open-source nouveau driver)",
                    "Nvidia (proprietary)"
                ]
            }
        },
```


Torxed:
This would be huge! Just the multi-select on it's own would be awesome hehe.
I gotta test this thurrowly once I'm done releasing v2.5.1!

svartkanin:
I'll let you know when it's ready ;)

@Torxed but as a question which I'm currently unsure about is how should we handle which greeter to install? Desktop profiles define lightdm and ssdm at the moment. Any suggestions here?

Torxed:
> I'll let you know when it's ready ;)
> 
> @Torxed but as a question which I'm currently unsure about is how should we handle which greeter to install? Desktop profiles define lightdm and ssdm at the moment. Any suggestions here?

That's a good question.
It would be nice if the profiles defined a meta-class or something that we can use to determinate collisions, something along the lines of what you've already got, but make a `Required` structure as well as `IncompatibleWith` value that defines what absolutely doesn't work together.

In this case I would say it would be a collision and the two desktop profiles won't work side by side.

Or we'd need to defined a `Decativate` function/class of profiles so that we can undo any setups done, like disabling greeters. And then the user could select one or simply, we pick a default and pre-configure that - and the user will have to do *some* work to get the other DE up and running? :)

Torxed:
The list of issues this PR would fix <3

svartkanin:
@Torxed I think this is ready for a review now
I have tested it to a degree but it would be better to get someone new to jump on it :)

Torxed:
I'll give it some testing during the week!

svartkanin:
@Torxed now that the new release is out, any chance to start merging the open PRs :) ?

Torxed:
I'm full speed ahead! Feels like a classic but it's been a lot lately. But I'm in merge-mode and I'm planning on getting this in within an hour or so. Testing it now :)

Torxed:
Any reason why we removed `/examples/`?
I get that they moved into scripts, but the main reason for the `/examples/` folder was to served as coding examples not just script entry-points. Could we move them back or would that be a horrific amount of change?

VexfulNiko:
Tested it, seems to not like updating the arch keyring as it spits out an error.

svartkanin:
@VexfulNiko interesting, would you mind sharing the steps and possibly error?

@Torxed any more thoughts on this PR otherwise or can we aim to get it into the next release 😀

VexfulNiko:
@svartkanin Just downloaded the Arch ISO and ran the Arch install, and it created errors saying it couldn't update the arch keyring. I tried to ping to ensure I had internet, and I did.

svartkanin:
This PR isn't even merged so I don't think it has anything to do with the problem you encountered

VexfulNiko:
An testing Archiso was created with this pr request if you go into actions

VexfulNiko:
I did! Said it was corrupted lol.

VexfulNiko:
Any update for merging this?

atillabirer:
> Ahh I see, regardless I don't think it has anything to do with the changes in this PR.
> 
> Have you tried running `pacman -Sy archlinux-keyring`?

The latest archiso was built with this bug. For now doing rm -rf /etc/pacman.d/gnupg then pacman-keyring init pacman-keyring --populate archlinux fixes it. Also lsblk gives an error. 

I am sorry but you fucked up the installer. It's not a shame to make a mistake but it's a shame to deny it. Please fix it.

svartkanin:
@atillabirer How would this PR have "fucked up the installer" if the PR isn't even merged yet?

svartkanin:

Issue #1444, opened 2022-08-30
AttributeError 'str' objet has no attribute 'display_name' when loading JSON --conf without --silent


Torxed:
![screenshot](https://user-images.githubusercontent.com/861439/187468530-ebf2f6fa-772f-4139-9aad-c93ca98548e0.png)

It has to do with the language percentage counter I believe.
Not sure but in case of loading menu items via `--conf` they come in as a string and not a built menu object that has `.display_name`.

Torxed:
This is btw the final blocker for releasing `v2.5.1` IMO as it's a reproducible issue. I'd be happy to solve #1192 as well but it's very sporadic.

PR #1446, opened 2022-08-30
Homogenize language option during handling

Torxed:
- This fix issue: #1444 

## PR Description:

During normal operation and "vanilla" startup of `archinstall`, the language option of `archinstall` is prepared and structured via `translationhandler.py` which populates the menu entry with `Language()` options, that can be represented in different ways.

But loading a `--conf` externally, meaning launching `archinstall --conf ...` will cause the `Language` field to be populated with a `str` object. Which isn't handled and thus the crash mentioned in #1444.

There was two solutions to this, either to centralize cleaning of the specific variables early in the `__init__` process.
Or to deal with it at a local level. Seeing as the language structure might change by improving upon it.. I thought it was best to prepare a solution more close to that logic.

So the `lambda` handling of the `Language` variable now has a "cleaning function" called `display_language()` which displays the proper language regardless if the input is a `Language()` object or a `str` object.

## Tests and Checks

### DIFF
--- a/archinstall/lib/menu/global_menu.py
+++ b/archinstall/lib/menu/global_menu.py
@@ -40,10 +40,6 @@ from ..output import FormattedOutput
 if TYPE_CHECKING:
 	_: Any
 
-def DEBUG(x):
-	print(x)
-	return x.display_name
-
 class GlobalMenu(GeneralMenu):
 	def __init__(self,data_store):
 		self._disk_check = True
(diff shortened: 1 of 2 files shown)


svartkanin:
Hmm I think it might be better to parse that from the input of the config instead

Torxed:
That is true :) And one of the things I was worried about. However I wasn't sure where to put the parsing in such a case. So this was the easiest and less messy way for me.

I'd welcome a better change and I'd be happy to merge it before we release.
I just pushed RC1 for v2.5.1 in hope that we'll at least get it on the October ISO :)

svartkanin:
Yeah that's fine for now, I can have a look at it, but I definitely agree that this fix is good enough for now since it alleviates the problem at hand :D

Torxed:
Hehe, then I'm looking forward to the upcoming release party! And nice of you to bring up the external deps. I think it was about time we re-visited that topic :)

PR #1456, opened 2022-09-05
Move deserialization into init

svartkanin:
## In this PR

* Addressed possible problems discussed in the comment https://github.com/archlinux/archinstall/pull/1446#issuecomment-1232170915
* Fix percentage display which can be > 100% if the `base.pot` file is out of sync


Torxed:
This looks like a really solid 1-Up over my solution.
Clean and elegant, I'll run a few tests right away as this would be nice to get in to v2.5.1.

Torxed:
Works, but I have a weird artifact which I don't know if it stems from this PR:
archinstall

svartkanin:
Hmm I don't get that behaviour

svartkanin:
I pulled in latest master as well but still I can't reproduce this :O

Torxed:
What happens if you (in a VM obviously) run `archinstall --conf https://0x0.st/ofza.json --disk-layout https://0x0.st/ofzB.json --creds https://0x0.st/ofzM.json`?

This should reproduce it : )

codefiles:
I can reproduce the same thing but looks like it is happening in master too.

Torxed:
> I can reproduce the same thing but looks like it is happening in master too.

Might be something we introduced earlier then.
Which needs to be solved before I can tag the actual release. We've got plenty of time so I'm not stressed yet.
I'll see if I can find out why, it's probably just a `\n` being appended from one of the translation files that we've merged recently.

codefiles:
@Torxed, Here is what introduced that bug: https://github.com/archlinux/archinstall/commit/dba30fdcf0e4e0cf3ef66d26a182c3852f3ee8f0#diff-d2120481fa970bf42b179322acfa63f24ad884c6b56f24a9f7f0166cdf1dfe78R158
The line added to `archinstall/locales/languages.json` in that commit causes the bug.

Torxed:
Thank you for finding it. I wonder if it's because the locale causes terminal issues?

svartkanin:
Very interesting, let me have a look at this today (just got up ;) )

Torxed:
I'm merging this, as it's a great fix and does nothing for the issue I found (as it's in `master` hehe).

svartkanin:
Wow so funny enough, I can reproduce it in a VM, but on my local machine everything works just fine

svartkanin:
There's also `?` displayed for greek, urdu etc. 
This is getting quite a bit messy now, since we have to display all of them at the same time in the selection I'm not sure if there's a font covering them all available on the ISO, I'll dig into this a bit

svartkanin:
Okay according to https://wiki.archlinux.org/title/Fonts#Greek there are different fonts for certain languages required. 
Some of them are only available as AUR though...

Torxed:
Hum, the AUR ones we can skip. But we can most likely package a font on the ISO if that's required :) Should count as accessibility which is a high priority.

PR #1461, opened 2022-09-07
Fix translation fonts

svartkanin:
## Translation fixes

I have switched to set the `LatGrkCyr-8x16` font via the installer which covers Cyrillic and Greek at the same time. 

Unfortunately I wasn't able to fix the problem completely, I have checked the fonts in `/usr/share/kbd/consolefonts` but there doesn't seem to be any available that cover `Tamil` nor `Urdu`. So for now these have been excluded since they won't work anyways. 

For `Tamil` I could only find a font in the AUR https://aur.archlinux.org/packages/ttf-tamil (I have not actually tried it out) and for `Urdu` there's this one https://aur.archlinux.org/packages/ttf-urdufonts. I don't thing there will be an option of including those in the ISO.

Covering languages for which there's no font available on the ISO will also require manual work in the future though, so I'm a bit unsure at the moment if it might be considerable to exclude certain support?

### DIFF
--- a/archinstall/lib/translationhandler.py
+++ b/archinstall/lib/translationhandler.py
@@ -46,8 +46,8 @@ class TranslationHandler:
 	_languages = 'languages.json'
 
 	def __init__(self):
-		# to display cyrillic languages correctly
-		self._set_font('UniCyr_8x16')
+		# to display latin, greek, cyrillic characters
+		self._set_font('LatGrkCyr-8x16')
 
 		self._total_messages = self._get_total_active_messages()
 		self._translated_languages = self._get_translations()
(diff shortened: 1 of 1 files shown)


Torxed:
This looks good. I think we can leave the translations there, but maybe hide them unless we find the two packages installed?
That way the users can install the AUR packages and get their languages if they desire, or when the packages gets adopted eventually.

But as it stands, yes, those cannot be included in the ISO reliably.

svartkanin:
That's an interesting point, I wonder if this could be displayed in the menu somehow

Torxed:
I know how I'd solve it in a GUI, with a hover over an exlamation point ^^ TUI is trickier!

Issue #1436, opened 2022-08-28
Better description what buttons u can use with the script.


eskiq:
There is no information about space for checking boxes. Also when someone will try to set for example more than 1 mirror region while in search mode gl to him. I just figure out that I can press TAB to check boxes in search mode. Better description would help.

Torxed:
I think this should go into the `man` page or `--help` if anywhere.
The menu system is getting quite full of text as it is IMHO. Would that be good enough?

eskiq:
I think that not everybody is using man pages or project pages to check basic stuff like this while script is already loaded. There is already partial legend but not full. Man page for archinstall is really long. Maybe it would be good to add a line to the beginning of the script with basic information of how to use this script containing a link to man page for further information.

The other and probably a better way to improve things would be by using more info in the same line. So for example instead of:

Use ESC to skip
Use CTRL+C to reset current selection

something like this with color:
[ESC](#) to skip &ensp; [CTRL+C ](#) to reset current selection

That would make it look more compact and there is still place to add for ex. SPACE and TAB in search mode if those options could be used.

I would also change skip to something different like maybe 'go back' or 'main menu'

BTW I shouldn't wrote it here but when I use ESC for skip in Drive(s) it would go outside of the script. Any other option with skip is working correctly but not this one.

Torxed:
> Maybe it would be good to add a line to the beginning of the script with basic information of how to use this script containing a link to man page for further information.

I'm personally quite against this. I want to keep the installer to a workflow that is non-interruptive. It's already a bit of a PITA that the user creation isn't set to sudo by default and that you need to jump down to "confirm" instead of being set to the username. It's small stuff, but overall it creates a bit of friction and frustration.
Adding a welcome screen will only be useful once, but will be there always - if that makes sense.

> something like this with color:
> [ESC](https://github.com/archlinux/archinstall/issues/1436#) to skip [CTRL+C ](https://github.com/archlinux/archinstall/issues/1436#)to reset current selection

That could work, but again. It would only be useful once and after that it just takes up space :)

Lets try it and see, maybe I'm wrong.
But I personally don't have the time to submit such a PR as there's more pressing things. Like updating the `man` and `--help` pages that you pointed out is way to long hehe.

> BTW I shouldn't wrote it here but when I use ESC for skip in Drive(s) it would go outside of the script. Any other option with skip is working correctly but not this one.

I think this is a known issue, but I'll see if we have a ticket otherwise I'll create one.

[...]

PR #1462, opened 2022-09-07
Update the action text

svartkanin:
## Action text improvements 

Fixes https://github.com/archlinux/archinstall/issues/1436

Some minor changes to the action help description depending on what is possible

Actions: Skip
`ESC to skip`

Actions: Skip and Reset
`ESC to skip, CTRL+C to reset`

Actions: Skip and Reset and Multi selection
`ESC to skip, CTRL+C to reset, TAB to select`

image

### DIFF
--- a/archinstall/lib/menu/menu.py
+++ b/archinstall/lib/menu/menu.py
@@ -162,14 +162,17 @@ class Menu(TerminalMenu):
 
 		action_info = ''
 		if skip:
-			action_info += str(_("Use ESC to skip"))
+			action_info += str(_('ESC to skip'))
 
 		if self._raise_error_on_interrupt:
-			if len(action_info) > 0:
-				action_info += '\n'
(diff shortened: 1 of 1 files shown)


Torxed:
Today I learned that you can select with tab hehe. I always use space ^^

svartkanin:
Yes so did I, but as pointed out in the issue Space won't work anymore if you use the filter so I just opted for Tab to cover all cases ;)

Torxed:
Aaaah!

eskiq:
That solution is better than I expected. Now I do realize that there is also one missing from the list. Very basic one. Enter to save changes. Arrows aren't either on that list, but I wouldn't like them to be there. I think Enter is worth of consideration.

svartkanin:
I think enter is a fairly universally known key for confirmation. I don't think it's necessary to explicitly state it's functionality. Same goes for the arrow keys. If this is an issue for people with accessibility then we can revisit the instructions again

PR #1467, opened 2022-09-10
Update handling of unsupported translations

svartkanin:
This is an update for the previous PR https://github.com/archlinux/archinstall/pull/1461.

It provides the option to mark languages that need external font support via the `languages.json` file. 
When a language was marked with that, it will not be selectable in the menu but it will indicate that it needs a manual installation of the font

image

If someone wants to provide the font manually they can simply save it as `/usr/share/kbd/consolefonts/archinstall_font.psfu.gz` and it will be automatically picked up by the installer. 

I've also updated the README to clarify that some languages may not work without a manual font installation.


Torxed:
This looks like a clean way of doing it. I'll run some tests and do a merge run soon. Great work!

archinstall/locales/README.md
```diff
@@ -1,8 +1,26 @@
 # Nationalization
 
-Archinstall supports multiple languages, which depend on translations coming from the community :) 
+Archinstall supports multiple languages, which depend on translations coming from the community :)
 
-New languages can be added simply by creating a new folder with the proper language abbrevation (see list `languages.json` if unsure).  
+## Important Note
+Before starting a new language translation be aware that a font for that language my not be
```
codefiles, line 6:
```suggestion
Before starting a new language translation be aware that a font for that language may not be
```

codefiles:
Should the use of first-person (we) and second-person (yourself) be avoided in documentation, including comments? That is what is typical for formal writing.

svartkanin:
@codefiles yeah i think that makes sense, i can reformulate it :)

Torxed:
Nice work on this one, I really like how it's descriptive and allowing people to use it. Keep up the good work!

Issue #1447, opened 2022-08-30
External Python dependencies


svartkanin:
I'm putting this here as a discussion point and/or exploration if it's possible at all. 

Would there be a way for us to use external dependencies? 
I don't know how the current release loop works or what the blockers/challenges are to accomplish this. 

I'm not keen on pulling in heaps of external dependencies but things like `pyparted` might be extremely useful and would probably prevent a lot of bugs from reinventing the wheel :)

Torxed:
&lt;background-rant&gt;
I'm quite sure that if we dig through the archives of comments I have historically been quite against this.
The main reason for not doing it earlier was that archinstall was more or less a single-file library that could be downloaded with wget and be executed. I later transitioned to compiling it with nuitka3 instead. But this is no longer the state of archinstall.. It's now a official application/library that users don't have to source themselves.
&lt;/background-rant&gt;

I have some reservations still, but they're minor:
1. The library must be well maintained and have a large traction *(`pyparted` is a good example)*. This minimizes the risk of sudden abandonment and deprecation.
2. It must be a Arch Linux package, this is to simplify packaging for upstream [package](https://archlinux.org/packages/extra/any/archinstall/)
3. Preferably the library should evict/replace large chunks of our own inventions *(this is to avoid npm hell where we have large amount of dependencies for very small tasks)*
4. The library should be either extremely small *(to avoid ISO bloating)* or we should use a lot of the library's potential *(to justify the somewhat large library)*
5. *(Bonus point) it would be nice if the library's introduction was easy to revert in case of point 1 or 4* 

If we go down this route, I suspect it might be a perfect opportunity to do large scale rewrites. As a lot of what archinstall both suffers from, but does ok and has a lot of helper functions for - is disk management. `pyparted` does a lot of it better for sure. But that would mean a lot of rewriting code and thus perhaps allow for a v2.6.0 or even v3.0 release if we bunch up a lot of other rewriting *(that would make quality of code a lot better.. We still have a lot of legacy code from like v1.0 that is far from best practice sadly)*.

So yea, tl;dr I'm in favor these days with careful consideration on the introduction :)

svartkanin:
I do agree on that only libraries that give a lot of benefit should be included. 

> The library must be well maintained and have a large traction (pyparted is a good example, I think.. Even tho it was 2 years between the last two releases heh). This minimizes the risk of sudden abandonment and deprecation.

[...]

PR #1478, opened 2022-09-21
Support external dependencies

svartkanin:
This should enable support for external dependencies, fix https://github.com/archlinux/archinstall/issues/1447

Since the project is using `flit` as a install/build/publish utility, we should no longer need the legacy `setup.{py,cfg}` files since everything is managed in the `pyproject.toml`. 

Now, `flit` has some downsides unfortunately, like it can't install "only" dependencies from a pyproject.toml file and it also doesn't allow to specify a target installation directory.
Since https://wiki.archlinux.org/title/Python_package_guidelines#Installation_methods mentions that installation has to be done into a temporary directory `flit` can't achieve this and therefore can't be used in the PKGBUILD file. 
Looking at other projects, a fair amount of them are using pip to install both, the dependencies and the package, into the temporary directory, so I borrowed the approach from those

I'm not an expert in the PKGBUILD space so @Torxed please advise if this is a bad idea :)


Torxed:
Interesting, I'll have a gander at this :) Would this solve https://github.com/pypa/flit/issues/509? :)

svartkanin:
@Torxed I haven't tried that actually, I'll test that tomorrow

svartkanin:
I do remember though that in the reworked profile PR https://github.com/archlinux/archinstall/pull/1434, I had removed the links since they caused some issues in various IDEs and I wasn't sure about the usage of them. 
Are those symlinks necessary to keep around and if so is there another use case except having them in the root for more visibility?

svartkanin:
@Torxed I run it on my local this is the resulting structure

```
$ ls pkg/archinstall/usr/lib/python3.10/site-packages/archinstall/
drwxr-xr-x  3 dan dan 4.0K Sep 22 11:37 examples
-rw-r--r--  1 dan dan  12K Sep 22 11:37 __init__.py
drwxr-xr-x 10 dan dan 4.0K Sep 22 11:37 lib
drwxr-xr-x 16 dan dan 4.0K Sep 22 11:37 locales
-rw-r--r--  1 dan dan  481 Sep 22 11:37 __main__.py
drwxr-xr-x  4 dan dan 4.0K Sep 22 11:37 profiles
drwxr-xr-x  2 dan dan 4.0K Sep 22 11:37 __pycache__
```

when running master with 

```python setup.py install```

it results in 
```
$ ls /usr/lib/python3.10/site-packages/archinstall-2.5.1rc1-py3.10.egg/archinstall/
drwxr-xr-x  3 root root 4.0K Sep 22 11:47 examples
-rw-r--r--  1 root root  13K Sep 22 11:47 __init__.py
drwxr-xr-x 10 root root 4.0K Sep 22 11:47 lib
drwxr-xr-x 20 root root 4.0K Sep 22 11:47 locales
-rw-r--r--  1 root root  481 Sep 22 11:47 __main__.py
drwxr-xr-x  4 root root 4.0K Sep 22 11:47 profiles
drwxr-xr-x  2 root root 4.0K Sep 22 11:47 __pycache__
```

which seems to be the same

**Comment by svartkanin** [comment]
I'll take this as a reference https://github.com/archlinux/archinstall/issues/1477#issuecomment-1265519354 :) 

Anything I can help here with further?
If we're certain that we're moving forward with external dependencies I'd be keen on starting implementation with pyparted meanwhile:)

**Comment by Torxed** [comment]
Let me check some things first, especially around packaging :)

File: pyproject.toml [hunk 1]
```diff
@@ -11,16 +11,18 @@ authors = [
...
     "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
     "License :: OSI Approved :: GNU General Public License v3 (GPLv3)",
     "Operating System :: POSIX :: Linux",
 ]
+dependencies = [
+    "pyparted==3.12.0",
+    "simple-term-menu==1.5.0",
+]
```
Torxed, line 25: If we update the PR to only include this change, after merging `upstream master` into this PR branch, I think we're good to merge this.

I've replaced `flit` and the build process in #1655

Torxed (changes requested):

svartkanin:
This work has all been merged with another PR so I'll close this PR

Issue #1229, opened 2022-05-21
We should exclude the read-only Arch ISO media from the list of volumes that can be used for install


dylanmtaylor:

Torxed:
Yes! maybe we exclude any RO media in general?

dylanmtaylor:
Is this really complete? I didn't see a change related to this

Torxed:
Oh, thought this fixed it: https://github.com/archlinux/archinstall/pull/1275

Issue #563, opened 2021-06-10
The installation media should not even be offered as a choice to install Arch Linux to


dylanmtaylor:
I saw someone hit this last night (Eastern time) in the help channel on Discord.

Scenario:
* Arch ISO burned to USB drive mounted on /dev/sda
* /dev/sda chosen as target, and user chose to format /dev/sda and install there thinking it was their hard drive (their real drive was an NVMe SSD, on nvme0n1 - they didn't notice the sizes)
* parted mklabel command failed due to mounted partitions
* Manual unmount of /dev/sda partitions failed due to it being the Arch ISO

I propose that, if we can somehow detect the installer is running off of a block device, not listing that device as an option in guided.

Torxed:
We should be able to map `/dev/loop0` (when ISO is used) or `/dev/sda` (if USB is used) against the logical device and filter the real device out. I believe `archinstall.Partition('/run/archiso/airrootfs').real_device` will turn out to be `/dev/sda` in this case, in which case we can filter out `/dev/sda` as it maps against the airrootfs device.

Pesudo code but `.real_device` exists in the library and could potentially be utilized.

PR #1481, opened 2022-09-22
Fix 1075

svartkanin:
Fixes https://github.com/archlinux/archinstall/issues/563
Fixes https://github.com/archlinux/archinstall/issues/1229

This should by default exclude archiso related devices from the selectable blockdevices


Torxed:
Nice! Thank you for these fixes.

PR #1492, opened 2022-09-27
Fix Polish language name

MedzikUser:
Changed from "Polskie" to "Polski"


Torxed:
Thank you for this fix! :)

PR #1494, opened 2022-09-27
Fix languages and delete whitespace

NPastorale:
## PR Description:

This is a small correction of the languages. As you've gone with the approach of writing them as it's written in each original language I can only correct the ones I'm knowledgeable of and confident it is written a certain way.

Additionally, there was a lot of unnecessary whitespace, which I have removed


Torxed:
I added back `Français` instead of `Française` as the former is a singular feminine word if my google-fu is correct, and `Française` is plural. If that's incorrect do correct me. But I had to merge upstream into this PR as there were some other changes.

PR #1496, opened 2022-09-29
Fixed spelling mistakes

benni347:
This pr fixed some spelling mistakes in `archlinux/__init__.py` doc strings for methods.

### DIFF
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -156,12 +156,12 @@ def get_arguments() -> Dict[str, Any]:
 	Is done on following steps:
 	0) we create a dict to store the arguments and their values
 	1) preprocess.
-		We take those arguments which use Json files, and read them into the argument dict. So each first level entry becomes a argument un it's own right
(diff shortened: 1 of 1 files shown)


archinstall/__init__.py
```diff
@@ -156,12 +156,12 @@ def get_arguments() -> Dict[str, Any]:
 	Is done on following steps:
 	0) we create a dict to store the arguments and their values
 	1) preprocess.
-		We take those arguments which use Json files, and read them into the argument dict. So each first level entry becomes a argument un it's own right
+		We take those arguments which use json files, and read them into the argument dict. So each first level entry becomes a argument on it's own right
```
Torxed, line 159: Could we change it to `JSON` as I think that's the correct naming convention as it's a shortening?
Other than that, great work!

benni347: Yes JSON is the correct convention.

Torxed:
Thank you for this one :)

PR #1498, opened 2022-10-01
Add Network tools xfce4.py

vinothkumar80:
Adding Network tools for connection  of network...........

- This fix issue: 

## Tests and Checks

### DIFF
--- a/profiles/xfce4.py
+++ b/profiles/xfce4.py
@@ -12,6 +12,8 @@ __packages__ = [
 	"lightdm-gtk-greeter",
 	"gvfs",
 	"xarchiver"
+	"networkmanager"
+	"network-manager-applet"
 ]
 
 


Torxed:
Can you change it so that it detects if the user selected the network manager option from the main menu instead? And then only install the applet, thank you!

vinothkumar80:
> Can you change it so that it detects if the user selected the network manager option from the main menu instead? And then only install the applet, thank you!

The Arch install script used by beginners and some experts and the network manager is essential package in xfce4.Adding the network tools its Helpful for beginners. Doesn't a optional tools...

svartkanin:
If this is installed by default then the option to setup the network would be pointless. 
The reason there is a separate option for network configuration is to give a user the flexibility to set that up independently from the profiles. So if someone didn't select the NetworkManager option as a network why would they want the applet installed?

vinothkumar80:
🙂

Torxed:
I'll re-open this if the before mentioned fix is implemented :)

Issue #1495, opened 2022-09-28
[256]: b'mount: /mnt/archinstall/dev: udev already mounted on /mnt/archinstall/dev


erikdubois:
Have a look at this error message.

It may tell you what went wrong and a clue how to improve the code

https://youtu.be/XxIh-B4Ol1g

Second install in same video worked again.

![test](https://user-images.githubusercontent.com/10594806/192700526-9edfdbc5-719a-4fc0-94d1-cbb4c37d5951.png)

Torxed:
Yea this one is a fun one :) It's a stack trace of a forked process.
I've seen it happen a few times, I have a hunch of what's wrong.

I'm assuming since you build the ISO in the video that you used the latest package 2.5.1, and it's not old footage you just uploaded?

erikdubois:
sorry about not mentioning the numbers.

I have build the Arch Linux iso my self - https://youtu.be/XxIh-B4Ol1g

looking at the iso I got archinstall 2.5.1-1

Torxed:
> idea: why no let the script start with the version number so it can be in the video

Thought about it when I was asking hehe.
Normally we ask people to submit the `install.log` where that information exists tho :) But perhaps we can put it in a corner or something.

acfuns:
same error

VexfulNiko:
Same here
Dell Optiplex 3080

nonwibb:
I'm getting the same error. I tried it on two different drives and even reflashed my usb. It's the Oct 1 arch iso release. I used the same settings and hardware with an older iso and it worked.

Torxed:
Can anyone submit your conf and disk layout?

https://archinstall.readthedocs.io/help/issues.html

Torxed:
![screenshot](https://user-images.githubusercontent.com/861439/194324716-21a4a1aa-6cdc-4ed1-a8d6-abe08c593ade.png)

I suspect this change to be a potential candidate for the root cause.
Because we don't update it in real time, but rely on static information:  https://github.com/archlinux/archinstall/blob/977976f88e761ced7d5bc9f39b32e2296ee91806/archinstall/lib/disk/partition.py#L35-L38

Torxed:
Temporary hotfix, in `partition.py` add:
```python
def mount(self, target :str, fs :Optional[str] = None, options :str = '') -> bool:
	SysCommand(f"umount {self._path}")
	if not self._partition_info.get_first_mountpoint():
```
And that should at least unmount the device first.

[...]

PR #1508, opened 2022-10-06
Updating Partition() information after mount and unmount.

Torxed:
- This fix issue: #1495 

## PR Description:

As described in the issue, we forgot to update partition information after `mount` and `umount` calls.
This leads `archinstall` to believe the partition isn't mounted anywhere.

This PR will:

1) Make sure the mount information (and general partition information) is updated after every `mount` or `umount` call
2) Add better error handling to `Partition().raw_parted` so that we can get better debug information out of it
3) Adds back `lsblk` information retrieval whenever `PartitionInfo()` is instantiated, making sure information such as `PARTUUID` and `UUID` is always fetched properly. If `PartitionInfo()` is instantiated too quickly after a `partprobe` there was no way for that information to be repopulated.

## Tests and Checks

Tested with 2 disk layouts:
```json
{
    "/dev/sda": {
        "partitions": [
            {
                "boot": true,
                "encrypted": false,
                "filesystem": {
                    "format": "fat32"
                },
                "mountpoint": "/boot",
                "size": "512MiB",
                "start": "1MiB",
                "type": "primary",
                "wipe": true
            },
            {
                "btrfs": {
                    "subvolumes": [
                        {
                            "compress": false,
                            "mountpoint": "/",
                            "name": "@",
                            "nodatacow": false
                        },
                        {
                            "compress": false,
                            "mountpoint": "/home",
                            "name": "@home",
                            "nodatacow": false
                        },
                        {
                            "compress": false,
                            "mountpoint": "/var/log",
                            "name": "@log",
                            "nodatacow": false
                        },
                        {
                            "compress": false,
                            "mountpoint": "/var/cache/pacman/pkg",
                            "name": "@pkg",
                            "nodatacow": false
                        },
                        {
                            "compress": false,
                            "mountpoint": "/.snapshots",
                            "name": "@.snapshots",
                            "nodatacow": false
                        }
                    ]
                },
                "encrypted": true,
                "filesystem": {
                    "format": "btrfs",
                    "mount_options": [
                        "compress=zstd"
                    ]
                },
                "generate-encryption-key-file": true,
                "mountpoint": null,
                "size": "100%",
                "start": "513MiB",
                "type": "primary",
                "wipe": true
            }
        ],
        "wipe": true
    }
}
```
```json
{
    "/dev/sda": {
        "partitions": [
            {
                "boot": true,
                "encrypted": false,
                "filesystem": {
                    "format": "fat32"
                },
                "format": true,
                "mountpoint": "/boot",
                "size": "513MiB",
                "start": "1MiB",
                "type": "primary"
            },
            {
                "encrypted": false,
                "filesystem": {
                    "format": "ext4",
                    "mount_options": []
                },
                "mountpoint": "/",
                "size": "10GiB",
                "start": "513MiB",
                "type": "primary",
                "wipe": true
            }
        ],
        "wipe": true
    }
}
```

Both work.

### DIFF
--- a/archinstall/lib/disk/partition.py
+++ b/archinstall/lib/disk/partition.py
@@ -37,7 +37,11 @@ class PartitionInfo:
 		if not all([self.partuuid, self.uuid]):
 			for i in range(storage['DISK_RETRY_ATTEMPTS']):
 				lsblk_info = SysCommand(f"lsblk --json -b -o+LOG-SEC,SIZE,PTTYPE,PARTUUID,UUID,FSTYPE {self.device_path}").decode('UTF-8')
-				lsblk_info = json.loads(lsblk_info)
+				try:
+					lsblk_info = json.loads(lsblk_info)
+				except json.decoder.JSONDecodeError:
+					log(f"Could not decode JSON: {lsblk_info}", fg="red", level=logging.ERROR)
+					raise DiskError(f'Failed to retrieve information for "{self.device_path}" with lsblk')
 
 				if not (device := lsblk_info.get('blockdevices', [None])[0]):
 					raise DiskError(f'Failed to retrieve information for "{self.device_path}" with lsblk')
@@ -182,8 +186,11 @@ class Partition:
 			output = error.worker.decode('UTF-8')
 
 		if output:
-			lsblk_info = json.loads(output)
-			return lsblk_info
+			try:
+				lsblk_info = json.loads(output)
+				return lsblk_info
+			except json.decoder.JSONDecodeError:
+				log(f"Could not decode JSON: {output}", fg="red", level=logging.ERROR)
 
 		raise DiskError(f'Failed to read disk "{self.device_path}" with lsblk')
 


Torxed:
Another issue is that the NVMe drives are really quick, so we'll end up here 5 times: https://github.com/archlinux/archinstall/blob/5c8eb7144d9583a8a117a4e892b1e78be1324a28/archinstall/lib/disk/filesystem.py#L271-L274

And then crash at: https://github.com/archlinux/archinstall/blob/5c8eb7144d9583a8a117a4e892b1e78be1324a28/archinstall/lib/disk/filesystem.py#L282-L285

Where the `New partitions:` will show:
```
New partitions: {None, 12312-512512-123123}
```
Indicating that the `Partition()` object never got it's information refreshed properly.

Torxed:
So for my own sanity. The loop above, it re-creates all the `Partition()` objects on each loop.
So there's no need to call `Partition()._fetch_information()` additionally.

But why the `Partition.part_uuid` returns `None` is beyond me. Especially after 5 loops with 1, 2, 3 and 4 seconds of delay between each loop. There's no reasonable way a NVMe wouldn't have returned a PARTUUID by that time. While SSD's do..

Torxed:
Reintroduced the logic of re-trying getting `PARTUUID` and `UUID` in the PartitionInformation model.
We'll have to clean up the logic of calling `_fetch_information()` on several places, but this is the most reliable way of making sure the PartitionInformation contains the information we want.

Torxed:
This fix has been confirmed to work. There's still the underlying issue of `archinstall` becoming confused if all `PARTUUID`'s are `None`. In which case it doesn't know what to do and it shouldn't continue either. But some how it has mixed up two partitions for being the same and that is bad. But at least this addresses the immediate issue of people not being able to partition at all.

jkbbwr:
Can confirm fix worked on my disk setup as well.

PR #1513, opened 2022-10-07
Clean up `add_systemd_bootloader()`

codefiles:
## PR Description:

- Clean up `add_systemd_bootloader()`
- Deduplicate loader entry content creation
- Use a try ... except statement rather than checking if a file exists before reading it to prevent introducing a race condition
- Fix the glob pattern for the default option in `/boot/loader/loader.conf`

## Tests and Checks
- [x] I have tested the code!

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -805,96 +805,125 @@ class Installer:
 		# points towards the same disk and/or partition.
 		# And in which case we should do some clean up.
 
+		boot_mountpoint = boot_partition.mountpoints[0]
+		esp = boot_mountpoint.root / boot_mountpoint.relative_to(self.target)
+
 		# Install the boot loader
 		try:
-			SysCommand(f'/usr/bin/arch-chroot {self.target} bootctl --path=/boot install')
+			SysCommand(f'/usr/bin/arch-chroot {self.target} bootctl --esp-path={esp} install')
 		except SysCallError:
 			# Fallback, try creating the boot loader without touching the EFI variables
-			SysCommand(f'/usr/bin/arch-chroot {self.target} bootctl --no-variables --path=/boot install')
-
-		# Ensure that the /boot/loader directory exists before we try to create files in it
-		if not os.path.exists(f'{self.target}/boot/loader'):
-			os.makedirs(f'{self.target}/boot/loader')
+			try:
+				SysCommand(f'/usr/bin/arch-chroot {self.target} bootctl --no-variables --esp-path={esp} install')
+			except SysCallError:
+				raise RequirementError('installing systemd-boot failed')
 
 		# Modify or create a loader.conf
-		if os.path.isfile(f'{self.target}/boot/loader/loader.conf'):
-			with open(f'{self.target}/boot/loader/loader.conf', 'r') as loader:
-				loader_data = loader.read().split('\n')
-		else:
-			loader_data = [
-				f"default {self.init_time}",
-				"timeout 15"
+		loader_dir = boot_mountpoint / 'loader'
+		loader_conf = loader_dir / 'loader.conf'
+
+		# Use a try ... except statement rather than checking if the file exists
+		# before reading it to prevent introducing a race condition
+		try:
+			with loader_conf.open('r') as loader:
+				loader_lines = loader.readlines()
+		except FileNotFoundError:
+			loader_lines = [
+				'default {}_{}.conf\n'.format(self.init_time, self.kernels[0]),
+				'timeout 15\n'
 			]
+		else:
+			for index, line in enumerate(loader_lines):
+				if line.split()[0] in '#timeout':
+					# Uncomment the default timeout to support dual-boot
+					loader_lines[index] = line[1:]
 
-		with open(f'{self.target}/boot/loader/loader.conf', 'w') as loader:
-			for line in loader_data:
-				if line[:8] == 'default ':
-					loader.write(f'default {self.init_time}_{self.kernels[0]}\n')
-				elif line[:8] == '#timeout' and 'timeout 15' not in loader_data:
-					# We add in the default timeout to support dual-boot
-					loader.write(f"{line[1:]}\n")
-				else:
-					loader.write(f"{line}\n")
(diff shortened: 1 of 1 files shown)


archinstall/lib/installer.py
```diff
@@ -844,63 +847,77 @@ def add_systemd_bootloader(self, boot_partition :Partition, root_partition :Part
...
+		comments = [
+				'# Created by: archinstall\n',
+				'# Created on: {}\n'.format(self.init_time)
+		]
+
+		contents = [
+				'title   Arch Linux ({kernel})\n',
+				'linux   /vmlinuz-{kernel}\n',
+				'initrd  /initramfs-{kernel}.img\n'
+		]
```
Torxed, line 859: Would it be cleaner to do:
```python
contents = [
		'title   Arch Linux ({kernel})',
		'linux   /vmlinuz-{kernel}',
		'initrd  /amd-ucode.img' if vendor == 'AuthenticAMD' else '',
		'initrd  /intel-ucode.img' if vendor == 'GenuineIntel' else '',
		'initrd  /initramfs-{kernel}.img'
]
```
And also do `'\n'.join(contents).format(kernel=kernel)` instead to remove the multiple `\n`?

archinstall/lib/installer.py
```diff
@@ -820,21 +820,24 @@ def add_systemd_bootloader(self, boot_partition :Partition, root_partition :Part
 		if not os.path.exists(f'{self.target}/boot/loader'):
 			os.makedirs(f'{self.target}/boot/loader')
 
+		loader_conf = f'{self.target}/boot/loader/loader.conf'
```
Torxed, line 823: Maybe we should use `pathlib.Path()` here?
```python
loader_conf = pathlib.Path(f'{self.target}/boot/loader/loader.conf')
```
And later use `with loader_conf.open("r") as loader` instead?
I think I prefer `if loader_conf.exists()` instead of a `try` & `catch` scenario.

Is there a reason we you'd prefer try and except?

svartkanin: I'd second to use Path over strings

archinstall/lib/installer.py
codefiles: There are issues with that solution. Here is a demonstration:

```
>>> val = 'B'
>>> l = [
...     'a',
...     'b' if val == 'B' else '',
...     'c' if val == 'C' else '',
...     'd'
... ]
>>> print(l)
['a', 'b', '', 'd']
>>> print('\n'.join(l))
a
b

d
>>> val = 'X'
>>> l = [
...     'a',
...     'b' if val == 'B' else '',
...     'c' if val == 'C' else '',
...     'd'
... ]
>>> print(l)
['a', '', '', 'd']
>>> print('\n'.join(l))
a


d
```

The portion of code that currently takes care of enabling microcode also serves to call `log()` when a match for microcode is not found.

The intention behind using `\n` in the list items is to be explicit that it is a list of lines.

archinstall/lib/installer.py
codefiles: > Is there a reason we you'd prefer try and except?

@Torxed , I'm glad you asked so that I can give an explanation for that change. There is a race condition with the current solution and the one you are proposing that the `try` ... `except` statement fixes.

archinstall/lib/installer.py
Torxed: That's true. I was mainly trying to conserve code complexity and amount of code since this PR would clean up a lot of the legacy nastiness that we have.

The `log()` would be hard to get around, but the empty newlines could be circumvented with `print('\n'.join(x for x in l if x))`. But that would add complexity I suppose.

It's good enough for me, just bouncing ideas :) Great work!

archinstall/lib/installer.py
Torxed: Oh yea that is True, never crossed my mind. I mean in most cases that would be an issue. For us that isn't really the case unless someone would go out of their way in a separate shell to destroy something. Which has been one of the major reasons for archinstall being single threaded *(with the slight exception of forks being used in certain places)* to avoid issues like those.

But I don't mind adhering to best practices. Maybe stick a short note there explaining why if possible (same in the other PR) because generally speaking `try-catch` statements are slower and less clean. Maybe that's just a fix idea I have and I need to change it. But a little note would go a long way to please me xD

codefiles: I was not aware that the `try` ... `except` statement was slower. My intention was bug free code.

codefiles:
Did a test install with this and saw no issues.

codefiles:
Through my testing I discovered that the original code had a bug with the default entry option in the `loader.conf` file. The default option takes a glob pattern [1] so not ending the default with `*` or being explicit and ending with `.conf` results in a failed match. It appears the first entry displayed will be the default selection without a match occurring. Currently the default entry is not configurable with archinstall but matching will now be fixed for if that ever happens.

The basic loader configuration file `/usr/share/systemd/bootctl/loader.conf` provided by the Arch Linux systemd package [2] has this same problem. The pattern should be `arch.conf` rather than just `arch` since the Arch Linux systemd package [3] provides the example entry is `/usr/share/systemd/bootctl/arch.conf`.

[1] https://man.archlinux.org/man/loader.conf.5#OPTIONS
[2] https://github.com/archlinux/svntogit-packages/blob/22c9777ede1f8752e8cd9a736b87d35d58952e2a/trunk/loader.conf
[3] https://github.com/archlinux/svntogit-packages/blob/22c9777ede1f8752e8cd9a736b87d35d58952e2a/trunk/arch.conf

codefiles:
Could maybe drop `efibootmgr` since it is not needed unless troubleshooting. [1]
https://github.com/archlinux/archinstall/blob/5c8eb7144d9583a8a117a4e892b1e78be1324a28/archinstall/lib/installer.py#L804

On another note, I do not care for the comments that are added to the top of the loader entry files.

[1] https://wiki.archlinux.org/title/Systemd-boot#Manual_entry_using_efibootmgr

codefiles:
This pull request has been superseded by the following:

- #1854
- #1866
- #1907
- #1909
- #1911

Issue #1493, opened 2022-09-27
Enabling Zram support should disable Zswap


Dungeonseeker:
According to the Arch Wiki, when using Zram you have to disable Zswap since it is enabled by default in the Linux Kernel.

See - [https://wiki.archlinux.org/title/Improving_performance#zram_or_zswap](url)

> Since it is enabled by default, [disable zswap](https://wiki.archlinux.org/title/Zswap#Toggling_zswap) if you decide to use zram to avoid zswap acting as a swap cache in front of it. Having both enabled also results in incorrect [zramctl(8)](https://man.archlinux.org/man/zramctl.8) statistics as zram remains mostly unused; this is because zswap intercepts and compresses memory pages being swapped out before they can reach zram.

Thought I'd bring it up since archinstall uses Zram for swap by default.

spaceoden:
Check the .conf file(s) in `/boot/loader/entries/`. Is `zswap.enabled=0` among the kernel parameters? If so, archinstall is successfully disabling zswap.

Dungeonseeker:
> Check the .conf file(s) in `/boot/loader/entries/`. Is `zswap.enabled=0` among the kernel parameters? If so, archinstall is successfully disabling zswap.

I'm using grub but the answer is no, I had to add it manually.

codefiles:
I'll add this to #1517 .

svartkanin:
This should be fixed now

PR #1517, opened 2022-10-08
Rework `add_grub_bootloader()`

codefiles:
- This fixes issue: #1493 (#881)

## PR Description:

- Rework `add_grub_bootloader()`
- Replace `sed` approach with Pythonic style
- Disable zswap if zram is enabled

## Tests and Checks
- [ ] I have tested the code!

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -907,69 +907,74 @@ class Installer:
 	def add_grub_bootloader(self, boot_partition :Partition, root_partition :Partition) -> bool:
 		self.pacstrap('grub')  # no need?
 
-		root_fs_type = get_mount_fs_type(root_partition.filesystem)
-
-		_file = pathlib.Path(f'{self.target}/etc/default/grub')
-
-		# Use a try ... except statement rather than checking if the file exists
-		# before reading it to prevent introducing a race condition
-		try:
-			with _file.open('r') as fh:
-				contents = fh.readlines()
-		except FileNotFoundError:
-			log(f"Could not configure GRUB, file not found: '{_file}'.", level=logging.DEBUG)
-			return False
-
-		options = [f'rootfstype={root_fs_type}']
-
-		if self._zram_enabled:
-			options.insert(0, 'zswap.enabled=0')
-
-		if real_device := self.detect_encryption(root_partition):
-			root_uuid = SysCommand(f"blkid -s UUID -o value {real_device.path}").decode().rstrip()
-
-			log(f"Using UUID {root_uuid} of {real_device} as encrypted root identifier.", level=logging.INFO)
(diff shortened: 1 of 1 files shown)


archinstall/lib/installer.py
```diff
@@ -888,19 +888,48 @@ def add_grub_bootloader(self, boot_partition :Partition, root_partition :Partiti
...
+		options = [f'rootfstype={root_fs_type}']
+
+		if self._zram_enabled:
+			options.insert(0, 'zswap.enabled=0')
+
 		if real_device := self.detect_encryption(root_partition):
 			root_uuid = SysCommand(f"blkid -s UUID -o value {real_device.path}").decode().rstrip()
-			_file = "/etc/default/grub"
-			add_to_CMDLINE_LINUX = f"sed -i 's/GRUB_CMDLINE_LINUX=\"\"/GRUB_CMDLINE_LINUX=\"cryptdevice=UUID={root_uuid}:cryptlvm rootfstype={root_fs_type}\"/'"
-			enable_CRYPTODISK = "sed -i 's/#GRUB_ENABLE_CRYPTODISK=y/GRUB_ENABLE_CRYPTODISK=y/'"
```
Torxed, line 895: I haven't used GRUB in a millennia. But I do agree that explicitly creating a config file is probably better.
But are there other options in the default conf that we would lose if we don't use the `sed` approach?

Way cleaner code the way this PR suggests, just checking in if we're losing anything by doing it?

codefiles: I also do not use GRUB. I did a minimal amount of testing on the code using the `grub.default` file, which is distrusted as `/etc/default/grub`, from the package source for GRUB on Arch Linux.

This should produces an identical copy of the `/etc/default/grub` file as the `sed` approach does and is just the pythonic way of doing it.

Dungeonseeker:
Couldn't you simply echo the parameter into "/etc/sysctld.d/99-zram.conf"? That way it would work regardless of what bootloader the user is running.

codefiles:
That is certainly a valid option and one I had not considered. I was simply replicating the solution that was merged for systemd-boot.

If the pull requests that I have open for boot loaders are merged I plan to then submit a pull request to move any of the code that is duplicated in the boot loader functions to the `add_bootloader()` function. This would include moving kernel parameters to `add_bootloader()` and adding them as a parameter to the boot loader functions.

Torxed (approved):

Torxed:
Looks good and the future plans are a nice direction. Just needs the latest `rework-grub` fix and conflicts sorted and I'll merge in!

codefiles:
I plan to close this once the kernel parameters are taking care of in a separate pull request since most of this was superseded by:

- #1962
- #1968

codefiles:
In addition to the list given in the last comment, also superseded by:
- #2079

Issue #1340, opened 2022-06-20
Select encryption partitions earlier


svartkanin:
When selecting an encryption password and then going to `Install` it prompts to select the paritions to be encrypted 

![image](https://user-images.githubusercontent.com/19396770/174587548-36f61332-8634-4f8e-b261-6cc9811ccd4b.png)

this selection should be prompted when entering an encryption password rather

svartkanin:
This has been resolved in some previous changes

Issue #1369, opened 2022-07-10
[Feature Request] Add Full disk encryption


Durko222:
Hello.

Please add the ability to use full disk encryption for grub. You can use both LUKS1 and add experimental [LUKS2](https://wiki.archlinux.org/title/GRUB#LUKS2) at the user's choice. Just in case, let me remind you that fde implies encryption of all partitions, including boot.

Thank you in advance.

z36r2850o:
I fully support!

Zbergen-cli:
It would be great!

PR #1520, opened 2022-10-09
New encryption menu

svartkanin:
# Sub menu for encryption related things

Since the current global menu got quite big, I grouped the encryption related things together into a sub-menu. 

image

image

This should also make things a bit easier to start working on https://github.com/archlinux/archinstall/issues/1369

Currently the type is only selectable as `Partition` but for the above ticket it would be extended to `Full disk encryption` and other ones in the future, hopefully :)

## New dataclass to hold all encryption related settings

Moving away from the primitive data types, this dataclass will hold all information related to the encryption settings. 
It will replace the `!encryption-password`, `hsm` and also the encryption setting in the disk layout menu.

```
@dataclass
class DiskEncryption:
	encryption_type: EncryptionType = EncryptionType.Partition
	encryption_password: str = ''
	partitions: List[str] = field(default_factory=list)
	hsm_device: Optional[Path] = None

```

### Table menu selection

A new `TableMenu` class has been added which allows to display tables as selection menus, including multi select support
```
class TableMenu(Menu):
	def __init__(
		self,
		title: str,
		data: List[Any] = [],
		table_data: Optional[Tuple[List[Any], str]] = None,
		custom_menu_options: List[str] = [],
		default: Any = None,
		multi: bool = False
	):
...
```

image

## Fixes 
https://github.com/archlinux/archinstall/issues/1340


svartkanin:
@Torxed this is ready for a review now :)

Torxed:
Awesome work as always! If your name was Harry I'd call you a Wizzard!

archinstall/lib/models/disk_encryption.py
```diff
@@ -0,0 +1,43 @@
+from dataclasses import dataclass, field
+from enum import Enum, auto
+from typing import Optional, List, Dict, TYPE_CHECKING, Any
+
+from archinstall.lib.hsm.fido import Fido2Device
```
Torxed, line 5: @svartkanin can we some how configure VSCode to not behave like this? Because VSCode really screw `archinstall` over as a library when I miss these and they get merged in.

svartkanin: Yeah sorry about that, it's my Pycharm that's messing it up. I'll have a look at the settings to see if things can be tweaked

svartkanin: @torxed what exactly is the issue that is happening with absolute paths btw? 
Maybe we can see if this could be fixed differently as well. 
There are some good arguments for using absolute paths in general https://softwareengineering.stackexchange.com/questions/159503/whats-wrong-with-relative-imports-in-python/159505#159505 :) but this doesn't mean I'm necessarily in favor of picking one over the other:)

PR #1530, opened 2022-10-20
Retry lsblk (partition._call_lsblk) when needed

cfinnberg:
## PR Description:

Modify lib.disk.partition._call_lsblk() function to make it more reliable. I was getting many errors trying automated install. Tracked the problem to lsblk giving incomplete information (like missing partition UUID) that made partition creation and mounting very unreliable. Test environment was Virtualbox in several different physical computers (Linux and Windows).

## Tests and Checks
- [X ] I have tested the code manually in a Virtualbox virtual machine

### DIFF
--- a/archinstall/lib/disk/partition.py
+++ b/archinstall/lib/disk/partition.py
@@ -171,7 +171,7 @@ class Partition:
 
 	def _call_lsblk(self) -> Dict[str, Any]:
 		self.partprobe()
-		for _ in range(storage['DISK_RETRY_ATTEMPTS']):
+		for _ in range(storage.get('DISK_RETRY_ATTEMPTS', 3)):
 			try:
 				output = SysCommand(f"lsblk --json -b -o+LOG-SEC,SIZE,PTTYPE,PARTUUID,UUID,FSTYPE {self.device_path}").decode('UTF-8')
 			except SysCallError as error:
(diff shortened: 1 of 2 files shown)


cfinnberg:
It was not my intention to add the other commit (new partition check in filesystem.py) but it ended here by mistake. I will leave it anyway as it is related to the original PR.

archinstall/lib/disk/partition.py
```diff
@@ -171,26 +171,32 @@ def __dump__(self) -> Dict[str, Any]:
...
+				continue
+			
 			try:
 				lsblk_info = json.loads(output)
-				return lsblk_info
 			except json.decoder.JSONDecodeError:
 				log(f"Could not decode JSON: {output}", fg="red", level=logging.ERROR)
+				# This is unexpected, better to raise an error
+				break
+			if not lsblk_info['blockdevices'][0]['partuuid']:
```
Torxed, line 193: Could we use `lsblk_info.get('blockdevices', [{'partuuid' None}])[0].get('partuuid')` here? To avoid `IndexError` and `KeyError` when the path doesn't exist?

Torxed (changes requested):

archinstall/lib/disk/partition.py
cfinnberg: Of course

svartkanin:
This PR is quite outdated so I'll go ahead and close it

Issue #1532, opened 2022-10-21
README "Live ISO Image" instructions fail to setup 


deobald:
The ISO does not have `python-setuptools' installed by default, so `python setup.py install` in step 5 fails.

eskiq:
#1526

martinstark:
Ran into the same issue when I wanted to try the 2.5.2 release with the NVME fixes

svartkanin:
`setup.py` has been removed and we're only using `pyproject.yaml` now. The README instructions are also up to date and should work as described

PR #1533, opened 2022-10-21
The Live ISO image does not have 'python-setuptools' installed

deobald:
- This fix issue: #1532 

## PR Description:

Corrects README instructions to help beginners install `python-setuptools`

## Tests and Checks

- [x] I have tested the added instruction

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,7 @@ you can replace the version of archinstall with a new version and run that with 
 5. Enter the repository with `cd archinstall`
    *At this stage, you can choose to check out a feature branch for instance with `git checkout v2.3.1-rc1`*
 6. Build the project and install it using `python setup.py install`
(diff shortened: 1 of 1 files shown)


Torxed:
Will it fit within the initramfs*(/squashfs)* that is loaded into RAM?
Historically speaking installing new packages into root is dicy. If it works I'll go ahead and merge this with some other PR's this weekend :)

deobald:
It did for me, but I was only trying to stumble my way through the
instructions to work around #1495.  :) Is there some way this could work
without `python-setuptools`?

martinstark:
Installing python-setuptools worked for me too, without it the "Live ISO Image" instructions would not work.

(I need to run 2.5.2 for the nvme fixes, so I tried this first over building my own ISO)

PR #1534, opened 2022-10-21
Add Chinese translation

yetist:
Please review this PR, thanks.


svartkanin:
@yetist thanks for the contribution, could you run the script mentioned in the reae again to generate the .mo file and commit it :)

Also as mentioned in the Readme https://github.com/archlinux/archinstall/blob/master/archinstall/locales/README.md languages with a different character set will not work out of the box but require the installation of 3rd party fonts.

yetist:
Done.

Issue #1540, opened 2022-10-28
intel_pstate kernel parameter not needed


mrgian:
`intel_pstate=no_hwp` is set as kernel parameter, but it's not needed as stated [here](https://bbs.archlinux.org/viewtopic.php?id=266794)

Also, this kernel parameter prevents the `power-profiles-daemon` package to work, a package needed to change the power mode in GNOME.

PR #1541, opened 2022-10-28
Remove intel_pstate kernel parameter

mrgian:
- This fix issue: #1540

## PR Description:

Removes `intel_pstate=no_hwp` from the kernel parameters

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -865,9 +865,9 @@ class Installer:
 				root_fs_type = get_mount_fs_type(root_partition.filesystem)
 
 				if root_fs_type is not None:
-					options_entry = f'rw intel_pstate=no_hwp rootfstype={root_fs_type} {" ".join(self.KERNEL_PARAMS)}\n'
+					options_entry = f'rw rootfstype={root_fs_type} {" ".join(self.KERNEL_PARAMS)}\n'
 				else:
-					options_entry = f'rw intel_pstate=no_hwp {" ".join(self.KERNEL_PARAMS)}\n'
+					options_entry = f'rw {" ".join(self.KERNEL_PARAMS)}\n'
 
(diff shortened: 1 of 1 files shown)


henriqueffc:
Wouldn't it be better to use the `intel_pstate=hwp_only` parameter. Register intel_pstate as the scaling driver only if the hardware-managed P-states (HWP) feature is supported by the processor.
[https://www.kernel.org/doc/html/v4.19/admin-guide/pm/intel_pstate.html](https://www.kernel.org/doc/html/v4.19/admin-guide/pm/intel_pstate.html)

mrgian:
It's more of a thing the user wants to set by himself, especially on GNOME, since its power modes don't work with this enabled.
This should be removed if archinstall wants to follow Arch Linux principles.

Torxed:
> This should be removed if archinstall wants to follow Arch Linux principles.

Since we're commited to this it's an easy choice :) Well done and a good PR.
The reason for this being here is purely legacy, back in 2013 that was needed to even boot on the laptop I was working on and since I was the only one using it - it got hard-coded in.

Just for fun, I'll attach the original `archinstall.sh` where this originates from, so I can find it later as it's fun to remissness on: https://gist.github.com/Torxed/c4d66b51c912cd7decbb7dbe34113c88

```bash
cat <<- EOF > /boot/loader/entries/arch.conf
	title "${BOOT_OPTION_TITLE}"
	linux /vmlinuz-linux
	initrd /initramfs-linux.img
	options cryptdevice=UUID="${ENC_UUID}:${ENC_DMAPPING}" root=/dev/mapper/"${ENC_DMAPPING}" rw "${INTEL_PSTATE}"
	# If not encrypted device:
	#options root=UUID="${DEV}" rw "${INTEL_PSTATE}"
EOF
```

PR #1542, opened 2022-10-29
Unify all lsblk calls

svartkanin:
## PR

Unify all calls to `lsblk` into it's own module and use the previously introduced dataclass to pass the information around.   
This should:
- Reduce redandency
- Handle errors in a single place
- Safer handling of types


Torxed:
Touching this logic is inherently dangerous, as this is the root cause of the regression in v2.5.0 and a lot of previous versions where I've tried to optimize these things.

We can merge this one, but we'll need to allot a serious amount of time for testing. As we'll need to pool resources from users in the wild to ensure certain NVMe harddrives etc don't bug out again. Every time disk issues happen, I always say to myself "This shouldn't be able to happen" ._.

Thoughts? I like the code change, do we want to commit to proper testing? :)

svartkanin:
This doesn't really introduce any new logic IIRC but just pulls out all the calls from the different places into a single file and uses a dataclass for safety. 

I'm happy to test it further if you have any specific desires about what exactly we would need to to test

svartkanin:

PR #1544, opened 2022-10-31
Move bootloader to dataclass

svartkanin:
### Update bootloader configuration

- Move bootloader config to a dedicated dataclass
- Align bootloader selection with other menu selections


Torxed:
Nice! This one I'll have to test out properly so the values dump correctly and gets loaded in via `--conf` etc. But it looks to be good stuff!

svartkanin:
Sure thing, I did test the configuration and also made sure it's backwards compatible

Issue #1509, opened 2022-10-06
Unable to resize font in archinstall


Torxed:
A user noted that while setting font size in arch releng iso, it doesn't translate into `archinstall`.
This is probably because we loaded a specific font with a specific font size to cover cyrillic issues etc.

https://discord.com/channels/726808923662712894/726809631342592081/1026892316268560404

svartkanin:
Yes that's right, it's due to the custom font setting :(

codefiles:
Suppose a conditional was added to the `else` statement in the code below to make it an `elif` that checked if the current font was the default ISO font. That way if the user set a font before `archinstall` it would not match the default ISO font and `_set_font()` would not run.

https://github.com/archlinux/archinstall/blob/5c8eb7144d9583a8a117a4e892b1e78be1324a28/archinstall/lib/translationhandler.py#L53-L56

svartkanin:
The problem here is that to be able to display a fair amount of languages correctly in the language change menu we need to set a font that supports cyrillic.

I see some options here:
- if custom font was set outside archinstall, don't override it and just break some language displays
- stop displaying languages in their own translation and revert back to displaying all in English and only set a font if the language was selected

codefiles:
> The problem here is that to be able to display a fair amount of languages correctly in the language change menu we need to set a font that supports cyrillic.
> 
> I see some options here:
> 
>     * if custom font was set outside archinstall, don't override it and just break some language displays
> 
>     * stop displaying languages in their own translation and revert back to displaying all in English and only set a font if the language was selected
@svartkanin, could another option be to change the console font temporarily when displaying the language menu?

In my suggestion I didn't consider if a user sets a console font they may do it without checking/knowing if the font supports the language they plan to use. I was going by if they manually set it that is what they want to use.

svartkanin:
Yeah that would work as well actually, sorry i read over your comment too quickly :)

svartkanin:
@codefiles actually I'm not entirely sure we can detect if a custom font was set. Last time I tried I wasn't able to retrieve the currently set font

codefiles:
I see what you mean.

svartkanin:
I gave this another go, but I wasn't able to determine the currently set font. 
In that case, i'd vote for reverting back to displaying all languages in English until one is selected. If the selected language has characters that are outside the default font support, we'll load another one. This way only a smaller subset of languages will potentially overwrite the custom setting.

Issue #1551, opened 2022-11-06
Selection of options looks small


castillofrancodamian:
Current Release: 2022.11.01

![Screenshot_20221106_023737](https://user-images.githubusercontent.com/47670442/200155999-92718dab-f0b5-40dc-b4e9-48b52808b37b.png)
![Screenshot_20221106_023757](https://user-images.githubusercontent.com/47670442/200156000-8ef639e1-230c-4205-935c-4f97cf12e7ea.png)
![Screenshot_20221106_023833](https://user-images.githubusercontent.com/47670442/200156002-3a46c760-a8ae-41c5-99aa-3426d53edb01.png)

svartkanin:
What exactly looks too small? The screenshots look as expected

svartkanin:
We're currently setting a custom font in the installer which may explain the small text. 

I'll raise a PR to only set it if a different language is selected

castillofrancodamian:
These screenshots are from the `archlinux-2022.09.03-x86_64.iso` image and the options take up the entire screen.
![Screenshot_20221106_153717](https://user-images.githubusercontent.com/47670442/200189000-e3c81f7c-cad8-4154-a4c5-76949f49f327.png)
![Screenshot_20221106_153741](https://user-images.githubusercontent.com/47670442/200189002-163e5052-091a-4616-af0b-7e3a311ae435.png)
![Screenshot_20221106_153754](https://user-images.githubusercontent.com/47670442/200189003-d98b6c11-69fc-4613-913e-c1509a96d8b5.png)

svartkanin:
I see, thanks for reporting. 
I'll have a look and see what changed

svartkanin:
This is resolved with the above PR

PR #1552, opened 2022-11-06
Remove custom font setting

svartkanin:
This PR fixes https://github.com/archlinux/archinstall/issues/1509 and https://github.com/archlinux/archinstall/issues/1551

In a previous PR https://github.com/archlinux/archinstall/pull/1467 a handler for unsupported languages as well as setting a new default font was introduced. 
Setting a new font via archinstall solved the problem of being able to display a different set of languages inside archinstall. 

However, for users who want to use archinstall in English that had some major drawbacks, they were no longer able to set their own font in the console which should be applied to archinstall. 

This font/language feature has caused a multitude of different problems in the past. Considering the large limitations of archiso and the minimal things shipped with it, there doesn't seem to be a font available that can handle a large variety of languages properly in the console. Archinstall trying to solve it is a bit of a lost cause unfortunately. 

I opted for a approach that will work for the majority of the users in this case: 
- Remove displaying languages in their own tongue in the menu
- Don't set any custom fonts via archinstall as that will break something somewhere 
- Display a message that should guide a user who wants to change the language on what they will have to do to get a different font working

image


Alexmelman88:
Maybe you should mark these strings as translatable:

> NOTE: If a language can not displayed properly, a proper font must be set manually in the console.
> All available fonts can be found in "/usr/share/kbd/consolefonts"
> e.g. setfont LatGrkCyr-8x16 (to display latin/greek/cyrillic characters)

svartkanin:
I don't think that would be good as at this stage some languages are not translatable due to the default font setting.

Therefore, it's probably better to have it displayed in English so it doesn't get lost:)

Torxed:
I agree, at least that string will be safe if left as it is :) Great work!
I'm trying to muster up some energy and time to merge all the PR's that's in the pipe.

Expected to take some time as some of them overlap and touches the same files.

PR #1559, opened 2022-11-13
Remove HSM from global menu

svartkanin:
@Torxed small follow up PR for the merged https://github.com/archlinux/archinstall/pull/1520

Forgot to remove these lines, they were responsible for enabling the HSM entry in the main menu but that got moved into the submenu now so they are not needed and will cause an error if `advanced` is enabled.


Torxed:
Ah good! :D Thanks for quick fix!

PR #1579, opened 2022-11-29
Adding a DependencyCheck runner. 

Torxed:
Mainly for future prep, when we start using external dependencies.
This should at least give us a basic safety check against some known issues.

I might have misunderstood DependencyCheck support for Python?: https://github.com/jeremylong/DependencyCheck/pull/5025

### DIFF
--- a/.github/workflows/dependencycheck.yaml
+++ b/.github/workflows/dependencycheck.yaml
@@ -1,0 +1,26 @@
+on: [ push, pull_request ]
+jobs:
+  depchecktest:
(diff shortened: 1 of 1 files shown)


svartkanin:
@Torxed is this something we need?

Torxed:
It's meant to alert us if we need to urgently bump a version.
But if we use renovate to always stay up to date - that could work too as there's nothing more we *can* do besides upgrading.

Torxed:
I think this is redundant now that we have renovate, so I'll close it :)

PR #1580, opened 2022-11-29
Add guarddog runner

Torxed:
Mainly for future prep, when we start using external dependencies.
This should at least give us a basic safety check against some known issues.


Torxed:
> Currently the project uses `flit` as a build/install tool handling the `pyproject.toml` file. In that case the dependencies should probably live inside the `pyproject.toml` as well

I agree, since `requirements.txt` is legacy anyway.
I'll figure out a way to do `xargs` over `grep` or something :)

svartkanin:
Something like this probably 
https://github.com/archlinux/archinstall/pull/1478/files#diff-c13dbcca92f9ff12cd26ecce958be3f9ee8563baace04f7a463a6d2dd4252e0bR46

`flit` recently had a new release that added a new argumen
`flit install --only-deps` which will only install dependecies. So it may be easier(?) to run that and do a pip freeze, but not sure

Torxed:
> Something like this probably https://github.com/archlinux/archinstall/pull/1478/files#diff-c13dbcca92f9ff12cd26ecce958be3f9ee8563baace04f7a463a6d2dd4252e0bR46
> 
> `flit` recently had a new release that added a new argumen `flit install --only-deps` which will only install dependecies. So it may be easier(?) to run that and do a pip freeze, but not sure

That would be a lot easier for installing the dependencies for sure.
The grep magic would have to be done so that guarddog is run on all dependencies tho, separate from installing.

So you're absolutly right with the oneliner:
```
grep -oP '^ *"[\s\S]+?[=><]+[\s\S]+?"' pyproject.toml > requirements.txt && sed -i 's|"||g' requirements.txt
xargs guarddog scan --exit-non-zero-on-finding <requirements.txt
```

I don't mind if it ends up on the disk between runs, but if we could boil it down to a one-liner that's helpful too :) I just don't have the necessary bash magic within me to do so ^^
And the `sed` magic confuses me too but I'll trust it.

svartkanin:
Yeah it doesn't have to be fancy as it's just sitting in the runner.

I'm working on the libparted implementation and I'll be able to share the migration soon(TM)

Torxed:
> Yeah it doesn't have to be fancy as it's just sitting in the runner.
> 
> I'm working on the libparted implementation and I'll be able to share the migration soon(TM)

Awesome! It's going to be a game changer and we'll probably need to bounce ideas on it at some point :)

PR #1445, opened 2022-08-30
Fixing issue where blkid causes SysCallException

Torxed:
- This adds debug information for issue: #1192 
- It solves some other issues surrounding partitions not showing up during partitioning

## PR Description:

Somewhere along the lines of trying to get disk information when the disk is new, a `blkid` crash popes out through numerous `exception` handlers. And this PR aims to address that.

## Tests and Checks
- [x] I have tested the code!<br>

PR #1582, opened 2022-11-30
Give lsblk x chances to respond

stuartcallum:
I see this sleep was removed '[for testing purposes](https://github.com/archlinux/archinstall/pull/1445)'

This is causing a DiskError on my poor slow USB stick.

```
...line 195, in _call_lasblk
...Failed to read disk "/dev/sdb1" with lsblk
```
I had a crack at implementing the retry system suggested on that fork. Any feedback would be appreciated ^_^

The changes look more drastic than they actually are because I added tabs for my while loop.


stuartcallum:
Thanks for getting back to me so quickly :-)

I appreciate your feedback. I'm a fan of your project and had fun understanding how it works. Best of luck with your bigger changes!

PR #1583, opened 2022-12-01
Generate -fallback variant of boot entries for systemd-boot

### LINKED ISSUE #1572: Installation didn't create a fallback initramfs boot entry for systemd-boot
I recently tried to recommend my friends to use Hyper-V USB passthrough and archinstall to try out Arch Linux in an easy and harmless way on their spare USB HDD. But, although Ubuntu Live CD and setup a bootable USB HDD this way, the archinstall script failed to do so because in the default `mkinitcpio` procedure, the `autodetect` hook is placed before `block` and `keyboard` hook.

There are two ways to fix this, the optimal way is to let me teach them to chroot into the new installation, modify `mkinitcpio.conf` then rebuild the `initramfs`. But I also think add the wiki-recommend `arch-fallback.conf` boot entry to help people got caught off guard.

escape0707:
This PR addresses one of the possible improvements mentioned in #1572 and generate `-fallback` variant boot entries for `systemd-boot`.

## PR Description:

As in title, this simply add another level of loop to create `-fallback` variant of boot entries for systemd-boot. This might be helpful for users installed Arch with archinstall into an external USB drive and boot it on another machine, or plan to swap / upgrade hardware.

## Tests and Checks


Torxed:
Nice! Sadly it won't fix #1572 but it does improve archinstall in terms of functionality.
Because the issue in #1572 as I understand it is that the `EFI` partition is to small (usually due to windows creating it) and thus both variants doesn't fit into the partition size wise.

And `-fallback` can never be created. Which will be solved by a separate partition for `/boot` and `/efi` :) But great PR for functionality increases!

Torxed:
Tested and looks good, great work!
Merging with this result:
screenshot

screenshot

Issue #1557, opened 2022-11-12
Failed to read disk "/dev/sdaX" with lsblk


Torxed:
Unsure of how or why, but a user reports issues with `/dev/sda1` being called onto when it doesn't appear to exist.

Stack trace:
```
Exit code for blkid -p -o export was: 512
Not a loop device, trying uevent rules.
Could not decode JSON: lsblk: /dev/sda1: not a block device
{
   "blockdevices": [

   ]
}

Traceback (most recent call last):
  File "/usr/bin/archinstall", line 8, in <module>
    sys.exit(run_as_a_module())
  File "/usr/lib/python3.10/site-packages/archinstall/__init__.py", line 311, in run_as_a_module
    script.execute()
  File "/usr/lib/python3.10/site-packages/archinstall/lib/profiles.py", line 195, in execute
    self.spec.loader.exec_module(sys.modules[self.namespace])
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/usr/lib/python3.10/site-packages/archinstall/examples/guided.py", line 301, in <module>
    ask_user_questions()
  File "/usr/lib/python3.10/site-packages/archinstall/examples/guided.py", line 107, in ask_user_questions
    global_menu.run()
  File "/usr/lib/python3.10/site-packages/archinstall/lib/menu/selection_menu.py", line 336, in run
    if not self._process_selection(value):
  File "/usr/lib/python3.10/site-packages/archinstall/lib/menu/selection_menu.py", line 353, in _process_selection
    return self.exec_option(config_name, selector)
  File "/usr/lib/python3.10/site-packages/archinstall/lib/menu/selection_menu.py", line 373, in exec_option
    result = selector.func(presel_val)
  File "/usr/lib/python3.10/site-packages/archinstall/lib/menu/global_menu.py", line 79, in <lambda>
    lambda preset: self._select_harddrives(preset),
  File "/usr/lib/python3.10/site-packages/archinstall/lib/menu/global_menu.py", line 345, in _select_harddrives
    harddrives = select_harddrives(old_harddrives)
  File "/usr/lib/python3.10/site-packages/archinstall/lib/user_interaction/system_conf.py", line 52, in select_harddrives
    hard_drives = all_blockdevices(partitions=False).values()
  File "/usr/lib/python3.10/site-packages/archinstall/lib/disk/helpers.py", line 275, in all_blockdevices
[...]

PR #1585, opened 2022-12-05
Revise packages in Budgie profile

serebit:
## PR Description:

This PR revises the list of packages included in the Budgie profile.

- The Arc GTK theme and Papirus icon theme are the schema defaults for the `budgie-desktop` package (Adwaita does not support Budgie), and thus including their packages sets them in the desktop automatically
- The `gnome` group has been replaced with Nemo and MATE Terminal, as most of that group was unnecessary to include in a minimal Budgie installation
- `budgie-desktop-view` provides desktop icon support for users who want the feature

These changes result in a Budgie installation more reflective of current upstream recommendations.

## Tests and Checks

### DIFF
--- a/profiles/budgie.py
+++ b/profiles/budgie.py
@@ -4,12 +4,16 @@ import archinstall
 
 is_top_level_profile = False
 
-# "It is recommended also to install the gnome group, which contains applications required for the standard GNOME experience." - Arch Wiki
 __packages__ = [
+	"arc-gtk-theme",
 	"budgie-desktop",
-	"gnome",
+	"budgie-desktop-view",
 	"lightdm",
 	"lightdm-gtk-greeter",
+	"mate-terminal",
+	"network-manager-applet",
+	"nemo",
+	"papirus-icon-theme",
 ]
 
 
--- a/profiles/budgie.py
+++ b/profiles/budgie.py
@@ -6,12 +6,10 @@ is_top_level_profile = False
 
 __packages__ = [
 	"arc-gtk-theme",
-	"budgie-desktop",
-	"budgie-desktop-view",
+	"budgie",
 	"lightdm",
 	"lightdm-gtk-greeter",
 	"mate-terminal",
-	"network-manager-applet",
 	"nemo",
 	"papirus-icon-theme",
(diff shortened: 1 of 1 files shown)


Torxed:
Thanks for the PR.
I assume this is per discussions I saw the other day in Arch internal?

I'd be happy to merge this, but would you mind if I tweak it a bit so that `network-manager-applet` only gets installed if NM was the selected network choice in the global menu? :)

serebit:
Yeah, this is a result of that discussion.

That sounds ideal, go right ahead!

dylanmtaylor:
I also think that network manager should be dropped. I'm not sure about some of the others, it feels a bit opinionated. Should we ship the optional desktop icons? Which terminal/file manager should we use? Why not gnome-terminal and nautilus, as mate is based off of a really old codebase? Overall, this is probably a step in the right direction.

serebit:
Point by point:

- The network manager applet should be installed if NetworkManager is selected by the user, but not otherwise. I wasn't aware that this was possible to automate, but I'm in favor of it.
- Technically budgie-desktop-view is optional for Budgie to function, which is why it's an optional dep in the package, but including them allows the user to simply enable desktop icons if they want them and it doesn't take up much disk space. It's in service of providing a more complete Budgie with the profile.
- It is opinionated, but it sort of has to be. I'm one of the developers of Budgie, and there is no officially recommended file manager or terminal—downstreams are free to choose what they feel is most appropriate. For example, last I checked, Solus ships Nautilus and GNOME Terminal, but Fedora's work-in-progress Budgie spin developed by the Budgie BDFL will ship Caja and MATE Terminal instead. I believe Nemo and MATE Terminal are more reasonable defaults for the Budgie profile than Nautilus and GNOME Terminal given that they both use GTK3 and support GTK widget themes like the rest of the desktop.

serebit:
Updated and tested the code by modifying the profile .py file in a live environment. Works perfectly.

Torxed:
Thank you for these fixes, apologies that it took so long to merge.
Wanted to get #1557 out of the way first to make sure we're back to stable again.

PR #1563, opened 2022-11-13
Fix 1538

svartkanin:
According to the comments in the thread, this potentially fixes #1538

PR #1604, opened 2022-12-21
PyParted and a large rewrite of the underlying partitioning

svartkanin:
PyParted and large rewrite of the underlying partitioning

Unfortunately, I didn't get to wrap this up before the end of the year approached and will have to be completed in the next one. 

## PR prerequisites 
There are a fair amount of other PRs open at the moment so I've pulled in my previous ones as I didn't want to sort out the massive amount of merge conflicts that happen otherwise: 
- https://github.com/archlinux/archinstall/pull/1434
- https://github.com/archlinux/archinstall/pull/1478
- https://github.com/archlinux/archinstall/pull/1542
- https://github.com/archlinux/archinstall/pull/1544
- https://github.com/archlinux/archinstall/pull/1563

## New arhcinstall dependencies
As this PR uses external dependencies, they will have to be installed to be able to run `archinstall`:

```
flit install --only-deps
```

this will isntall the dependencies listed in `pyproject.toml`

```
dependencies = [
    "pyparted==3.12.0",
    "simple-term-menu==1.5.0",
]
```

### No more separate `disk_layout` files

There used to be 3 possible configuration files
- user_configuration.json
- user_disk_layouts.json
- user_credentials.json

now, there are only two supported ones

- user_configuration.json
- user_credentials.json

This change allowed to remove the separate, sort of redundant, `harddrive(s)` menu entry completely and only rely on disk layout configuration. 

The disk layout configuration has been moved to a data class defined as 

```
class DiskLayoutType(Enum):
	Default = 'default_layout'
	Manual = 'manual_partitioning'
	Pre_mount = 'pre_mounted_config'

@dataclass
class DiskLayoutConfiguration:
	layout_type: DiskLayoutType
	layouts: List[DeviceModification] = field(default_factory=list)
```

which will be saved into the `user_configuration.json` file as 

```
"disk_layouts": {
        "layout_type": "default_layout",
        "layouts": [
            {
                "device": "/dev/sdb",
                "partitions": [
                    {
                        "btrfs": [],
                        "flags": [
                            "Boot"
                        ],
                        "fs_type": "fat32",
                        "length": {
                            "sector_size": null,
                            "total_size": null,
                            "unit": "MiB",
                            "value": 512
                        },
                        "mount_options": [],
                        "mountpoint": "/boot",
                        "start": {
                            "sector_size": null,
                            "total_size": null,
                            "unit": "MiB",
                            "value": 1
                        },
                        "status": "create",
                        "type": "primary"
                    },
                    {
                        "btrfs": [],
                        "flags": [],
                        "fs_type": "ext4",
                        "length": {
                            "sector_size": null,
                            "total_size": null,
                            "unit": "GiB",
                            "value": 20
                        },
                        "mount_options": [],
                        "mountpoint": "/",
                        "start": {
                            "sector_size": null,
                            "total_size": null,
                            "unit": "MiB",
                            "value": 513
                        },
                        "status": "create",
                        "type": "primary"
                    },
                    {
                        "btrfs": [],
                        "flags": [],
                        "fs_type": "ext4",
                        "length": {
                            "sector_size": null,
                            "total_size": {
                                "sector_size": null,
                                "total_size": null,
                                "unit": "B",
                                "value": 250148290560.0
                            },
                            "unit": "Percent",
                            "value": 100
                        },
                        "mount_options": [],
                        "mountpoint": "/home",
                        "start": {
                            "sector_size": null,
                            "total_size": null,
                            "unit": "GiB",
                            "value": 20
                        },
                        "status": "create",
                        "type": "primary"
                    }
                ],
                "wipe": true
            }
        ]
    }
```

This change is at this moment not backwards compatible, theoretically it can be made backwards compatible, but I'd really not do that as it may cause bad behavior as in the new configuration the mode of disk layout configuration (default, manual or pre-mounted) is stored in the config as well. Which means that parsing an old version, we have to guess what it was. 

### UI changes

No more harddrive selection 
image

New sub-menu disk layout selection 
image

New device selection menu 
image

New manaul partitioning menu (similar to the user configuration menu) 
image

## Current state
What works at the moment
- New UI menu should work as expected
- On clicking `Install`
  - Partitioning works as expected
  - Formatting works as expected
 
Still missing
- The entire `installation.py` file still has to be updated to work with all the new components


Torxed:
Huge, good job! I'll probably isolate some components and try them for a quick patch before this weekend. Just to try and fix the lsblk issue.

bd-g:
Updates? I'd love to help if I can to get this closer to a release

svartkanin:
Sorry didn't have too much time recently but the majority of the flow is working. I'll have to address the Btrfs stuff, hopefully next week, now as it's quite different from the rest

bd-g:
> Sorry didn't have too much time recently but the majority of the flow is working. I'll have to address the Btrfs stuff, hopefully next week, now as it's quite different from the rest

Sounds good. I have a spare laptop and I've been running multiple installs a day just testing out different configs, so if you need any help on that end let me know.

svartkanin:
@bd-g that would be fantastic actually, I'll give you shout when things are ready for some testing

bd-g:
One note, we need to update the build dependencies now that `archinstall` is using some python modules (like pyparted). 

The README should be updated under "Using a Live ISO Image". One potential revision to step 2 is as follows (including a link to Arch Wiki to resize the root partition as gcc is quite large):

Install the build requirements with `pacman -Sy; pacman -S archlinux-keyring; pacman -S git python-pip python-flit pkg-config gcc`
   *(note that this may or may not work depending on your RAM and current state of the squashfs maximum filesystem free space, see [Arch Wiki entry](https://wiki.archlinux.org/title/archiso#Adjusting_the_size_of_root_partition_on_the_fly) on increasing size of filesystem on the fly)*

bd-g:
I've been testing this on various devices with various configurations throughout the morning - excited for your btrfs changes whenever those come!

The one bug I found was in the new way to store users and the pipewire profile.

The old pipewire application profile was a plugin that defined `on_user_created`, thus only running when a user was created. The new pipewire profile runs `_enable_pipewire_for_all` when the pipewire audio option is selected. However, this fails when only a root password is supplied and no additional users are added.

In this case, `users` on line 34 of `pipewire.py` is `[{}]`, an array of a single empty dictionary, instead of an array of dictionaries containing per-user data. It loops once over this object, and tries to access the username field of the empty dict, and crashes.

We probably still want to allow installation of pipewire even without non-root users, so we can probably just include a small sanity check in `pipewire.py`. Something like the below

```
def _enable_pipewire_for_all(self, install_session: 'Installer'):
    users: Union[User, List[User]] = archinstall.arguments.get('!users', None)
    if not isinstance(users, list):
        users = [users]

    for user in users:
        if user:
	    install_session.arch_chroot('systemctl enable --user pipewire-pulse.service', run_as=user.username)
```

svartkanin:
Thanks for debugging this already :) 
I noticed some pipewire issues tbis week as well, I'll address those of course, and when everything is ready all docs, examples, schemas, configa will be updated as well

svartkanin:
Good news, I found a great python package for Btrfs handling https://github.com/kdave/btrfs-progs/tree/master/libbtrfsutil%2Fpython
It seems to be well maintained well and supported.

Bad news, it has not been published to PyPi and the way to use it is by running `pacman -S btrfs-progs` which will install the module globally on the system. Meaning that it can't be used in a virtual environment that way.

Now, i have some ideas how to get it to work, not sure how happy we'd be with it however. 

The best thing I van think of is to clone the repo and build the wheel which is very simple. We could then include the wheel as an external dependency in archinstall. flit doesn't seem to have a way of installing such packages so it will have to be handled explicitly with pip.

Another solution is to switch to poetry which has a lot more functionality and would allow to point to a git repo directly or install a wheel.

Torxed:
> `pacman -S btrfs-progs` which will install the module globally on the system. Meaning that it can't be used in a virtual environment that way.

That is more than fine. Again, remember that the preferred way of packaging is through `pacman` and not `pip` as we'd like to source dependencies through [PKGBUILD](https://wiki.archlinux.org/title/PKGBUILD) anyway :)

svartkanin:
That's great to hear, then we just need a "decent" way to have it integrated in our dev flow, I'll see what the smoothest way would be

svartkanin:
@Torxed @bd-g Alrighty I think this is ready for some testing now. I run it a few times with both `ext4` and `btrfs` as well as encryptions and the installations went fine. 

I'll start pulling in master changes as well now and add the relevant changes that were merged

archinstall/lib/installer.py
```diff
@@ -1070,6 +1070,26 @@ def install_profile(self, profile: Profile):
 		self.log(f'Installing archinstall profile(s) {absolute_name}', level=logging.INFO)
 		profile.install(self)
 
+	def install_audio(self, audio: str, enable_for_users: List[User] = []):
```
Torxed, line 1073: I like how clean this function is, but there is a risk `Installer()` grows out of control. Which is why I would argue we keep the `/applications/<app>` handlers for specific applications.

Thoughts?

svartkanin: My though on this was that the installer is a dedicated class that can handle all installation specific things by simply calling a function which defined what will be installed, e.g.
```
installer.install_profile(...)
installer.install_audio(...)
installer.install_(...)
installer.install_bootloader(...)
```

This would make documentation quite simple and in case of users wanting to script their own installation it will make things less disconnected. The installer is self-contained and no installer instance or session needs to be passed around which also helps with the dependency hell :)

I'm not too worried about the installer size tbh, I think it'll need a cleanup all together and all exposed functions that are handling specific installation steps can be nicely highlighted. But I'm happy to discuss it :)

svartkanin: I don't mind splitting out single things into a separate class as well if that helps, but just looking at the pipewire setup, I'm not entirely sure I fully understand why it was actually implemented as a profile (maybe that was also me when doing the profile rewrite, it was a long time ago...)

Torxed: I don't mind if we have a wrapper in installer for those things.
But the bulk task of performing actions could still be in the application profile.

I also do outside of installations:
```
import archinstall
archinstall.Application(installation, 'postgresql').install()
```

These serve as helpers outside of installations as well.
For instance for post-installation scripts and what not.

I get that it might seem weird considering the profiles use `archinstall.storage['installation_session']` quite extensively. But that's something that can be injected at runtime with dummy instances as well as something I was hoping we could move away from some how.

Torxed: Another reason why I like dividing them up into their own little files is that it's easier to find, add and modify the specific application steps. Digging through `Installer()` is probably a bit intimidating.

svartkanin: Teah I'm not a fan of throwing the session around and storing in it global places that very much defeats the purpose of isolating logic and providing a interface. 

I'm happy to revert this for now and keep it as a profile, we can look at this in a follow up

Torxed: Sounds good! Let's pool our experiences together and make a kick ass fix for the profiles after this one.

Makes this PR a bit easier to test too so thank you, you're doing awesome with this massive PR

Torxed:
`python-pyparted` should show up in `[community-testing]` soon if all goes well: https://archlinux.org/packages/?Community-Testing&q=pyparted

That means we're one step closer to being able to properly integrate this into the release process with proper dependencies :D

svartkanin:
Thanks for sharing @Torxed , could you elaborate on what that process is?
Would this not be installed via the pypi dependencies?

codefiles:
In `archinstall/lib/disk/device_handler.py` there are a few cases of conditionals that check the exit code of a `SysCommand()` call. If the conditional is checking for a non-zero exit code then the code that depends on the conditional being true will never be executed because a `SysCallError` will be raised.

https://github.com/archlinux/archinstall/blob/8f6cc07062968b259bebd346521ef685c16f89dc/archinstall/lib/general.py#L280-L281

Torxed:
> In `archinstall/lib/disk/device_handler.py` there are a few cases of conditionals that check the exit code of a `SysCommand()` call. If the conditional is checking for a non-zero exit code then the code that depends on the conditional being true will never be executed because a `SysCallError` will be raised.
> 
> https://github.com/archlinux/archinstall/blob/8f6cc07062968b259bebd346521ef685c16f89dc/archinstall/lib/general.py#L280-L281

Correct, that is the old 2021-2022 way of checkning. There's still a few places that use the old method.

svartkanin:
@Torxed Yeah I figured the build is done via PKGBUILD, but looking at how PKGBUILD works, I was under the impression that the function `package()` would be used to install dependencies, which I've updated with 

```
package() {
  cd "$pkgname-$pkgver"

  #install the dependencies
  PIP_CONFIG_FILE=/dev/null flit install --only-deps
  # install archinstall
  PIP_CONFIG_FILE=/dev/null pip install --root="${pkgdir}" --ignore-installed dist/*.whl
  # install the docs
  install -vDm 644 docs/_build/man/archinstall.1 -t "$pkgdir/usr/share/man/man1/"
}
```

This will install all the dependencies necessary listed in the `pyproject.toml` under 
```
dependencies = [
    "pyparted==3.12.0",
    "simple-term-menu==1.6.1",
]
```

Would that not work?

Torxed:
> Would that not work?

Would work, but it would also create multiple ownership of the pyparted source files under `/usr/lib/python3.X/site-packages/pyparted`.

Not a huge problem, maybe.
Unless people start installing stuff outside of the ArchISO :)

So it's always best to rely on `PKGBUILD`'s dependency listing, so I pushed to get [python-pyparted](https://archlinux.org/packages/community-testing/x86_64/python-pyparted/) in :)

Torxed:
Great work! I'll chime in this weekend as well and test.
I'll *probably* postpone this on purpose to 2023-04-02 and drop this as it's own release just after the upcoming smaller release, to not mix in too many changes at once.

svartkanin:
@Torxed maybe I'm confusing things now, but I thought in the package function it would install tge dependencies into a virt env. And then when when installing the actual wheel it will install pyparted as part of the archinstall package. That would mean the exact version specified in the pyproject.toml. whereas now we're served the version installed globally or not? 

Also does that mean any future dependencies must be available in the arch repo not only on PyPi? How would versions be handled?

 I've also included the `simple-term-menu` as a dependency and removed the manual copy, would that work then with the build?

svartkanin:
@bd-g @Torxed yep after April is fine, there's some bits and pieces still left to cleanup and I'll update all the documents and examples as well

svartkanin:
I think I had these removed here already as we only require the pyproject.toml :)
I'm very keen on seeing this getting merged 😁

Torxed:
> I'm very keen on seeing this getting merged

Me too ^^ I'm devoted to the next release revolving around this one! :D

RotaN8:
Oh, nice! I was about to surrender and I have seen this new PR and I couldn't be more happy. I'm trying to automate the installation of ArchLinux on my PC (only for fun). To test it I'm using a VM from now, and I achieved to install it with a dual boot with Windows 10, but I'm having problems with the PARTUUID that it's suposed to not be required, but it actually is because otherwise it doesn't continue and complains about it. I have the theory that is because the partitions are already existent so it uses it to verify which is each one. And because every time you create a partition the UUID changes, so you always end by doing it manually with the UI instead of running the installation with the JSON configs. 

I also loved the part where it says that the `disk_layouts.json` will disappear and finally will be present inside the `user_configuration.json`. That's a great improvement!

Thanks for all! I will be waiting for the PR (Is it still planned for tomorrow?) and I will test it and give some feedback if I find problems ;)

Btw, how can I update it on the _archiso_? I'm new to the ArchLinux world...

Torxed:
> Is it still planned for tomorrow?

Testing is yes, this one is huge and will require at least a week or so of thorough testing :)

> Btw, how can I update it on the archiso? I'm new to the ArchLinux world...

The easiest is to download the ISO that is generated on each commit/change: https://github.com/archlinux/archinstall/actions/workflows/iso-build.yaml

Currently this branch isn't being correctly built do to `pyproject.toml` issues. But we'll sort that out soon enough.

Torxed:
I think I merged the conflicts correctly with the new path-hint patch from #1659:
```
<<<<<<< dev-pyparted
	dest_path = prompt_dir(str(_('Enter a directory for the configuration(s) to be saved: ')))

	if options['user_config'] == choice.value:
		config_output.save_user_config(dest_path)
	elif options['user_creds'] == choice.value:
		config_output.save_user_creds(dest_path)
	elif options['all'] == choice.value:
		config_output.save_user_config(dest_path)
		config_output.save_user_creds(dest_path)
=======
	dirs_to_exclude = [
		'/bin',
		'/dev',
		'/lib',
		'/lib64',
		'/lost+found',
		'/opt',
		'/proc',
		'/run',
		'/sbin',
		'/srv',
		'/sys',
		'/usr',
		'/var',
	]
	...
	if save_dirs is not None:
		for save_dir_str in save_dirs:
			save_dir = Path(save_dir_str)
			if options['user_config'] == choice.value:
				config_output.save_user_config(save_dir)
			elif options['user_creds'] == choice.value:
				config_output.save_user_creds(save_dir)
			elif options['disk_layout'] == choice.value:
				config_output.save_disk_layout(save_dir)
			elif options['all'] == choice.value:
				config_output.save_user_config(save_dir)
				config_output.save_user_creds(save_dir)
				config_output.save_disk_layout(save_dir)
>>>>>>> master
```

RotaN8:
> > Is it still planned for tomorrow?
> 
> Testing is yes, this one is huge and will require at least a week or so of thorough testing :)
> 
> > Btw, how can I update it on the archiso? I'm new to the ArchLinux world...
> 
> The easiest is to download the ISO that is generated on each commit/change: https://github.com/archlinux/archinstall/actions/workflows/iso-build.yaml
> 
> Currently this branch isn't being correctly built do to `pyproject.toml` issues. But we'll sort that out soon enough.

Thanks for answering that fast! I managed to update the archinstall package by doing: `pacman -S archinstall` inside the _archiso_ I didn't know that was that simple lol. But it's nice to know about the existence of the Actions section ISOs I will be tracking them to make tests :-)

And don't worry I'm not in a hurry, I was only asking to know if I will be able to keep testing it tomorrow ;-)

Thanks for all!

PD: Which package manager are you using?

Torxed:
> PD: Which package manager are you using?

For AUR? `yay` or manual builds :) Otherwise `pacman` heh.

RotaN8:
> > PD: Which package manager are you using?
> 
> For AUR? `yay` or manual builds :) Otherwise `pacman` heh. We hang out on Discord and IRC btw if you want to ask general questions :)

I'm dumb, I actually wanted to ask about the python dependency manager you guys are using for this project. But I have confused the terms XD. Too many hours in front of the computer hahaha.

Don't worry, I'll ask about it on Discord, I don't want to make this PR even longer :)

svartkanin:
Thanks @Torxed I'll address the conflicts and failing checks

svartkanin:
Okay I fixed some issues with the save_config, not sure if that was from the merged prev PR or due to the rebase, anyhow it should be good now. 
Not sure why the build check fails, there seems to be some issue with the pyproject.toml atm

RotaN8:
> Okay I fixed some issues with the save_config, not sure if that was from the merged prev PR or due to the rebase, anyhow it should be good now. Not sure why the build check fails, there seems to be some issue with the pyproject.toml atm

```
ValueError: invalid pyproject.toml config: 'project'.
configuration error: 'project' must not contain {'license_files', 'long_description_content_type', 'long_description'} properties
```

These three options are maybe at the wrong place?

Torxed:
> or due to the rebase

Yea that one has tripped me up a few times. Hard to find where and why as well. And don't even get me started on `git stash` issues hehe.

I can help sort out the pyproject, should be an easy contribution to this PR.

Torxed:
I'll continue testing, because it builds now but some how this differs vastly from the `master` build result.

This `pyproject.toml` produces:
```
anton@lappy ~ $ ls -l /usr/lib/python3.10/site-packages/archinstall/
total 16
-rw-r--r-- 1 root root 11300 Apr  4 09:41 __init__.py
-rw-r--r-- 1 root root   481 Apr  4 09:41 __main__.py
drwxr-xr-x 1 root root    96 Apr  4 09:41 __pycache__
```

`master`'s produces:
```
anton@lappy ~ $ ls -l /usr/lib/python3.10/site-packages/archinstall/
total 20
-rw-r--r-- 1 root root 12692 Apr  4 09:51 __init__.py
-rw-r--r-- 1 root root   481 Apr  4 09:51 __main__.py
drwxr-xr-x 1 root root    96 Apr  4 09:51 __pycache__
drwxr-xr-x 1 root root   144 Apr  4 09:51 examples
drwxr-xr-x 1 root root   440 Apr  4 09:51 profiles
```

Both have issues tho since `archinstall` is not present.
I'll correct this in the new couple of hours :)

Torxed:
I know there's gone a lot of work into re-working this, but `setuptools` appears to have support for symlinks now *(was broken in earlier versions)*.
If we want to keep the `/profiles/` rather than `/archinstall/profiles` layout in github to make it easier for users to find them, it's now possible.

I'm ok either way as people will probably find them :)
Logically they belong where they've been put in this PR - but ease of access *miiiiiight* have been easier before? I dunno, but tested it now while building and noticed that symlink works.

svartkanin:
> I'll continue testing, because it builds now but some how this differs vastly from the `master` build result.
> 
> This `pyproject.toml` produces:
> 
> ```
> anton@lappy ~ $ ls -l /usr/lib/python3.10/site-packages/archinstall/
> total 16
> -rw-r--r-- 1 root root 11300 Apr  4 09:41 __init__.py
> -rw-r--r-- 1 root root   481 Apr  4 09:41 __main__.py
> drwxr-xr-x 1 root root    96 Apr  4 09:41 __pycache__
> ```
> 
> `master`'s produces:
> 
> ```
> anton@lappy ~ $ ls -l /usr/lib/python3.10/site-packages/archinstall/
> total 20
> -rw-r--r-- 1 root root 12692 Apr  4 09:51 __init__.py
> -rw-r--r-- 1 root root   481 Apr  4 09:51 __main__.py
> drwxr-xr-x 1 root root    96 Apr  4 09:51 __pycache__
> drwxr-xr-x 1 root root   144 Apr  4 09:51 examples
> drwxr-xr-x 1 root root   440 Apr  4 09:51 profiles
> ```
> 
> Both have issues tho since `archinstall` is not present. I'll correct this in the new couple of hours :)

How did you install it? 
I run `pip install .` inside the project and it created the following 

```
dan@mordor:~/.local/lib/python3.10/site-packages $ ls archinstall
total 36K
drwxr-xr-x  6 dan dan 4.0K Apr  6 14:20 default_profiles
-rw-r--r--  1 dan dan  12K Apr  6 14:20 __init__.py
drwxr-xr-x 10 dan dan 4.0K Apr  6 14:20 lib
drwxr-xr-x 24 dan dan 4.0K Apr  6 14:20 locales
-rw-r--r--  1 dan dan  481 Apr  6 14:20 __main__.py
drwxr-xr-x  2 dan dan 4.0K Apr  6 14:20 __pycache__
drwxr-xr-x  3 dan dan 4.0K Apr  6 14:20 scripts
```

svartkanin:
> 

Yes I did go with the logical approach here to put them under the `archinstall` folder. IMHO they are part of archinstall itself and linking them into the root may not be the thing as there's always risks with symlinks. 

I'm happy to reference them in the README and point to the location or even put in examples in the `examples/` folder with some profiles.

Torxed:
> I run `pip install .` inside the project and it created the following

It's:
```
python -m build
```
And then:
```
python -m installer dist/*.whl
```
As `pip` is no longer the build mechanic for PEP517, and there's really no point in using pip unless you want to for installing the `*.whl` although it should be able to. But in my experience pypa is messing around with PEP517 causing a lot of tool breakage and odd behaviors.. So I'm sticking with their reference implementations `build` and `installer` for now.

> and that resulted in the same

Is this not the desired structure?
I got a bit confused by `default_profiles` but realized this was a change in the PR and that `profiles/` and `examples/` no longer should be included.

svartkanin:
Right then it should be the right output except for the examples missing, I'll fix that

svartkanin:
@Torxed any chance to get this merged? 
I feel other contributors may try to fix the existing code pointlessly as it may be replaced anyways due to this rework :/

Torxed:
if you feel ready sure :)

svartkanin:
I wouldn't mind someone running some tests for it ;) to confirm things are working, at least for the base installation cases

Torxed:
I'll run a bunch of tests in ~6h and if nothing surprising shows up I'll merge this beast of a PR.

bd-g:
I'm getting an error when trying to save configuration files - "TypeError: Object of type PosixPath is not JSON serializable"

Additionally, if there is a failed attempt to encrypt the hard drive (like interrupting setting an encryption password), any subsequent  encryption commands freeze. Haven't looked into it yet, but I've had similar problems when working with crypsetup manually - a single failed command locks down everything, and it forces a reboot to let you work with it again. Might be something similar happening here (and probably isn't unique to this PR, just food for thought)

archinstall/lib/luks.py
```diff
@@ -97,120 +83,163 @@ def encrypt(self, partition :Partition,
...
-					for child in children:
-						# Unmount the child location
-						if child_mountpoint := child.get('mountpoint', None):
-							log(f'Unmounting {child_mountpoint}', level=logging.DEBUG)
-							SysCommand(f"umount -R {child_mountpoint}")
-
-						# And close it if possible.
-						log(f"Closing crypt device {child['name']}", level=logging.DEBUG)
-						SysCommand(f"cryptsetup close {child['name']}")
+			if err.exit_code == 256:
```
codefiles, line 106:
```suggestion
			if err.exit_code == 1:
```

This is a regression from f611310 which looks to be merged in from `master`.

archinstall/lib/installer.py
```diff
@@ -460,7 +477,32 @@ def genfstab(self, flags :str = '-pU') -> bool:
...
+									for sub_vol in part_mod.btrfs_subvols:
+										# We then locate the correct subvolume and check if it's compressed
+										if sub_vol.compress and str(sub_vol.mountpoint) == Path(mountpoint[0].strip()):
+											# We then sneak in the compress=zstd option if it doesn't already exist:
+											# We skip entries where compression is already defined
+											if ',compress=zstd,' not in line:
+												line = line.replace(subvoldef[0], f',compress=zstd{subvoldef[0]}')
+												break
+
+								fp.write(f'{line}\n')
```
codefiles, line 505:
```suggestion
				if part_mod.fs_type != disk.FilesystemType.Btrfs:
					continue
				
				fstab_file = Path(f'{self.target}/etc/fstab')
				with fstab_file.open('r') as fp:
					fstab = fp.readlines()

				# Replace the {installation}/etc/fstab with entries
				# using the compress=zstd where the mountpoint has compression set.
				for index, line in enumerate(fstab):
					# So first we grab the mount options by using subvol=.*? as a locator.
					# And we also grab the mountpoint for the entry, for instance /var/log
					subvoldef = re.findall(',.*?subvol=.*?[\t ]', line)
					mountpoint = re.findall('[\t ]/.*?[\t ]', line)

					if not subvoldef or not mountpoint:
						continue

					for sub_vol in part_mod.btrfs_subvols:
						# We then locate the correct subvolume and check if it's compressed,
						# and skip entries where compression is already defined
						# We then sneak in the compress=zstd option if it doesn't already exist:
						if (sub_vol.compress and str(sub_vol.mountpoint) == Path(mountpoint[0].strip())
								and ',compress=zstd,' not in line):
							fstab[index] = line.replace(subvoldef[0], f',compress=zstd{subvoldef[0]}')
							break

				with fstab_file.open('w') as fp:
					fp.writelines(fstab)
```

This should accomplish the same task but reduce the amount of indentation.

archinstall/lib/luks.py
svartkanin: Potentially not rebased yet, I'll have a look

archinstall/lib/disk/device_handler.py
```diff
@@ -0,0 +1,593 @@
...
+				if recursive:
+					command += ' -R'
+
+				try:
+					SysCommand(f'{command} {mountpoint}')
+				except SysCallError as err:
+					# Without to much research, it seams that low error codes are errors.
+					# And above 8k is indicators such as "/dev/x not mounted.".
+					# So anything in between 0 and 8k are errors (?).
+					if err and err.exit_code and 0 < err.exit_code < 8000:
```
codefiles, line 521: See f611310

archinstall/lib/disk/device_model.py
```diff
@@ -0,0 +1,1034 @@
...
+		retry = 1
+
+	result = None
+
+	for i in range(retry):
+		try:
+			result = SysCommand(f'lsblk --json -b -o+{lsblk_fields} {dev_path}')
+		except SysCallError as error:
+			# It appears as if lsblk can return exit codes like 8192 to indicate something.
+			# But it does return output so we'll try to catch it.
```
codefiles, line 984: See f611310

svartkanin:
@bd-g I realized that I accidentally committed the commenting out of the save config file, I fixed that now. For the
> I'm getting an error when trying to save configuration files - "TypeError: Object of type PosixPath is not JSON serializable"

I can't seem to be able to reproduce this, could you try again and see if the latest push may have fixed it? 

Regarding the encryption, did you get the first error due to a bug? For the subsequent failure until restart I'd need a way to reproduce it. Could you share the steps you took to get there?

archinstall/lib/luks.py
svartkanin: @codefiles I think the merge conflicts may have been resolved incorrectly, thanks for finding this! I keeps getting more difficult the longer the PR is open to keep it up to date with master :(

archinstall/lib/installer.py
svartkanin: @codefiles thanks, that does look much better. I just update the formatting a bit and committed it

archinstall/lib/disk/device_handler.py
svartkanin: @codefiles thanks, I updated it to align with the master version

archinstall/lib/disk/device_model.py
svartkanin: @codefiles thanks, updated

bd-g:
Having trouble getting the Arch ISO to work, so can't check the configuration file problem right now.

> Regarding the encryption, did you get the first error due to a bug? For the subsequent failure until restart I'd need a way to reproduce it. Could you share the steps you took to get there?

1) Launch archinstall and select disk configuration
2) Choose any valid setup
3) Select disk encryption
4) When entering the disk encryption password, CTRL+C to exit archinstall. 
5) Repeat steps 1-3. Then reselect disk encryption, and if you try to enter a password, it just resets everytime you try and enter a password.

Like I said, the same thing happens when working with cryptsetup directly. If you run `cryptsetup` to encrypt a partition, and you send a keyboard interrupt when entering or confirming the password, any subsequent calls to `cryptsetup` will fail to encrypt that same partition.

Because it's a bug I see with cryptsetup too, this might not be anything you can fix at the level of `archinstall`, but maybe something to be aware of and put in a troubleshooting section.

archinstall/lib/installer.py
```diff
@@ -178,146 +173,168 @@ def __exit__(self, *args :str, **kwargs :str) -> None:
...
+		We need to wait for it before we continue since we opted in to use a custom mirror/region.
+		"""
+		log('Waiting for automatic mirror selection (reflector) to complete...', level=logging.INFO)
+		while service_state('reflector') not in ('dead', 'failed'):
+			time.sleep(1)
 
-				if not os.path.isdir(f"{self.target}/{os.path.dirname(absolute_logfile)}"):
-					os.makedirs(f"{self.target}/{os.path.dirname(absolute_logfile)}")
+		log('Waiting pacman-init.service to complete.', level=logging.INFO)
+		while service_state('pacman-init') not in ('dead', 'failed'):
```
codefiles, line 194: See 3f5170f

This is the second commit to `master` that I have discovered to not be merged in. Any commit to `master` on code that you have moved to a different or new file may not be successfully merged.

archinstall/lib/installer.py
```diff
@@ -178,146 +173,168 @@ def __exit__(self, *args :str, **kwargs :str) -> None:
...
+			time.sleep(1)
 
-				shutil.copy2(absolute_logfile, f"{self.target}/{absolute_logfile}")
+		log('Waiting Arch Linux keyring sync (archlinux-keyring-wkd-sync) to complete.', level=logging.INFO)
+		while service_state('archlinux-keyring-wkd-sync') not in ('dead', 'failed'):
+			time.sleep(1)
```
codefiles, line 199: See 3f5170f

This is the second commit to `master` that I have discovered to not be merged in. Any commit to `master` on code that you have moved to a different or new file may not be successfully merged.

codefiles:
@bd-g that sounds like an issue unrelated to this pull request. If you are interrupting during the encryption password prompt then `cryptsetup` is not being called. At that point you would be in the `get_password()` function inside `lib/user_interaction/utils.py` coming from `lib/disk/encryption_menu.py`. I do not see any of that logic being changed with this pull request.

bd-g:
> @bd-g that sounds like an issue unrelated to this pull request. If you are interrupting during the encryption password prompt then `cryptsetup` is not being called. At that point you would be in the `get_password()` function inside `lib/user_interaction/utils.py` coming from `lib/disk/encryption_menu.py`. I do not see any of that logic being changed with this pull request.

I didn't have this problem with the old setup, only with the new disk updates has this problem surfaced. Hence my confusion. Probably a bug to document and fix another day, nothing to hold up the PR for. 

Also is this an AI bot?

codefiles:
My assumption that the issue is unrelated to the pull request is wrong if the issue does not occur with `master`.

> Also is this an AI bot?

Me?

codefiles:
I built this branch as a package and added it to a local repo. I then used Archiso, configured for the local repo and for the addition of the package, to build an ISO. Testing that ISO in a VM produces the following error when `archinstall` is executed on the command line.

```
  File "/usr/lib/python3.10/site-packages/archinstall/__init__.py", line 5, in <module>
    from .lib import disk
  File "/usr/lib/python3.10/site-packages/archinstall/lib/disk/__init__.py", line 1, in <module>
  File "/usr/lib/python3.10/site-packages/archinstall/lib/disk/device_handler.py", line 572, in <module>
  File "/usr/lib/python3.10/site-packages/archinstall/lib/disk/device_handler.py", line 31, in __init__
    self.load_devices()
  File "/usr/lib/python3.10/site-packages/archinstall/lib/disk/device_handler.py", line 41, in load_devices
    disk = Disk(device)
  File "/usr/lib/python3.10/site-packages/parted/decorators.py", line 42, in new
    ret = fn(*args, **kwds)
  File "/usr/lib/python3.10/site-packages/parted/disk.py", line 52, in __init__
    self.__disk = _ped.Disk(device.getPedDevice())
_ped.DiskLabelException: Can't have a partition outside the disk!
```

Output of `lslkb`:

```
NAME  MAJ:MIN RM   SIZE RO TYPE MOUNTPOINTS
loop0   7:0    0 684.1M  1 loop /run/archiso/airootfs
sr0    11:0    1 804.3M  0 rom  /run/archiso/bootmnt
vda   254:0    0    20G  0 disk 
```

Output of `fdisk -l`:

```
Disk /dev/vda: 20 GiB, 21474836480 bytes, 41943040 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes


Disk /dev/loop0: 684.15 MiB, 717381632 bytes, 1401136 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
```

svartkanin:
@bd-g this doesn't seem to be related to cryptsetup as that would only be called during the installation process, so it's probably a menu issue. I'll have a look at it


@codefiles thanks for testing, I'll debug the issue

codefiles:
In `device_handler.py` the `DeviceHandler.load_devices()` function is initializing devices with the `parted.Disk()` class. This is raising `_ped.DiskLabelException: Can't have a partition outside the disk!` for the CD-ROM device and `_ped.DiskLabelException: /dev/vda: unrecognised disk label` for the unformatted disk. Maybe don't initialize CD-ROM devices and use `parted.freshDisk()` for unformatted disks.

archinstall/lib/disk/device_handler.py
```diff
@@ -0,0 +1,584 @@
...
+
+	@property
+	def devices(self) -> List[BDevice]:
+		return list(self._devices.values())
+
+	def load_devices(self):
+		block_devices = {}
+
+		for device in parted.getAllDevices():
+			disk = Disk(device)
```
codefiles, line 41:
```suggestion
			try:
				disk = Disk(device)
			except DiskLabelException as error:
				if 'unrecognised disk label' in getattr(error, 'message', str(error)):
					disk = freshDisk(device, 'gpt')
				else:
					continue
```

This is a possible way to fix https://github.com/archlinux/archinstall/pull/1604#issuecomment-1509924019

If you use this don't forget the imports:

```diff
-from parted import Disk, Geometry, FileSystem, PartitionException
+from parted import Disk, freshDisk, Geometry, FileSystem, PartitionException
+from _ped import DiskLabelException
```

archinstall/lib/installer.py
svartkanin: I did try to merge things in quite carefully, but some may have slipped through the net :(

svartkanin:
@bd-g I followed the steps but I can't seem to find the problem you described 

https://user-images.githubusercontent.com/19396770/232266596-c8e086ee-7a9c-4967-9686-833a213c93db.mp4

archinstall/lib/disk/device_handler.py
svartkanin: I managed to reproduce the error and this seems to fix it indeed! 
I guess if the label is not recognizable it is save to fallback to a fresh disk :thinking:

Torxed:
So the workflow has changed a lot and I'm struggling with piecing it all together.
This used to be two different workflows that was working:
```
# cd github/archinstall-svartkanin
# python -m archinstall
```
But also
```
# cd github/archinstall-svartkanin
# cp examples/guided.py ./
# python guided.py
```
Obviously the later has changed location so I did:
```
# cd github/archinstall-svartkanin
# cp archinstall/scripts/guided.py ./
# python guided.py
```

But this now causes:
```
  File "/home/anton/github/archinstall-svartkanin/guided.py", line 7, in <module>
    from ..lib import disk
ImportError: attempted relative import with no known parent package
```
`guided.py` has always been seen as a standalone executable, something people would look at to get inspiration of how to write a installer script. It was never part of the library.

I know I've mentioned to not do `import archinstall` in the library, but `guided.py` and the other examples would be the exception. I'm ok with them living inside the library folder as we use them as `python -m <entrypoint>`.

Torxed:
Another minor change:
I cannot back out of "Please chose which greeter to install"

Neither ESC or Ctrl+C works.

Torxed:
Note sure how this happened but was using pretty much default partition suggestions.
`ext4` + encryption:

screenshot

 * Config: https://0x0.st/H8S4.json
 * Log: https://0x0.st/H8SJ.log
 * cmd log: https://0x0.st/H8Sy.txt

svartkanin:
@Torxed 
> guided.py has always been seen as a standalone executable, something people would look at to get inspiration of how to write a installer script. It was never part of the library.
I know I've mentioned to not do import archinstall in the library, but guided.py and the other examples would be the exception. I'm ok with them living inside the library folder as we use them as python -m <entrypoint>.

I'm happy to move the import structure for the scripts to absolute paths, doesn't make any difference for me. 

I generally would like to move ALL imports to absolute imports rather than relative ones. You mentioned before that there are issues using absolute imports, would you mind sharing what those are? 


>  I cannot back out of "Please chose which greeter to install"

Fixed :) 

> Note sure how this happened but was using pretty much default partition suggestions.

Have to debug this...

svartkanin:
@Torxed regarding the error it doesn't seem to have anything to do with the actual config or archinstall (at least not from what I can see from the error message regarding `sr0`. I would imagine if you run `partprobe` in the console it would probably produce the same error. 

I remove the actual disk error now, which should log if partprobe fails but still continue. I'm not sure if that will potentially cause issues though :/ 

Also we're sending in the path now from the luks call which should definitely eliminate the place you've seen it pop up

codefiles:
While the dependency simple-term-menu is in the dependencies list in `pyproject.toml` it is missing when archinstall is packaged. What is the intended way to include this dependency when archinstall is packaged? I've resolved the error below for myself by packaging python-simple-term-menu and adding it to the archinstall PKGBUILD depends list.

This is the error that occurs when this branch is packaged and archinstall is executed at the command line:

```
  File "/usr/lib/python3.10/site-packages/archinstall/__init__.py", line 5, in <module>
    from .lib import disk
  File "/usr/lib/python3.10/site-packages/archinstall/lib/disk/__init__.py", line 3, in <module>
    from .filesystem import FilesystemHandler
  File "/usr/lib/python3.10/site-packages/archinstall/lib/disk/filesystem.py", line 13, in <module>
    from ..menu import Menu
  File "/usr/lib/python3.10/site-packages/archinstall/lib/menu/__init__.py", line 1, in <module>
    from .abstract_menu import Selector, AbstractMenu, AbstractSubMenu
  File "/usr/lib/python3.10/site-packages/archinstall/lib/menu/abstract_menu.py", line 6, in <module>
    from .menu import Menu, MenuSelectionType
  File "/usr/lib/python3.10/site-packages/archinstall/lib/menu/menu.py", line 6, in <module>
    from simple_term_menu import TerminalMenu
ModuleNotFoundError: No module named 'simple_term_menu'
```

svartkanin:
Good question, I was initially under the impression that the dependencies will be specified in the pyproject.toml and then potentially installed via setuptools in the installation step in PKGBUILD. As @Torxed mentioned for pyparted it will ve added to the iso preinstalled packages directly and therefore already be available.


For any additional abd future dependencies I'm not sure what the preferred way is atm. I'm personally keen on specifying them in the pyproject as that contains all.other documentation as well

codefiles:
I also figured simple-term-menu would get installed into the package directory with the package step in the PKGBUILD since it is listed as a dependency in the pyproject.toml. That does not appear to be so or at least not with the current configuration.

archinstall/lib/installer.py
codefiles: I guess it is harder to merge when code has been moved to different or new files.

archinstall/lib/disk/device_handler.py
codefiles: I think a device not having a label means that it does not have a partition table and/or is not formatted but I did not look into it. Using `parted.freshDisk()` does not appear to touch the disk so the fallback should be safe.

codefiles:
Everything so far is looking good in my testing. Disk operations are faster, great job @svartkanin!

Torxed:
> I also figured simple-term-menu would get installed into the package directory with the package step in the PKGBUILD since it is listed as a dependency in the pyproject.toml. That does not appear to be so or at least not with the current configuration.

So I've applied and am now a approved Trusted User (aka packager) and I can try to package `simple-term-menu` along being co-maintainer of `python-pyparted`. It will take a few days while i sort out PGP key issues.

Until then we have two options:
1. pip install simple-term-menu
2. add `supplies=('simple-term-menu')` as well as package it in `archinstall`'s `PKGBUILD` to make sure any future packaging doesn't cause a conflict.

Third option is the best, where a separate package is created to package and distribute the lib – and source it as a depenency like any other package.

Torxed:
> I remove the actual disk error now, which should log if partprobe fails but still continue. I'm not sure if that will potentially cause issues though :/

This one has bitten me in the ass probably 10+ time now over 2+ years. Lets go for it and put `archinstall` in testing for a solid time before releasing it in the official ISO so people can chip in with bug hunting :)

codefiles:
> Third option is the best, where a separate package is created to package and distribute the lib – and source it as a depenency like any other package.

That is the way I am doing it locally for testing and the way I am in support of. I asked about the intended way because it was missing when archinstall was built with the PKGBUILD and I did not think python-simple-term-menu would get picked up into the official rep from the AUR. That is different now with you as Trusted User. Congrats on the additional responsibility.

archinstall/lib/disk/device_model.py
```diff
@@ -0,0 +1,1033 @@
...
+			'nodatacow': self.nodatacow
+		}
+
+	def as_json(self) -> Dict[str, Any]:
+		return {
+			'name': self.name,
+			'mountpoint': str(self.mountpoint),
+			'compress': self.compress,
+			'nodatacow': self.nodatacow
+		}
```
codefiles, line 450:
```suggestion
	def __dump__(self) -> Dict[str, Any]:
		return {
			'name': str(self.name),
			'mountpoint': str(self.mountpoint),
			'compress': self.compress,
			'nodatacow': self.nodatacow
		}

	def as_json(self) -> Dict[str, Any]:
		return {
			'name': str(self.name),
			'mountpoint': str(self.mountpoint),
			'compress': self.compress,
			'nodatacow': self.nodatacow
		}
```

Fixes https://github.com/archlinux/archinstall/pull/1604#issuecomment-1510363959

archinstall/lib/disk/device_model.py
```diff
@@ -0,0 +1,1033 @@
...
+	def is_root(self, relative_mountpoint: Optional[Path] = None) -> bool:
+		if self.mountpoint:
+			if relative_mountpoint is not None:
+				return self.mountpoint.relative_to(relative_mountpoint) == Path('.')
+			return self.mountpoint == Path('/')
+		return False
+
+	def __dump__(self) -> Dict[str, Any]:
+		return {
+			'name': self.name,
```
codefiles, line 438:
```suggestion
			'name': str(self.name),
```

Fixes https://github.com/archlinux/archinstall/pull/1604#issuecomment-1510363959

archinstall/lib/disk/device_model.py
```diff
@@ -0,0 +1,1033 @@
...
+			'mountpoint': str(self.mountpoint),
+			'compress': self.compress,
+			'nodatacow': self.nodatacow
+		}
+
+	def as_json(self) -> Dict[str, Any]:
+		return {
+			'name': self.name,
```
codefiles, line 446:
```suggestion
			'name': str(self.name),
```

Fixes https://github.com/archlinux/archinstall/pull/1604#issuecomment-1510363959

svartkanin:
@codefiles thanks for testing I'll have a look at the btrfs issue

@Torxed from this PR perspective do we need anything for the dependency packaging? I didn't fully fet the 3rd option, will you be making that change in a different PR and prepping that?

codefiles:
> @codefiles thanks for testing I'll have a look at the btrfs issue

You see my suggestions above that contain a fix, right? 🤪

svartkanin:
Ahh I missed those, only on the app which isn't the best on displaying things

svartkanin:
Alrighty, I finally managed to get the build flow working and also to execute archinstall which should make sure that it starts up without issues. 

@Torxed Do you think we could consider merging this so people could test it on master rather and start contributing towards it? As @codefiles rightfully mentioned, it gets more difficult keeping this massive monster as a shadow of master and for people to actually add changes

Torxed:
@svartkanin Consider it done! Huge thank you to you and the work you've done with this so far!

svartkanin:
Thanks heaps for the patience on this, i know it was a massive change and as with all such changes I'm certain it's not perfect, so I'll keep improving it and making sure it doesn't cause any havoc :)

hkrutzer:
I just tried this and got:
```
Unmounting all partitions: /dev/nvme0n1
Wiping partitions and metadata: /dev/nvme0n1
Creating partitions: /dev/nvme0n1
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/root/archinstall/archinstall/__main__.py", line 19, in <module>
    archinstall.run_as_a_module()
  File "/root/archinstall/./archinstall/__init__.py", line 265, in run_as_a_module
    importlib.import_module(mod_name)
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/root/archinstall/./archinstall/scripts/guided.py", line 274, in <module>
    fs_handler.perform_filesystem_operations()
  File "/root/archinstall/archinstall/lib/disk/filesystem.py", line 55, in perform_filesystem_operations
    device_handler.partition(mod, partition_table=partition_table)
  File "/root/archinstall/archinstall/lib/disk/device_handler.py", line 469, in partition
    self._perform_partitioning(part_mod, modification.device, disk, requires_delete=requires_delete)
  File "/root/archinstall/archinstall/lib/disk/device_handler.py", line 347, in _perform_partitioning
    raise DiskError(f'Unable to determine new partition uuid: {part_mod.dev_path}')
archinstall.lib.exceptions.DiskError: Unable to determine new partition uuid: /dev/nvme0n1p2
```


Disk config is:
```json
{
        "device_modifications": [
            {
                "device": "/dev/nvme0n1",
                "partitions": [
                    {
                        "btrfs": [],
                        "flags": [
                            "Boot"
                        ],
                        "fs_type": "fat32",
                        "length": {
                            "sector_size": null,
                            "total_size": null,
                            "unit": "MiB",
                            "value": 512
                        },
                        "mount_options": [],
                        "mountpoint": "/boot",
                        "obj_id": "3b064b30-ec25-4814-948e-3edf6fd26ca4",
                        "start": {
                            "sector_size": null,
                            "total_size": null,
                            "unit": "MiB",
                            "value": 1
                        },
                        "status": "create",
                        "type": "primary"
                    },
                    {
                        "btrfs": [],
                        "flags": [],
                        "fs_type": "ext4",
                        "length": {
                            "sector_size": null,
                            "total_size": {
                                "sector_size": null,
                                "total_size": null,
                                "unit": "B",
                                "value": 500107862016
                            },
                            "unit": "Percent",
                            "value": 100
                        },
                        "mount_options": [],
                        "mountpoint": "/",
                        "obj_id": "56c07541-60d8-44c2-b77b-760f829f9340",
                        "start": {
                            "sector_size": null,
                            "total_size": null,
                            "unit": "MiB",
                            "value": 513
                        },
                        "status": "create",
                        "type": "primary"
                    }
                ],
                "wipe": true
            }
        ]
    }
```
I also got a partitioning related error prior to this change and decided to wait for this to be merged. I'm not sure if it was the same error though.

```bash
# git describe
v2.5.4-19-g00b0ae7b
```

svartkanin:
Thanks for reporting this, I'll have a look.
@hkrutzer would you be able to run a `lsblk -J -O` after you got the error and paste the output

svartkanin:
> Consider it done! Huge thank you to you and the work yo

@Torxed send you a message on discord instead :)

Torxed:
> > Consider it done! Huge thank you to you and the work yo
> 
> @Torxed send you a message on discord instead :)

awesome! I'll be back infront of my computer after today :D

hkrutzer:
Made a new issue https://github.com/archlinux/archinstall/issues/1759 as to not turn this comment section into an issue :)

PR #1608, opened 2022-12-27
French translation update

roxfr:
Hello,
Here is the updated French translation.
Regards,
Roxfr


Torxed:
Thank you for the updated language!

PR #1617, opened 2023-01-09
Use partx instead of partprobe to avoid the kernel removing a blockdevice entry

lavafroth:
- This fix issue: #1557 

## PR Description:

Occasionally `partprobe` causes the kernel to remove a partition entry from its list of block devices.
This commit replaces the use of the `partprobe` command with `partx`. Refer to the following for more details:
 - https://access.redhat.com/solutions/510843
 - https://serverfault.com/questions/501493/partprobe-will-not-work-i-have-tried-everything

## Tests and Checks

### DIFF
--- a/archinstall/lib/disk/partition.py
+++ b/archinstall/lib/disk/partition.py
@@ -420,9 +420,9 @@ class Partition:
 	def partprobe(self) -> bool:
 		try:
 			if self.block_device:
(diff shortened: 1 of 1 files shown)


Torxed:
Any reasons why you closed this?
I think perhaps investigating the use of `partx` in general might be a good idea?

lavafroth:
I heard the /dev/sdaX issue was fixed. It is true, partx does not seem to be affected by the partprobe issue.

PR #1626, opened 2023-01-21
add polkit to packages since it won't launch without it

zkade0:
I believe polkit is required for sway to launch properly if using the sway profile.

### DIFF
--- a/profiles/sway.py
+++ b/profiles/sway.py
@@ -15,6 +15,7 @@ __packages__ = [
 	"slurp",
 	"pavucontrol",
 	"foot",
+	"polkit", ## Sway wont launch and throw errors if polkit is not installed. 
 ]
 
 


Torxed:
My only issue with adding this, is that it won't allow users to choose between `seatd` or `polkit` for the `seat`: https://wiki.archlinux.org/title/Sway#Starting

Do users care a lot or can we default to one of them without it being a touchy subject? :)

bd-g:
> My only issue with adding this, is that it won't allow users to choose between `seatd` or `polkit` for the `seat`: https://wiki.archlinux.org/title/Sway#Starting
> 
> 
> 
> Do users care a lot or can we default to one of them without it being a touchy subject? :)

Considering the general user base for Arch Linux, I'd say they care 😄 I know lots that use both. Perhaps having two profiles, sway-polkit and sway-seatd?

bd-g:
> My only issue with adding this, is that it won't allow users to choose between `seatd` or `polkit` for the `seat`: https://wiki.archlinux.org/title/Sway#Starting
> 
> 
> 
> Do users care a lot or can we default to one of them without it being a touchy subject? :)

Considering the general user base for Arch Linux, I'd say they care 😄 I know lots that use both. Perhaps having two profiles, sway-polkit and sway-seatd?

Torxed:
> Perhaps having two profiles, sway-polkit and sway-seatd?

Technically speaking that could work, but I think minimizing entries is still a desire for us. So we need a menu option in `_prep_function` that asks which one would be nice :)

Torxed:
Closing this in favor of #1661

Issue #1621, opened 2023-01-10
Remove i3-wm and i3-gaps configuration choice


ettoc00:
Hi, I wanted to let you know that i3-wm and i3-gaps have been merged: https://github.com/i3/i3/pull/5241. This means that the choice between the two versions is no longer necessary and should be removed. Also, this would resolve #611.

Torxed:
Fixed in #1635

PR #1635, opened 2023-01-31
Removed i3-gaps configuration choice

ettoc00:
Fix to issue #1621

Added i3-wm to the profile packages and removed i3-gaps configuration choice.


Torxed:
Thank you for this contribution, I really appreciate it!

PR #1636, opened 2023-01-31
Resolving Failed to read disk "/dev/sdaX" with lsblk.

damianodoug:
- This fix issue:  #1557 for instance.

## PR Description:
When we are running the installation an error appears, preventing the installation.
```
Could not decode JSON: lsblk: /dev/sda1: not a block device
{
   "blockdevices": [

   ]
}
```
If we run the command `lsblk` we get can get a similar result:
```
NAME        MAJ:MIN RM   SIZE RO TYPE MOUNTPOINTS
loop0         7:0    0 705.6M  1 loop /run/archiso/airootfs
sda           8:0    1  29.9G  0 disk
└─sda2        8:2    1    15M  0 part
nvme0n1     259:0    0 476.9G  0 disk
├─nvme0n1p1 259:1    0   100M  0 part
├─nvme0n1p2 259:2    0   128M  0 part
├─nvme0n1p3 259:3    0   475G  0 part
└─nvme0n1p4 259:4    0   1.7G  0 part
```
**sda** is our installation disk, the **MAJ** value is **8** and, it's the problem, in my code I have add a option to run the `lsblk`excluding this device.

## Tests and Checks
Maybe it's not the best way to resolve it, but for this moment is working.

### DIFF
--- a/archinstall/lib/disk/blockdevice.py
+++ b/archinstall/lib/disk/blockdevice.py
@@ -137,6 +137,8 @@ class BlockDevice:
 		# Doesn't harm us, but worth noting in case something weird happens.
 		try:
 			output = SysCommand(f"parted -s --machine {self._path} print free").decode('utf-8')
+			if not output:
+				output = SysCommand(f'lsblk -a -e 8 --json -b -o+SIZE,PTTYPE,ROTA,TRAN,PTUUID {self._path}').decode('UTF-8')
 			if output:
 				free_lines = [line for line in output.split('\n') if 'free' in line]
 				sizes = []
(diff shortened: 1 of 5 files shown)


Torxed:
I don't mind patching this in any way shape or form at the moment hehe.
I get that it will exclude [SCSI block devices](https://www.kernel.org/doc/Documentation/admin-guide/devices.txt) but won't it also hide the partitions (sharing the same MAJOR code)?

I suspect this fix will be very unique to those having issues on SCSI-based devices, and as long as they won't install to one of those this should work.

But I'm still contemplating fixing whatever the issue is without excluding things.
A workaround would be to exclude the MAJOR 8 with a MINOR of 0, and preserve the partitions if possible.

From all the logs (unless I missed one) the only issue should be https://github.com/archlinux/archinstall/blob/126f56169d6faacd492b0b0f10c18762cf42a2ba/archinstall/lib/disk/partition.py#L181

So perhaps limiting the patch to that section would avoid issues elsewhere?
Either way, getting back into combating this issue so had a look on this PR and it appears to work for those that have the issue - but I'm wondering if it will stop working for the other half that isn't having the issue?

Torxed:
Closing this as it would ignore more devices than it should. Another PR (https://github.com/archlinux/archinstall/pull/1645) has been introduced which simply returns empty information for devices unknown to `lsblk`.

PR #1655, opened 2023-02-27
Replacing flit

Torxed:
- This replaces: #1478

## PR Description:

As discussed in https://github.com/pypa/flit/issues/522 and https://github.com/archlinux/archinstall/issues/832 the future of flit doesn't appear appealing to me and others.

So while we're ahead, I'm replacing `flit` with just `setuptools` and relying on it to build for us.
This allows for external dependencies as well via:
```toml
[project]
dependencies = [
    "pyparted",
]
```
or `setup.cfg`
```toml
[options]
install_requires =
    pyparted
```

And if it's a build-only specific dependency put it in:
```toml
[build-system]
requires = ["pyparted"]
```

A good reference on how to include the profiles and stuff was from here: https://godatadriven.com/blog/a-practical-guide-to-setuptools-and-pyproject-toml/

## Tests and Checks


Torxed:
I've verified that both `/examples/` and `/profiles/` contain all the profiles.

svartkanin:
@Torxed I didn't realize this happened until now that I was trying to merge in master into the pyparted branch. Isn't this sort of a move backwards? Wouldn't it rather make sense to use `poetry` instead as it has full support for `pyproject.toml` and comes with a lot of other neat futures as well we could utilize?

Torxed:
> with a lot of other neat futures as well we could utilize?

Until we do, the move backwards is just a simplification which would hopefully be less prone to regressions and issues as it's `pypa`'s reference implementations *(knock on wood)*. I simply chose `python-setuptools` as the PEP517 backend because it's the simplest one and also to keep it in the [pypa](https://github.com/pypa) family:
 * https://github.com/pypa/build
 * https://github.com/pypa/installer
 * https://github.com/pypa/setuptools

Anything specific we'd like from poetry? I think we evaluated poetry in the very beginning :)

svartkanin:
It makes sense to have them all operating from the same set of tools. Some issues I have experienced with `setuptools` were primarily related to dependency management. It run into conflicts and wasn't really able to handle different versions of the same packages properly. Setuptools is a more complex tool to handle things which needs some familiarity  as well probably.

svartkanin:
Maybe I'm missing something but do we actually need a `setup.cfg` file? `setuptools` seems to be able to handle `pyproject.toml` just fine, so why use them both https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html ? 

AFAIK setuptools is not able to install dependencies only, so it will be (a bit) more cumbersome for devs to get started as they have to dig out the required packages from the toml and then install them. 

#### Setup with Poetry
```
git clone...
poetry env use python3.10  # use or create new virt env
poetry install   # install all dependencies in current virt env
```

#### Build project
```poetry build  # builds wheel```

#### Install project
```poetry install```

:)

PR #1657, opened 2023-02-28
Use setdefault for sys-language and sys-encoding

lavafroth:
….

- This does not fix an issue 

## PR Description:
For `sys-encoding` and `sys-language`, the setdefault function of the arguments dictionary is used instead of manually checking if key-value pair exists. Minimal changes.

## Tests and Checks


Torxed:
Thank you for this quick tweak!

PR #1659, opened 2023-02-28
Make it easier to save configuration files

### LINKED ISSUE #1652: "Save configuration" Improve UX
*Feature Request*

I have some time in the next week, I may try and implement this depending on how many roadblocks I hit. Currently, to save a configuration, you are prompted to enter a directory. If it isn't valid, you have to scratch your head wondering what the exact paths are, because you can't discover any directory without exiting `archinstall`. 

It would be great if there was a sort of tab completion, or even filesystem traversal similar to `ncdu` to find a directory to save the configuration in.

bd-g:
This addresses resolves #1652 

## PR Description:

This PR seeks to made it easier to save configuration files. Previously, one had to enter a directory path by memory. This PR simply uses `find` to recursively look up possible save locations, and then uses the built in multi-select menu to allow the user to select where they want to save the files.

## Outstanding Concerns

1. Is `dirs_to_exclude` complete enough? I based it off a standard archlinux filesystem, but there may be other options that are worth including.
2. If the user has a mounted filesystem with a large amount of directories, the find command could run for a while finding all the options. The user's only option to exit during this process would be to exit the entire program (Ctrl-C), potentially losing their progress. 
\
We could intercept a SIGINT signal and kill the underlying `find` process and return to the main menu, but this may be premature optimization, considering I tested on a few filesystems with several thousand directories and it didn't take more than a couple seconds for the command to run. Let me know your thoughts.

## Tests and Checks
- [x] I have tested the code using manual tmp functions I wrote!
- [ ] I have tested the code using the built ISO!


archinstall/lib/user_interaction/save_conf.py
```diff
@@ -1,9 +1,13 @@
 from __future__ import annotations
 
+import archinstall
```
Torxed, line 3: Avoid importing like this, as it's not a relative import - it might import from outside sources.

Torxed (changes requested):

bd-g:
Three changes (see commit messages for details), the first commit addresses your requested changes.

archinstall/lib/user_interaction/save_conf.py
Torxed: Yea, I know vscode and other IDE's love to add `import archinstall` because it auto detects that this is where `log` is exposed hehe.

bd-g:
Updates? (Only asking since contributing guidelines say to recomment if thread goes stale)

Torxed:
ETA this weekend before i can merge bigger PR's :)

Torxed:
Looks good, but there's a slight tweak needed.
If the user press `Esc` to back out of saving the config the chosen location becomes `None` (maybe even `"None"`?).
screenshot

And if manually writing `/home/test/` it also becomes `None` because the location doesn't exist (which to be fair, it doesn't). We should allow generic input I think, as there's no "create folder".

Torxed:
hmm, I might be stuck in the old times.
It used to be that the folder structure was created on input, this does not appear to be the case any longer:
 * https://github.com/archlinux/archinstall/blob/c3862c5779194f5e93f9fd2518bb15706c93ad2b/archinstall/lib/configuration.py#L113-L120

That means you can probably skip my comment about creating a path, but the `Esc` functionality needs to be implemented so that it doesn't trigger the question above.

bd-g:
Updates? Linting should be fixed in #1676

Torxed (approved):

Torxed:
> Updates? Linting should be fixed in #1676

Having guests over this weekend so most work will be done on wednsday.
But IIRC I tested this pretty well last week prior to releasing the latest version so I merged the changes from upstream `master` just so that linting works.

Great work and merged! :)

svartkanin:
@Torxed @bd-g this runs quite long for me until I get a selection, is it necessary to to eagerly query all these directories?

svartkanin:
Also I can't really choose `/tmp` anymore as the filter isn't really able to filter that out properly...
I'm not sure but to me it seems a bit of an overkill to literally put every single filesystem directory into the otpions

bd-g:
> @Torxed @bd-g this runs quite long for me until I get a selection, is it necessary to to eagerly query all these directories? 

One of my original questions on the PR was asking for suggestions on directories to exclude. Do you have other suggestions?

I originally implemented it using fzy, so that you could begin searching while results were still being found. However, I dropped the dependency in order to use the native menu.

svartkanin:
Hmm I'm not entirely sure if excluding is a good way though, as it will basically remove options for the user to choose from. 

I wonder how difficult it would be to do an autocomplete instead, basically emulate the tab-tab in a way and then give suggestions

bd-g:
Not terrible, just maybe not possible to do (efficiently) within the existing menus. That was the constraint I was trying to stay within, but if we remove that constraint we could use other purpose built tools. 

I know an original selling point for the repo was "no external dependencies", so I tried to stay within that. However, now that we have introduced pyparted, maybe that requirement has loosened.

Torxed:
It feels like a trivial thing to sort out tho without a dependency? I'm not against dependencies but if the task is small, it feels like more maintanance to adopt a library

bd-g:
> It feels like a trivial thing to sort out tho without a dependency? I'm not against dependencies but if the task is small, it feels like more maintanance to adopt a library

I agree the added maintenance is not what we want. I'm just thinking edge cases, not sure how gnarly it gets there, I just have gone down rabbit holes with filesystem issues. I'll take a stab at it sometime in the next few days. 

Do you have any preferences or suggestions on the UI?

Torxed:
I would almost hand over the UI suggestions to @svartkanin, he got a black belt in UI work at this point :D

svartkanin:
This is me just DDG and the first entry https://stackoverflow.com/questions/5637124/tab-completion-in-pythons-raw-input

PR #1661, opened 2023-03-01
Sway profile - select seat

bd-g:
This is perhaps a more fleshed out option to resolve #1626 

## PR Description:

This PR adds a selection menu when selecting the sway profile that allows the user to either install polkit or seat. One of the two is required for sway to run, so we should prompt the user to choose one when installing sway.

## Tests and Checks
- [x] I have tested the code! I ran on the v2.5.3 ISO and testing the following installation flows:
    - Installing sway profile with polkit
    - Installing sway profile with seatd
    - Selecting sway profile with seatd/polkit, then clearing selection
    - Selecting sway profile with seatd/polkit, then clearing selection and selecting KDE

### DIFF
--- a/archinstall/lib/user_interaction/general_conf.py
+++ b/archinstall/lib/user_interaction/general_conf.py
@@ -174,7 +174,10 @@ def select_profile(preset) -> Optional[Profile]:
 			storage['profile_minimal'] = False
 			storage['_selected_servers'] = []
 			storage['_desktop_profile'] = None
+			storage['sway_sys_priv_ctrl'] = None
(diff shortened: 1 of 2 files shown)


bd-g:
Updates? (Only asking since contributing guidelines say to recomment if thread goes stale)

Torxed:
I've been meaning to test this after the discussions in #1626
ETA this weekend :)

Torxed:
Tested and it works.
I have no idea how to get out of sway so that was an interesting test :P

Torxed:
Great work btw!

PR #1668, opened 2023-03-06
Added parted sector unit

StrangeGirlMurph:
## PR Description

Added the sector unit for parted. See: https://www.gnu.org/software/parted/manual/html_node/unit.html
The sector unit is IMO actually the most useful one. It makes creating partitions way easier. Instead of having to calculate the MB (i.e.) for the start and end location of the partition one can just work with the already given start and end sectors of the other partitions.

Maybe @Torxed left them out because he thought that sectors were the default unit.
But actually is kB the default (at least if not set differently) https://users.cs.fiu.edu/~downeyt/cgs3767/parted.html

Another alternative instead of adding the sector unit there would be setting the default unit to sectors in parted which is probably what you had in mind in the first place and what I would also recommend!

This PR partly also relates to https://github.com/archlinux/archinstall/issues/1669

## Todo
Also I think a better prompt text would be very helpful. I had to look up a lot of stuff before understanding what was going on. The "block number" was what confused me. I thought that meant the sector.

Currently ([code](https://github.com/archlinux/archinstall/blob/master/archinstall/lib/user_interaction/partitioning_conf.py#L211)):
> Enter the start sector (percentage or block number, default: {}):
> Enter the end sector of the partition (percentage or block number, ex: {}):

My idea:
> Enter the start sector (in parted units: 's', 'GB', '%', etc. ; default: {}):
> Enter the end sector of the partition (in parted units: 's', 'GB', '%', etc. ; ex: {}):

I am very new to the code base and have no idea how one would go about changing that in the locales and everything. If you like my idea or some variation of it please just commit it :)

## Tests and Checks

### DIFF
--- a/archinstall/lib/disk/validators.py
+++ b/archinstall/lib/disk/validators.py
@@ -7,10 +7,12 @@ def valid_parted_position(pos :str) -> bool:
 	if pos.isdigit():
 		return True
 
-	if pos.lower().endswith('b') and pos[:-1].isdigit():
+	pos_lower = pos.lower()
+
+	if (pos_lower.endswith('b') or pos_lower.endswith('s')) and pos[:-1].isdigit():
 		return True
 
-	if any(pos.lower().endswith(size) and pos[:-len(size)].replace(".", "", 1).isdigit()
+	if any(pos_lower.endswith(size) and pos[:-len(size)].replace(".", "", 1).isdigit()
 			for size in ['%', 'kb', 'mb', 'gb', 'tb', 'kib', 'mib', 'gib', 'tib']):
 		return True
 


Torxed:
Great catch, indeed I forgot/missed/didn't think sectors were used that much in the wild.
Can't remember honestly because it was years ago that code started to take form *(even tho it's been tweaked over the years)*.

Thank you for adding it in!

Torxed:
Using sectors as default feels like a risky move, at least all the people I've worked with would expect MB or GB as the default :)

StrangeGirlMurph:
Okay :) And the prompt text?
I feel like "Enter the start sector" basically implicates that the default unit is sectors. I think my proposal would clarify that. Should I create a separate Issue for that?

Torxed:
"Enter start location" would work in my world, but input is welcome here :)

StrangeGirlMurph:
I mentioned my idea in the [initial comment](https://github.com/archlinux/archinstall/pull/1668#issue-1611804594). I think just mentioning some of the units would make it clear what archinstall is expecting.
> Enter the start location (in parted units: 's', 'GB', '%', etc. ; default: {}):

Torxed:
As long as it fits on the screen I'm all for it!
And I guess screen real estate is not a huge issue these days after the new TUI was implemented :D

Mind throwing a PR with the updated/suggested text? :)

StrangeGirlMurph:
Okay, cool! 
No but like I said in the initial comment...
> I am very new to the code base and have no idea how one would go about changing that in the locales and everything. If you like my idea or some variation of it please just commit it :)

I would but I don't know how. Someone else would have to do that or explain it to me.

Torxed:
No worries, I can do it in a bit :)

tl;dr tho if you're curious:
`_("Enter start location")` will automatically make it translatable.

After that, it's not *necessary* to put the string in the translation files, archinstall will work either way. But it's appreciated if it is. But language contributors are usually pretty quick on finding and updating newly added strings :)

StrangeGirlMurph:
Thank you!

Issue #1673, opened 2023-03-12
archinstall won't start due to TypeError


Lautron:
# Error
When I execute `archinstall` I get the following error:

```
...
File "/usr/lib/python3.10/site-packages/archinstall/lib/packages/packages.py", line 115, in installed_package
  return LocalPackage(**package_info)
TypeError: LocalPackage.__init__() got an unexpected keyword argument 'warning'
```
# What I tried
I looked at the code of `installed_package` and found it executes `pacman -Q --info {package}`, so I tried to execute that with archlinux-keyring to see what the output was like.

I executed `pacman -Q --info archlinux-keyring` and got the following output
```
warning: config file /etc/pacman.conf, line 19: directive 'SyncFirst' in section 'options' not recognized
Name 	: archlinux-keyring
Version : 20230225-1
...
```
# Why this seems to be happening
## Code
https://github.com/archlinux/archinstall/blob/8f6cc07062968b259bebd346521ef685c16f89dc/archinstall/lib/packages/packages.py#L105-L115
## Explanation 
Because the line `warning: config file /etc/pacman.conf, line 19: directive 'SyncFirst' in section 'options' not recognized` I get as part of the output of `pacman -Q --info {package}` has a colon it is being interpreted as a key value pair.

# Possible fix
Ignore all lines that start with 'warning'

Torxed:
Nice catch, I proposed a more aggressive fix in your PR. Would like to hear your thoughts on it :)

PR #1674, opened 2023-03-12
Check if line begins with 'warning' before adding it to package info

Lautron:
- This fix issue: #1673

## PR Description:
Skip lines from `pacman -Q --info` that begin with 'warning' before adding them to package info to avoid getting a TypeError.
The problem is explained with more detail on #1673 

## Tests and Checks
  - Tested on live ISO

### DIFF
--- a/archinstall/lib/packages/packages.py
+++ b/archinstall/lib/packages/packages.py
@@ -106,7 +106,7 @@ def installed_package(package :str) -> LocalPackage:
 	package_info = {}
 	try:
 		for line in run_pacman(f"-Q --info {package}"):
-			if b':' in line:
+			if b':' in line and not line.startswith(b'warning'):
 				key, value = line.decode().split(':', 1)
 				package_info[key.strip().lower().replace(' ', '_')] = value.strip()
 	except SysCallError:
--- a/archinstall/lib/packages/packages.py
+++ b/archinstall/lib/packages/packages.py
@@ -1,3 +1,4 @@
+import dataclasses
 import json
 import ssl
 from typing import Dict, Any, Tuple, List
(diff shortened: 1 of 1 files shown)


Torxed:
Nice catch. I think a long term solution would be to modify the fix to fix all future possible "odd outputs" from `pacman`.
Would you mind changing the fix to:

```diff
- return LocalPackage(**package_info)
+ import dataclasses
+ return LocalPackage({field.name: package_info.get(field.name) for field in dataclasses.fields(LocalPackage)})
```

I'm no expert, but I think this would solve it.
It would iterate the fields we've defined in `LocalPackage` and only populate those fields. Any additional fields (added inte future in the pacman project) would be ignored, including `warning:` or `info:` or whatever it outputs :)

Lautron:
> Nice catch. I think a long term solution would be to modify the fix to fix all future possible "odd outputs" from `pacman`. Would you mind changing the fix to:
> 
> ```diff
> - return LocalPackage(**package_info)
> + import dataclasses
> + return LocalPackage({field.name: package_info.get(field.name) for field in dataclasses.fields(LocalPackage)})
> ```
> 
> I'm no expert, but I think this would solve it. It would iterate the fields we've defined in `LocalPackage` and only populate those fields. Any additional fields (added inte future in the pacman project) would be ignored, including `warning:` or `info:` or whatever it outputs :)
> 
> I can manually change the patch but to keep the credit of finding and fixing the bug you're free to change if you want to :)

Hi there! Thanks for your feedback. 

I think your suggestion to make the fix more future-proof is great. I will definitely keep that in mind for future contributions to open source.

I have just made the changes you suggested in a new commit. Thank you again for your help and for being such a supportive member of the open source community.

Torxed (approved):

Torxed:
Thank you for this one! Merged! :)

PR #1677, opened 2023-03-14
Add gwenview spectacle and kcalc to "KDE"

kassindornelles:
## PR Description:
We are adding those in this PR for the "KDE" option
Gwenview - is the default KDE software to open image files
Spectacle - is the default KDE software to make screenshots (without it "printscreen" key does nothing at all)
Kcalc - is the default KDE "calculator"

Always good to have basic apps pre-installed

## Tests and Checks

### DIFF
--- a/profiles/kde.py
+++ b/profiles/kde.py
@@ -13,6 +13,9 @@ __packages__ = [
 	"sddm",
 	"plasma-wayland-session",
 	"egl-wayland"
+	"spectacle"
+	"gwenview"
+	"kcalc"
 ]
 
 


svartkanin:
I'd argue that these should not be part of the basic installation. The packages listed for each profile are the bare minimum for it to work and that's all that it should do.

As you said, it's applications therefore there's no need for them to ve present on a new installation and some users may not want them. If some do, they can easily be added as additional packages in the menu

Torxed:
I agree, the exceptions being a terminal and file handler to get users going. At least that's my view on what the generic user (advanced and "beginner") would need to move around and add the things they need.

I personally prefer maim for screenshots, but it's not one of the first things i need to get going.

I appreciate the thought behind this PR and I don't want to come off as brisk or negative, but I will close this PR as I don't consider these applications essential.

kassindornelles:
> I agree, the exceptions being a terminal and file handler to get users going. At least that's my view on what the generic user (advanced and "beginner") would need to move around and add the things they need.
> 
> I personally prefer maim for screenshots, but it's not one of the first things i need to get going.
> 
> I appreciate the thought behind this PR and I don't want to come off as brisk or negative, but I will close this PR as I don't consider these applications essential.

i thought this PR would be acceptable since the GNOME options comes with a bunch of apps pre-installed, so maybe it might need a fix too

kassindornelles:
> It does, GNOME has been one of those thorns in the side where the default package/group contains way to many things. And we try our best to use default packages rather than maintaining application lists as far as possible. KDE > GNOME in that regard.

any special reason for why vim and vim-runtime are included in every single installation?

Torxed:
> any special reason for why vim and vim-runtime are included in every single installation?

Either it's because `vim` is a dependency of another package that was chosen as it's a dependency for 133 other packages. Or when installing the desktop profile we do give a minimal help package in the shape of text editors. You'll also find `nano` installed as we don't want to get into the whole crusade about which editor is best: https://github.com/archlinux/archinstall/blob/0fa52a5424e28ed62ef84bdc92868bbfc434e015/profiles/desktop.py#L17-L28

There's also the `python -m archinstall minimal` example that installs `nano` and serves only as a example: https://github.com/archlinux/archinstall/blob/870da403e79ab50350803b45f200e0b272334989/examples/minimal.py#L30

I would like to convert the `desktop` packages to a smart-detect list, for instance `iwd` isn't always needed - and we could probably detect if wifi is even present. Either way, that is the direction we're taking but not there yet :)

PR #1679, opened 2023-03-14
Fix `exit_code`

### LINKED ISSUE #1678: Exit code of `SysCommand()` is exit status indication
The `exit_code` of a `SysCommand()` call is not an exit code but an exit status indication. [1] An exit status indication is "a 16-bit number, whose low byte is the signal number that killed the process, and whose high byte is the exit status (if the signal number is zero); the high bit of the low byte is set if a core file was produced." [2]

Below is an example of how to use `os.waitstatus_to_exitcode()` [3] to obtain the exit code. In the example, `grep` is executed without any arguments and so it returns an exit code of `2`. [4]

```python
import os
import pty

pid, fd = pty.fork()

if pid == 0:
    os.execv('/usr/bin/grep', ['grep'])
else:
    pid, wait_status = os.waitpid(pid, 0)
    exit_code = os.waitstatus_to_exitcode(wait_status)
    print(f'wait status: {wait_status}')
    print(f'exit code: {exit_code}')
```

**Output:**

```
wait status: 512
exit code: 2
```

### Findings

- Relevant lines of code in archinstall:

https://github.com/archlinux/archinstall/blob/f0a6adb96dd388d62b2f6683709a8e310e6727c8/archinstall/lib/general.py#L357-L361

- Exit code in archinstall error message:

```
>>> import archinstall
>>> archinstall.SysCommand('grep')
  File "<stdin>", line 1, in <module>
...
archinstall.lib.exceptions.SysCallError: ['/usr/bin/grep'] exited with abnormal exit code [512]: b"Usage: grep [OPTION]... PATTERNS [FILE]...\r\nTry 'grep --help' for more information.\r\n"
```

- Exit code on the command line:

```
$ grep
Usage: grep [OPTION]... PATTERNS [FILE]...
Try 'grep --help' for more information.
$ echo $?
2
```

[1] https://docs.python.org/3/library/os.html#os.waitpid
[2] https://docs.python.org/3/library/os.html#os.wait
[3] https://docs.python.org/3/library/os.html#os.waitstatus_to_exitcode
[4] https://man.archlinux.org/man/grep.1#EXIT_STATUS

codefiles:
Fixes #1678

### DIFF
--- a/archinstall/lib/general.py
+++ b/archinstall/lib/general.py
@@ -355,10 +355,12 @@ class SysCommandWorker:
 			if self.ended or (got_output is False and pid_exists(self.pid) is False):
 				self.ended = time.time()
 				try:
-					self.exit_code = os.waitpid(self.pid, 0)[1]
+					wait_status = os.waitpid(self.pid, 0)[1]
+					self.exit_code = os.waitstatus_to_exitcode(wait_status)
 				except ChildProcessError:
 					try:
-						self.exit_code = os.waitpid(self.child_fd, 0)[1]
+						wait_status = os.waitpid(self.child_fd, 0)[1]
+						self.exit_code = os.waitstatus_to_exitcode(wait_status)
 					except ChildProcessError:
 						self.exit_code = 1
 


Torxed:
Works like a charm! Thank you for this one!

PR #1685, opened 2023-03-19
Change exit status indications to exit codes

### LINKED ISSUE #1684: Exit status indications require change to exit codes
With the `exit_code` fix in #1679 merged, any reference to or use of an exit status indication (wait status) requires a change to exit code.

### Example

This is an example to demonstrate the relation of exit code and wait status. It passes `exit` with an exit code to `os.system()` and is returned a wait status. [1] Then passes a wait status to `os.waitstatus_to_exitcode()` and is returned an exit code.

```
>>> import os
>>> os.system('exit 1')
256
>>> os.system('exit 2')
512
>>> os.system('exit 32')
8192
>>> os.waitstatus_to_exitcode(256)
1
>>> os.waitstatus_to_exitcode(512)
2
>>> os.waitstatus_to_exitcode(8192)
32
```

### Relevant code

- https://github.com/archlinux/archinstall/blob/79eb6bba627c4c70f24a265ac0b2ef9cbfbbd211/archinstall/lib/disk/helpers.py#L233-L236

  The exception here would be caught if the `SysCommand()` call in the `blkid()` function returns a non-zero exit code. If that occurs then the exit code is being checked for either `512` or `2`. The `blkid` man page contains documentation of the exit code `2` but has no mention of `512`. [2] An exit code of `2` produces a wait status of `512`. If this check for an exit code of `512` arose from `exit_code` containing a wait status before the fix then `512` can safely be removed from the check.

- https://github.com/archlinux/archinstall/blob/79eb6bba627c4c70f24a265ac0b2ef9cbfbbd211/archinstall/lib/disk/partition.py#L186-L190

  The comment here should be updated. An exit code of `32` produces a wait status of `8192`. The `lsblk` man page explains an exit code of `32` as "none of specified devices found". [3]

- https://github.com/archlinux/archinstall/blob/79eb6bba627c4c70f24a265ac0b2ef9cbfbbd211/archinstall/lib/disk/partition.py#L635-L642

  This section of code contains a logic error. If the `umount` command in the `SysCommand()` call returns a non-zero exit code then a `SysCallError` will be raised before any code below it can be executed. The exit code conditional that follows will therefore only check when the exit code was zero. The particular `SysCallError` that is inside the conditional will then never get raised and is not serving the intended purpose.

  The exit code is checked to be less than `8000` which appears to have arose from `exit_code` reporting a wait status. The man page for `umount` provides no details on the exit code of the command but a review of the `util-linux` code confirms that it uses the same return codes as `mount`. [4] The highest exit code documented by the `mount` man page is `64` and a review of the `umount` code reveals that `int main()` returns at max `255`. [5][6]

  This could be reduced to just the `SysCommand()` call since any non-zero exit code will already raise a `SysCallError`.

- https://github.com/archlinux/archinstall/blob/79eb6bba627c4c70f24a265ac0b2ef9cbfbbd211/archinstall/lib/luks.py#L109-L118

  If the `cryptsetup` command in the `SysCommand()` call returns a non-zero exit code a `SysCallError` will be raised and caught by the exception that follows. If this occurs a conditional then checks for an exit code of `256`. The `cryptsetup` man page section on exit codes does not mention an exit code of `256`. [7] An exit code of `1` produces a wait status of `256` so this may have arose from `exit_code` reporting a wait status.

- https://github.com/archlinux/archinstall/blob/79eb6bba627c4c70f24a265ac0b2ef9cbfbbd211/archinstall/lib/systemd.py#L96-L101

  The comment here if uncommented would check for an exit code of `256` when a `SysCallError` has been raised by the `SysCommand()` call that is passed a `systemd-run` command. The `systemd-run` man page section on exit codes does not mention an exit code of `256`. [8] An exit code of `1` produces a wait status of `256` so this appears to have arose from `exit_code` reporting a wait status. The commented lines do not look relevant anymore and should be deleted if they are not.

### References

[1] https://docs.python.org/3/library/os.html#os.system
[2] https://man.archlinux.org/man/blkid.8#EXIT_STATUS
[3] https://man.archlinux.org/man/lsblk.8#EXIT_STATUS
[4] https://github.com/util-linux/util-linux/blob/8cf6c5075780598fe3b30e7a7753d8323d093e22/libmount/src/libmount.h.in#L244-L304
[5] https://man.archlinux.org/man/mount.8#EXIT_STATUS
[6] https://github.com/util-linux/util-linux/blob/8cf6c5075780598fe3b30e7a7753d8323d093e22/sys-utils/umount.c#L648
[7] https://man.archlinux.org/man/core/cryptsetup/cryptsetup.8.en#RETURN_CODES
[8] https://man.archlinux.org/man/systemd-run.1#EXIT_STATUS

codefiles:
Fixes #1684

### DIFF
--- a/archinstall/lib/disk/partition.py
+++ b/archinstall/lib/disk/partition.py
@@ -186,8 +186,7 @@ class Partition:
 			try:
 				output = SysCommand(f"lsblk --json -b -o+LOG-SEC,SIZE,PTTYPE,PARTUUID,UUID,FSTYPE {self.device_path}").decode('UTF-8')
 			except SysCallError as error:
-				# It appears as if lsblk can return exit codes like 8192 to indicate something.
-				# But it does return output in stderr so we'll try to catch it minus the message/info.
+				# Get the output minus the message/info from lsblk if it returns a non-zero exit code.
 				output = error.worker.decode('UTF-8')
 				if '{' in output:
 					output = output[output.find('{'):]
@@ -632,14 +631,7 @@ class Partition:
 		return False
 
 	def unmount(self) -> bool:
-		worker = SysCommand(f"/usr/bin/umount {self._path}")
-		exit_code = worker.exit_code
-
-		# Without to much research, it seams that low error codes are errors.
-		# And above 8k is indicators such as "/dev/x not mounted.".
-		# So anything in between 0 and 8k are errors (?).
-		if exit_code and 0 < exit_code < 8000:
-			raise SysCallError(f"Could not unmount {self._path} properly: {worker}", exit_code=exit_code)
+		SysCommand(f"/usr/bin/umount {self._path}")
 
 		# Update the partition info since the mount info has changed after this call.
 		self._partition_info = self._fetch_information()
--- a/archinstall/lib/systemd.py
+++ b/archinstall/lib/systemd.py
@@ -97,8 +97,6 @@ class Boot:
 			shutdown = SysCommand(f'systemd-run --machine={self.container_name} --pty shutdown now')
 		except SysCallError as error:
 			shutdown_exit_code = error.exit_code
-			# if error.exit_code == 256:
-			# 	pass
 
 		while self.session.is_alive():
 			time.sleep(0.25)
--- a/archinstall/lib/disk/helpers.py
+++ b/archinstall/lib/disk/helpers.py
@@ -233,7 +233,7 @@ def get_blockdevice_info(device_path, exclude_iso_dev :bool = True) -> Dict[str,
 			information = blkid(f'blkid -p -o export {device_path}')
 			return enrich_blockdevice_information(information)
 		except SysCallError as ex:
-			if ex.exit_code in (512, 2):
+			if ex.exit_code == 2:
 				# Assume that it's a loop device, and try to get info on it
 				try:
 					resolved_device_name = device_path.readlink().name
--- a/archinstall/lib/luks.py
+++ b/archinstall/lib/luks.py
@@ -115,7 +115,7 @@ class luks2:
 			if cmd_handle.exit_code != 0:
 				raise DiskError(f'Could not encrypt volume "{partition.path}": {b"".join(cmd_handle)}')
 		except SysCallError as err:
-			if err.exit_code == 256:
+			if err.exit_code == 1:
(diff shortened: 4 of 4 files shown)


archinstall/lib/disk/partition.py
```diff
@@ -186,7 +186,7 @@ def _call_lsblk(self) -> Dict[str, Any]:
 			try:
 				output = SysCommand(f"lsblk --json -b -o+LOG-SEC,SIZE,PTTYPE,PARTUUID,UUID,FSTYPE {self.device_path}").decode('UTF-8')
 			except SysCallError as error:
-				# It appears as if lsblk can return exit codes like 8192 to indicate something.
+				# It appears as if lsblk can return exit codes like 32 to indicate something.
```
codefiles, line 189: The exit status of 32 is "none of specified devices found". [1] Should this line and maybe even the following line then be deleted?

[1] https://man.archlinux.org/man/lsblk.8#EXIT_STATUS

codefiles, line 189: The exit status of 32 is "none of specified devices found". [1] I recommend deletion of this line, what do you think?

[1] https://man.archlinux.org/man/lsblk.8#EXIT_STATUS

archinstall/lib/systemd.py
```diff
@@ -97,7 +97,7 @@ def __exit__(self, *args :str, **kwargs :str) -> None:
 			shutdown = SysCommand(f'systemd-run --machine={self.container_name} --pty shutdown now')
 		except SysCallError as error:
 			shutdown_exit_code = error.exit_code
-			# if error.exit_code == 256:
+			# if error.exit_code == 1:
 			# 	pass
```
codefiles, line 101: Maybe just delete these lines?

archinstall/lib/disk/partition.py
```diff
@@ -186,7 +186,7 @@ def _call_lsblk(self) -> Dict[str, Any]:
...
 				# But it does return output in stderr so we'll try to catch it minus the message/info.
```
codefiles, line 190: The exit status of 32 is "none of specified devices found". I recommend deletion of this line, what do you think? Maybe delete the line that follows as well?

https://man.archlinux.org/man/lsblk.8#EXIT_STATUS

archinstall/lib/systemd.py
codefiles, line 101: Maybe just delete these lines?

Torxed (approved):

Issue #1600, opened 2022-12-18
nvidia open source driver and zen kernel


aucub:
I selected the zen kernel and nvidia open source drivers, but installed nvidia-open instead of nvidia-open-dkms.

lavafroth:
Can confirm. You have to manually add nvidia-open-dkms to the additional packages list. Perhaps we can check if a selected kernel is custom and add necessary dkms packages implicitly?

Torxed:
> Perhaps we can check if a selected kernel is custom

I think this sounds like a good approach!

lavafroth:
I have a couple ideas on implementing this when selecting nvidia drivers.
* Have a list of custom kernels (like zen, hardened, etc.) and checking if any of them are selected.
* Have a class Kernel and an Enum attribute (with enum types like Mainline, Custom) and check that attribute while selecting a nvidia drivers.

Although I prefer the latter way, it might be an overkill for something this simple. Suggestions?

Torxed:
> * Have a list of custom kernels (like zen, hardened, etc.) and checking if any of them are selected.

This one we have already I believe. It's a list within `Installer`.

PR #1698, opened 2023-03-29
Adds kernel dependent hook when installing nvidia driver on custom kernel.

lavafroth:
- This fix issue: #1600 

## PR Description:
Adds a function `kernel_hook` in `nvidia_utils.py` to be invoked after prompting the user for input during guided installation. When a custom kernel is detected, the corresponding packages names are appended with '-dkms'.
Feel free to suggest a better place to invoke this function.

## Tests and Checks


archinstall/lib/nvidia_utils.py
```diff
@@ -0,0 +1,15 @@
+import archinstall
```
Torxed, line 1: From within the `archinstall` library use relative imports instead.

lavafroth: Hi, yes. I have fixed the code. It now uses relative imports and only imports `storage` from the `storage` module.

lavafroth:
@Torxed, any other suggestions?

Torxed:
Nothing to complain about, I gotta test this and see how the position of the hook calling behaves, other than that I don't have any larger objections. No ETA on when I can test it out tho sadly.

Issue #1702, opened 2023-03-30
Several `SysCommand()` calls contain logic error


codefiles:
When a command executed by `SysCommand()` returns a non-zero exit code a `SysCallError` is raised.

https://github.com/archlinux/archinstall/blob/4f539a193146b67e122d30732294df44eb826f13/archinstall/lib/general.py#L283-L284

This means that using the `exit_code` of a `SysCommand()` call in a boolean condition creates a logic error.

Code in the repository along the lines of the "bad" examples below require updating.

### Examples

❌ **Bad**

```python
def test_cmd(cmd: str) -> None:
    if (output := SysCommand(cmd)).exit_code != 0:
        raise SystemError(f'Non-zero exit code returned: {output}')
```

If the command executed by `SysCommand()` returns a non-zero exit code then a `SysCallError` will be raised rather than the intended `SystemError`.

✔️ **Good**

```python
def test_cmd(cmd: str) -> None:
    try:
        SysCommand(cmd)
    except SysCallError as error:
        raise SystemError(f'Non-zero exit code returned: {error}')
```

❌ **Bad**

```python
def test_cmd(cmd: str) -> bool:
    return SysCommand(cmd).exit_code == 0
```

If the command executed by `SysCommand()` returns a non-zero exit code then a `SysCallError` will be raised rather than the intended return of `False`.

✔️ **Good**

```python
def test_cmd(cmd: str) -> bool:
    try:
        SysCommand(cmd)
        return True
    except SysCallError:
        return False
```

PR #1703, opened 2023-03-30
Update `SysCommand()` calls in `installer.py`

codefiles:
Update `SysCommand()` calls in `installer.py` for #1702


Torxed:
Awesome work, thank you!

PR #1707, opened 2023-03-30
Update `SysCommand()` calls in remaining files

### LINKED ISSUE #1743: Bug with `cryptsetup` retries
There is a logic error in the `luks2.encrypt()` function where the `DISK_RETRY_ATTEMPTS` do not occur if the first `SysCommand()` call for `cryptsetup` fails. A non-zero exit code will raise a `SysCallError`, which will be caught by the exception, leaving the retry loop.

https://github.com/archlinux/archinstall/blob/5689d4f59f4f08465d2ef72412f5819e9df87430/archinstall/lib/luks.py#L105-L117

I have an open pull request (#1707) that includes a fix for the logic error as I saw it related to issue #1702 (should still be open). @Torxed, would you prefer it as a separate pull request?

codefiles:
Update `SysCommand()` calls in remaining files for #1702 and also fixes #1743

### DIFF
--- a/archinstall/lib/disk/device_handler.py
+++ b/archinstall/lib/disk/device_handler.py
@@ -139,19 +139,18 @@ class DeviceHandler(object):
 			log(f'Failed to read btrfs subvolume information: {err}', level=logging.DEBUG)
 			return subvol_infos
 
-		if result.exit_code == 0:
-			try:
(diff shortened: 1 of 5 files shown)


Torxed:
Awesome job! Apologies for the delays as mentioned in the other PR.

PR #1711, opened 2023-03-31
Removing legacy build files setup.py and setup.cfg

Torxed:
This would make `archinstall` fully rely on `pyproject.toml` which is the future.
This should work out of the box assuming it being packaged on:
 * python >=3.11.2
 * python-build >= 0.10.0
 * python-installer >= 0.7.0
 * python-setuptools >= 67.5.0

Tested locally with:
 * PKGBUILD: https://github.com/archlinux/archinstall/blob/3f5e63ca45b5adfdf8817fe6a5551f21e3434e85/PKGBUILD#L54-L66
 * pyproject.toml: https://github.com/archlinux/archinstall/blob/3f5e63ca45b5adfdf8817fe6a5551f21e3434e85/pyproject.toml#L29-L40

It relies on [dynamic metadata](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata).


pyproject.toml
```diff
@@ -1,14 +1,15 @@
...
 
 [project]
 name = "archinstall"
-dynamic = ["version", "entry-points", "license"]
+dynamic = ["version", "entry-points"]
 description = "Arch Linux installer - guided, templates etc."
 authors = [
     {name = "Anton Hvornum", email = "anton@hvornum.se"},
 ]
+license = {text = "GPLv3"}
```
dvzrv, line 12: Is it GPL-3.0 or GPL-3.0-or-later?

Torxed: Updating, it should be `GPL-3.0-only`. Couldn't remember what to search for but SPDX license identifiers was it, thanks! :D

pyproject.toml
```diff
@@ -26,11 +27,24 @@ Home = "https://archlinux.org"
...
+
 [project.scripts]
 archinstall = "archinstall:run_as_a_module"
 
-[project.optional-dependencies]
-doc = ["sphinx"]
+[tool.setuptools]
+packages = ["archinstall", "profiles", "examples"]
+
+[tool.setuptools.package-data]
```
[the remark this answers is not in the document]
Torxed: It's to avoid `setuptools` trying to auto-detect sub-packages etc. It would constantly detect the locales as a subpackage as `archinstall.locales` for instance. This should avoid that from happening. We can probably even remove `profiles` and `examples` but I wanted to be sure those got included in the packaging, even tho `package-data` should be responsible for that part.

pyproject.toml
```diff
@@ -26,11 +27,24 @@ Home = "https://archlinux.org"
...
+archinstall = [
+    "examples/*.py",
+    "profiles/*.py",
+    "profiles/applications/*.py"
+]
+
+[tool.setuptools.dynamic]
```
svartkanin, line 46: @Torxed this section was already defined on line 46

Torxed: Yea the merge from `master` did'nt take that into account so it created double entries. I manually consolidated them.

PR #1736, opened 2023-04-09
Update Brazilian Portuguese translation

rffontenelle:


Torxed:
Thank you for this update!

PR #1752, opened 2023-04-18
Remove superfluous time sync steps

### LINKED ISSUE #1751: Superfluous steps in `Installer.activate_time_syncronization()`
The only necessary step for systemd-timesyncd is to enable the service and that is already being done here.

https://github.com/archlinux/archinstall/blob/98f13ece6e99e32604e915c1e5e43527a7578346/archinstall/lib/installer.py#L532

Removing both portions of code below and creating an installation with the ntp option set to True will still start systemd-timesyncd on boot and produce a successful NTP synchronization using the default Arch servers demonstrating that these steps are superfluous.

- https://github.com/archlinux/archinstall/blob/98f13ece6e99e32604e915c1e5e43527a7578346/archinstall/lib/installer.py#L534-L537

  This was added in the commit 7e85002dcbc913c117c9a341026283cce2dc0787 with the intent to add a "default config that's 'sane'" but on Arch Linux the systemd package comes with these NTP servers as defaults. [1]

  Below is a quote from the ArchWiki on systemd-timesyncd configuring [2], emphasis mine.

  > To **add** [time servers](https://wiki.archlinux.org/title/Network_Time_Protocol_daemon#Connection_to_NTP_servers) or **change** the **provided ones**, uncomment the relevant line and list their host name or IP separated by a space.

  Without adding or changing the provided ones editing the `/etc/systemd/timesyncd.conf` file is unnecessary. For this piece of code I assume the ArchWiki was followed but this distinction was missed. This matches the example configuration given on the ArchWiki that also includes a FallbackNTP of `0.fr.pool.ntp.org` (`fr` is the pool zone for France).
  
  Running the `systemd-timesyncd.service` unit with the default configuration file the Arch Linux systemd package provides and checking the output of the command `timedatectl show-timesync --property=ServerName` should verify that the default Arch servers are being used without the need to edit this configuration file.

- https://github.com/archlinux/archinstall/blob/98f13ece6e99e32604e915c1e5e43527a7578346/archinstall/lib/installer.py#L539-L541

  The command `timedatectl set-ntp true` is unnecessary because it enables and starts the systemd-timesyncd service [3] when enabling it is already taken care of in the code a few lines above and starting the service fails anyway. The service does not work within a systemd-nspawn container and that is the method used here in an effort to start it. A NTP synchronization will not occur during or at the end of the installation process but on boot as it would occur with only the service enabled. Chrooting in directly after the installation and checking the log (`journalctl -u systemd-timesyncd`) reveals that a network time synchronization did not occur from the attempt to start the service.

  ```
  Network Time Synchronization was skipped because of an unmet condition check (ConditionVirtualization=!container).
  ```

  Deleting this portion of code should make a slight performance improvement because time is being spent booting the installation as a systemd-nspawn container, running the command in the container, and shutting down the container without the intended effect.

  The ArchWiki states this command is used to start and enable systemd-timesyncd under the usage heading [2] so maybe that was confused for meaning the command must be run to get this working.

[1] https://github.com/archlinux/svntogit-packages/blob/0eba6dd91509b12c13a6bb77baaaee70f31a026b/trunk/PKGBUILD#L99
[2] https://wiki.archlinux.org/title/systemd-timesyncd
[3] https://man.archlinux.org/man/systemd-timesyncd.service.8.en

codefiles:
Resolves #1751

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -531,15 +531,6 @@ class Installer:
 		self.log('Activating systemd-timesyncd for time synchronization using Arch Linux and ntp.org NTP servers.', level=logging.INFO)
 		self.enable_service('systemd-timesyncd')
 
-		with open(f"{self.target}/etc/systemd/timesyncd.conf", "w") as fh:
-			fh.write("[Time]\n")
-			fh.write("NTP=0.arch.pool.ntp.org 1.arch.pool.ntp.org 2.arch.pool.ntp.org 3.arch.pool.ntp.org\n")
-			fh.write("FallbackNTP=0.pool.ntp.org 1.pool.ntp.org 0.fr.pool.ntp.org\n")
-
-		from .systemd import Boot
-		with Boot(self) as session:
-			session.SysCommand(["timedatectl", "set-ntp", 'true'])
-
 	def enable_espeakup(self) -> None:
 		self.log('Enabling espeakup.service for speech synthesis (accessibility).', level=logging.INFO)
 		self.enable_service('espeakup')


Torxed:
Thank you, trying to find the wiki page where I copied that from and can't even find it anymore.

codefiles:
https://wiki.archlinux.org/title/systemd-timesyncd

Torxed:
That's the one! I assume ntp works without those steps despite us using `systemd-timesyncd`.

codefiles:
With the way the ArchWiki is written for that page it can be easy to mistaken the steps as required. The steps are there to demonstrate how to set servers other than the defaults already provided and enable and start ntp the service if it has not already been enabled.

PR #1754, opened 2023-04-20
Fix logic errors in `_fetch_lsblk_info()`

### LINKED ISSUE #1753: Logic errors in `_fetch_lsblk_info()`
The for loop here will cause the `lsblk` command to be called as many times as specified by the retry parameter. It does not break out of the loop if there is a success. In addition, if all tries fail the last try does not need a sleep.

https://github.com/archlinux/archinstall/blob/00b0ae7ba439a5a420095175b3bedd52c569db51/archinstall/lib/disk/device_model.py#L979-L989

The `get_lsblk_info()` function calls `_fetch_lsblk_info()` without the retry argument.

https://github.com/archlinux/archinstall/blob/00b0ae7ba439a5a420095175b3bedd52c569db51/archinstall/lib/disk/device_model.py#L1004-L1008

The retry parameter will then default to 3.

https://github.com/archlinux/archinstall/blob/00b0ae7ba439a5a420095175b3bedd52c569db51/archinstall/lib/disk/device_model.py#L967

This means that when `get_lsblk_info()` is called, if the `lsblk` command is working fine (exit code of 0) then the command will be called unnecessarily 2 additional times.

codefiles:
Fixes #1753


codefiles:
@Torxed, is there anything holding this back from being merged?

codefiles:
Respectfully, I do not know why this has not been merged.

There is more I would like to contribute but I do not want to keep stacking pull requests that do not get merged. I've tried to keep my pull request small so they would be easier to review. @Torxed, I appreciate that you volunteer and I understand that, so maybe it is a lack of time available for the project? I can do more if there is anything you would like to delegate to me. Any suggestions are welcome.

Torxed:
@codefiles The reason is the huge #1604 which was substantial. It not only took a lot of time code reviewing but also would cause a few bugs most likely. So to minimize the clutter of which bug belonged to which merges I've been holding off on other merges - as it could also impact code causing merge conflicts and getting it in was a priority.

I'm also human so nedded some rest after testing and reviewing it for weeks at end.
And once that was done, Python 3.11 was in the pipeline of being released so had to prioritize testing against it otherwise anything I merge/add would be pointless if nothing works.

That said, I really do feel the frustration of having to wait, as these are my problems - not yours.. and you have unused capacity at the moment. I'm working on that as well by finalizing the TU onboarding and looking into have collaborators that have a history of commitment in this repo - so I can offload some of this work.

And with that out of the way, I'll begin merging these now :)

Torxed:
Thank you for taking the time to make this, nice cleanup!

PR #1755, opened 2023-04-20
Fix insufficient documentation in the README.md

uranderu:
- This fix issue: #1756

## PR Description:

I tried building from source following the instructions. But, it seems there are a few small things missing from it and so the build fails. This PR adds the missing pieces so that other users don't run into the same errors.

## Tests and Checks
This PR does not change any code, so this section is irrelevant.


Torxed:
may I ask what gcc is needed for? Have'nt built from source in a while so curious what requires it?

svartkanin:
@Torxed it's only needed if pyparted isn't installed yet. Installing it will have to build the c lib bindings unfortunately.

uranderu:
> @Torxed it's only needed if pyparted isn't installed yet. Installing it will have to build the c lib bindings unfortunately.

Which to be clear, is not the case using the latest Arch ISO (2023.04.01) :)

svartkanin:
No it isn't, because there was a new change merged this week which makes use of the mentioned dependency :)

Torxed:
I'll tweak this PR a bit, it's a good step in the right direction but I'll also swap out `pip` entirely in the build and installation process (as a previous PR fixed)

uranderu:
> I'll tweak this PR a bit, it's a good step in the right direction but I'll also swap out `pip` entirely in the build and installation process (as a previous PR fixed)

If there is a way for me to help, let me know!

Torxed:
Thanks for this one, tweaked it a bit more as there were more changes to pip in the last few weeks.

PR #1768, opened 2023-04-25
Fix README.md links

codefiles:
Fix broken links and link to desktop and server profiles rather than list them all out.


Torxed:
Wise move to consolidate the desktop listing, thank you!

Issue #1759, opened 2023-04-21
"Unable to determine new partition uuid" on NVMe storage


hkrutzer:
As described earlier here https://github.com/archlinux/archinstall/pull/1604#issuecomment-1514704715:

```
Unmounting all partitions: /dev/nvme0n1
Wiping partitions and metadata: /dev/nvme0n1
Creating partitions: /dev/nvme0n1
Traceback (most recent call last):
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/root/archinstall/archinstall/__main__.py", line 19, in <module>
    archinstall.run_as_a_module()
  File "/root/archinstall/./archinstall/__init__.py", line 265, in run_as_a_module
    importlib.import_module(mod_name)
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/root/archinstall/./archinstall/scripts/guided.py", line 274, in <module>
    fs_handler.perform_filesystem_operations()
  File "/root/archinstall/archinstall/lib/disk/filesystem.py", line 55, in perform_filesystem_operations
    device_handler.partition(mod, partition_table=partition_table)
  File "/root/archinstall/archinstall/lib/disk/device_handler.py", line 469, in partition
    self._perform_partitioning(part_mod, modification.device, disk, requires_delete=requires_delete)
  File "/root/archinstall/archinstall/lib/disk/device_handler.py", line 347, in _perform_partitioning
    raise DiskError(f'Unable to determine new partition uuid: {part_mod.dev_path}')
archinstall.lib.exceptions.DiskError: Unable to determine new partition uuid: /dev/nvme0n1p2
```

<details><summary>Disk config</summary>
<p>

```json
{
        "config_type": "default_layout",
        "device_modifications": [
            {
[...]

PR #1770, opened 2023-04-27
Attempt to fetch partuuid multiple times

svartkanin:
This may fix issue #1759. 
Attempt to fetch the partuuid of the newly created partition multiple times

### DIFF
--- a/archinstall/lib/disk/device_handler.py
+++ b/archinstall/lib/disk/device_handler.py
@@ -352,14 +352,15 @@ class DeviceHandler(object):
 		attempts = 3
 		info: Optional[LsblkInfo] = None
 
+		self.partprobe(path)
 		for attempt_nr in range(attempts):
-			self.partprobe(path)
+			time.sleep(attempt_nr + 1)
 			info = get_lsblk_info(path)
 
 			if info.partuuid:
 				break
 
-			time.sleep(attempt_nr + 1)
+			self.partprobe(path)
 
 		if not info or not info.partuuid:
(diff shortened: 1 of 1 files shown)


archinstall/lib/disk/device_handler.py
```diff
@@ -341,16 +341,34 @@ def _perform_partitioning(
...
 			part_mod.uuid = info.uuid
 		except PartitionException as ex:
 			raise DiskError(f'Unable to add partition, most likely due to overlapping sectors: {ex}') from ex
 
+	def _fetch_partuuid(self, path: Path) -> LsblkInfo:
+		attempts = 3
+		info: Optional[LsblkInfo] = None
+
+		for attempt_nr in range(attempts):
+			self.partprobe(path)
```
Torxed, line 356: Can we move this `partprobe()` out right before the loop?
AFAIK, calling `partprobe()` once will initiate the kernel to look for new ID's, and it will take X amount of time and the kernel will not stop partprob-ing until all ID's are populated. Calling `partprobe()` again will just re-start that process.

svartkanin: Right, so we want to kick off partprobe then wait for x seconds and the run lsblk?

Torxed: That's the curse of `partprobe()`, it doesn't block.
Which means you can't rely on "when partprobe is done, all the data will be there".

At least that's my experience with it.

svartkanin: In that case I'll update the PR

PR #1777, opened 2023-05-03
Full mypy compliance and small fixes

svartkanin:
This is the last puzzle piece for the full mypy compliance. 
Along the way some things got fixed as well:

-  Move the gernation for the systemd config for network mangaer configuration file into the NetworkConfiguration file
- Remove backwards compability support in NetworkConfiguration; this has been in there for quite a while now and it is to assume that these will by now most likely have been migrated to new config style by users
- Fix cursor positioning for menu entries on pre-selection and default values

### DIFF
--- a/.github/workflows/mypy.yaml
+++ b/.github/workflows/mypy.yaml
@@ -15,4 +15,4 @@ jobs:
             # one day this will be enabled
               # run: mypy --strict --module archinstall || exit 0
             - name: run mypy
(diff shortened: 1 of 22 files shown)


Torxed:
Nicely done! I did not testrun this but looked through it and couldn't see any breaking changes. I'll merge some other stuff and perform a bigger all around test instead :)

PR #1729, opened 2023-04-07
Remove obsolete enabling of NTP in ISO

codefiles:
Resolves #1570

Also includes a small improvement to the wait for a time sync.

PR #1788, opened 2023-05-05
Fix NTP option

codefiles:
I missed that a mistake was made when upstream master was merged in #1729. The intention of that pull request was to remove enabling NTP in the ISO/live environment since it is already activated on boot of that environment. That was successful but the merge of upstream master broke the option in the guided installer to disable/enable use of systemd-timesyncd (NTP) in the installation (installation is an entirely separate environment from ISO/live environment with its own systemd-timesyncd). The option is stuck on enable and can not be disabled. This restores that function.


Torxed:
I thought I might have missed some parts while moving the logic, apologies for that and thank you for cleaning it up!

codefiles:
No problem.

PR #1797, opened 2023-05-06
Refactor `build_iso.sh` added packages

codefiles:
Refactor `build_iso.sh` to use an array and loop for adding packages to the archiso profile pacakges. This makes it easier to: review what packages are being added to the ISO, add or remove packages, and maintain.


Torxed:
100% easier to read ^^

PR #1800, opened 2023-05-06
Fix default value idx in `_determine_cursor_pos()`

codefiles:
The following are examples of messages that can be observed in the log when interacting with a `Menu` that is initialized with a `default_option` such as the guided installer menus for swap, NTP, and audio.
- `Error finding index of yes: ['yes (default)', 'no']`
- `Error finding index of No audio server: ['No audio server (default)', 'pipewire', 'pulseaudio']`

~~This is fixed by increasing the priority of a default value over that of preset values in `_determine_cursor_pos()`.~~

### DIFF
--- a/archinstall/lib/menu/menu.py
+++ b/archinstall/lib/menu/menu.py
@@ -337,7 +337,10 @@ class Menu(TerminalMenu):
 					if '|' in p:
 						p = p.replace('|', '\\|')
 
-					idx = self._menu_options.index(p)
+					if p in self._menu_options:
+						idx = self._menu_options.index(p)
+					else:
+						idx = self._menu_options.index(self._default_menu_value)
 					indexes.append(idx)
 				except (IndexError, ValueError):
 					log(f'Error finding index of {p}: {self._menu_options}', level=logging.DEBUG)


svartkanin:
I agree that it's a bug but moving the default up the rank isn't fixing the underlying problem and it gives a weird user experience: 
- select `profile -> server` 
- select `audio -> pipewire` 
- go back to audio
- cursor is set to default option

The problem with the bug is that the preset flow doesn't cater for `preset == default`, so I think changing line 340 to 
```
if p in self._menu_options:
	idx = self._menu_options.index(p)
else:
	idx = self._menu_options.index(self._default_menu_value)
```
						
should do the trick

codefiles:
I missed that this change would always put the cursor to the default option even when a selection has been made as you have explained. Thanks for looking this over and pointing that out. How about I close the pull request, create an issue, and you can submit a pull request that closes the issue with that fix?

svartkanin:
Sure, but I don't mind if you just wanna update your PR with it

Torxed:
Awesome teamwork! keep it up!

PR #1801, opened 2023-05-07
Cleanup imports and unused code

svartkanin:
- Remove asterisk imports from __init__.py as they are very dangerous and mypy will not be able to resolve issues
- Create a `SysInfo` class in `hardware.py` to encapsulate common functionality through a single interface
- Remove unused functions and code

The main reason for grouping modules and imports together in the __init__.py file is that it will allow for a clear documentation of the API archinstall is able to provide for users who choose to code their own installer


archinstall/lib/exceptions.py
```diff
@@ -27,33 +24,13 @@ def __init__(self, message :str, exit_code :Optional[int] = None, worker :Option
...
 
 class PackageError(BaseException):
 	pass
-
-
-class TranslationError(BaseException):
-	pass
-
-
-class Deprecated(BaseException):
```
Torxed, line 58: Could we keep this one? It feels like this might have been removed by an automatic tool because it's currently not used anywhwre?

Torxed (changes requested):

archinstall/lib/exceptions.py
svartkanin: I searched the codebase and couldn't find any usages, bu I'm happy to keep it

Torxed: It's used whenever we deprecate something, so it's used from time to time - and people can use it in custom scripts to detect deprecations easier :)

svartkanin:
@Torxed I put the exception back :)

svartkanin:
@Torxed hope you don't mind but I pushed some more cleanup in here instead of raising another one. 
The logging is simplified into 4 additional functions 
```
def info(*msgs: str):
	log(*msgs, level=logging.INFO)


def debug(*msgs: str):
	log(*msgs, level=logging.DEBUG)


def error(*msgs: str):
	log(*msgs, level=logging.ERROR, fg='red')


def warn(*msgs: str):
	log(*msgs, level=logging.WARNING, fg='yellow')

```

the original `log(...)` is still there as well, but those should make it easier to use at least with less typing :)

Torxed:
That's totally fine, took me a while to check the last few changes because they looked horrible while code-reviewing in the phone hehe. The diff files were a mess.

The only thing I'll miss a bit is the (optional) journald logging:
```diff
- class Journald:
```

The reason being, setting up [systemd-journal-remote.service](https://www.freedesktop.org/software/systemd/man/systemd-journal-remote.service.html) logging .
I some times use that on remote installations to get the log locally sent when deploying machines for instance. That way anything logged through `archinstall.output.log()` would also call `Journald.log(string, level=int(str(kwargs.get('level', logging.INFO))))` for instance.

Could we keep that or is it a lot of hassle?

svartkanin:
@Torxed I'm happy to keep it but I didn't think that it's actually working? Ir was trying to import a module that didn't exist, maybe I missed something

svartkanin:
I was probably confused as we have a module called systemd which I assumed was the import source.
Happy to put it back in and potentially rename the existing one

svartkanin:
@Torxed updated :)

Torxed:
Thank you for putting in the effort!

Hehe, yea the mysterious import was the only good way i could solve the check if the lib exists as optimal dep ^^

Torxed (approved):

Torxed:
There's a few changes here that might cause some disturbances.
But we've got other stuff in the pipeline (https://github.com/archlinux/archinstall/issues/1807) waiting for this so merging :)

Great work!

svartkanin:
> There's a few changes here that might cause some disturbances.
> But we've got other stuff in the pipeline (https://github.com/archlinux/archinstall/issues/1807) waiting for this so merging :)
> 
> Great work!

Anything you're concerned about or something I can update?

Torxed:
Nothing major, there were some connection points that might cause logic issues I haven't thought of :D But so far nothing has popped up :D

PR #1805, opened 2023-05-07
Rework `build_iso.sh`

codefiles:
Move the development message to motd. Move archinstall build from zprofile to a script. Create a service that is enabled on boot to run the script. Wait for completion of the service at root login shell (zprofile).

Currently the build of archinstall in the ISO environment is output to the console. This is problematic because there is not a scrollback buffer in the console to review the output. The changes in this pull request will allow the review of the build with `journalctl -u build-archinstall`.

Another issue with the current implementation is changing tty or executing tmux and opening panes and windows will read and execute the zprofile causing archinstall to be built and installed again. This pull request does not have this problem because the script to build and install archinstall is started by the service.

### DIFF
--- a/build_iso.sh
+++ b/build_iso.sh
@@ -1,8 +1,16 @@
 #!/bin/bash
 
-packages_file="/tmp/archlive/packages.x86_64"
+set -eu
 
-# Packages to add to the archiso profile packages
+readonly base_profile="/usr/share/archiso/configs/releng/"
+readonly custom_profile="/tmp/archlive"
+readonly packages_file="$custom_profile/packages.x86_64"
+readonly airootfs="$custom_profile/airootfs"
+readonly archinstall_dir="/root/archinstall"
+readonly build_script="/root/build-archinstall.sh"
+readonly build_service="/etc/systemd/system/build-archinstall.service"
+
+# Packages to add to the archiso custom profile
 packages=(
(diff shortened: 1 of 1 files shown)


Torxed:
The script was never really meant to be used on a broader scale. It's only real purpose was to run in https://github.com/archlinux/archinstall/blob/1df53c1406960b5d7b17bb7ee84431f8de7bfbdf/.github/workflows/iso-build.yaml#L35

That said the improvements might be quite useful. Let me test this out before merging :)

codefiles:
Understood. After some more thought on the matter I feel this pull request should not be merged. At least not yet.

build_iso.sh
```diff
@@ -13,34 +21,79 @@ packages=(
...
-	python -m build --wheel --no-isolation
-	pip install dist/archinstall*.whl
-
-	echo "This is an unofficial ISO for development and testing of archinstall. No support will be provided."
-	echo "This ISO was built from Git SHA $GITHUB_SHA"
-	echo "Type archinstall to launch the installer."
-_EOF_
-
 pacman -Sy
 pacman --noconfirm -S git archiso
```
lavafroth, line 25: Perhaps you can add the `--disable-download-timeout` flag here. It helps with errors like `Operation too slow`.

PR #1811, opened 2023-05-12
Install the package `sof-firmware` if required

### LINKED ISSUE #1807: Autodetection of certain hardware requirements
I just ran archinstall on a laptop that uses tigerlake audio, which requires you to install sof-firmware in order to have the audio work, it would be nice if tigerlake audio was automatically detected if you choose to setup audio.

codefiles:
Resolves #1807

"[sof-firmware](https://archlinux.org/packages/?name=sof-firmware) is required for some newer laptop models (mainly since 2019) ..." [1]

The 'Sound Open Firmware (SOF) Core' kernel module (`sof_snd`) should be automatically loaded when hardware that requires it is detected. [2] Determine if the package is required by checking if this kernel module is loaded. Install the package if it is required and an audio selection is present.

[1] https://wiki.archlinux.org/title/Advanced_Linux_Sound_Architecture#ALSA_firmware
[2] https://wiki.archlinux.org/title/Kernel_module#Automatic_module_loading

### DIFF
--- a/archinstall/lib/hardware.py
+++ b/archinstall/lib/hardware.py
@@ -1,7 +1,7 @@
 import os
 from functools import cached_property
 from pathlib import Path
-from typing import Optional, Dict
+from typing import Optional, Dict, List
 
 from .general import SysCommand
 from .networking import list_interfaces, enrich_iface_types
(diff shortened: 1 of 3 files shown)


Torxed:
This is a really cool conceptc that could be used in many situations :D

Issue #1435, opened 2022-08-28
More options to pick than one in "Locale language" and "Locale encoding".


eskiq:
If I want en language and other one with special letters there is no option to do that. I would suggest to add check boxes just like in "Mirror region".

Torxed:
I believe this is what https://github.com/archlinux/archinstall/pull/1423 is aiming to do? Or at least to some extent.
There's no reason why we don't have multi-select, good call! :)

eskiq:
Yes, that would probably solve the problem in half bcs "Locale encoding" also need that ability to make special characters working for specific language . Otherwise there would be post installation configuration for the same thing when somebody need those special letters for some languages.

codefiles:
The plan is to have multi-select and #1423 sets up what is needed to implement this into the guided installer. This will be in a follow up pull request. @Torxed, you can assign this to me if you would like since this is what I'm working towards.

codefiles:
Waiting for #1423 and #1458 to be merged before submitting a pull request for this.
![locales](https://user-images.githubusercontent.com/11915375/189510847-ada4f25a-2e39-43d1-abf6-f8f1a400d16a.gif)

Torxed:
@eskiq any reason for closing this? As there's no PR for it ready yet AFAIK.

eskiq:
@Torxed I was thinking that it would be not useful to keep that thread open. Forgive me.  : D

codefiles:
@eskiq would you please reopen this issue since it has not been resolved yet?

PR #1814, opened 2023-05-13
Move locales and cleanup menu

svartkanin:
To be able to add new global menu items such as `Optimizations` and potentially `Advanced` more space is needed :) 
This PR:
- Grouped the locales configuration (sys-lang, sys-enc and kb-layout) into a submenu selection
- Removed the `set:` prefix from the menu (@Torxed it feels more lightweight this way but if it looks odd then happy for suggestions)

image


codefiles:
I still have plans to combine sys-lang and sys-enc into a single option, see https://github.com/archlinux/archinstall/issues/1435#issuecomment-1242876835. The changes to locales here interferes with and ignores a pull request I currently have open.

In 1604 you were failing to properly merge in commits from master and in 1801 you moved the locale file. I have been holding off on trying to get my locale changes merged because of this. I have been working to fix bugs and implement requests in the meantime.

Would you please put more work into partitioning? Partitioning is still lacking when compared with `cfdisk`. There are also still some bugs from 1604. For example, in the guided installer the audio option is inaccessible without selecting a profile.

svartkanin:
I don't think it's a blocker for yours as the main goal here is to put the locals into a submenu.to create space in the global one

Torxed:
It looks good to me! Do you mind fixing the conflicts?
I am 100% sure I will mess it up as it's entangled in the menu code :)

svartkanin:
@Torxed should be ready now

PR #1815, opened 2023-05-13
installer: add Limine bootloader

Andy-Python-Programmer:
This pull request adds support for the limine bootloader to archinstall :) Limine is a modern, advanced, portable, multiprotocol bootloader.

## References

[Limine GitHub](https://github.com/limine-bootloader/limine) 
[Limine Arch Wiki](https://wiki.archlinux.org/title/Limine)

## Tests and Checks
- [x]  `x86_64` BIOS
- [x] `x86_64` UEFI

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -8,6 +8,8 @@ import time
 from pathlib import Path
 from typing import Any, List, Optional, TYPE_CHECKING, Union, Dict, Callable
 
+from ..lib.disk.device_model import get_lsblk_info
+
 from . import disk
 from .exceptions import DiskError, ServiceException, RequirementError, HardwareIncompatibilityError, SysCallError
 from .general import SysCommand
@@ -858,6 +860,10 @@ class Installer:
 		self._pacstrap('limine')
 		info(f"Limine boot partition: {boot_partition.dev_path}")
 
+		# XXX: We cannot use `root_partition.uuid` since corresponds to the UUID of the root
+		#      partition before the format.
+		root_uuid = get_lsblk_info(root_partition.safe_dev_path).uuid
+
 		device = disk.device_handler.get_device_by_partition_path(boot_partition.safe_dev_path)
 		if not device:
 			raise ValueError(f'Can not find block device: {boot_partition.safe_dev_path}')
@@ -925,7 +931,7 @@ Description = Deploying Limine after upgrade...
 When = PostTransaction
 # XXX: Kernel name descriptors cannot be used since they are not persistent and
 #      can change after each boot.
-Exec = /bin/sh -c \\"/usr/bin/limine bios-install /dev/disk/by-uuid/{root_partition.uuid} && /usr/bin/cp /usr/share/limine/limine-bios.sys /boot/\\"
+Exec = /bin/sh -c \\"/usr/bin/limine bios-install /dev/disk/by-uuid/{root_uuid} && /usr/bin/cp /usr/share/limine/limine-bios.sys /boot/\\"
 			""")
 
 		# Limine does not ship with a default configuation file. We are going to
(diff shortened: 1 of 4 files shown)


Torxed:
Interesting, nicely added!
The only suggestion I have is if we should check if the selected filesystem in use is any of the supported ones? I noticed BTRFS is not supported for instance, and it's a quite popular one and I can see a future where we'll get a log of support tickets if they use a combination of this and BTRFS.

Andy-Python-Programmer:
> The only suggestion I have is if we should check if the selected filesystem in use is any of the supported ones? I noticed BTRFS is not supported for instance, and it's a quite popular one and I can see a future where we'll get a log of support tickets if they use a combination of this and BTRFS.

I have added the compatibility checks in https://github.com/archlinux/archinstall/pull/1815/commits/54e815a9bd2b1f5e05bbea357f888958ca6a4ea8.

image

Andy-Python-Programmer:
https://github.com/archlinux/archinstall/pull/1815/commits/4d71d8838b0f373be16e6696281f6ae8e8c6880d and https://github.com/archlinux/archinstall/pull/1815/commits/0d48af015834979bc47edd4063f7600ed36fa339 should've fixed all of the CI errors in https://github.com/archlinux/archinstall/pull/1815/commits/5a84781e97c8df6a06d227a6601b1b156d88dced.

Andy-Python-Programmer:
To ensure Limine is deployed and updated correctly, it is necessary to have a pacman hook in place (refer to https://wiki.archlinux.org/title/Limine#pacman_hook). Should I just create the hook by running `echo "...." > /etc/pacman.d/hooks/liminedeploy.hook` in `_add_limine_bootloader` or is there a specific API that I should use to assist in creating the hook.

Andy-Python-Programmer:
cc @Torxed?

mintsuki:
@Andy-Python-Programmer this needs to be updated for Limine 5.x :)

Andy-Python-Programmer:
Apologies for the late reply!

I have updated it to Limine 5.x and now it installs the pacman hooks aswell. Tho, now I am trying to debug the issue where `root_partition.uuid` is different then the actual root partition UUID. Not sure where the issue is.

Here is the reported UUID:
image
From the mountpoint `/mnt/archinstall` (got that from the logs):
image
The actual UUID (got by chrooting into the installation):
image

Manually fixing the UUID makes it boot; not sure why its reporting the wrong UUID. This is not reproducible with Grub since it makes use of `grub-probe` to get the UUID (not sure what the `root_partition.uuid` corresponds to because it certainly is not the root partition UUID).

Andy-Python-Programmer:
After a bit more debugging, I found the following:

UUID reported before the installation:
image

UUID reported after the installation:
image

The UUID that `root_partition.uuid` corresponded to:
image

As you can see, the `root_partition.uuid` corresponded to the UUID before the formatting. So, `root_partition.uuid` needs to update.

Andy-Python-Programmer:
The bug is fixed in https://github.com/archlinux/archinstall/pull/1815/commits/611ce938491ebccea3014fdc551d0a6a3935fb28 :^)

archinstall/lib/installer.py
```diff
@@ -1017,6 +1126,7 @@ def add_bootloader(self, bootloader: Bootloader):
 		Archinstall supports one of three types:
 		* systemd-bootctl
 		* grub
+		* limine
```
Torxed, line 1129: Maybe swap this to `* limine (beta)`, just for a bit?

Andy-Python-Programmer: Sure!

Andy-Python-Programmer: Fixed in https://github.com/archlinux/archinstall/pull/1815/commits/88123bcdf5dbfc965655fbc345c65b416799fe76.

svartkanin:
I left some comments, would you mind wrapping the displayed text with the translation wrapper

PR #1818, opened 2023-05-15
Allowing for fg='color' in info, warn, error and debug.

Torxed:
This is a complementary patch to https://github.com/archlinux/archinstall/pull/1801
Otherwise things like this won't work: https://github.com/archlinux/archinstall/blob/6b367c7a1cc6011847daeb6b52d1593ac4b6b87a/archinstall/lib/configuration.py#L166

Technically yea, that maybe could be something other than `red` in the example. But could be purple or anything, issue is that keywords got lost :)

### DIFF
--- a/archinstall/lib/output.py
+++ b/archinstall/lib/output.py
@@ -238,20 +238,34 @@ def _stylize_output(
 	return f'\033[{ansi}m{text}\033[0m'
 
 
-def info(*msgs: str):
-	log(*msgs, level=logging.INFO)
+def info(*msgs: str, **kwargs: dict):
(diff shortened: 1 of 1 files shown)


svartkanin:
The reason I didn't put those arguments into the dedicated info, debug and error was that they should be interpreted from the name, otherwise there isn't any difference between log and info if you can change the color regardless.

A message should be red if it's an error, and not a info, so providing the ability to change the color leads to inconsistencies.

svartkanin:
If there is a need to explicitly change the color one can still use the log as was.
@Torxed any thoughts?

PR #1822, opened 2023-05-17
Detect more modules requiring `alsa-firmware`

codefiles:
In #1812 detection for the module `snd-emu10k1` was added, this adds detection for more modules.

I was unable to find a list of loadable kernel modules that on Arch Linux require firmware provided by the `alsa-firmware` package so I created the following Python script.

- [script](https://github.com/codefiles/examples/blob/main/src/alsa-firmware.py)
- [output](https://github.com/codefiles/examples/actions/runs/4998487791/jobs/8953940446#step:9:1)

Only a few firmware files provided by the package were not matched to any of the sound loadable kernel modules provided by the `linux` kernel package.

### DIFF
--- a/archinstall/lib/hardware.py
+++ b/archinstall/lib/hardware.py
@@ -191,4 +191,33 @@ class SysInfo:
 
 	@staticmethod
 	def requires_alsa_fw() -> bool:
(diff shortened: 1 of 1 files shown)


Torxed:
Great work! Sorry for the delay, some personal stuff happened in life.

PR #1824, opened 2023-05-17
Renamed hyperland to hyprland, fixed seatd via post_installation and added waybar-hyprland

Sxmourai:
- Fixes: #2007
- Fixes: #2027

## PR Description:
Renamed the "hyperland" profile to "hyprland".
Fix seatd (when I test hyprland seatd always fails on opening seat)
Install waybar-hyprland by default, (waybar-hyprland and not waybar for the workspaces)

## Tests and Checks


svartkanin:
@Sxmourai is the configuration actually needed to get hyprland working? Or is it a better default?
 
Archinstall shouldn't configure anything beyond the minimum setup otherwise it's a rabbit hole with no return

dylanmtaylor:
I agree with @svartkanin on this one - as is, it looks too opinionated. We try not to configure too much for the user other than the bare minimum for the desktop to be usable out of the box. If it's not usable without a bunch of custom profiles, etc. perhaps it is a candidate for removal and they can set it up w/ the generic xorg profile.

dylanmtaylor (changes requested):
After thinking this over, I belive all of the sample configuration would need to be removed before this profile is suitable for archinstall.

Sxmourai:
@svartkanin Yes it is a better default...
It sets up stuff that I think everyone will want. Waybar etc is used by a lot of persons, and there a good practices that aren't made by default by Hyprland like trying to kill all other desktop portals, etc
We could configure automatically the keyboard, and I think launching kitty by default is interesting if people have issues with the keyboard shortcuts for Hyprland.

But the automatic support for nvidia could be interesting

dylanmtaylor:
-1 as is. While I appreciate that the configs will likely lead to a better/more optimal out of the box experience, as mentioned, changing defaults is a rabbit hole, users should configure their own system, and we don't really want to deviate from default configuration in the installer. I think the configurations should be dropped entirely unless the environment actually doesn't work at all without them due to the aforementioned reasons.

svartkanin:
If we're making this configurable then so would the other profiles have to be. This is a personalised configuration, even if the majority of users are using it, it's still a configuration on top of the default out of the box one.
We would make this into maintenance hell having to update and add new tiny little things to it to cater for everyone's desires.  That is not the purpose of the installer. It's supposed to install a working system and everything beyond that is up to the user to configure. I would imagine some people to have their own scripts running afterwards anyways to set things up.

Torxed:
I agree with @dylanmtaylor and @svartkanin here.
One route would be configuration via the menu - but the more sustainable option is to only add things that is required during installation, anything that can be safely configured during first boot or later can be left out for now.

If this project ever reaches Red Hat level of maintainers maybe we can reconsider it, otherwise this is best left up to the user with traditional tooling :)

Appreciate the work tho! But I'll merge this when the above mentioned changes are implemented and then we can revisit the other things at a later time :)

Torxed:
I think I've reverted the changes mentioned in the comments.
I don't mind the menu option to choose a shell, as some people will want to run `zsh` for instance.

But that should be a separate PR, otherwise this PR will do many things and while I allow it in some situations I would prefer if this PR was regarding hyprland only (and ideally just the renaming since that's what the title is hehe).

But I kept the renaming, and then added a seat selector as both seatd and polit appears to be options for Hyprland.

Note that I have never used hyprland tho and feedback on usage is welcome.

Torxed:
I've had issues on some other wayland profiles as well, but the greeter throws me back to the login after attempting to start Hyprland:

screenshot

vaxerski:
@Torxed likely missing gpu accel, since it's a VM. Either enable it or pass `WLR_RENDERER_ALLOW_SOFTWARE=1` to Hyprland at launch

Sxmourai:
Yes sorry ! I coded this mostly in the car while going on holidays 😅
The zsh part should definitely be in another PR. I can work on the configuration part if you want... 
And for the nvidia part I think we could make it work for sway right ? I had the issue with my arch installation, and it took me ~1 hour to make wayland work with a nvidia card so if we could merge this it would be cool too

Torxed (approved):

Torxed:
@svartkanin I think this is the one to merge, of all the Hyprland renames heh.

svartkanin:
The class name is still set to `Hyperland` so that needs to be updated and do we really need `kitty` to get it to work? That seems more like a subjective customization

Torxed:
Removed `kitty` and renamed the class! Thanks for noting it!

Torxed:
Ah, maybe we add it in then. As I assume it will be hard graphically to do anything cmd related otherwise?

vaxerski:
well, not hard, impossible lol

image

These are all the default binds

Torxed:
> well, not hard, impossible lol
> 
> image
> 
> These are all the default binds (ignoring workspace manip / moving windows)

I think it makes sense to have kitty, dolphin and wofi in this case.

dylanmtaylor:
> > well, not hard, impossible lol
> > 
> > image
> > 
> > These are all the default binds (ignoring workspace manip / moving windows)
> 
> I think it makes sense to have kitty, dolphin and wofi in this case.

I agree with this. If the original developers are that opinionated on what tools are used, we need to provide a working out of the box installation.

Torxed:
I opened a PR to add them, but need at least one approval in order to merge.
This is a new limitation we set in order to allow people that are not (yet?) Arch Linux staff to merge code. This unfortunately blocks devs like myself as well - so awaiting approval in #2092 :)

vaxerski:
> I agree with this. If the original developers are that opinionated on what tools are used, we need to provide a working out of the box installation.

I have a "welcome"/"initial setup" type of utility planned for hyprland, but it's, as I said, just planned.

Torxed:
> > I agree with this. If the original developers are that opinionated on what tools are used, we need to provide a working out of the box installation.
> 
> I have a "welcome"/"initial setup" type of utility planned for hyprland, but it's, as I said, just planned. For now, I just made those binds a year ago resembling what I was using and it kinda stuck around.

Ah you're the creator, just noticed. Thanks for the feedback here and cool to see everyone get involved :) hope the profile looks good.

A setup would be handy I recon, but with any project.. there's always more urgent tasks hehe

PR #1831, opened 2023-05-21
Fix 1830

### LINKED ISSUE #1830: Error on opening log file
At commit 5276d95339368210e75791e2b88c1bf5aca4517b, it would seem this commit is broken for me.

On a fresh VM, it complains about having a partition outside of the disk.

```bash
This is an unofficial ISO for development and testing of archinstall. No support will be provided.
This ISO was built from Git SHA 5276d95339368210e75791e2b88c1bf5aca4517b
Type archinstall to launch the installer.
root@archiso ~/archinstall-git # archinstall
  File "/usr/lib/python3.11/site-packages/archinstall/lib/disk/device_handler.py", line 49, in load_devices
    disk = Disk(device)
           ^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/parted/decorators.py", line 42, in new
    ret = fn(*args, **kwds)
          ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/parted/disk.py", line 52, in __init__
    self.__disk = _ped.Disk(device.getPedDevice())
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ped.DiskLabelException: Can't have a partition outside the disk!

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.11/site-packages/archinstall/__init__.py", line 8, in <module>
    from .lib import disk
  File "/usr/lib/python3.11/site-packages/archinstall/lib/disk/__init__.py", line 1, in <module>
  File "/usr/lib/python3.11/site-packages/archinstall/lib/disk/device_handler.py", line 603, in <module>
                     ^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/archinstall/lib/disk/device_handler.py", line 38, in __init__
    self.load_devices()
  File "/usr/lib/python3.11/site-packages/archinstall/lib/disk/device_handler.py", line 54, in load_devices
    debug(f'Unable to get disk from device: {device}')
  File "/usr/lib/python3.11/site-packages/archinstall/lib/output.py", line 259, in debug
    log(*msgs, level=level, fg=fg, bg=bg, reset=reset, font=font)
  File "/usr/lib/python3.11/site-packages/archinstall/lib/output.py", line 304, in log
    with open(absolute_logfile, 'a') as fp:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/var/log/archinstall/install.log'
```

_Originally posted by @Reflux0301 in https://github.com/archlinux/archinstall/issues/1828#issuecomment-1555386857_

svartkanin:
Should fix #1830

### DIFF
--- a/archinstall/lib/output.py
+++ b/archinstall/lib/output.py
@@ -137,26 +137,28 @@ class Journald:
 
 def check_log_permissions():
 	filename = storage.get('LOG_FILE', None)
+	log_dir = storage.get('LOG_PATH', Path('./'))
 
 	if not filename:
-		return
+		raise ValueError('No log file name defined')
 
-	log_dir = storage.get('LOG_PATH', Path('./'))
-	absolute_logfile = log_dir / filename
+	log_file = log_dir / filename
 
 	try:
 		log_dir.mkdir(exist_ok=True, parents=True)
-		with absolute_logfile.open('a') as fp:
+		log_file.touch(exist_ok=True)
+
+		with log_file.open('a') as fp:
 			fp.write('')
 	except PermissionError:
 		# Fallback to creating the log file in the current folder
-		fallback_log_file = Path('./').absolute() / filename
-		absolute_logfile = fallback_log_file
-		absolute_logfile.mkdir(exist_ok=True, parents=True)
-		storage['LOG_PATH'] = Path('./').absolute()
+		fallback_dir = Path('./').absolute()
+		fallback_log_file = fallback_dir / filename
 
-		err_string = f"Not enough permission to place log file at {absolute_logfile}, creating it in {fallback_log_file} instead."
-		warn(err_string)
+		fallback_log_file.touch(exist_ok=True)
+
+		storage['LOG_PATH'] = fallback_dir
+		warn(f'Not enough permission to place log file at {log_file}, creating it in {fallback_log_file} instead')
 
 
 def _supports_color() -> bool:
@@ -248,6 +250,7 @@ def info(
 ):
 	log(*msgs, level=level, fg=fg, bg=bg, reset=reset, font=font)
 
+
 def debug(
 	*msgs: str,
 	level: int = logging.DEBUG,
@@ -257,6 +260,7 @@ def debug(
 	font: List[Font] = []
 ):
 	log(*msgs, level=level, fg=fg, bg=bg, reset=reset, font=font)
+
 
 def error(
 	*msgs: str,
@@ -295,14 +299,10 @@ def log(
 	if _supports_color():
 		text = _stylize_output(text, fg, bg, reset, font)
 
-	# If a logfile is defined in storage,
-	# we use that one to output everything
-	if filename := storage.get('LOG_FILE', None):
-		log_dir = storage.get('LOG_PATH', Path('./'))
-		absolute_logfile = log_dir / filename
+	log_file: Path = storage['LOG_PATH'] / storage['LOG_FILE']
 
-		with open(absolute_logfile, 'a') as fp:
-			fp.write(f"{orig_string}\n")
+	with log_file.open('a') as fp:
+		fp.write(f"{orig_string}\n")
 
 	Journald.log(text, level=level)
 


archinstall/lib/output.py
```diff
@@ -137,26 +137,28 @@ def log(message: str, level: int = logging.DEBUG) -> None:
 
 def check_log_permissions():
 	filename = storage.get('LOG_FILE', None)
+	log_dir = storage.get('LOG_PATH', Path('./'))
```
codefiles, line 140:
```suggestion
	log_dir = storage.get('LOG_PATH', Path().cwd())
```

archinstall/lib/output.py
```diff
@@ -137,26 +137,28 @@ def log(message: str, level: int = logging.DEBUG) -> None:
...
-	absolute_logfile = log_dir / filename
+	log_file = log_dir / filename
 
 	try:
 		log_dir.mkdir(exist_ok=True, parents=True)
-		with absolute_logfile.open('a') as fp:
+		log_file.touch(exist_ok=True)
+
+		with log_file.open('a') as fp:
 			fp.write('')
```
codefiles, line 152:
```suggestion

```

Is this still needed now that `log_file.touch(exist_ok=True)` has been added just above this?

archinstall/lib/output.py
```diff
@@ -137,26 +137,28 @@ def log(message: str, level: int = logging.DEBUG) -> None:
...
 	except PermissionError:
 		# Fallback to creating the log file in the current folder
-		fallback_log_file = Path('./').absolute() / filename
-		absolute_logfile = fallback_log_file
-		absolute_logfile.mkdir(exist_ok=True, parents=True)
-		storage['LOG_PATH'] = Path('./').absolute()
+		fallback_dir = Path('./').absolute()
```
codefiles, line 155:
```suggestion
		fallback_dir = Path().cwd()
```

The path object returned by `cwd()` will contain an absolute path.
https://docs.python.org/3/library/pathlib.html#pathlib.Path.cwd

codefiles:
Checking the log file permissions just once on initialization (`__init__.py`) and not each time it is opened in append mode for a write is an example of a race condition. While the program is running the permissions for the log file could be changed and without it being checked if it is inaccessible the program will crash.

Torxed:
I agree that it is a theoretical possibility. It would have to be done purposefully tho as we don't change any permissions during runtime.

I'm fine merging this knowing this risk, as 99.9% of live ISO environments most likely won't be affected. And users running custom scripts most likely won't modify log file permissions during execution either.

It's up to you @svartkanin if you wanna merge the suggested changes or not, they're valid points but in practicality I don't think we'll ever hit that race condition and the impact is single-user only.

I'm ready when I get a GO :)

svartkanin:
@svartkanin I don't see a reason that this might be a concern in practice. As you said we're not modifying any permissions and if someone is trying to do it deliberately it's kind of their responsibility anyways

codefiles:
> I'm fine merging this knowing this risk, as 99.9% of live ISO environments most likely won't be affected. And users running custom scripts most likely won't modify log file permissions during execution either.

> @svartkanin I don't see a reason that this might be a concern in practice. As you said we're not modifying any permissions and if someone is trying to do it deliberately it's kind of their responsibility anyways

Why would both of you wager against Murphy's law when I have offered a viable solution?

svartkanin:
Because it's an overkill to repeatedly check file permissions. There should be an initial state of the program that is expected and which is the starting point. After that validation point, if the user is doing something unexpected they should be informed about that

codefiles:
> Because it's an overkill to repeatedly check file permissions. There should be an initial state of the program that is expected and which is the starting point. After that validation point, if the user is doing something unexpected they should be informed about that

If the log was written to a thousand times during execution of the program your optimization would save in the range of a thousandth of a second.

Torxed:
> Why would both of you wager against

Not against, just asking if svartkanin wanted to tweak the code or if this should be merged as is.

I'll merge this and the enhancement can be a separate PR.

Torxed:
Also, I appreciate the work you both put in! You've both accelerated the development pace and keep pushing the quality forward! Keep it up!

PR #1839, opened 2023-05-26
Update i3.py

Tr3bleee:
Bug with the installation of the i3 tiling manager because of an extra comma in the file

- This fix issue:

## PR Description:

Removed the comma after the "dmenu" package, since it is the final package

```
	@property
		return [
			'i3-wm',
			'i3lock',
			'i3status',
			'i3blocks',
			'xterm',
			'lightdm-gtk-greeter',
			'lightdm',
			'dmenu' # Removed comma
		]
```

## Tests and Checks

### DIFF
--- a/archinstall/default_profiles/desktops/i3.py
+++ b/archinstall/default_profiles/desktops/i3.py
@@ -21,7 +21,7 @@ class I3wmProfile(XorgProfile):
 			'xterm',
 			'lightdm-gtk-greeter',
 			'lightdm',
-			'dmenu',
+			'dmenu'
 		]
 
 	@property


Tr3bleee:
Python 3.10.10 is preinstalled in archiso, and apparently in this version, it reacts to this comma, because when I removed this comma, my installation started

svartkanin:
Might it be that you were running an older version of archinstall, we did have a comma issue in i4 but that got resolved already

Torxed:
> Might it be that you were running an older version of archinstall, we did have a comma issue in i4 but that got resolved already

This is what I was thinking as well, and the comma was in a different spot then. So updating would help.

PR #1848, opened 2023-06-07
Add some german translations

Kefaku:
## PR Description:

Added some more german translations to `/locales/de/LC_MESSAGES/base.po`.

Also, I noticed various keys like `msgid "Defined"` appear multiple (mostly 3) time only in `/locales/de/LC_MESSAGES/base.po`, but not in other languages. I didn't change anything about that, cuz idk whether that's some sort of weird system I don't understand or something went wrong with a merge in the past.

## Tests and Checks
- [ ] I have tested the code!


Torxed:
That's mostly my conflict resolution that added those and can be safely removed.

Thanks for this PR tho and the work!

PR #1854, opened 2023-06-08
Update bootctl --path option to --esp-path

codefiles:
Since systemd v242 the bootctl `--path` option has been a compatibility alias and is therefore no longer documented.

### References
- https://man.archlinux.org/man/bootctl.1#OPTIONS
- https://github.com/systemd/systemd/blob/0b2aa2064f93cc154e704b262713de4fc395ce56/src/boot/bootctl.c#L231-L232
- https://github.com/systemd/systemd/commit/d25d289ae2e72be97d794d2895ea016410f9c1e0

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -793,10 +793,10 @@ class Installer:
 
 		# Install the boot loader
 		try:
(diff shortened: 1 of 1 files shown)


Torxed:
Awesome! Merged!

Issue #1832, opened 2023-05-23
Install Archlinux on a single ext4 partition


medmedin2014:
I don't use any secure boot so I'm used to installing Ubuntu Mate and Manjaro KED on a single partition ext4 (the others 3  partitions are only NTFS and contain Windows 10 and personal data), I simply set "/" to that partition then start the installation process, but with `archinstall` in manual partitioning I set mountpoint to `/` and boot to `True` and wipe to `True` and format to `ext4`, but after clicking on install it gives kind of error `! Formatting...Could not find... in disk after 5 retries...` then it gets stack at `Setting boot on on (parted) partition index 1`. I'm using archlinux-2023.05.03-x86_64.iso which I put on a USB using Ventoy.

jothi-prasath:
I am also experiencing the same error

svartkanin:
There have been some major upgrades which are not released yet. If you have a chance you could try to run the latest master

Issue #1650, opened 2023-02-23
Set /efi partition as default ESP mountpoint (separate /boot and /efi)


kojq:
Referencing https://wiki.archlinux.org/title/EFI_system_partition#Typical_mount_points:

> /efi is a replacement[[5]](https://github.com/systemd/systemd/pull/3757#issuecomment-234290236) for the historical and now discouraged ESP mountpoint /boot/efi.

Thus, could we consider `/efi` as the default `ESP` mountpoint? I know this was mentioned a few times, such as in /pull/1583. However, here is an issue dedicated to discussing separate partitions for `/boot` and `/efi`.

Torxed:
100% agree with this, as mentioned in:
 * https://github.com/archlinux/archinstall/issues/645
 * https://github.com/archlinux/archinstall/issues/1072
 * https://github.com/archlinux/archinstall/issues/1189

I'll keep this issue open to track the other 3 as the title is a better description than the other ones.

uhthomas:
@Torxed I'm confused, was this issue resolved? I don't believe `/efi` is the default mount point.

Issue #1597, opened 2022-12-16
Partition /mnt/archinstall/boot too full


Confuze:
I'm trying to multiboot arch linux with manjaro and windows but archinstall keeps failing when it tries to install some packages.

Here's whole error:
https://imgur.com/a/B004woX

I'm using v2.5.0 because the latest version (v2.5.2) has some other issues with my nvme drive. (Described it in #1593)

Here's my drive setup: (photo from v2.5.2 but I have the exact same setup in v2.5.0)
https://imgur.com/a/s0mP7Cz

And here's my whole config: (again, different version but config is the same)
https://imgur.com/a/UD1hd1p

I tried unchecking multilib and additional packages but it still tries to install a lot of stuff in the boot partition (whcich I explicitly set to be mounted to /boot and have another one mounted to /)

I'd love to provide the whole log file but I have no idea how to save it. Any tips?

lucasvs:
I also got this error today, i typed yes each time the error popped up and somehow the last time it worked.

Confuze:
> I also got this error today, i typed yes each time the error popped up and somehow the last time it worked.

Lol that's one way to fix problems for sure. Gonna try screaming yes at it then. Thanks.

Torxed:
I'll post a new release and then i'll start working on https://github.com/archlinux/archinstall/issues/1650 which should resolve this.

Confuze:
Awesome, thanks!

Issue #1306, opened 2022-06-04
Supporting single root partition schemes


JoelLarson:
Attempting to use archinstall on a single root partition scheme does not complete the installation and fails with the following message:

```
# Moved here for readability
$root=Partition(path=/dev/sda2, size=420.3, PARTUUID=bb94b01a-d29a-4feb-a650-c1aad564a828, fs=ext4, mounted=/mnt/archinstall)

Could not detect root ($root) or boot (None) in /mnt/archinstall based on: dict_values([Partition(path=/dev/sda2, size=420.3, PARTUUID=bb94b01a-d29a-4feb-a650-c1aad564a828, fs=ext4, mounted=/mnt/archinstall)])
```

Since boot is not found because this partition scheme does not use it, the install fails. [This code is responsible](https://github.com/archlinux/archinstall/blob/master/archinstall/lib/installer.py#L1020-L1030) as it makes an assumption that `/boot` will always be a partition.

I would be interested in implementing this fix if supporting single root partitions should be possible with the installer. Otherwise I would like to add a message explaining this problem when it is detected.

Log file:
```
Hardware model detected: Dell Inc. OptiPlex 990; UEFI mode: False
Processor model detected: Intel(R) Core(TM) i5-2400 CPU @ 3.10GHz
Memory statistics: 11687152 available out of 12133084 total installed
Could not detect virtual system: ['/usr/bin/systemd-detect-virt'] exited with abnormal exit code [256]: b'none\r\n'
System is not running in a VM: ['/usr/bin/systemd-detect-virt'] exited with abnormal exit code [256]: b'none\r\n'
Virtualization detected: None; is VM: None
Graphics devices detected: dict_keys(['Intel Corporation 2nd Generation Core Processor Family Integrated Graphics Controller (rev 09)'])
[...]

PR #1859, opened 2023-06-12
Enable separate /boot and /boot/esp via XBOOTLDR in systemd-boot

Torxed:
This PR will solve:
 * https://github.com/archlinux/archinstall/issues/1832
 * https://github.com/archlinux/archinstall/issues/1072
 * https://github.com/archlinux/archinstall/issues/1650
 * https://github.com/archlinux/archinstall/issues/1597
 * https://github.com/archlinux/archinstall/issues/1306
 * https://github.com/archlinux/archinstall/issues/1072
 * https://github.com/archlinux/archinstall/issues/862 *(partially addresses this)*
 * https://github.com/archlinux/archinstall/issues/645

This PR partially initiates the need for the boot partition to be explicitly located at `/boot` as well. However we have a bunch of hard coded places where `/boot` is used. So this will by no means make this dynamic on its own.

But it opens up to implement more dynamic boot partitions in the future.

# Breaking changes:

 * `is_boot()` now returns True if either `Boot` or `XBOOTLDR` is set in the declarative configuration *(as both are part of the boot process)*. Further filtration has to be done on the user end depending on their needs. We do that in `get_efi_partition()` and `get_boot_partition()`.
 * `Installer._add_boot_partition()` parameter order is changed, the first parameter is now `boot_partition` to reflect the order of the other two functions.


Torxed:
Note to self, do we need to worry about https://github.com/dcantrell/pyparted/blob/4a197317794d22be3cab3177f1cc660df9e664f2/src/_pedmodule.c#L650 ?

Torxed:
screenshot

Partitions appear to be setup correctly, and boot loader is installed successfully.
But it refuses to boot for some reason.

codefiles:
Are you trying with systemd-boot? There are other location in `_add_systemd_bootloader()` that have `/boot` hardcoded. See #1866

Dazzlly:
I'm confused, I'm having the same problem as everyone else when installing arch with other systems in Dual boot, it doesn't show the option to disable ESP for a certain partition, but reading all these articles I don't know what I should call the partition that contains Grub UEFI, call /efi or /boot/efi or /boot/esp or all with comma separation?? what about my other partition just in /boot how do i disable the ESP flag without disabling Boot? because I use these two separate partitions due to the dual boot, I also have separate /home and swap partitions, how should I mount them? /home is apparently correct but for example recently one of the errors I had in / (root) said that something like /sysroot was missing ???

Torxed:
> I'm confused, I'm having the same problem as everyone else when installing arch with other systems in Dual boot, it doesn't show the option to disable ESP for a certain partition, but reading all these articles I don't know what I should call the partition that contains Grub UEFI, call /efi or /boot/efi or /boot/esp or /esp or all with comma(, ) separation?? what about my other partition just in /boot how do i disable the ESP flag without disabling Boot? because I use these two separate partitions due to the dual boot, I also have separate /home and swap partitions, how should I mount them? /home is apparently correct but for example recently one of the errors I had in / (root) said that something like /sysroot was missing ???

So, not until 5 minutes ago, there was no way in archinstall to separate `/efi` and `/boot`. And `archinstall` has historically had a combined `ESP` partition for both `efi` and `boot` data.

With this PR, we can now do:
```json
"disk_config": {
    "device_modifications": [
        {
            "device": "/dev/sda",
            "partitions": [
                {
                    "btrfs": [],
                    "flags": [
                        "Boot"
                    ],
                    "fs_type": "fat32",
                    "length": {
                        "sector_size": null,
                        "total_size": null,
                        "unit": "MiB",
                        "value": 512
                    },
                    "mount_options": [],
                    "mountpoint": "/boot/EFI",
                    "obj_id": null,
                    "start": {
                        "sector_size": null,
                        "total_size": null,
                        "unit": "MiB",
                        "value": 1
                    },
                    "status": "create",
                    "type": "primary"
                },
                {
                    "btrfs": [],
                    "flags": [
                        "XBOOTLDR"
                    ],
                    "fs_type": "fat32",
                    "length": {
                        "sector_size": null,
                        "total_size": null,
                        "unit": "MiB",
                        "value": 1024
                    },
                    "mount_options": [],
                    "mountpoint": "/boot",
                    "obj_id": null,
                    "start": {
                        "sector_size": null,
                        "total_size": null,
                        "unit": "MiB",
                        "value": 513
                    },
                    "status": "create",
                    "type": "primary"
                },
                {
                    "btrfs": [],
                    "flags": [],
                    "fs_type": "ext4",
                    "length": {
                        "sector_size": null,
                        "total_size": {
                            "sector_size": null,
                            "total_size": null,
                            "unit": "B",
                            "value": 16106127360
                        },
                        "unit": "Percent",
                        "value": 100
                    },
                    "mount_options": [],
                    "mountpoint": "/",
                    "obj_id": null,
                    "start": {
                        "sector_size": null,
                        "total_size": null,
                        "unit": "MiB",
                        "value": 1537
                    },
                    "status": "create",
                    "type": "primary"
                }
            ],
            "wipe": true
        }
    ]
},
```

The notable parts being:

EFI partition
-------------
This partition can be re-used *(for instance, a pre-created partition from Windows)*

```json
"flags": [
    "Boot"
],
"fs_type": "fat32",
"mountpoint": "/boot/EFI",
```

XBOOTLDR partition
------------------
```json
"flags": [
    "XBOOTLDR"
],
"fs_type": "fat32",
"mountpoint": "/boot",
```
As mentioned in https://wiki.archlinux.org/title/Systemd-boot#Installation_using_XBOOTLDR, `fat32` is a safe bet for the `XBOOTLDR` partition but depending on your UEFI implementation others might work too.



I currently have not modified the GRUB code to support this however.
As I rarely (if ever) use GRUB this will be a bit of a wiki-reading challenge for me, and hope that it works.

archinstall/lib/disk/device_model.py
```diff
@@ -765,8 +779,16 @@ def device_path(self) -> Path:
 	def add_partition(self, partition: PartitionModification):
 		self.partitions.append(partition)
 
+	def get_efi_partition(self) -> Optional[PartitionModification]:
+		"""
+		Returns the first EFI partition marked for boot.
+		Perhaps we could mark it more clearly in the PartitionModification() model instead.
+		"""
+		liltered = filter(lambda x: x.fs_type == FilesystemType.Fat32 and x.is_boot() and str(x.mountpoint) == '/boot/EFI', self.partitions)
```
Torxed, line 787: Before merging, I will remove `/boot/EFI` check, and make sure that we check for the `PARTTYPE` identifier `bc13c2ff-59e6-4262-a352-b275fd6f7172` instead, and rename it to `get_xbootldr_partition` as that's really what we're checking here.
`efi` was incorrectly chosen as a placeholder name - as /boot can be an "EFI" partition as well.

archinstall/lib/installer.py
```diff
@@ -698,13 +709,17 @@ def _add_systemd_bootloader(self, root_partition: disk.PartitionModification):
...
 		try:
-			SysCommand(f'/usr/bin/arch-chroot {self.target} bootctl --esp-path=/boot install')
+			SysCommand(f"/usr/bin/arch-chroot {self.target} bootctl {' '.join(bootctl_options)} install")
 		except SysCallError:
 			# Fallback, try creating the boot loader without touching the EFI variables
-			SysCommand(f'/usr/bin/arch-chroot {self.target} bootctl --no-variables --esp-path=/boot install')
+			SysCommand(f"/usr/bin/arch-chroot {self.target} bootctl --no-variables {' '.join(bootctl_options)} install")
 
 		# Ensure that the /boot/loader directory exists before we try to create files in it
 		loader_dir = self.target / 'boot/loader'
```
codefiles, line 725:
```suggestion
		loader_dir = self.target / efi_partition / 'loader'
```

Do not forget to update this.

codefiles: The comment above it and the comment one the line before `entries_dir = loader_dir / 'entries'` should be updated too since they contain `/boot`. Maybe change `/boot` to `esp` in those comments.

Torxed: That's wise! Will do!

Torxed: hmm, isn't the path the same regardles of `efi_partition` or not? because `efi_partition` will be `/boot/EFI` in one case but `boot_partition` will be `/boot` in both. So I think keeping it makes sense. As the loader configuration will remain in `/boot/loader/entries/` in both instances?

codefiles: Take a look at how the ArchWiki refers to the path: https://wiki.archlinux.org/title/Systemd-boot#Loader_configuration

It uses <code><i>esp</i>/loader</code>, that is the source for my suggestion.

Torxed: https://wiki.archlinux.org/title/Systemd-boot#Installation_using_XBOOTLDR mentions:
> XBOOTLDR partition to `/mnt/boot`

And then to configure as normal, so `/boot/loader/entries` should naturally be on the `boot_partition` as XBOOTLDR partition will be found in `boot_partition`?

I might be confused tho, been bouncing between wiki entries and external sources on what the XBOOTLDR partition should or shouldn't be.

codefiles: At the top of the section you gave, emphasis mine:

> A separate [/boot partition](https://wiki.archlinux.org/title/Partitioning#/boot) of type "Linux extended boot" (XBOOTLDR) can be created to keep the **kernel** and **initramfs** _separate_ from the **ESP**.

codefiles: The ESP can be mounted at `/efi`, `/boot`, or `/boot/efi`. Using `/efi` is the most modern way and what I would suggest.

Torxed: I thought it was specifically for the EFI binary blobs, and it in turn checked `/boot/loader/loader.conf`. As the boot process worked for me I saw that as confirmation. So this is not the case?

Torxed: I can see why it would work on my test, as the loader *can* init via the EFI but finds no entries.

On the flip side, it could find entries but not load them?

codefiles: > - On disks with GPT (GUID Partition Table)
>
>   - The EFI System Partition (ESP for short) — a partition with a GPT type GUID of c12a7328-f81f-11d2-ba4b-00a0c93ec93b — may be used as one of two locations for boot loader menu entries.
>
>   - Optionally, an Extended Boot Loader Partition (XBOOTLDR partition for short) — a partition with GPT type GUID of bc13c2ff-59e6-4262-a352-b275fd6f7172 — may be used as the second of two locations for boot loader menu entries. This partition must be located on the same disk as the ESP.
    
https://uapi-group.org/specifications/specs/boot_loader_specification/#the-partitions

> It is recommended to mount $BOOT to /boot/, and the ESP to /efi/. If $BOOT and the ESP are the same, then either a bind mount or a symlink should be established making the partition available under both paths.

https://uapi-group.org/specifications/specs/boot_loader_specification/#mount-points

> A boot loader locates the XBOOTLDR partition and the ESP, then simply reads all the files /loader/entries/*.conf in them, and populates its boot menu (and handle gracefully if one of the two partitions is missing).

https://uapi-group.org/specifications/specs/boot_loader_specification/#locating-boot-entries

I would recommend doing a thorough read of this documentation and then read it once again. Any question you have on this topic should be answered in it.

Torxed: Sure, I can read more documentation.
But this expert:

> Optionally, an Extended Boot Loader Partition (XBOOTLDR partition for short) — a partition with GPT type GUID of bc13c2ff-59e6-4262-a352-b275fd6f7172 — may be used as the second of two locations for boot loader menu entries. This partition must be located on the same disk as the ESP.

In my mind confirms the current implementation. Where `XBOOTLDR` would be represented by the `boot_partition` in the code, *may* be used for boot loader menu entries.

The only check we might not do currently is to verify that the partition exists on the same disk as the ESP/EFI partition.

Or am I reading that incorrectly?

codefiles: It looks like when an XBOOTLDR is to be used, it is the correct location for the boot entries. I would still recommend against hard coding `/boot` at this location and in the comments.

codefiles:
In the example above the mount point `/boot/EFI` is incorrect, it should be `/boot/efi`. This could be the source of your issue from the discussion in the review above:

> I can see why it would work on my test, as the loader _can_ init via the EFI but finds no entries.
> 
> On the flip side, it could find entries but not load them?

Avoid using the mount point `/boot/efi` anyway.

> It is recommended to mount `$BOOT` to `/boot/`, and the ESP to `/efi/`.

> Mounting the ESP to `/boot/efi/`, as was traditionally done, is not recommended.

https://uapi-group.org/specifications/specs/boot_loader_specification/#mount-points

In the man page for bootctl under the `--esp-path=` option.

> It is recommended to mount the ESP to /efi/, if possible.

https://man.archlinux.org/man/bootctl.1#OPTIONS

archinstall/lib/installer.py
Torxed: Reading it a third time, I am still convinced that the implementation is correct.
Namely this:

> If the ESP is separate from $BOOT it may also contain a /loader/entries/ directory, where the boot loader should look for boot entry snippets, as an additional source. The boot loader should enumerate both directories and present a merged list to the user. Note that this is done for compatibility only: while boot loaders should look in both places, **OSes should only add their files to $BOOT.**

ESP *(`efi_partition` in the code)* is separated from $BOOT *(`boot_partition`)*, seeing as both folders are enumerated, and that the recommendation is that the OS *(that would be us?)* should only add to $BOOT *(`boot_partition`)*, aka XBOOTLDR.

I still feel like you're questioning the implementation but the more I read it I start to believe that the implementation is sane?

Torxed:
> Avoid using the mount point `/boot/efi/` anyway and use `/efi/` instead.

Yepp, the `/boot/efi` was just to make sure custom paths worked that weren't in the auto detection paths by `bootctl` as mentioned in the commit https://github.com/archlinux/archinstall/pull/1859/commits/f356246b5bc906e0385ec713b66275deafb9fa83, I'll move to `/efi` now :D

> In the example above the mount point /boot/EFI is incorrect, it should be /boot/efi

That might be it. As I didn't get it working with `/boot/efi` at first.
Did a bunch of changes and got it working and thought `/boot/EFI` would be correct.

I'll tweak around and make sure `/efi` works.

codefiles:
Another item to note is the directory `/boot` is provided by the filesystem package but `/efi` is not. This should mean the `/efi` directory will need to be created before it can be utilized.

Torxed:
> Another item to note is the directory `/boot` is provided by the filesystem package but `/efi` is not. This should mean the `/efi` directory will need to be created before it can be utilized.

I believe it's created by `bootctl`, as it worked:
screenshot

I'm not usually in favor of relying on external applications, but perhaps this is good enough to get into `master` so we can work on tweaks separably? I kinda want to get this in for a `[testing]` release.

archinstall/lib/installer.py
Torxed: One reason why it perhaps would make sense to leave loader entries in `$BOOT/loader/entries` is that they're related to the installation, which `boot_partition` is *(it also happens to work out in our favor as the code can be simplified to deal with bot cases, `ESP` + `XBOOTLDR` and in the case of one united partition)*

codefiles: In my last comment I was agreeing with you. I also agree with your comments above, the implementation is sane.

This particular line of code we are commenting on should still have the `boot` portion of the string changed to a variable and not be hard coded. If an XBOOTLDR exists the variable should point to the mount point of it and if it does not is should point to the mount point of the ESP. Even if this ends up being `boot` for both or all cases let us not hard code it and acquire the path from a variable. For the comments that I referenced as needing to be updated, I recommend just dropping the `/boot/` prefix from the paths in them.

Torxed: > In my last comment I was agreeing with you.

Oh, I misunderstood it, my bad!
I think I put too much thought into this kinda small PR because it's been on the TODO since 2021 :S

> the variable should point to the mount point of it 

That's a good idea!

Dazzlly:
Have there been any changes? Have you updated? Because I have a dual boot installation on efi with 3 other systems, And that's why I would like to know how I'm going to mount the boot partition where my Grub is because in the other installations I used the installation in efi, And the last time I tried to install the system with ArchInstall in /boot/efi and in /efi the installation breaks Grub and also doesn't install ArchLinux correctly because it doesn't find boot files

Torxed:
We have't pushed this release yet but we should have it fixed in `master`, where `/efi` and `/boot` is working as long ad they have the correct PARTGUID's

PR #1863, opened 2023-06-12
Bump Python in workflow

codefiles:


codefiles:
Do not merge yet since this change causes the workflow to fail at the _Install build dependencies_ step.

https://github.com/archlinux/archinstall/actions/runs/5243024919/jobs/9467166843?pr=1863#step:6:1

Torxed:
Thanks! This was on my TODO before releasing a new version.
We also need to hold off until `simple_term_menu` is in the arch package listing, which most likely will happen when my signing key is fully through the on-boarded with the TU team.

I'm trying to squeeze in a few more features before my key is on-boarded.

Torxed:
[python-simple-term-menu](https://archlinux.org/packages/extra/any/python-simple-term-menu/) is now a official package, together with [python-pyparted](https://archlinux.org/packages/extra/x86_64/python-pyparted/) that concludes our dependencies for a release.

Torxed:
Created a PR to bump PKGBUILD etc: https://github.com/archlinux/archinstall/pull/1901

Torxed:
Not sure why we're getting: https://github.com/archlinux/archinstall/actions/runs/5401038764/jobs/9810298743?pr=1863#step:6:17:
> /__w/_temp/da4bc36b-a35c-4d76-bd39-074500d8d7eb.sh: /__t/Python/3.11.4/x64/bin/pip: /opt/hostedtoolcache/Python/3.11.4/x64/bin/python: bad interpreter: No such file or directory

codefiles:
It is probably not appropriate or necessary to use `actions/setup-python@v4` since there is also an instance of Python in the `archlinux:latest` container. That could be the source of the issue.

codefiles:
There is also the `name` and `with` associated with the step that you should remove.

Torxed:
Lets see if I got it right this time :)

Torxed:
Was missing pip, and I added in the dependencies via `pacman` instead of `pip` just to make sure we test against the packaged versions of those two dependencies. pip might be ahead of us in some instances.

Torxed:
That did the trick.
Any feedback before merging?

codefiles:
LGTM

Torxed:
Boom! In it goes! Prepping for release candidate :)

PR #1866, opened 2023-06-12
Use pathlib in `_add_systemd_bootloader()`

codefiles:


Torxed:
Thank you, this will make cleaning up the boot code a bit easier

PR #1870, opened 2023-06-13
Fixes 1861

### LINKED ISSUE #1861: Translation - UnicodeDecodeError 'utf-8' codec can't decode byte
screenshot

svartkanin:
- Fixes #1861 
- Add workflow to verify translation files


Torxed:
The workflow normally only gets triggered after it's been merged for some reason. So i'll merge and we'll evaluate!

The other issues are separate yes and stem from a previous merge.

PR #1877, opened 2023-06-14
Update budgie.py

Tr3bleee:
Bug with the installation of the budgie because of an extra comma in the file

- This fix issue: 

## PR Description:

Removed the comma after the "papirus-icon-theme" package, since it is the final package

```
	@property
		return [
			"arc-gtk-theme",
			"budgie",
			"mate-terminal",
			"nemo",
			"papirus-icon-theme" # Removed comma
		]
```

## Tests and Checks
- [🮱] I have tested the code!<br>

### DIFF
--- a/archinstall/default_profiles/desktops/budgie.py
+++ b/archinstall/default_profiles/desktops/budgie.py
@@ -18,7 +18,7 @@ class BudgieProfile(XorgProfile):
 			"budgie",
 			"mate-terminal",
 			"nemo",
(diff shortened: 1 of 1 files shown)


Torxed:
I would prefer if we stop falsely advertising bugs when this is a linting issue.
It causes a bit of confusion for me when verifying these things.

Other than that, I appreciate the effort in cleaning up the code and I'm merging this! :)

Issue #1862, opened 2023-06-12
[master] bring back ability to set start of partition in MB size


Torxed:
I know we had a whole discussion about sectors being the default value.
But explicitly dis-allowing `MB` as a unit feels like a step backwards after using it now:

![rec](https://github.com/archlinux/archinstall/assets/861439/7e8c97be-2a3d-4ae4-bebe-8c3eb35e4cd1)

Personally at least, I can't be bothered knowing what sector correlate to 512MB, but I know certain USB devices/memory cards require a 512MB offset, and just working in human readable units is pretty convenient when you just quickly need to offset a partition or two :)

And if nothing else, not being allowed to use `%` as a unit on the first step, but is allowed on the second is a bit frustrating hehe.

svartkanin:
This is meant to be made more flexible with the new partitioning, I just didn't get to it yet :)

We have a size converter already implemented that allows for easy switcheroo between sizes so this can be integrated quite nicely now

Torxed:
> This is meant to be made more flexible with the new partitioning, I just didn't get to it yet :)

Ah ^^ I'll hold off on tweaking here :D 
I did a rudimentary "sector -> mb" estimation when fiddling with the PR #1859

svartkanin:
You can use the `Size` object defined, it's used pretty much everywhere in the partitioning code. It comes with functions for conversations

PR #1884, opened 2023-06-18
Fix 1862

svartkanin:
This fixes https://github.com/archlinux/archinstall/issues/1862 and brings back the ability to enter values with unit suffixes.


Torxed:
This is a life saver for me! I'd buy you a cake if I could!

PR #1885, opened 2023-06-18
Make Gfx driver handling saver

svartkanin:
Use a dedicated Enum class for both the Gfx driver definitions and the respective packages. 
This will make the code safer as it's verifiable by mypy


Torxed:
Cool way of improving the gfx definitions! I'll have to play around with this a bit.

PR #1891, opened 2023-06-21
Ability to add other DKMS packages.

lavafroth:
This does not fix an issue 

## PR Description:

- A new method `into_dkms` is added to `GfxPackage` to convert a variant to its respective DKMS packages.
- A new method `dkms_packages` is added to `GfxDriver` to get a package list with the non-DKMS packages replaced with their DKMS counterparts.

This should simplify the logic for installing GFX drivers and make it easier for any future DKMS packages to be added.

## Tests and Checks

### DIFF
--- a/archinstall/lib/hardware.py
+++ b/archinstall/lib/hardware.py
@@ -26,7 +26,7 @@ class GfxPackage(Enum):
 	Xf86VideoNouveau = 'xf86-video-nouveau'
 	Xf86VideoVmware = 'xf86-video-vmware'
 
-	def into_dkms(self) -> GfxPackage:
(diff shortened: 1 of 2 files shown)


lavafroth:
The first commit may potentially be a bugfix. On [this line](https://github.com/archlinux/archinstall/blob/72661dbf9bfbd12e42425066f9d5ce7c887f40c7/archinstall/lib/profile/profiles_handler.py#LL209), simply using `driver_pkg` in an f-string will result in `GfxPackage.Nvidia-dkms` instead of `nvidia-dkms` because Python will use `Enum.Variant` as the string representation by default.

lavafroth:
Superseded by #1895

PR #1895, opened 2023-06-23
Refactor installer and general design patterns

lavafroth:
- This is a refactor
## PR Description:
This is an intensive refactor of the existing design patterns. The following are some of the refactor patterns:
- Use of short circuit logic, i.e., returning early for the error path and returning late for the happy path of the code.
- Use of the walrus operator to create a temporary variable in a conditional context.
- Employing sets instead of `list(filter(...))` for finding common elements in lists.
- Reading and writing with `read_text` and `write_text` methods when a `pathlib.Path` is available.
- Custom class to handle pacman related transactions like `sync` and `strap`.
- Replacement of multiple f-string interpolation with a single `pathlib.Path` when applicable for paths like `pacman.conf` or `fstab`.
- Custom class to keep track of changes to `pacman.conf` and apply them in a single pass. This also includes a handy method to persist the changes on the fresh installation.
- Use of iterator patterns when a matched line and its subsequent line has to be uncommented.
- Better documentation for some functions, given the above changes.

## Tests and Checks


svartkanin:
This is really nice, thanks for cleaning it up!

Torxed:
I appreciate the work! As we're doing some changes to the core code we should probably get this into 2.6.0 as well. Since we announced that this one would be breaking a few changes.

lavafroth:
No worries, I have tested the changes in the live environment. I'll merge or rebase off the changes to the core if and when necessary.
Looking forward to getting this merged in 2.6.0 😄

Torxed:
Let me know when you consider this to be done, and I'll have a proper review/test with this and merge it. I'm aiming to test archinstall properly today and tomorrow and hopefully release a new version at the end of tomorrow (Assuming the small fixes I have in mind don't escalate).

lavafroth:
I think I'm pretty happy with how this has turned out. I'll stop adding more commits (unless I find a bug while testing) and you may review it.

archinstall/lib/configuration.py
```diff
@@ -170,27 +161,23 @@ def preview(selection: str):
...
 		save_confirmation = Menu(prompt, Menu.yes_no(), default_option=Menu.yes()).run()
 		if save_confirmation == Menu.no():
 			return
 
 		debug(
-			_("Saving {} configuration files to {}").format(
-				list(options.keys())[list(options.values()).index(str(save_choice.value))],
-				dest_path.absolute(),
-			)
+			_("Saving {} configuration files to {}").format(options[str(save_choice.value)], dest_path.absolute())
```
svartkanin, line 171: The translation here can actually be removed as otherwise we end up with translated text in the log file which isn't good 

```
debug("Saving {} configuration files to {}".format(options[str(save_choice.value)], dest_path.absolute()))
```

lavafroth: Thanks 😄

Torxed:
Great work! I wanted to merge this before it got out of hand too much hehe.
Lets begin testing and I/We can iron out the remaining issues, that way I can push a release with the help of @grazzolini soon hopefully.

Issue #1734, opened 2023-04-08
Encryption partition not loaded from config (>= 2.5.4)


SimonASK8317:
Hi everybody,

when I tried to re-install Arch with saved configs from 2.5.1, I noticed that my configured disk encryption was not applied (everything else was fine).

Digging a little deeper I noticed that apparently, since 2.5.4, the disk encryptions settings are not loaded from the configuration files.

How to reproduced:

1. Boot the April 2023 install ISO (e.g. with VirtualBox)
2. Start `archinstall`
3. Select /dev/sda as the installation drive
4. In the disk layout, select the default BTRFS settings
5. In the disk encryption, set the entryption password and select the previously configured BTRFS partition
6. Set a root password
7. Save all configs (e.g. to /tmp) and copy them to a save place.
8. Install, then reboot. You will be asked for the decryption password, so far so good.
9. Boot the installation ISO again
10. Copy the configuration files you previously saved back to /tmp.
11. Start archinstall `archinstall --config /tmp/user_configuration.json --disk_layout /tmp/user_disk_layout.json --creds /tmp/user_credentials.json`
12. You will notice now, that in the section for Disk encryption, the "Partitions: 1 selected" part is missing
14. If you hit install without further changes OR if you start archinstall with the config files in non-interactive mode (`archinstall --config /tmp/user_configuration.json --disk_layout /tmp/user_disk_layout.json --creds /tmp/user_credentials.json --silent`), if you reboot after the installation finishes, you will not be asked for a decryption password since the disk was not encrypted. During the installation you can also see that there is no "encrypting disk" part.

Disk encryption screen after partition is manually selected:

![encryption_selected_partition](https://user-images.githubusercontent.com/68168003/230744378-914cb047-5f3c-452a-b5f2-0fe55b8b682f.png)

Disk encryption screen when archinstall is started with previously saved configuration files where the partition was selected:

![encryption_missing_partition](https://user-images.githubusercontent.com/68168003/230744379-98a73cb4-2de7-4c9d-a6ed-8f1b194d7f63.png)

I was able to reproduce this behavior in version 2.5.4 which is shipped with the April 2023 installation ISO as well as the version 2.5.5 after updating it via pacman `pacman -Sy archinstall`.

I also noticed that when starting the installation in the error case, archinstall displays the disk_encryption.partitions dict as an empty object. Could be related to the changed made regarding the disk encryption declarations in 2.5.4 (#1672 ), am not that good in Python though to confirm this.

[...]

PR #1897, opened 2023-06-25
Document new pip install flag

rbroggi:
See [this thead](https://github.com/archlinux/archinstall/issues/1734#issuecomment-1605709792) 

Including `--break-operating-system` flag in `pip install` instructions in doc.

- This fix issue: 

## Tests and Checks

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ you can replace the version of archinstall with a new version and run that with 
 4. Now clone the latest repository with `git clone https://github.com/archlinux/archinstall`
 5. Enter the repository with `cd archinstall`
    *At this stage, you can choose to check out a feature branch for instance with `git checkout v2.3.1-rc1`*
-6. Build the project and install it using `pip install`
+6. Build the project and install it using `pip install --break-operating-system .`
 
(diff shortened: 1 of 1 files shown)


Torxed:
Thank you the update!

C0rn3j:
Please revert this and instead suggest users to use a venv, to my knowledge there is no reason to use the breakage flag and it is normalizing its usage for no reason.

The venv change also needs to be done in two more places.

https://github.com/archlinux/archinstall/tree/v2.6.0#installation--usage

https://github.com/archlinux/archinstall/tree/v2.6.0#without-a-live-iso-image
The existing change also broke formatting.

Torxed:
There is a clear reason for suggesting this, and that is because you're installing the source using pip rather than with pacman. The library is also intended to be run in the live ISO environment so no harm should be possible by suggesting this.

It also helps to keep the guide clean as otherwise `/usr/bin/archinstall` wouldn't be accessible via the normal PATH variable.

The dot is on purpose, as you're supposed to build from the source directory. Or did I miss something regarding the dot concern?

Valid concern, but I don't think it's any danger here.

PR #1901, opened 2023-06-28
Adding python-simple-term-menu to the dependency list

Torxed:
This PR addresses some new changes to upstream packaging.
Namely that we now can add dependencies to `python-simple-term-menu` as well as some other `depends` to binaries that we utilize *(this will hopefully enable us to get notifications if and I really mean **if** (as it's unlikely atm) binaries such as `ps` disappears)*

@svartkanin & @IngoMeyer441 *(of the upstream package https://github.com/IngoMeyer441/simple-term-menu)*: the AUR package is now removed and merged into `extra` repo. I appreciate the effort of maintaining the AUR package and the introduction of it into `archinstall` :)

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -13,6 +13,15 @@ license=(GPL3)
 depends=(
   'python'
   'systemd'
+  'procps-ng'
+  'pciutils'
+  'arch-install-scripts'
+  'e2fsprogs'
+  'util-linux'
+  'coreutils'
+  'cryptsetup'
+  'kbd'
+  'btrfs-progs'
   'python-pyparted'
   'python-simple-term-menu'
 )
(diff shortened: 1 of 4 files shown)


svartkanin:
Awesome! Looking forward to the new release

svartkanin:
@Torxed I don't think we need btrfs-progs, unless it's a dependency of something else?

Torxed:
> @Torxed I don't think we need btrfs-progs, unless it's a dependency of something else?

It comes as a dependency due to us calling this for instance: https://github.com/archlinux/archinstall/blob/1ae1f2ff1144d502830834ba5a64262f7d195e91/archinstall/lib/disk/device_handler.py#L137

svartkanin:
Interesting so this was just never listened before as it's not anything new?

Torxed:
Me being a peasant and just relying solely on luck. They've always been part of the ISO so never thought about adding it throughout the years.

I finally had a moment of clarity and did what I should have done years ago ^^

PR #1902, opened 2023-06-28
Created dwm.py

Dark-Kernel:
Created dwm.py for suckless users.

- Added new dwm.py at archinstall/default_profiles/desktops/.

## PR Description:

This will make easier for the dwm users to setup their suckless environment.

## Tests and Checks

### DIFF
--- a/archinstall/default_profiles/desktops/dwm.py
+++ b/archinstall/default_profiles/desktops/dwm.py
@@ -1,0 +1,35 @@
+from typing import List, Optional, Any, TYPE_CHECKING
+
+from archinstall.default_profiles.profile import ProfileType, GreeterType
+from archinstall.default_profiles.xorg import XorgProfile
+
+if TYPE_CHECKING:
+	_: Any
+
+
+class BspwmProfile(XorgProfile):
+	def __init__(self):
+		super().__init__('Bspwm', ProfileType.WindowMgr, description='')
+
+	@property
+	def packages(self) -> List[str]:
+		return [
+			'dwm',
+			'st',
+			'dmenu',
+			'picom',
+			'xorg-xsetroot',
+      'xorg-xset',
+      'pywal',
+      'xdo',
+      'dunst',
+      'slock'
+		]
+
+	@property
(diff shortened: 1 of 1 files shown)


lavafroth (approved):
Looks good!

Torxed:
hmm indentation error, didn't caught that on the phone.

codefiles:
Suckless software is intended to be customized through modification of the source code and built by the user. This is why `dwm` and `st` are not in the Arch Linux official repositories but can be found in the AUR. Since these packages are not in the Arch Linux repos and the AUR is not supported I recommend reverting.

lavafroth:
@codefiles is correct. I looked into the [dwm ArchWiki](https://wiki.archlinux.org/title/Dwm) and it seems that `dwm` is only available through the AUR. Since we don't have or plan to employ an AUR helper, we might want to revert this.

It's better `dwm` users consider a minimal installation and build it from source or use an AUR helper like `paru`.

Dark-Kernel:
Sorry I didn't noticed 😅, And thank you for your consideration. 

But, why we couldn't use AUR?
Yes, I know suckless software is intended to be installed from source, but we can do that thing for the users? 
And after installation they can manually updated and patch their `dwm` and other suckless tools according to them. 
My idea was to giving the base to the user, which can make easier for them.

Anyway Thanks.

lavafroth:
If we try to package an AUR helper alongside the installer, it would open up a whole new can of worms for the "additional packages" section. Sure, users would then be able to install `dwm` and `dwm-git` but they'd probably also be tempted to install proprietary packages like Spotify or Google Chrome without realizing that those AUR packages are probably maintained by third parties and individuals.

I'm not dunking on third party AUR maintainers by any means but I think that new users should at least be aware of what they're installing. Final verdict goes to @Torxed, maybe he can explain this better than me.

Dark-Kernel:
Thanks, understood.

Torxed:
I agree, and I missed that it was an AUR package. But now that it's mentioned I recall we had discussions about this profile before and agreed to not include it. I'll revert, apologies!

PR #1907, opened 2023-06-30
Refactor `_add_systemd_bootloader()` entries creation

codefiles:


archinstall/lib/installer.py
```diff
@@ -734,61 +734,72 @@ def _add_systemd_bootloader(self, root_partition: disk.PartitionModification):
...
+				)
+
+		# blkid doesn't trigger on loopback devices really well,
+		# so we'll use the old manual method until we get that sorted out.
+
+		options_entry = f'rw rootfstype={root_partition.safe_fs_type.fs_type_mount} {" ".join(self._kernel_params)}'
+
+		for sub_vol in root_partition.btrfs_subvols:
+			if sub_vol.is_root():
+				options_entry = f"rootflags=subvol={sub_vol.name} " + options_entry
```
Torxed, line 763: Very minor question, but would it make sense to add a `break` here?

Torxed:
Looks good to me, just the small question of a `break`.
I don't see how two root subvolumes could ever co-exist, but would logically feel good if the break was there hehe.

codefiles:
I think that is out of the scope of this pull request since the intent here is to not change any of the logic.

archinstall/lib/installer.py
codefiles: I think so.

Torxed:
Totally fine! Great work, keep it up!

PR #1908, opened 2023-06-30
global_menu: wrap invalid config error message with the translation shim

Andy-Python-Programmer:
cc https://github.com/archlinux/archinstall/pull/1815#issuecomment-1614276956

### DIFF
--- a/archinstall/lib/global_menu.py
+++ b/archinstall/lib/global_menu.py
@@ -363,7 +363,7 @@ class GlobalMenu(AbstractMenu):
 			return text[:-1]  # remove last new line
 
 		if error := self._validate_bootloader():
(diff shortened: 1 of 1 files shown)


Torxed:
@svartkanin is the translation clever enough to deal with f-strings? I believe the translation has to be done on the `{error}` strings themselves right?

Meaning these: https://github.com/archlinux/archinstall/blob/d89983402c8c816a5a7ddb1c1d726c5e8a652b32/archinstall/lib/global_menu.py#L351

svartkanin:
Yes it is but we used to use the old fashioned way

```
"example: {}".format(str(_("Some error description")))
```

PR #1918, opened 2023-07-04
 Move preview_text method to XorgProfile instead of subclasses implementing it

lavafroth:
- This does not fix an issue 

## PR Description:

The same as described in #1912

## Tests and Checks


Torxed:
Nice and clean!

PR #1920, opened 2023-07-06
Fix 1916

### LINKED ISSUE #1916: Log messages interfere with menus
If a call to a log function is made while a menu is being displayed the log message will be written on top of the menu.

This came to my attention with 748f03c where an instance of the bug became possible inside the menu created by `GlobalMenu._disk_encryption()`. 

Below are two screenshots; the first shows this menu displayed as expected and the second shows this menu when the libfido2 package is not installed.

image

image

The following `systemd-cryptenroll --fido2-device=list` command results in an exception being raised when the libfido2 package is not installed and so both of the error log messages get triggered.

https://github.com/archlinux/archinstall/blob/64fc585fa7624060d406b9fcb877baf968ff0f24/archinstall/lib/disk/fido.py#L39-L46

In this case the log messages are written to `stdout` since the log level is `logging.ERROR`.

https://github.com/archlinux/archinstall/blob/64fc585fa7624060d406b9fcb877baf968ff0f24/archinstall/lib/output.py#L324-L326

svartkanin:
@codefiles this should fix #1916 for both producing stdout while a menu is active as well as the particular case of HSM. 

When the menu is triggered it will set itself to active and when it's done it will reset it. Any output that happens during that period is only written to the log file. 

In addition, the HSM libfido2 problem is handled more gracefully now. I think it's a fair trade-off as it will show a visual indication of the problem in the installer. 
If anyone wants to use the function in their own custom script, then it should throw a exception to give a straight forward indication of what's wrong and we can probably mention in some docs that the lib is required for using the HSM capability.

### DIFF
--- a/archinstall/lib/disk/encryption_menu.py
+++ b/archinstall/lib/disk/encryption_menu.py
@@ -71,6 +71,7 @@ class DiskEncryptionMenu(AbstractSubMenu):
 				description=_('Use HSM to unlock encrypted drive'),
 				func=lambda preset: select_hsm(preset),
 				display_func=lambda x: self._display_hsm(x),
+				preview_func=self._prev_hsm,
 				dependencies=['encryption_password'],
 				default=self._preset.hsm_device,
 				enabled=True
@@ -93,8 +94,6 @@ class DiskEncryptionMenu(AbstractSubMenu):
 		if device:
 			return device.manufacturer
 
-		if not Fido2.get_fido2_devices():
-			return str(_('No HSM devices available'))
 		return None
 
 	def _prev_disk_layouts(self) -> Optional[str]:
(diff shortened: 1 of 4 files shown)


codefiles (changes requested):
This crashes with the following:

```
archinstall.lib.exceptions.SysCallError: ['/usr/bin/systemd-cryptenroll', '--fido2-device=list'] exited with abnormal exit code [1]: FIDO2 token support is not
installed.
```

The given suggestion fixes that. Below is a screenshot of the encryption menu with the suggestion applied that show the message produced when the libfido2 package is not installed.

image

Thanks friend.

Torxed:
Awesome colab! Merged it! Unless someone tags me in something during the day, I'll begin updating the language files and start the process of tagging a release no later than tomorrow.

PR #1921, opened 2023-07-09
Fix network settings loading from config file

svartkanin:
# Problem
I have stumbled across a but when saving blank configuration file and then loading it:
- Start archinstall 
- Go to save configuration 
- Will save the network configuration as 
```
    "nic": 'None',
```
- when loading the configuration it will cause an error

# Solution
- Rewrote the network configuration data class to align better with existing patterns which also simplifies the code quite a bit
- This creates a new configuration format such as 

```
    "network_config": {
        "type": "iso|nm"
    }
```

or for manual setup
```
    "network_config": {
        "nics": [
            {
                "dhcp": false,
                "dns": [
                    "3.3.3.3"
                ],
                "gateway": "2.2.2.2",
                "iface": "enp0s31f6",
                "ip": "1.1.1.1"
            }
        ],
        "type": "manual"
    },
```

### DIFF
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -225,10 +225,9 @@ def load_config():
 	if arguments.get('servers', None) is not None:
 		storage['_selected_servers'] = arguments.get('servers', None)
 
-	if arguments.get('nic', None) is not None:
-		handler = models.NetworkConfigurationHandler()
-		handler.parse_arguments(arguments.get('nic'))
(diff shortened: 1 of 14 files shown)


Torxed:
This is good, the phrasing is more clear as `nic` was a bit ambiguous.
We'll probably introduce networkmanager options later on too, so that way those options can live under there together with `type: nm`.

Issue #1930, opened 2023-07-16
[improve-boot-and-esp/master] "None" instead of null when configuration is saved


Torxed:
Started happening after running:
```
sh -c "cd archinstall-git; cp archinstall/scripts/guided.py ./; time python guided.py --conf https://hvornum.se/user_configuration.json --creds https://hvornum.se/user_credentials.json --silent";
```

![screenshot](https://github.com/archlinux/archinstall/assets/861439/ecff854a-942b-4611-b675-f8f0c079f67b)

It happens during saving of configuration, it looks like it's saving the translated string rather than the value.
This is after a run where the config was saved (but due to the branch being broken it didn't finish installation.. just to show that it wasn't me modifying it):
![screenshot](https://github.com/archlinux/archinstall/assets/861439/23a823c2-01cd-4a0f-89af-e3dca2d1b77f)

Torxed:
It's the `JSON` serializer:
```python
>>> json.dumps({"some-key": None}, cls=archinstall.JSON)
'{"some-key": "None"}'
```

PR #1871, opened 2023-06-14
Simplify object serialization before JSON encoding

lavafroth:
- This fix issue: #1860

## PR Description:

The `JsonEncoder` class with its static methods is replaced by a simple `jsonify` function. It recursively walks through a data structure or an instance to create nested dictionaries, excluding values whose keys begin with a bang (`!`) when the safe property is `True`. Unhashable types and types incompatible with `json.dumps` are avoided from being used as dictionary keys.

For non-dictionary objects, systematically check for methods like `json`, `__dump__`, `__dict__` before defaulting to returning the object as is.

The `JSON` and `UNSAFE_JSON` thunk classes simply use the `jsonify` function and encode the returned dictionary.

## Tests and Checks
- [x] I have tested the code!<br>

PR #1932, opened 2023-07-17
removed str() around objects to be serialized. As None should stay intact rather than become 'None'

Torxed:
- This fix issue: #1930 

## PR Description:

After #1871 the `else` clause causes a last effort attempt to cast everything to a string, causing `None` to become `'None'` etc.

I missed this when merging so I'm correcting it in this PR.

### DIFF
--- a/archinstall/lib/general.py
+++ b/archinstall/lib/general.py
@@ -79,7 +79,8 @@ def jsonify(obj: Any, safe: bool = True) -> Any:
 		return str(obj)
 	if hasattr(obj, "__dict__"):
 		return vars(obj)
-	return str(obj)
+
+	return obj
 
 class JSON(json.JSONEncoder, json.JSONDecoder):
 	"""


Torxed:
screenshot

This appears to do the trick, I haven't run through all the config scenarios so we'll have to double check that all variations of configurations can be saved safely *(if only we introduced those tests way back when)*.

lavafroth:
Apologies for introducing the bug. Would it be better to check `if obj is None` and return `None` since `str` can act as a sort of catchall?

Torxed:
I think I would rather the encoder crashed, than allow everything to sneak by.
If we're trying to pass something to the JSON encoder that doesn't have `.json()` or is within the normal types I think it should fail.

Otherwise we'll have to do a `if obj is <?>` for *all* python types, not only `None` but also `False`, `True`, etc.

Torxed:
Also no worries hehe, that's on me! :)

PR #1939, opened 2023-07-19
Uncomment `/etc/locale.gen` entry and use first column for `LANG` variable

### LINKED ISSUE #1937: Locale entry in `/etc/locale.gen` is appended rather than uncommented
This deviates from the ArchWiki _Installation guide_ instructions which states to uncomment.

https://wiki.archlinux.org/title/Installation_guide#Localization

https://github.com/archlinux/archinstall/blob/e41900a25cfbf35e550863a8415a0c7a29c9e9b2/archinstall/lib/installer.py#L389-L390

### LINKED ISSUE #1938: Set `LANG` variable not always consistent with ArchWiki instructions
Following the ArchWiki, the `LANG` variable is set to:

> ... the **first column** of an uncommented entry in `/etc/locale.gen`

https://wiki.archlinux.org/title/Locale#Setting_the_system_locale

Locale entries in `/etc/locale.gen` may or may not include the encoding in the first column. For example, the first column for the locale entry `en_US.UTF-8 UTF-8` contains the encoding while `en_ZM UTF-8` does not. The implementation to set the `LANG` variable will deviate from the ArchWiki instructions when the encoding was not included in the first column of the entry since the encoding is always being included.

https://github.com/archlinux/archinstall/blob/e41900a25cfbf35e550863a8415a0c7a29c9e9b2/archinstall/lib/installer.py#L392

codefiles:
Fixes #1937, fixes #1938

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -366,7 +366,7 @@ class Installer:
 		with open(f'{self.target}/etc/hostname', 'w') as fh:
 			fh.write(hostname + '\n')
 
-	def set_locale(self, locale_config: LocaleConfiguration):
+	def set_locale(self, locale_config: LocaleConfiguration) -> bool:
 		modifier = ''
 		lang = locale_config.sys_lang
 		encoding = locale_config.sys_enc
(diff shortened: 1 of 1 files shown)


Torxed:
Quick work, thank you!

PR #1942, opened 2023-07-21
Fix boot partition regression

### LINKED ISSUE #1941: Boot partition regression
The change to the `DeviceModification.get_boot_partition()` function in afaf42e means that it will return `None` when an ESP is used and an XBOOTLDR is not present. Before this change in that given scenario the function would return the ESP. Anywhere the boot partition is accessed when there is only an ESP is going to cause an issue.

The code below will prevent proceeding with the installation by selecting _Install_.

https://github.com/archlinux/archinstall/blob/69c37e7c79b8756e300438fd2d25f6d40db5e04b/archinstall/lib/global_menu.py#L350-L351

The screenshot below demonstrates that this occurs with the _Disk configuration_ option _Use a best-effort default partition layout_.

image

The code below will cause a crash and prevent the boot loader from being installed if providing a configuration with the described setup and then using `--silent`.

https://github.com/archlinux/archinstall/blob/69c37e7c79b8756e300438fd2d25f6d40db5e04b/archinstall/lib/installer.py#L1101-L1102

```
ValueError: Could not detect boot at mountpoint /mnt/archinstall
```

codefiles:
Fixes #1941

### DIFF
--- a/archinstall/lib/disk/device_model.py
+++ b/archinstall/lib/disk/device_model.py
@@ -789,8 +789,8 @@ class DeviceModification:
 		"""
 		Similar to get_boot_partition() but excludes XBOOTLDR partitions from it's candidates.
 		"""
-		fliltered = filter(lambda x: x.is_boot() and x.fs_type == FilesystemType.Fat32 and PartitionFlag.XBOOTLDR not in x.flags, self.partitions)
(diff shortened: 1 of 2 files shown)


Torxed:
Great fix! I think we're getting close to tagging the release!

PR #1943, opened 2023-07-22
Update Simplified Chinese translation

yetist:
## PR Description:

Update Simplified Chinese translation

### DIFF
--- a/archinstall/locales/zh-CN/LC_MESSAGES/base.po
+++ b/archinstall/locales/zh-CN/LC_MESSAGES/base.po
@@ -3,13 +3,13 @@ msgstr ""
 "Project-Id-Version: \n"
 "POT-Creation-Date: \n"
 "PO-Revision-Date: \n"
-"Last-Translator: wc7086 <190022474@qq.com>\n"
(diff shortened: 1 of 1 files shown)


.github/workflows/translation-check.yaml
```diff
@@ -9,14 +9,18 @@ name: translation file checks
 jobs:
     translation-check:
         runs-on: ubuntu-latest
-        container:
-            image: archlinux:latest
```
Torxed, line 13: I don't think this is desirable.
I get that it might look strange that we're running arch in ubuntu, but it's nice to test and run all code related things on a arch machine - to catch breaking changes in the OS.

yetist: I totally understand what you're thinking, but I found it impossible to run `git diff --name-only` in a docker environment

yetist: I droped my changes of `translation-check.yaml`, this PR only updates the po file.

Torxed: The main issue regarding the check appers to be the path. I'll fix it in another PR :)

Torxed:
Thank you for this update!

PR #1944, opened 2023-07-22
Let the Chinese translation show in the menu item

yetist:
## PR Description:

Even though the translation files exist, we still can't find Simplified or Traditional Chinese translations from the language menu, this patch fixes that.


Torxed:
Thank you for this PR, it will be part of the 2.6 release!

PR #1945, opened 2023-07-22
Improve the Chinese, Japanese and Korean text menu layout

yetist:
## PR Description:

Improve the Chinese, Japanese and Korean text menu layout

Before this patch, menus in Korean language would not be aligned:

After apply this patch, menus in Korean language are aligned:

### DIFF
--- a/archinstall/lib/menu/abstract_menu.py
+++ b/archinstall/lib/menu/abstract_menu.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+import unicodedata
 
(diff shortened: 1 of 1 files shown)


yetist:
Simplified Chinese

Before:


After:

Torxed:
Thank you for putting this work in, this looks way better.
Let me just run through this code a couple of installs and I'll merge it.

Torxed:
Thanks for your efforts in this area, it really helps!

PR #1946, opened 2023-07-23
Fix 1934 - audio server regression

### LINKED ISSUE #1934: Audio option regression
The Audio option is inaccessible without selecting a profile first.

svartkanin:
This fixes #1934 

- Make audio server always selectable regardless of the profile selection by decoupling the setting completely from the profile
- Introduce a new `AudioConfiguration` wrapper that makes it easier to extend in the future and to facilitate a single installation logic 

New audio configuration setting `audio_config` 
```
"audio_config": { "audio": "pulseaudio"}
```


Torxed:
Awesome work!

Issue #1955, opened 2023-07-29
[master] xorg profile missing xorg packages


Torxed:
![screenshot](https://github.com/archlinux/archinstall/assets/861439/cc0d62bf-6f28-4c3a-9ecd-2ebebea7d10e)

It installs, but won't run.
Need to check over all packages.

PR #1956, opened 2023-07-29
Adding pack packages to profiles after they were removed

Torxed:
This will fix: #1955 

Marking it as WIP as a collection patch for all profiles

 * Added back xinit for awesome, since it can be used without a greeter, as well as other useful tools we've had in previous releases


Torxed:
hyprland appears a bit wonky, it kicks me back to the greeter.
No idea what needs to be done with it so moving on.

rec

Torxed:
Qtile also can't get past greeter, but I also know that might need extra config.

screenshot

Torxed:
Other than those two profiles, this should be all for profiles.

PR #1961, opened 2023-07-30
Show translated languages in menu items

yetist:
## PR Description:

Show translated languages in menu items whenever possible, they are defined in `archinstall/locales/linguals.json` file.


svartkanin:
Can we please not do this, there was a reason we have removed this before. It will not show the languages whose fonts are not supported by default.

cc @Torxed

yetist:
ok

PR #1972, opened 2023-08-01
Fix and update Portuguese translation

hugok79:
* Portuguese translation completed
* Fix lang code


Torxed:
Cheers for the update!

PR #1981, opened 2023-08-03
update Simplified Chinese mo file

yetist:
## PR Description:

This commit 355a0dbe only updated the po file, and forgot to update the mo file.


svartkanin:
@yetist thanks, we are currently working on a solution not to commit the blobs anymore as their causing too many issues and conflicts. We will then generate the translations when a new release gets out
cc @Torxed

Issue #1974, opened 2023-08-01
btrfs manual partitioning 'could not set compress attribute at @home'


VRof:
archinstall 2.6.0 from the newest arch iso

[my config file](https://0x0.st/H2Wn.json)

![Screenshot_1](https://github.com/archlinux/archinstall/assets/101021359/29deae7f-0b9d-4d1b-ad3d-bbff4126ffd3)

lavafroth:
This appears related to the `nodatacow` attribute set to true for `@home`. The [BTRFS docs](https://btrfs.readthedocs.io/en/latest/btrfs-man5.html#btrfs-specific-mount-options) states,
> Nodatacow implies nodatasum, and disables compression

In the configuration file you have shared, the `nodatacow` attribute conflicts with the `compress=zstd` mount option. I think the installer should check the `nodatacow` flag and only enable compression when it is false.

lavafroth:
I have been able to reproduce this issue, I think the script is supposed to have `compress` and `nodatacow` for a `btrfs` subvolume as mutually exclusive. For the time being, you would have to choose either of the flags but not both during manual partitioning.

svartkanin:
The fix was merged

PR #1998, opened 2023-08-09
fix: make nodatacow and compress mutually exclusive

lavafroth:
- This fix issue: #1974 

## PR Description:

This PR removes the ability to select both the `nodatacow` and `compress` flags for a BTRFS subvolume.  

## Tests and Checks


Torxed (approved):

lavafroth:
I synced this branch with master. Should be ready for merging 😄.

svartkanin:
@lavafroth thanks, this should fix the selection in the menu. However, there is one problem left that one could load a config with both enabled, could we fix that as well https://github.com/svartkanin/archinstall/blob/31bd5e80e3a1fa6c99159415981b33d43461ee24/archinstall/lib/disk/device_model.py#L419-L419. 

We could prefer one over the other when both are specified

lavafroth:
How about we error out if both of them are set to true?

Torxed:
That might be better, so they don't think it silently worked.
But I foresee a storm of "oh, why didn't you just add a menu option asking what we want" if we error out.

But tbh, it is a faulty configuration IMO so.. I'm good with it erroring out :)

lavafroth:
I'm making the code error out for now. We could move the compress / nodatacow / unspecified options under the btrfs compression settings. I'll make a future PR if this idea sounds favorable.

svartkanin:
@lavafroth according to the docs 
> Most mount options apply to the whole filesystem and only options in the first mounted subvolume will take effect. This is due to lack of implementation and may change in the future. This means that (for example) you can’t set per-subvolume nodatacow, nodatasum, or compress using mount options. This should eventually be fixed, but it has proved to be difficult to implement correctly within the Linux VFS framework.

So I think it'd make sense to have these options as a global Btrfs option rather than a subvolume one. To fix the issue at hand I'll merge the PR for now and then we can refactor the code to ask the question on a btrfs creation, which may need some thinking on how to accomplish that on the manual partitioning

Issue #1111, opened 2022-05-01
insecure password storage


jamincollins:
It would appear that all user passwords are stored in plaintext in file `/var/log/archinstall/user_credentials.json`.  Which is accessible to any user on the system (mode=644).

```
# ls -l /var/log/archinstall/user_credentials.json 
-rw-r--r-- 1 root root 94 May  1 18:48 /var/log/archinstall/user_credentials.json
```

At minimum suggest changing the access permissions on this file to mode=600, user=root, group=root.

Better yet, store the password hash.  I see that there is another issue (#1029) for allowing passwords to provided in a hashed manner.  And in this issue there is some concern over supporting multiple potentially disparate hashing requirements.  Given that the entry is already stored as a JSON object (python dict), one solution would be to extend the object to indicate which subsystem the entry is for, such as:

```json
{
    "!superusers": {
        "wilbur": {
            "!password-shadow": "$6$xyz$ZmArb33D/r5Aftt/vkeJD.eOocWEgln8gW2Jn/71EUup1LDTJv0P8nL76leql7HDvsy2ObrAT6k.jQnol2pfV0"
        }
    }
}
```
or
```json
{
    "!superusers": {
        "wilbur": {
            "!password": {
                "shadow": "$6$xyz$ZmArb33D/r5Aftt/vkeJD.eOocWEgln8gW2Jn/71EUup1LDTJv0P8nL76leql7HDvsy2ObrAT6k.jQnol2pfV0"
            }
        }
    }
}
```
Personally, I prefer the second form.  IMO, it is easily extensible for any additional passwords that need to be convey.

Torxed:
I do not disagree that we could improve on this. And i'm not saying hashes change frequently. But storing a hash would at some point become equally insecure when it's deprecated. And if someone gets a hold of the hash it's a small bump in the road for a local attack.

I think the chmod is a solid approach and increases security significantly for me to consider this done for now.

I'll submit a PR with it together with the /tmp password and issue surrounding passwords being stored in disk_layout.json

Torxed:
I think I've figured out a clean way to approach this.
As storing hashes presents the issue of not being able to rely on toolings like `passwd` for setting the password.
Obviously not a huge obstacle but we'll need to transition to setting it with opening `/etc/shadow` manually.

Something in the lines of:
```python
salt = base64.b64encode(os.urandom(12))
password = base64.b64encode(hashlib.pbkdf2_hmac('sha512', b'password', salt, 200000, 64))

[...]

PR #2011, opened 2023-08-14
Replace encryption password

abulvenz:
- This fixes issue: #1111 

## PR Description:
The encryption password should not be copied to the target system log. Passing it to `Journald` in plain text should be OK on the installation system.

This is achieved by replacing the password when writing the log file.

## Tests and Checks

### DIFF
--- a/archinstall/lib/output.py
+++ b/archinstall/lib/output.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import sys
+import re
 from enum import Enum
 
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


svartkanin:
@abulvenz the log function is a generic function to log text. Adding a custom filter for strings in there isn't the right way to go, but we should rather track down which caller of this function is actually trying to log the encryption password and prevent it there

svartkanin:
In addition, this doesn't really solve the issue linked as the problem there is the credentials configuration file NOT any logging of passwords, which we're not actually doing

PR #2012, opened 2023-08-15
Support CJK text alignment

yetist:
## PR Description:

Support CJK text alignment

### DIFF
--- a/archinstall/lib/menu/abstract_menu.py
+++ b/archinstall/lib/menu/abstract_menu.py
@@ -1,31 +1,14 @@
 from __future__ import annotations
-import unicodedata
 
 from typing import Callable, Any, List, Iterator, Tuple, Optional, Dict, TYPE_CHECKING
 
 from .menu import Menu, MenuSelectionType
 from ..output import error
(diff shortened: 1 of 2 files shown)


yetist:
Before:
image

After:
image

Torxed (approved):

Torxed:
Well done on all of this!

svartkanin:
Should we maybe rename the `clk` to something more meaningful? If this turns out to affect other languages in the future then the name is not inclusive and both specific at the same time. 
Would `unicode` make more sense here as that's what it's doing?

yetist:
> Should we maybe rename the `clk` to something more meaningful? If this turns out to affect other languages in the future then the name is not inclusive and both specific at the same time. Would `unicode` make more sense here as that's what it's doing?

How about using the following names?
`_count_cjk_chars()` -> `_count_wchars()`
`cjk_ljust()` -> `unicode_ljust()`
`cjk_rjust()` -> `unicode_rjust()`

svartkanin:
Yes that'd be great

Torxed (approved):

Issue #2001, opened 2023-08-10
EFI system partition can't be mounted at /efi


aki-k:
If I create a partition setup as follows:

partition 1: esp fat32 mounted at /efi
partition 2: btrfs with three subvs:
arch_boot mounted at /boot
arch_root mounted at /
arch_home mounted at /home

Then select grub as boot loader; the installation starts, but fails spectacularly when it's installing grub. If I change the esp mount point to /boot, the installation succeeds.

https://wiki.archlinux.org/title/EFI_system_partition

4.1 Typical mount points

- mount the ESP to /boot

- mount the ESP to /efi

Weissnix4711:
[This line in installer.py](https://github.com/archlinux/archinstall/blob/master/archinstall/lib/installer.py#L908) is the culprit.

On efi systems with a different ESP mountpoint (other than `/boot`), it will attempt to use the `boot_partition` mountpoint (`/efi`) as the prefix directory.

roboboredom:
Same issue. Commenting for bump.

PR #2015, opened 2023-08-16
Fix GRUB with non-/boot ESP

Weissnix4711:
- This fix issue: #2001

## PR Description:

Changes `get_efi_partition()` logic to include partitions with ESP flag. Also had to change `get_boot_partition()` to prevent it returning the ESP even if that partition does not have the boot flag. Not that there currently is a way to set the ESP flag from the guided menu. But I hope that'll be added eventually.

Grub can now use a custom boot directory with UEFI, though this is untested and probably should never be done. Will actually use the correct ESP mount path now, and will make the config to the correct location.

All untested. There's probably better ways to do this too, but I don't have time to faff about with this, so if someone could take over from here, please do.

## Tests and Checks


Torxed:
Nicely done, and clean fix.
I suspect this might collide and cause some merge conflicts with your overhauls @svartkanin, just a heads up!

svartkanin:
@Weissnix4711 apologies for the delay, unfortunately the PR conflicts with some other changes that got merged. Would you mind resolving the conflicts then I think it's good to go

Weissnix4711:
Should be good now. :crossed_fingers:

Torxed (approved):

codefiles:
> Should be good now. 🤞

It is not; there are errors in mypy checks.

> archinstall/lib/installer.py:879: error: Item "None" of "Optional[PartitionModification]" has no attribute "dev_path"  [union-attr]
archinstall/lib/installer.py:884: error: Item "None" of "Optional[PartitionModification]" has no attribute "mountpoint"  [union-attr]
archinstall/lib/installer.py:899: error: Item "None" of "Optional[PartitionModification]" has no attribute "mountpoint"  [union-attr]
archinstall/lib/installer.py:901: error: Item "None" of "Optional[PartitionModification]" has no attribute "dev_path"  [union-attr]
archinstall/lib/installer.py:903: error: Item "None" of "Optional[PartitionModification]" has no attribute "safe_dev_path"  [union-attr]
archinstall/lib/installer.py:915: error: Item "None" of "Optional[PartitionModification]" has no attribute "dev_path"  [union-attr]
...
Found 6 errors in 1 file (checked 108 source files)
Error: Process completed with exit code 1.

You have the function parameters `boot_partition` and `efi_partition` typed as `Optional[disk.PartitionModification]` which means they could be `None`. In the case of `None` there would be no attributes so that needs to be checked for before attempting to accessing them.

Torxed:
@Weissnix4711 apologies for the delay in merging this.
Would you mind fixing the remaining mypy checks as well as merging in the latest changes from master?

That way we can get this in for the next release due in a couple of days :)

Torxed:
I merged the latest changes from master.
Once concern I have is the dismissal of the boot partition.

If `boot_partition` is `None`, `--boot-directory=` is skipped, leaving `--efi-directory` alone.
I would if `efi_partition` is empty, I'd expect the `--bot-directory` to be alone.

So I'm wondering if this makes more sense?
```python
f'--efi-directory={f"{efi_partition.mountpoint} " if efi_partition else f"{boot_partition.mountpoint}/efi"}'
f'--boot-directory={boot_partition.mountpoint} '
```

As I think `/boot` is required elsewhere, and `/efi` is an optional enhancement?

archinstall/lib/disk/device_model.py
```diff
@@ -800,20 +800,25 @@ def add_partition(self, partition: PartitionModification):
...
 		Returns the first partition marked as XBOOTLDR (PARTTYPE id of bc13c2ff-...) or Boot and has a mountpoint.
 		Only returns XBOOTLDR if separate EFI is detected using self.get_efi_partition()
+		Will return None if no suitable partition is found.
 		"""
 		if efi_partition := self.get_efi_partition():
 			filtered = filter(lambda x: x.is_boot() and x != efi_partition and x.mountpoint, self.partitions)
 			if boot_partition := next(filtered, None):
 				return boot_partition
-			return efi_partition
+			if efi_partition.is_boot():
```
codefiles, line 818: Why is this being checked when it was checked in `get_efi_partition()`?

Weissnix4711:
> If boot_partition is None, --boot-directory= is skipped

Yup. In this case grub-install defaults to `/boot`. This is presumably what we want if there is no boot partition (ie. `/boot` will be on root), no?

Not sure if defaulting to `{boot_partition.mountpoint}/efi` is a good idea when efi_partition is None. This might likely result in a situation where the firmware is unable to find the EFI binaries. Certainly it makes settings `--removable` redundant.

Weissnix4711:
The efi dir should really only be on some subdir of a suitable FAT32 partition. If efi_partition is None, that presumably means we found no suitable partition, so should error out or install only as legacy (i386-pc). At least that's my thinking.

It is possible the system may support other filesystems beyond the scope of the UEFI spec, and so the user may wish to proceed with setting the efi directory to something 'weird', but I don't see an easy way of achieving this customisability. Also, the user will realistically run through the manual install process anyway.

Keeping `--removable` would certainly be useful. In most 'normal' installations, where the efi dir is actually set to the mountpoint of the ESP, this removes any chance of efibootmgr messing up the boot entry and the firmware should just pick up the efi binaries regardless.

I will have it set boot_directory explicitly.

Weissnix4711:
As for typing fixes, I'm not 100% sure how to proceed. I'm not a python dev :)

Logically, at least one of boot or efi dir must be passed to the `_add_grub_bootloader` function. How this is done using python's type hints, I don't know.

I was originally hoping someone would just take over this PR for me lol

Torxed:
Hehe, I can fix the typing stuff in a couple of hours :)

As for all the other reasoning, I agree with it. I forgot this logic is specifically for EFI boot. And that we're really only need to consider that scenario :D

svartkanin:
I pushed the fixes for mypy and flake8 should be good now

Torxed:
Awesome, I'm on my phone so hard to check. But could we add an exception if both boot and efi is none? :)

svartkanin:
@Torxed I don't think the boot partition can be none actually, we're doing a validation here https://github.com/Weissnix4711/archinstall/blob/e87bf500c30eff5277be8cb2373b045e6a0d3888/archinstall/lib/installer.py#L1128-L1128 before calling the grub bootloader installation so that wouldn't pass

Torxed:
Fair enough, works for now. It's best to put a validation on the function performing the task tho since people might get creative and call the boot loader functions directly hehe.
But for this release this is fine :D

svartkanin:
Yeah that's fair, however, the method header does specify the boot partition argument as a object type not an optional object. So it would be a user error if someone calls it with none :)

HanM23:
Hi guys,

Could you please tell me if this has been tested with luks (luks2) encryption e.g
- grub as bootloader
- /efi fat32 partition
- encrypted btrfs filesystem with recommanded subvolume layout ?

Thanks

PR #2018, opened 2023-08-20
French language translation update

roxfr:
Hello,
Here is the update of the translation for the French language.
Regards,
Roxfr


roxfr:
I'm going to make another pass.

Issue #2025, opened 2023-08-25
`custom-commands` fails with no such file or directory


stephenhouser:
When using `-config` and specifying `custom-commands` a `no such file or directory` error is generated as the path to the script is built incorrectly in `lib/general.py`. It appears 2 months ago a refactor was done when this was introduced. The previous code looks fine. The errors introduced include:

- `chroot_path = installation.target / script_path` << not sure how to interpret this line
- `chroot_path.write_text(command)` << there is no such method on strings

svartkanin:
Closing as the fix was merged

PR #2026, opened 2023-08-25
fix custom-commands issue #2025

stephenhouser:
- This fix issue: #2025 

## PR Description:

Fixes broken refacotring in `lib/general.py` function `run_custom_user_commands` that caused custom-commands to fail.

## Tests and Checks
  
  I did manual testing but do not see a link (yet) for downloading an ISO to test with.

### DIFF
--- a/archinstall/lib/general.py
+++ b/archinstall/lib/general.py
@@ -460,10 +460,12 @@ def _pid_exists(pid: int) -> bool:
 def run_custom_user_commands(commands :List[str], installation :Installer) -> None:
 	for index, command in enumerate(commands):
 		script_path = f"/var/tmp/user-command.{index}.sh"
-		chroot_path = installation.target / script_path
+		chroot_path = f"{installation.target}/{script_path}"
 		
 		info(f'Executing custom command "{command}" ...')
-		chroot_path.write_text(command)
+		with open(chroot_path, "w") as user_script:
+			user_script.write(command)
+			
 		SysCommand(f"arch-chroot {installation.target} bash {script_path}")
 		
 		os.unlink(chroot_path)


Nicceboy:
I can confirm that the fix works. At least when used from `archinstall.run_custom_user_commands(custom_commands, installation)` on custom script.

Torxed (approved):

svartkanin:
Hmm I'm also a bit confused about this, the original code "should" have worked as we're using the `/` on a `Path`...
I wonder if the `installation.target` is a string for some reason actually. 
It would be good if the issue had a config file included as well

The change doesn't break anything so I reckon we can just go ahead with it

Torxed:
I wonder if it's a matter of different behavior on loaded conf vs menu walkthrough? That have in the past caused these kinds of issues as the conf never gets covered from str in the JSON file. But that shouldn't be the case if we use a dataclass-model with types

svartkanin:
Could be the case, the target isn't loaded from the config though it's hardcoded or from the settings which makes this a bit mysterious and I'd have liked to dig deeper but without config it's a bit tricky

Nicceboy:
For example this reproduced the issue:

```python
from pathlib import Path
from typing import TYPE_CHECKING, Any, Optional

import archinstall
from archinstall import Installer
from archinstall import profile
from archinstall.default_profiles.minimal import MinimalProfile
from archinstall import disk
from archinstall import models
from archinstall import locale
from archinstall import profile

fs_type = disk.FilesystemType("ext4")
# Virtual drive from Packer
device_path = Path("/dev/vda")

# get the physical disk device
device = disk.device_handler.get_device(device_path)

if not device:
    raise ValueError("No device found for given path")

# create a new modification for the specific device
device_modification = disk.DeviceModification(device, wipe=True)

# create a new boot partition
boot_partition = disk.PartitionModification(
    status=disk.ModificationStatus.Create,
    type=disk.PartitionType.Primary,
    start=disk.Size(1, disk.Unit.MiB),
    length=disk.Size(512, disk.Unit.MiB),
    mountpoint=Path("/boot"),
    fs_type=disk.FilesystemType.Fat32,
    flags=[disk.PartitionFlag.Boot],
)
device_modification.add_partition(boot_partition)

# create a root partition
root_partition = disk.PartitionModification(
    status=disk.ModificationStatus.Create,
    type=disk.PartitionType.Primary,
    start=disk.Size(513, disk.Unit.MiB),
    length=disk.Size(30, disk.Unit.GB),
    mountpoint=Path("/"),
    fs_type=fs_type,
    mount_options=[],
)
device_modification.add_partition(root_partition)


disk_config = disk.DiskLayoutConfiguration(
    config_type=disk.DiskLayoutType.Default, device_modifications=[device_modification]
)

# initiate file handler with the disk config and the optional disk encryption config
fs_handler = disk.FilesystemHandler(disk_config)

# perform all file operations
# WARNING: this will potentially format the filesystem and delete all data
fs_handler.perform_filesystem_operations(show_countdown=False)

mountpoint = Path("/mnt/archinstall")

testing = []
multilib = []

locale_config = locale.LocaleConfiguration("us", "en_US", "UTF-8")

with Installer(
    mountpoint, disk_config, disk_encryption=None, kernels=["linux"]
) as installation:
    installation.sanity_check()
    installation.mount_ordered_layout()
    installation.minimal_installation(
        testing=None,
        multilib=None,
        hostname="archlinux",
        locale_config=locale_config,
    )
    installation.add_bootloader(models.bootloader.Bootloader.Systemd)
    # ntp:true
    installation.activate_time_syncronization()
    installation.set_timezone("Europe/Helsinki")
    installation.minimal_installation(hostname="minimal-arch")
    installation.add_additional_packages(["nano", "wget", "git"])

    # audio
    audio_config = {"audio": "pipewire"}
    audio_config = models.AudioConfiguration.parse_arg(audio_config)
    audio_config.install_audio_config(installation)
    # Gnome desktop
    profile_config = {
        "gfx_driver": "",
        "greeter": "gdm",
        "profile": {"details": ["Gnome"], "main": "Desktop"},
    }
    profile_config = profile.ProfileConfiguration.parse_arg(profile_config)
    profile.profile_handler.install_profile_config(installation, profile_config)
    installation.set_keyboard_language(locale_config.kb_layout)
    core_packages = [
        "vim",
        "curl",
        "wget",
        "jq",
        "wezterm",
        "spice-vdagent",
        "docker",
        "git",
        "base-devel",
        "firefox",
        "grml-zsh-config",
        "mesa",
    ]
    installation.add_additional_packages(core_packages)
    services = ["docker"]
    installation.enable_service(services)

    # arch:arch sudo
    user = models.User("arch", "arch", True)
    installation.create_users(user)
    custom_commands = [
        "usermod -aG docker arch",
        "curl https://blackarch.org/strap.sh | sh",
    ]
    # Need to wait for merge: https://github.com/archlinux/archinstall/pull/2026
    # archinstall.run_custom_user_commands(custom_commands, installation)
    installation.genfstab()

```

It also happened with following JSON part:
```JSON
    "custom-commands": [
        "usermod -aG docker arch",
        "curl https://blackarch.org/strap.sh | sh"
    ]
```

PR #2030, opened 2023-09-03
Fix 1973 - pipewire

### LINKED ISSUE #1973: Install fails with "Pipewire-jack and Jack2 are in conflict"
With Archinstall 2.6, when installing pipewire, the installer always fails on the conflict between jack2 and pipewire-jack, with no way to confirm the overwrite of jack2 with pipewire-jack. I'm having this issue with the Hyprland desktop profile, I've had it with both Virtualbox and on my T470s.

### LINKED ISSUE #1980: Wireplumber and pipewire-media-session are in conflict
Trying to install KDE Plasma with pipewire
First jack2 has a compatibility issue as is already mentioned.
When I fixed that, this time wireplumber incompatibility.

### LINKED ISSUE #1989: Installation fails: "error: unresolvable package conflicts detected"
Not caused by additional packages I'm installing. I only requested `kwin`, `conky` and `xcursor-vanilla-dmz`.

Bottom of `install.log` (attached) shows:
```

resolving dependencies...
looking for conflicting packages...
:: wireplumber and pipewire-media-session are in conflict. Remove pipewire-media-session? [y/N] 
error: unresolvable package conflicts detected
error: failed to prepare transaction (conflicting dependencies)
:: wireplumber and pipewire-media-session are in conflict
[?25h==> ERROR: Failed to install packages to new root
```

[install.log](https://github.com/archlinux/archinstall/files/12267924/install.log)

### LINKED ISSUE #1990: Work around for:  Installation fails: "error: unresolvable package conflicts detected" issue
I do not exactly know why this is happening, but you can work around this issue by selecting no audio server. No pulse audio, no pipewire. Install it after installation. Hope this helps.

svartkanin:
This fixes: 
- Fix #1973 
- Fix #1980 
- Fix #1989 
- Fix #1990 

The problem is caused by `pipewire-jack` vs `jack2` and `pipewire-media-session` vs `wireplumber`, those packages cannot be installed alongside each other and a choice needs to be made when one of them is already installed. 
`pacstrap` will prompt for a auto resolution 
```
pipewire-jack and jack2 are in conflict (jack). Remove jack2? [y/N]
```

but the current implementation of SysCommand doesn't allow for any interactive behavior and also `pacstrap` dosn't have any flag to auto approve. 

However, by simply changing the order of installation such that pipewire (if chosen as a menu option) will be installed first and any other packages that depend on it and/or bring `jack2` into the mix can make use of the already installed packages.  
Main pointpoints seem to be `Gnome` and `firefox`. 

This isn't a perfect solution as the installation process shouldn't depend on the order of packages installed but it'll do the trick for now. A possible rewrite of `SysCommand` to allow for interactive behavior is probably a better way forward.

### DIFF
--- a/archinstall/scripts/guided.py
+++ b/archinstall/scripts/guided.py
@@ -166,9 +166,6 @@ def perform_installation(mountpoint: Path):
 				archinstall.arguments.get('profile_config', None)
 			)
 
-		if archinstall.arguments.get('packages', None) and archinstall.arguments.get('packages', None)[0] != '':
-			installation.add_additional_packages(archinstall.arguments.get('packages', None))
-
 		if users := archinstall.arguments.get('!users', None):
 			installation.create_users(users)
 
@@ -177,6 +174,9 @@ def perform_installation(mountpoint: Path):
 			audio_config.install_audio_config(installation)
 		else:
 			info("No audio server will be installed")
+
+		if archinstall.arguments.get('packages', None) and archinstall.arguments.get('packages', None)[0] != '':
+			installation.add_additional_packages(archinstall.arguments.get('packages', None))
 
 		if profile_config := archinstall.arguments.get('profile_config', None):
 			profile_handler.install_profile_config(installation, profile_config)
--- a/archinstall/scripts/swiss.py
+++ b/archinstall/scripts/swiss.py
@@ -230,9 +230,6 @@ def perform_installation(mountpoint: Path, exec_mode: ExecutionMode):
 					archinstall.arguments.get('profile_config', None)
 				)
 
-			if archinstall.arguments.get('packages', None) and archinstall.arguments.get('packages', None)[0] != '':
-				installation.add_additional_packages(archinstall.arguments.get('packages', []))
-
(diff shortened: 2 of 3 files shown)


Torxed:
> the current implementation of SysCommand doesn't allow for any interactive

It does, but it's a bit raw. But here's an example:
```python
>>> import archinstall
>>> worker = archinstall.SysCommandWorker("/bin/bash -c 'read -p \"Enter name: \" username; echo \"$username\"'")
>>> worker.poll()
>>> worker
b'Enter name: '
>>> worker.write(b'anton\n', line_ending=False)
6
>>> worker.poll()
>>> worker
b'Enter name: anton\r\nanton\r\n'
```

`SysCommand` is just a wrapper around `SysCommandWorker` for quicker things such as doing `ls` or `useradd`, whereas `SysCommandWorker` can be used for interactive things :)

Torxed (approved):

svartkanin:
Ahhh neat didn't know that's already working! I may do some rework on rhis fix later but I wanna get it in now to unblock users

PR #2034, opened 2023-09-04
Add romanian translation to archinstall

andrewKode:
- Created a manual translation with Poedit, updated to latest.
- 
## Tests and Checks
Tested with live ISO


Torxed:
Thank you for this contribution!
@svartkanin note to self, really need to block those binary blobs from being commited heh.

Torxed (approved):

svartkanin:
@Torxed yep lets add them to the ivnore file but also then we need to not forget to generate them at release. Is there a script where it should be included?

PR #2037, opened 2023-09-05
fix #1994 | Default to unknown on partition types

### LINKED ISSUE #1977: Archinstall fails to start
I'm trying for a whole month to install Archlinux on my laptop using archinstall tool, but every attempt finishes as failure, I don't know how can Manjaro and EndeavourOS which are based on Arch succeed in creating a graphical tools that never fail on any installation while Archlinux which is supposed to be superior with its large skilled devs can't do it. Now with the latest archlinux-2023.08.01-x86_64.iso archinstall (latest release) refuses even to start

### LINKED ISSUE #1994: archinstall  cant start
IMG_20230807_224800
since I use bare metal install this. cannot export the error to text.

svartkanin:
This fixes:
- Fix #1994 
- Fix #1977 

Instead of erroring when reading the existing partition types, we'll default to a unknown. This shouldn't impact anything else as we're hardcoding the assignment of the Primary type everywhere.

### DIFF
--- a/archinstall/lib/disk/device_model.py
+++ b/archinstall/lib/disk/device_model.py
@@ -20,6 +20,7 @@ from ..exceptions import DiskError, SysCallError
 from ..general import SysCommand
 from ..output import debug, error
 from ..storage import storage
+from ..output import info
 
 if TYPE_CHECKING:
 	_: Any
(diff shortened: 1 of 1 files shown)


Torxed:
Yea this makes sense, as it's unknown to us in a sense until we support more types.

Torxed (approved):

PR #2038, opened 2023-09-05
Fix github action

svartkanin:
The deploy github action is currently broken due to 
```
 Post job cleanup.
/usr/bin/docker exec  ef93653e6a12f7f62a1b7a063100ca046efcad7a7e53578789110c77096d9df5 sh -c "cat /etc/*release | grep ^ID"
/__e/node16/bin/node: /usr/lib/libc.so.6: version `GLIBC_2.38' not found (required by /usr/lib/libstdc++.so.6)
```

this PR should fix that.

In addition, upgraded all `action/checkout` to v4. 

@Torxed we may wanna merge this in first before all the other fixes

### DIFF
--- a/.github/workflows/bandit.yaml
+++ b/.github/workflows/bandit.yaml
@@ -6,7 +6,7 @@ jobs:
         container:
             image: archlinux:latest
         steps:
-            - uses: actions/checkout@v3
+            - uses: actions/checkout@v4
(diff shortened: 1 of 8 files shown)


Torxed:
Agreed, otherwise we'll get a bunch of notifications on failures when we merge in heh.

Torxed (approved):

PR #2039, opened 2023-09-06
updated deps added renovate

akac97:
Updated deps added renovate bot

### DIFF
--- a/.github/workflows/bandit.yaml
+++ b/.github/workflows/bandit.yaml
@@ -6,7 +6,7 @@ jobs:
         container:
             image: archlinux:latest
         steps:
-            - uses: actions/checkout@v3
+            - uses: actions/checkout@v4
             - run: pacman --noconfirm -Syu bandit
             - name: Security checkup with Bandit
               run: bandit -r archinstall || exit 0
--- a/.github/workflows/flake8.yaml
+++ b/.github/workflows/flake8.yaml
@@ -6,7 +6,7 @@ jobs:
         container:
             image: archlinux:latest
         steps:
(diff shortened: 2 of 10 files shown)


Torxed:
Never played around with renovatebot, thoughts from other contributors?
In general I'm a bit hesitant to tools that poke around too much, but perhaps this is fine? breakage concerns if versions are bumped?

svartkanin:
@Torxed we use renovate bot at our company so I can vouch for it. It's quite neat, it will simply scan the dependency modules we have (which aren't a lot) and then raise a PR to update the dependency to the latest version. This means that we won't fall behind on dependencies management. 

Looking at their page though https://github.com/renovatebot/tutorial, it seems we need a bit more setup to get it running

svartkanin:
That should be visible quite easily as the builds should fail. 
From what I could see there needs to be an app installed on the repo level

Torxed:
Ah yea, it needs a runner that checks continuously I suppose.
That's not something I can setup I'm afraid.

svartkanin:
@Torxed so is this a no-go or do we have to reach out to someone for it?

Torxed:
I've asked the devops to look in to it :)

klausenbusk:
Please checkout https://github.com/archlinux/archinstall/pull/2086.

Torxed:
I assume this means devops is working hard on it :D

Torxed:
With #2086, do we need this PR too?
If so, can we plug in the latest changes in `master`?

svartkanin:
The actions have all been update to v4 already, so I think everything from this PR is already in master

PR #2057, opened 2023-09-16
Added a forced buffer flush on each worker stdin write

Torxed:
This will fix a *(I think)* unknown issue where secondary disk encryption keys are not set properly, and archinstall hangs indefinitely at: https://github.com/archlinux/archinstall/blob/06d46ac762f58d051b2dba1aea50a2e3314ab403/archinstall/lib/luks.py#L214-L221

## PR Description:

`os.fsync()` ensures that the input given to `SysCommandWorker` is flushed immediately after writing it, so that no clever buffer management holds the input for a while.

### DIFF
--- a/archinstall/lib/general.py
+++ b/archinstall/lib/general.py
@@ -217,6 +217,7 @@ class SysCommandWorker:
 
 		if self.child_fd:
 			return os.write(self.child_fd, data + (b'\n' if line_ending else b''))
+			os.fsync(self.child_fd)
 
 		return 0
 


svartkanin:
Very sneaky bug

Torxed:
Yea it only triggers every now and then, I think it depends on the input length. If it's too short or not right the worker never gets the input, it's only shown in `stdout` for the worker (so if we look at it with peek_output it looks like it was written).. I think..
It behaves weirdly but this appears to fix it :)

I thought I'd add you as a reviewer and you could sign off on it to allow me to merge it @svartkanin

svartkanin:
It's a safe change so all good :)
It would be great if it wasn't a flaky thing and that the fix completely confirms it

svartkanin (approved):

Torxed:
Thanks for approving, the merge process is blocked otherwise (I can force-merge, but thought I'd test the workflow for myself too) ^^

codefiles:
You guys work really good together, I am jealous. Maybe one day I can be part of that and have my pull request merged too.

Torxed:
@codefiles we're trying a new workflow :) to eliminate me being a single point of failure and to avoid the backlog growing when I have a ton going on :)

It'll be a bit of catch-up now at the beginning :)

PR #2064, opened 2023-09-17
Add `_get_kernel_params()`

codefiles:
The kernel parameters used in `_add_systemd_bootloader()` and `_add_efistub_bootloader()` should be identical.

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -715,6 +715,44 @@ class Installer:
 				return root
 		return None
 
+	def _get_kernel_params(self, root_partition: disk.PartitionModification) -> List[str]:
+		kernel_parameters = []
+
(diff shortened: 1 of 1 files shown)


Torxed (approved):

svartkanin:
Looking at the efistub and systemd implementations they do have some differences 
* zram 
* subvols
* hsm

are we saying that those were originally missing from the efistub setup?

codefiles:
Yes, they **should** be identical and are not since those are missing. This is intended to fix that and keep the code from being duplicated.

svartkanin:
Gotcha, I see that grub also doesn't have them, are they missing their as well?

codefiles:
I plan to investigate application of this function to the other boot loaders. It may require additional changes to work with GRUB since by default `grub-mkconfig` determines the root filesystem.

PR #2069, opened 2023-09-19
Use `SUPPORTED` file for locale options and validation

### LINKED ISSUE #2041: locale language only en_US
archlinux-2023.09.01-x86_64.iso

codefiles:
Fixes #2041


Torxed:
Looks good, just add glibc to https://github.com/archlinux/archinstall/blob/master/PKGBUILD#L13 and I'll merge it to the [upstream package](https://gitlab.archlinux.org/archlinux/packaging/packages/archinstall/-/blob/main/PKGBUILD?ref_type=heads).

Torxed (approved):

codefiles:
Just a reminder that this has not been merged upstream. I also want to add that glibc should have already been in the dependency list of the PKGBUILD since `locale-gen`, which is from that package, is being used.

Torxed:
On it, thanks for the reminder :) Just got back from a trip, so will begin packaging (and updating upstream) soon.
And yea, there's a few missing direct deps in the PKGBUILD, I started making it more accurate by adding the direct commands we execute but haven't made the list perfect.
If we find any more, feel free to add them :)

PR #2076, opened 2023-09-20
Update instructions to run archinstall from ISO

svartkanin:
Small improvements to the instructions how to run a branch on a live ISO

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,8 +206,9 @@ If you want to test a commit, branch or bleeding edge release from the repositor
 you can replace the version of archinstall with a new version and run that with the steps described below.
 
 *Note: When booting from a live USB then the space on the ramdisk is limited and may not be sufficient to allow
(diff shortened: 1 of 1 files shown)


Torxed:
Perhaps we should mention `copytoram=y` which will use 75% of RAM to facilitate the tmpfs. Which should allow us to more reliably recommend upgrades and installs.

Docs: https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio-archiso/-/blob/master/docs/README.bootparams#L26

svartkanin:
Ahh nice I'll add that

codefiles:
This may also be relevant:
https://wiki.archlinux.org/title/Archiso#Adjusting_the_size_of_root_partition_on_the_fly

Torxed:
The later might be faster actually, that way the user don't need to copy the root content to the tmpfs initially.

svartkanin:
Added the on-the-fly instructions as well

PR #2079, opened 2023-09-21
Rework `_get_kernel_params()` for `_add_grub_bootloader()`

codefiles:
By default `grub-mkconfig` will determine the `root=` parameter (this appears to also detect and add the parameter for a default BTRFS subvolume when in use). The _id_root_ bool has been added to `_get_kernel_params()` to account for this and in the case that it is set to false, kernel parameters that are irrelevant in this condition will be omitted. Kernel parameters required for encryption (not including the `root=` parameter if _id_root_ is false) will be included in either case. The default is true, to add kernel parameters that identify the root. This could also be useful if support is added for systemd [GPT partition automounting](https://wiki.archlinux.org/title/Systemd#GPT_partition_automounting) where "if the root partition is automounted, then `root=` can be omitted from the kernel command line."

If the _partuuid_ bool is set to false `_get_kernel_params()` will use a UUID instead of a GPT partition UUID, which is the default.

The `_add_grub_bootloader()` function will now make use of kernel parameters for HSM and zram when necessary. The `_kernel_params` variable will also now be included.

Changed `cryptdevice=` parameter to end in _root_ following the colon instead of the previous _luksdev_ and _cryptlvm_, same with `root=/dev/mapper/` and those endings. This change is to conform with the ArchWiki guidance on [configuring the boot loader](https://wiki.archlinux.org/title/Dm-crypt/Encrypting_an_entire_system#Configuring_the_boot_loader) for _LUKS on a partition_.

Removed uncommenting of the line _#GRUB_ENABLE_CRYPTODISK=y_ in `/etc/default/grub` since archinstall does not support [encrypted /boot](https://wiki.archlinux.org/title/GRUB#Encrypted_/boot) as seen discussed here #1271.

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -881,9 +881,6 @@ class Installer:
 		grub_default = self.target / 'etc/default/grub'
 		config = grub_default.read_text()
 
-		if root_partition in self._disk_encryption.partitions:
-			config = re.sub(r'#(GRUB_ENABLE_CRYPTODISK=y\n)', r'\1', config, 1)
-
 		kernel_parameters = ' '.join(self._get_kernel_params(root_partition, False, False))
 		config = re.sub(r'(GRUB_CMDLINE_LINUX=")("\n)', rf'\1{kernel_parameters}\2', config, 1)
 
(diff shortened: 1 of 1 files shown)


Torxed (approved):

Torxed:
Excellent! This should make grub more robust!

PR #2080, opened 2023-09-21
Sort the profiles with ignore case

### LINKED ISSUE #2078: I3wm at the bottom of the list in the desktops
list is no longer alphabetical

image

Could you put i3 after hyprland?

svartkanin:
This fixes #2078 
Sorting the profiles alphabetically ignore the case


Torxed (approved):

Torxed:
Thanks for the quick turnaround, I think this is especially important for quicker usage with accessibility tools.

PR #2096, opened 2023-09-23
po: update Tamil translation

kbdharun:
## PR Description:

Update Tamil translations and add translations for untranslated strings.

## Tests and Checks


Torxed (approved):

Torxed:
@svartkanin not sure what's failing in the `local_generate` step, can't see any output of why it would fail. But approving as the PR appears good.

svartkanin:
Let me have a look, yeah the output is definitely not good, I'll update that

svartkanin:
@Torxed this should be clearer now 
image

kbdharun:
Seems like there are a few new strings, will create a PR for it soon.

PR #2102, opened 2023-09-24
Unify json functions

svartkanin:
Housekeeping PR to make the definitions for configurations less complex
- Remove `__dump__` function support
- Move all configuration relevant exports to `json()`
- Remove `json()` function support from the table formatter


Torxed (approved):
Just need to verify thoroughly that nothing gets lost in reading/writing configs. Doesn't look like anything would be affected, but let's keep an eye on it before releasing :)

svartkanin:
I'm happy to keep this back and only merge after release

Torxed:
Ah no go ahead, v2.6.1 is out so was more thinking to the next release, which likely isn't until next month :)

Issue #1837, opened 2023-05-24
No need to enable periodic TRIM for btrfs


kojq:
Btrfs async discard has been enabled since the 6.2 kernel. Consequently, I suggest not enabling the fstrim.timer (periodic TRIM) for btrfs.

lavafroth:
Thanks! This appears to be enabled by default. For everyone's reference, [here](https://elixir.bootlin.com/linux/latest/source/fs/btrfs/disk-io.c#L3688) is the source. Maybe someone can write a PR to change this behavior.

Torxed:
> Thanks! This appears to be enabled by default. For everyone's reference, [here](https://elixir.bootlin.com/linux/v6.4.11/source/fs/btrfs/disk-io.c#L3688) is the source. Maybe someone can write a PR to change this behavior.

We need to change it, only for BTRFS it seems.

Issue #897, opened 2022-01-24
`fstrim` does not trim encrypted partitions


rohandebsarkar:
PR #882 does provide support for `fstrim` to work for encrypted devices.

Encrypted LUKS devices would require `allow-discards` flags to be turned on for `fstrim` to work on them.

https://wiki.archlinux.org/title/Dm-crypt/Specialties#Discard/TRIM_support_for_solid_state_drives_(SSD)

PR #2108, opened 2023-09-25
fix: fstrim no longer needs to be manually enabled

lavafroth:
- This fix issue: #1837 

## PR Description:

This PR removes the code to manually enable the fstrim service since SSD optimizations are enabled by default in the kernel since version 6.2. Refer to the kernel source [here](https://elixir.bootlin.com/linux/v6.4/source/fs/btrfs/disk-io.c#L3688) which enables `discard=async`

This PR also marks #897 obsolete.

## Tests and Checks


lavafroth:
closing in favor of #2109

PR #2109, opened 2023-09-25
Disabled fstrim on BTRFS by default

### LINKED ISSUE #1837: No need to enable periodic TRIM for btrfs

### LINKED ISSUE #1841: i am encountering fstrim error
from everything i can read the testing repo causes this issue or a dirty drive and with a completely clean drive and not selecting testing i still hit this error

Torxed:
- Fixes #1837 
- Fixes #1841

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -587,7 +587,7 @@ class Installer:
 		hostname: str = 'archinstall',
 		locale_config: LocaleConfiguration = LocaleConfiguration.default()
 	):
-		_found_btrfs = False
+		_disable_fstrim = False
 		for mod in self._disk_config.device_modifications:
 			for part in mod.partitions:
 				if part.fs_type is not None:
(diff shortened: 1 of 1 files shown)


lavafroth (approved):
Looks great. I overlooked the case where the filesystem is not BTRFS. Good catch!

kojq (changes requested):
looks good; just heads up that this could go beyond `btrfs` as new filesystems such as `bcachefs`, `PuzzleFS`, and/or `SSDFS` could be implemented into the kernel later in some way, and some future filesystem(s) may not need periodic trim as well. You could consider changing the variable name to `_need_periodic_trim` or something else, and consider reversing the boolean values (this also remove the `not`).

I will approve otherwise, if desired.

Torxed:
Good idea, changed it to `_disable_fstrim` to be more generic. As we for most filesystems enable it, I thought of keeping the naming consistent with the if-logic and default value of being `False` for now.

kojq:
> Good idea, changed it to `_disable_fstrim` to be more generic. As we for most filesystems enable it, I thought of keeping the naming consistent with the if-logic and default value of being `False` for now.

Great. Also, is there a particular reason we want a variable, as I don't see anything technical like a scope issue with directly using `self.enable_periodic_trim()` when checking for btrfs?

Torxed:
Are you referring to why we don't do the check outside before calling enable_fstrim?
The only reason is to keep minimal_installation small and easy to read :)

kojq (approved):

PR #2113, opened 2023-09-25
Fix `mountpoint` for pre-mounted disk configuration

### LINKED ISSUE #2054: Pre-mounted disk configuration causing systemd and grub bootloaders errors and issues
partitions where mounted as following:

root - /mnt/archinstall
boot - /mnt/archinstall/boot

when ask root directory /mnt/archinstall

if choosing grub as bootloader, script will fail.. throwing some errors about boot

if systemd is chosen - script will install Arch linux. However, during boot, will have ERROR: device 'PARTUUID=None' not found. Solution is to edit the /boot/loader/entries/xxx.conf and add the UUID to make it permanent.

No issues when choosing use-a-best-effort default partition layout.

I have not tried the Manual Partitioning though

Id like to add that there are other partition on that testing... thats why I pre-mounted the partitions first

codefiles:
Fixes #2054

### DIFF
--- a/archinstall/lib/interactions/disk_conf.py
+++ b/archinstall/lib/interactions/disk_conf.py
@@ -139,7 +139,6 @@ def select_disk_config(
 
 				return disk.DiskLayoutConfiguration(
 					config_type=disk.DiskLayoutType.Pre_mount,
-					relative_mountpoint=path,
 					device_modifications=mods
 				)
 
--- a/archinstall/lib/disk/device_handler.py
+++ b/archinstall/lib/disk/device_handler.py
@@ -594,7 +594,9 @@ class DeviceHandler(object):
 					if is_subpath(mountpoint, base_mountpoint):
 						path = Path(part_info.disk.device.path)
 						part_mods.setdefault(path, [])
-						part_mods[path].append(PartitionModification.from_existing_partition(part_info))
+						part_mod = PartitionModification.from_existing_partition(part_info)
+						part_mod.mountpoint = mountpoint.root / mountpoint.relative_to(base_mountpoint)
+						part_mods[path].append(part_mod)
 						break
 
 		device_mods: List[DeviceModification] = []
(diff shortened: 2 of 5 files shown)


archinstall/lib/interactions/disk_conf.py
```diff
@@ -139,7 +139,6 @@ def select_disk_config(
 
 				return disk.DiskLayoutConfiguration(
 					config_type=disk.DiskLayoutType.Pre_mount,
-					relative_mountpoint=path,
```
svartkanin, line 142: I'm a bit confused, this is the path that was entered by the user as the pre-mount root path, how are now using the entered value?

codefiles: Just above that `path` is passed to `detect_pre_mounted_mods()` which requires a base mountpoint.

https://github.com/archlinux/archinstall/blob/b141609990fa4f7305443ee6ea6fe8796604c539/archinstall/lib/interactions/disk_conf.py#L137-L138

The _use a best-effort_ option by default will use `/mnt/archinstall` but the `DiskLayoutConfiguration` does not need `relative_mountpoint`. This lead me to think the `mountpoint` of each `PartitionModification` can be modified before the list of them goes to the `DiskLayoutConfiguration`.

https://github.com/archlinux/archinstall/blob/4274a9a5c02826b05a4c08f55646e23c3ea79c29/archinstall/lib/disk/device_handler.py#L597-L599

This fails when using pre-mounted BTRFS subvolumes. A slight change to the above code to account for subvolumes (not shown here but just a `mountpoint` conditional and a for loop over the subvolumes doing the same thing, updating the mountpoint to account for it being inside a chroot) should be enough. When I test the change I think is needed for subvolumes the root partition used by the boot loader has a PARTUUID that is None.

svartkanin:
I think it looks good, thanks for fixing this! 
And I'm also very happy to see that extra parameter vanish from the DiskConfiguration class

Torxed (approved):

Torxed:
Great work!

Any other things we want fixed before I drop `v2.6.2` tomorrow?

PR #2119, opened 2023-09-27
Fix 2118

svartkanin:
This will address the translation file issues (a bit).

- The script will now take a parameter of the language to be translated `bash locales_generate.sh se` this will only generate files for that language
- For @Torxed for a release there is the special param `bash locales_generate.sh all` which will generate all languages
- Diabled the translation check run as it doesn't give much value


Torxed (approved):
I like this one a lot :D Nicely done!

PR #2120, opened 2023-09-27
Extend the mypy checks

svartkanin:
This will add additional mypy validation


archinstall/lib/installer.py
```diff
@@ -237,7 +241,7 @@ def generate_key_files(self):
 			gen_enc_file = self._disk_encryption.should_generate_encryption_file(part_mod)
 
 			luks_handler = Luks2(
-				part_mod.dev_path,
```
Torxed, line 240: This is the only one I'm a bit worried about, but it's probably fine? `dev_path` can cause an exception, and that's maybe something we want to happen?

svartkanin: I looked at the code and at this point the partition must probably have a device path

svartkanin:
@Torxed if you're planning on a new release before the next ISO then lets wait with this not to break anything

archinstall/lib/installer.py
Torxed: In `guided.py` this is true, as the call comes after some sanity checks and stuff.
But if they were to copy this and run it super quick after the formatting step - I wonder if it might not properly report the right data?
We don't have any sanity checks in `Luks2` directly so this might get dicy?

Torxed:
> @Torxed if you're planning on a new release before the next ISO then lets wait with this not to break anything

Good judgement call! I'll release one today first, and then merge in :)

Torxed (approved):

PR #2121, opened 2023-09-27
Simplify SysCommand decoding

svartkanin:
This simplifies the return of SysCommand a bit
- `decode` will use utf-8 as a default arg
- `split` is always performed on decode but can be disabled via an arg


Torxed (approved):
I appreciate the improvements, we have to tripe-check all the logic using `SysCommand*` to make sure we don't hit any snags. But good work!

archinstall/lib/locale/locale.py
```diff
@@ -1,61 +0,0 @@
...
-
-def list_locales() -> List[str]:
-	locales = []
-
-	with open('/usr/share/i18n/SUPPORTED') as file:
-		for line in file:
-			if line != 'C.UTF-8 UTF-8\n':
-				locales.append(line.rstrip())
-
-	return locales
```
codefiles, line 21: This needs to be moved to `archinstall/lib/locale/utils.py`. This was merged recently to fix #2041, do not delete it.

svartkanin: Ahh my bad thanks for spotting this

archinstall/lib/general.py
```diff
@@ -430,10 +430,15 @@ def create_session(self) -> bool:
 
 		return True
 
-	def decode(self, *args, **kwargs) -> Optional[str]:
-		if self.session:
-			return self.session._trace_log.decode(*args, **kwargs)
-		return None
+	def decode(self, encoding: str = 'utf-8', errors='backlashreplace', strip: bool = True) -> str:
```
codefiles, line 433: I think `errors` can be dropped if it is not used otherwise please fix the typo in 'backslashreplace'.

svartkanin: I actually forgot to pass that into the `decode`

PR #2126, opened 2023-09-28
Fix mount command whitespace

codefiles:
Empty strings from `str_options` and/or `mount_fs` will cause excess whitespace in the command, this can be observed in the log.


Torxed:
It's not a breaking issue if understand the issue correctly?

Torxed (approved):

PR #2133, opened 2023-09-29
Add support for ESP partition flag

### LINKED ISSUE #2131: --efi-dir=None instead of being omitted
This appears to happen in v2.6.2

IMG_20230928_173654

Log: http://0x0.st/HV68.log

codefiles:
Fixes #2131

### DIFF
--- a/archinstall/lib/disk/device_model.py
+++ b/archinstall/lib/disk/device_model.py
@@ -658,7 +658,8 @@ class PartitionModification:
 	partuuid: Optional[str] = None
 	uuid: Optional[str] = None
 
-	_boot_indicator_flags = [PartitionFlag.Boot, PartitionFlag.XBOOTLDR]
+	_efi_indicator_flags = (PartitionFlag.Boot, PartitionFlag.ESP)
+	_boot_indicator_flags = (PartitionFlag.Boot, PartitionFlag.XBOOTLDR)
 
 	def __post_init__(self):
 		# needed to use the object as a dictionary key due to hash func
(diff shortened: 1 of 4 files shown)


Torxed (approved):

codefiles:
I thought that I should add support for the ESP partition flag to fix this when I looked over the logic. I ended up deciding on not making `PartitionModification.is_efi()` hard require the ESP partition flag in detection because then it would be needed in the configuration file when used with `--silent` or it would fail to get the ESP. Instead I made it depend on if either the boot flag or ESP flag are present. This can be changed to check for just the boot flag since it will be present when the ESP flag is. I am now wondering if there was value in adding the changes to `partitioning_menu.py` and `disk_conf.py`

The fix for the issue you were having is the mount point check that I added to `DeviceModification.get_efi_partition()`. You had more than one partition that matched the filter in `get_efi_partition()` and only the first one, which did not have a mount point, is returned.

Torxed:
> I would have confidence in it if you could confirm you tested it again and it does not fail with the configuration you were trying.

Tested and it works like a charm!

codefiles:
I want to revert the changes to `partitioning_menu.py` and `disk_conf.py`. I believe those changes were a mistake because the esp flag is an alias to the boot flag on GPT per parted documentation. That code would then not be needed since the boot flag was being set before this change and still is after it. This would also mean that `PartitionModification._efi_indicator_flags` can be removed and then the first check of `PartitionModification.is_efi()` can use the following change:

```diff
-any(set(self.flags) & set(self._efi_indicator_flags))
+PartitionFlag.Boot in self.flags
```

Considering this, the title used here is incorrect since the changes are to fix detecting the correct ESP as I mentioned in my last comment.

If you agree with this then I can submit a pull request for it immediately.

codefiles:
I now see that you made a release just before my comment was submitted. Oh well, I guess it does not matter anyway.

Torxed:
I missed it by a minute, is it a huge concern? If so I can revert it, because I haven't published to the Arch mirrors yet - so technically this release is not fully out there yet :)

codefiles:
Not a huge concern.

PR #2142, opened 2023-10-02
Fix ArchInstall ISO build

dylanmtaylor:
- This is an attempt to fix the issue of the Arch development ISOs not properly installing archinstall

## PR Description:

Add dependencies and a flag to make pip install the archinstall wheel despite Python being externally managed. This is fine because it's a live environment and these are development ISOs.

## Tests and Checks


dylanmtaylor:
Looks like this works
image

Torxed (approved):

Torxed:
Nice catch, thank you!

PR #2145, opened 2023-10-04
created kde-minimal

killertofus:
## PR Description:
Created kde-minimal.py for a more minimal Plasma installation

This will make easier for the kde users to setup their plasma environment.
## Tests and Checks

### DIFF
--- a/archinstall/default_profiles/desktops/kde-minimal.py
+++ b/archinstall/default_profiles/desktops/kde-minimal.py
@@ -1,0 +1,23 @@
+from typing import List, Optional, Any, TYPE_CHECKING
+
+from archinstall.default_profiles.profile import ProfileType, GreeterType
+from archinstall.default_profiles.xorg import XorgProfile
+
+if TYPE_CHECKING:
+	_: Any
+
+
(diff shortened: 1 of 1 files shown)


svartkanin:
Thus does feel a bit redundant, we do already have a KDE profile what exactly is not minimal about that one? 
All profiles should only install the bare minimum so they just work. If there is anything included in the existing one that is not essential we should rather update that one

Torxed:
I agree that the current profile could be looked over if it contains anything not KDE barbone.

We do ship with some default applications that KDE expects to be there when working with GUI tools. And the difference here in packages are very small:

```diff
return [
	"plasma-meta",
	"konsole",
+	"kwrite",
+	"dolphin",
+	"ark",
+	"plasma-wayland-session",
+	"egl-wayland"
]
```
Just for context, it's to give a graphical text editor, archive management and Wayland support (the last being hard to remove at all).

Torxed:
Closing this as there's been no response on our feedback.
And I'd recommend anyone wanting a smaller setup to feed a `--profile` parameter to a self hosted script (as it takes URL's and other goodies during installation).

PR #2151, opened 2023-10-08
Remove `select_language()` duplicate of `select_kb_layout()`

### LINKED ISSUE #2150: Functions `select_language()` and `select_kb_layout()` are identical
https://github.com/archlinux/archinstall/blob/edbc13590366e93bb8a85eacf104d5613bc5793a/archinstall/lib/interactions/general_conf.py#L88-L111

https://github.com/archlinux/archinstall/blob/edbc13590366e93bb8a85eacf104d5613bc5793a/archinstall/lib/locale/locale_menu.py#L132-L155

The function `select_language()` is not used.

The function `select_kb_layout()` is used once:

https://github.com/archlinux/archinstall/blob/edbc13590366e93bb8a85eacf104d5613bc5793a/archinstall/lib/locale/locale_menu.py#L90

codefiles:
Resolves #2150

### DIFF
--- a/archinstall/lib/interactions/general_conf.py
+++ b/archinstall/lib/interactions/general_conf.py
@@ -85,6 +85,19 @@ def ask_for_audio_selection(
 	return None
 
 
+def select_language(preset: Optional[str] = None) -> Optional[str]:
+	from ..locale.locale_menu import select_kb_layout
+	from ..exceptions import Deprecated
+
+	# We'll raise an exception in an upcoming version.
+	# raise Deprecated("select_language() has been deprecated, use select_kb_layout() instead.")
+
+	# No need to translate this i feel, as it's a short lived message.
+	warn("select_language() is deprecated, use select_kb_layout() instead. select_language() will be removed in a future version")
+
+	return select_kb_layout(preset)
+
+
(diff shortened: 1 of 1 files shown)


Torxed:
Should we add a deprecation warning and possible route users to the real function first? As we have a few library users who might use it without us being able to check if they do.

codefiles:
Makes sense. Could you proceed with that?

svartkanin:
There should be a deprecated decorator in the code somewhere to mark the functions

codefiles:
> There should be a deprecated decorator in the code somewhere to mark the functions

I thought I had seen a deprecated function with that or something like it used in the code somewhere before. I looked before submitting and could not find anything like that or I would have replicated it.

@Torxed or @svartkanin, would either of you submit a pull request to close the issue with the suggested deprecation procedures?

Torxed:
I was thinking of submitting a commit to this PR as this PR is pretty good. Give me a day :)

Torxed:
I believe this transparent call will suffice for now, with a `warn()` on future removal.
And later we'll raise an exception before finally removing the function alltogether.

PR #2154, opened 2023-10-09
Add handling of signal interrupt and EOF at input prompts

### LINKED ISSUE #1935: Signal interrupt results in a crash at input prompts
When a menu item prompts a user for input there is not a way to escape and return to the previous menu. Signal interrupt is not handled at these prompts and will cause a crash.

codefiles:
Fixes #1935


svartkanin:
This is awesome thanks

Torxed (approved):

PR #2168, opened 2023-10-15
Fix `MOUNT_POINT` for pre-mounted disk configuration

### LINKED ISSUE #2167: ERROR: /mnt/archinstall is not a directory
Hey!
So, I tried running archinstall after mounting my hard drives as /mnt /mnt/boot and /mnt/home
But it throws `/mnt/archinstall is not a directory`
Am I missing something, or is something wrong?

The logs and configs
[install.log](https://github.com/archlinux/archinstall/files/12908145/install.log)
[user_configuration.json](https://github.com/archlinux/archinstall/files/12908146/user_configuration.json)

codefiles:
Fixes #2167

### DIFF
--- a/archinstall/lib/interactions/disk_conf.py
+++ b/archinstall/lib/interactions/disk_conf.py
@@ -11,6 +11,7 @@ from ..menu import TableMenu
 from ..menu.menu import MenuSelectionType
 from ..output import FormattedOutput, debug
 from ..utils.util import prompt_dir
+from ..storage import storage
 
 if TYPE_CHECKING:
 	_: Any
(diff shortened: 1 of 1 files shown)


svartkanin:
I do agree that this will fix the bug, but I don't think it's good to override these static storage vars, I'm bit afraid that if we etart overriding these things will get quite messy potentially.

In the guided.py we're using this var to be passed to the installation, we should probably have some different logic there to check which one to use.

codefiles:
Monkey see, monkey do.

https://github.com/archlinux/archinstall/blob/3c2ed1c121f0cbe6340753f50e5c58b83675990b/archinstall/__init__.py#L255

This too.

https://github.com/archlinux/archinstall/blob/3c2ed1c121f0cbe6340753f50e5c58b83675990b/archinstall/lib/output.py#L167

svartkanin:
Ahh alright it's already like that, then let's get this in and put something in the backlog to address this

Torxed (approved):

Issue #2148, opened 2023-10-07
Auto updating the script when the ISO boots


RyzenDew:
Many people in the few discords that I’m a part of that are about linux we’re having issues with arch install and I think it would be a good idea that when you start Archinstall script it auto updates itself to avoid issues and always be on the newest release vs what the iso comes with .. so please consider how much this could help.

svartkanin:
Auto updating software isn't a good approach especially as it's a arch package or can be installed in other ways as well.

An option could be to display an initial message to inform the user that a newer version is available

ArchLinuxLover:
> Auto updating software isn't a good approach especially as it's a arch package or can be installed in other ways as well.
> 
> An option could be to display an initial message to inform the user that a newer version is available

If I have used the ISO before and it has been working for me  I dont want to use this as a "new" one without testing it.

Torxed:
> An option could be to display an initial message to inform the user that a newer version is available

This is something I've wanted to do, and try to get libalpm to be used. As it would be very beneficial to get C bindings to work. That way we could bypass a few more system calls. :)

PR #2169, opened 2023-10-15
User error info and version check

svartkanin:
* An error is now also written to the log file 
* If an error happens, there will be a more user friendly addiction to it to guide the user on where to report it and how to extract the file from the ISO
* Version check when starting archinstall to show the user that an update is available - Fix https://github.com/archlinux/archinstall/issues/2148

image
image

### DIFF
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -1,6 +1,7 @@
 """Arch Linux installer - guided, templates etc."""
 import importlib
 import os
+import traceback
 from argparse import ArgumentParser, Namespace
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Dict, Union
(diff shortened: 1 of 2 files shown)


Torxed (approved):
I like this a lot.
Note to the future, we should really start thinking about possibly using our own cache/db for archinstall related tasks.

There's two reasonings I find valid here:
1. This separates `archinstall` from the systems instance, meaning we can do `pacman -Sy` to get a fresh package list - but at the same time the user can outside of `archinstall` do `pacman -S <package>` and not get version conflicts.
2. When potentially using `archinstall` in the future to administrate package setups (lets say they want to post-install configure a desktop setup, that could become possible), then the same benefit from point 1 would apply :)

But this is a great enhancement, so lets get it in!

PR #2173, opened 2023-10-16
Refactor `_add_systemd_bootloader()` newlines

codefiles:


archinstall/lib/installer.py
```diff
@@ -836,41 +835,40 @@ def _add_systemd_bootloader(
...
 
 		# Ensure that the $BOOT/loader/entries/ directory exists before we try to create files in it
 		entries_dir = loader_dir / 'entries'
 		entries_dir.mkdir(parents=True, exist_ok=True)
 
 		comments = (
-			'# Created by: archinstall\n',
-			f'# Created on: {self.init_time}\n'
+			'# Created by: archinstall',
+			'# Created on: ' + self.init_time
```
Torxed, line 846: I think I prefer to keep the f-string. Even if I know init_time is a strong it feels better knowing f-string will ensure its a str() object. nitpicking from my side.

Torxed (approved):

Issue #2172, opened 2023-10-15
Exceptions that inherit from `BaseException` are not logged when raised


codefiles:
None of the exceptions in [`archinstall/lib/exceptions.py`](https://github.com/archlinux/archinstall/blob/c4c5cc42248645142828d09ec0322cdefa49209a/archinstall/lib/exceptions.py) are caught and logged in the following since they are all `BaseException` rather than `Exception`.

https://github.com/archlinux/archinstall/blob/c4c5cc42248645142828d09ec0322cdefa49209a/archinstall/__init__.py#L310-L324

Torxed:
Good catch, best practice would be to change all the exceptions to `Exception` as per hierarchy: https://docs.python.org/3/library/exceptions.html#exception-hierarchy

We'd just need to handle `KeyboardInterrupt` separately for instance

svartkanin:
Ahh right i missed this, I'll raise an update shortly

Issue #2171, opened 2023-10-15
Broken when archinstall is installed with pip


codefiles:
This is the output of `archinstall` when installed with pip rather than with pacman:

```
Checking version...
Traceback (most recent call last):
  File "/home/scripttest/archinstall/.venv/bin/archinstall", line 8, in <module>
    sys.exit(run_as_a_module())
             ^^^^^^^^^^^^^^^^^
  File "/home/scripttest/archinstall/archinstall/__init__.py", line 312, in run_as_a_module
    main()
  File "/home/scripttest/archinstall/archinstall/__init__.py", line 298, in main
    _check_new_version()
  File "/home/scripttest/archinstall/archinstall/__init__.py", line 284, in _check_new_version
    upgrade = Pacman.run("-Qu archinstall").decode()
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scripttest/archinstall/archinstall/lib/pacman/__init__.py", line 45, in run
    return SysCommand(f'{default_cmd} {args}')
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scripttest/archinstall/archinstall/lib/general.py", line 365, in __init__
    self.create_session()
  File "/home/scripttest/archinstall/archinstall/lib/general.py", line 414, in create_session
    with SysCommandWorker(
  File "/home/scripttest/archinstall/archinstall/lib/general.py", line 197, in __exit__
    raise SysCallError(
archinstall.lib.exceptions.SysCallError: ['/usr/bin/pacman', '-Qu', 'archinstall'] exited with abnormal exit code [1]: error: package 'archinstall' was not foun
d
warning: 'archinstall' is a file, you might want to use -p/--file.

```

Running `pacman -Qu archinstall` will result in an error and an exit code of 1 since archinstall will not be found in the local package database when it is not installed with pacman.

https://github.com/archlinux/archinstall/blob/c4c5cc42248645142828d09ec0322cdefa49209a/archinstall/__init__.py#L284

svartkanin:
Thanks for spotting! I'll raise a fix shortly

Issue #2170, opened 2023-10-15
Addition of `_check_new_version()` obsoletes `check_mirror_reachable()`


codefiles:
If `pacman -Sy` must be executed, there is no need to do it more than once. Both of these functions are called when `archinstall` is executed:

https://github.com/archlinux/archinstall/blob/c4c5cc42248645142828d09ec0322cdefa49209a/archinstall/__init__.py#L281-L289

https://github.com/archlinux/archinstall/blob/c4c5cc42248645142828d09ec0322cdefa49209a/archinstall/lib/networking.py#L35-L45

PR #2175, opened 2023-10-16
Fixes for prev PR

svartkanin:
@codefiles @Torxed fixes for the previous PR

- Fix https://github.com/archlinux/archinstall/issues/2170: Refactor the code a bit to only perform `pacman -Sy` once which made the check mirror function obsolete and introduce an official new archinstall parameter `--skip-version-check` 
- Fix https://github.com/archlinux/archinstall/issues/2171: Fail open in case we're dealing with a `pip` installed version of archinstall; the major relevance for the version check is for folks running it in a old ISO
- Fix https://github.com/archlinux/archinstall/issues/2172: Changed the inherritence of the custom Exceptions

### DIFF
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -41,7 +41,7 @@ if TYPE_CHECKING:
 	_: Any
 
 
-__version__ = "2.6.4"
+__version__ = "2.6.3"
 storage['__version__'] = __version__
 
 # add the custom _ as a builtin, it can now be used anywhere in the
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -41,7 +41,7 @@ if TYPE_CHECKING:
 	_: Any
 
 
-__version__ = "2.6.3"
+__version__ = "2.6.4"
 storage['__version__'] = __version__
 
 # add the custom _ as a builtin, it can now be used anywhere in the
@@ -89,6 +89,8 @@ def define_arguments():
 	parser.add_argument("--no-pkg-lookups", action="store_true", default=False,
 						help="Disabled package validation specifically prior to starting installation.")
 	parser.add_argument("--plugin", nargs="?", type=str)
+	parser.add_argument("--skip-version-check", action="store_true",
+						help="Skip the version check when running archinstall")
 
 
 def parse_unspecified_argument_list(unknowns: list, multiple: bool = False, err: bool = False) -> dict:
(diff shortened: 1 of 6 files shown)


archinstall/__init__.py
```diff
@@ -280,8 +282,20 @@ def plugin(f, *args, **kwargs):
...
+		debug(f'Failed to perform version check: {e}')
+		info(f'Arch Linux mirrors are not reachable. Please check your internet connection')
+		exit(1)
+
+	upgrade = None
+
+	try:
+		upgrade = Pacman.run("-Qu archinstall").decode()
+	except Exception as e:
+		debug(f'Failed determine pacman version: {e}')
```
Torxed, line 298: Could we do [pip check](https://stackoverflow.com/a/62571316) as backup?

codefiles: I think the stated intent is met without that.

> major relevance for the version check is for folks running it in a old ISO

In my development environment I have archinstall installed in [development mode](https://setuptools.pypa.io/en/latest/userguide/development_mode.html) (dependencies installed via pacman and substitue in `python -m venv --system-site-packages .venv` when creating venv). This is a pip install and I would rather not have to remember the new `--skip-version-check` argument to prevent sending a HTTP request to PyPI everytime I launched the installer to do testing.

archinstall/__init__.py
```diff
@@ -280,8 +282,20 @@ def plugin(f, *args, **kwargs):
...
-	upgrade = Pacman.run("-Qu archinstall").decode()
+
+	try:
+		Pacman.run("-Sy")
+	except Exception as e:
+		debug(f'Failed to perform version check: {e}')
+		info(f'Arch Linux mirrors are not reachable. Please check your internet connection')
+		exit(1)
+
+	upgrade = None
```
codefiles, line 293: Could delete this line and add an else statement to the try block below with the `upgrade` conditional inside it.

codefiles, line 293: Could delete this line and add an else statement to the try block below with the `upgrade` conditional inside it.

archinstall/__init__.py
Torxed: Fair enough, I think we could get around it by checking if we're [in venv](https://stackoverflow.com/a/1883251) or not.

But I think the main goal is achieved here by eliminating users running old versions on the official ISO's :)

svartkanin:
@Torxed you happy to merge it in?

Torxed (approved):

PR #2176, opened 2023-10-16
Added enum 'All open-source (default)' for compatibility with older config versions

d1nch8g:
Old configuration versions of archinstall when choosing graphics drivers look for `All open-source (default)` value. Current enum value has `All open-source` value.

Fix should make older versions of archinstall configurations compatible with new version of enum.

### DIFF
--- a/archinstall/lib/profile/profile_model.py
+++ b/archinstall/lib/profile/profile_model.py
@@ -32,6 +32,9 @@ class ProfileConfiguration:
 		greeter = arg.get('greeter', None)
 		gfx_driver = arg.get('gfx_driver', None)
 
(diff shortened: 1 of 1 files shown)


Torxed (approved):
I don't mind trying it :)

svartkanin:
This seems to be a bit of a dangerous syntax, we're making one of the enum types a tuple.

Can we handle this in a more elegantand explicit way

d1nch8g:
@svartkanin 

Changed to condition which updates value from old configuration files

d1nch8g:
@svartkanin 

Sorry for the inconvenience; should have created an issue highlighting an optional backward compatability addition for that field.

Issue #2194, opened 2023-10-28
Could not detect root at mountpoint /mnt/archinstall


XerwinXpl2:
hi, i tried setting up disk encryption in this script but it didnt worked so i tried doing it manually and then using archinstall.
according to arch wiki i did `cryptsetup -y -v luksFormat /dev/sdc5` (root partition) then `cryptsetup open /dev/sdc5 root`, `mkfs.ext4 /dev/mapper/root` and `mount /dev/mapper/root /mnt/archinstall` and everything worked, but then after mounting `/dev/sdc3` at `/mnt/archinstall/boot` and running archinstall (using already existing disc config or however that option was called i dont remember sorry, and typing /mnt/archinstall) it failed with `Could not detect root at mountpoint /mnt/archinstall`. here is log file: [install.log](https://github.com/archlinux/archinstall/files/13196587/install.log)

also everything was working before i tried to use it on encrypted partition

another edit: i tried both version installed from pacman and cloned github version

intentionally-left-nil:
I can confirm this is the case.

Looking at the details, the "manual partition" code doesn't detect any of the mounted btrfs volumes. It only sees the /boot/ partition that I mounted into that directory

intentionally-left-nil:
@E15dev I was able to get it to work with my setup using the linked PR. If you're comfortable & want to give it a shot I'd love to hear if it solves your issue as well

E15dev:
i managed to install grub manually, i dont have free disks now so i cant test.

svartkanin:
There has been significant changes since this was raised and may be fixed by now. 
I'll close this for now, if it reappears we can reopen it with a fresh log file

PR #2198, opened 2023-11-04
RFC: fix to detect mounted cryptfs btrfs subvolumes

intentionally-left-nil:
- This fix issue: #2194

## PR Description:
This is a prototype fix for detecting already-mounted btrfs volumes, when inside of a LUKS encrypted partition. Without this fix, the code only detects the LUKS partition and not its mounted counterpart.

This code adds the detection for this case, and then recurses down to the children to find the btrfs subvolumes. The reason this is a prototype (request for comment) is that I don't know what to do about the `partition` information. The subvolume is technically part of the LUKS partition, but e.g. the path is /dev/mapper/<name> not /dev/sda1 (which points to LUKS). etc.

This PR does seem? to work when running archinstall on my own setup, but I don't have enough knowledge of the codebase to understand what the correct solution is.

I'm happy to carry this PR forward, with guidance. Otherwise I'm happy to turn it over to @Torxed or others to implement a more robust solution

## Tests and Checks

### DIFF
--- a/archinstall/lib/disk/device_handler.py
+++ b/archinstall/lib/disk/device_handler.py
@@ -75,21 +75,28 @@ class DeviceHandler(object):
 				lsblk_info = get_lsblk_info(partition.path)
 				fs_type = self._determine_fs_type(partition, lsblk_info)
 				subvol_infos = []
+				path_override = None
 
 				if fs_type == FilesystemType.Btrfs:
 					subvol_infos = self.get_btrfs_info(partition.path)
-
-				partition_infos.append(
-					_PartitionInfo.from_partition(
-						partition,
-						fs_type,
-						lsblk_info.partn,
-						lsblk_info.partuuid,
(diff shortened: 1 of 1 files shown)


intentionally-left-nil:
The PR isn't quite right, now btrfs.progs isn't being installed. The attempted commit I pushed to fix this just made things worse...

Torxed:
I just wanted to swing by with a comment so you know I've seen it and will take a look :)

intentionally-left-nil:
Thanks!

I think there's a few pieces to get right:
* The UUID (especially when creating any boot entries)
* The filesystem type (btrfs) so that you end up with btrfs.progs installed
* The root subvolume id


and there are probably others that don't pertain to my specific scenario

svartkanin:
This PR is quite outdated so I'll close it for now. If you like to update it we can have another look at it

PR #2207, opened 2023-11-09
Fix logic error in `encrypt()` loop

### LINKED ISSUE #2206: Logic error in `encrypt()` loop
The loop in `Luks2.encrypt()` contains a logic error.

https://github.com/archlinux/archinstall/blob/c9755b35e7364d208d836ec051d365354be6dd63/archinstall/lib/luks.py#L93

If `DISK_RETRY_ATTEMPTS` is `0` then an initial attempt will not occur and no code in the loop will be executed.

The maximum number of retries that can occur is erroneously one less than indicated by `DISK_RETRY_ATTEMPTS`; the initial attempt is when `retry_attempt` is `0` and all others after it are retries.

codefiles:
Fixes #2206


Torxed (approved):

Torxed:
Nice catch!

Issue #2197, opened 2023-10-31
Limine bootloader error, unable to create directories 


spsf64:
I think title says it all, will attach image with the error....
I was able to install if I create the /EFI/BOOT directory in the /boot partiton before starting archinstall.

![limine](https://github.com/archlinux/archinstall/assets/1877391/5f8f9d90-83e2-4a00-91e4-8bb8778b7a11)

amebtw:
Tested today with arch iso 2023.11.01; still a problem.
I was not able to install by creating /EFI/BOOT.
[cmd_history.txt](https://github.com/archlinux/archinstall/files/13250609/cmd_history.txt)
[install.log](https://github.com/archlinux/archinstall/files/13250610/install.log)
[user_configuration.json](https://github.com/archlinux/archinstall/files/13250612/user_configuration.json)

amebtw:
This issue can be closed as #2216 fixed the issue.

spsf64:
Just tested and its working as expected.
Thanks!

PR #2216, opened 2023-11-16
Fix Limine bootloader deployment

48cf:
- This fix issue: #2197

## PR Description:

This PR fixes the deployment of the Limine bootloader on UEFI systems (previously broken) and fixes the post-install hooks on BIOS systems to deploy to the disk rather than the boot partition (that's wrong and also caused errors when previously upgrading).

## Tests and Checks
- [x] I have tested the code!

### DIFF
--- a/archinstall/lib/disk/device_handler.py
+++ b/archinstall/lib/disk/device_handler.py
@@ -133,6 +133,20 @@ class DeviceHandler(object):
 		lsblk = get_lsblk_info(dev_path)
 		return Path(f'/dev/{lsblk.pkname}')
 
+	def get_unique_path_for_device(self, dev_path: Path) -> Optional[Path]:
+		paths = Path('/dev/disk/by-id').glob('*')
+		linked_targets = {p.resolve(): p for p in paths}
+		linked_wwn_targets = {p: linked_targets[p] for p in linked_targets
+			if p.name.startswith('wwn-') or p.name.startswith('nvme-eui.')}
+
+		if dev_path in linked_wwn_targets:
+			return linked_wwn_targets[dev_path]
+
+		if dev_path in linked_targets:
+			return linked_targets[dev_path]
+
+		return None
+
 	def get_uuid_for_path(self, path: Path) -> Optional[str]:
 		partition = self.find_partition(path)
 		return partition.partuuid if partition else None
--- a/archinstall/lib/global_menu.py
+++ b/archinstall/lib/global_menu.py
@@ -364,10 +364,10 @@ class GlobalMenu(AbstractMenu):
 			return "Boot partition not found"
 
 		if bootloader == Bootloader.Limine:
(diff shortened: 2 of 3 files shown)


mintsuki:
LGTM.

This PR fixes a critical issue. The Limine bootloader option is completely broken as it stands without this patch.

Andy-Python-Programmer (approved):

mintsuki (approved):

archinstall/lib/disk/device_handler.py
```diff
@@ -133,6 +133,28 @@ def get_parent_device_path(self, dev_path: Path) -> Path:
 		lsblk = get_lsblk_info(dev_path)
 		return Path(f'/dev/{lsblk.pkname}')
 
+	def get_unique_path_for_device(self, dev_path: Path) -> Optional[Path]:
+		disks_by_id = os.listdir('/dev/disk/by-id')
```
svartkanin, line 137: Using `Path` instead will save some conversations
```
Path('/dev/disk/by-id').glob('*')
```
The glob can then also filter away `wwn-` and `nvme-eui.` right away

svartkanin: Basically this should do 
```
paths = Path('/dev/disk/by-id').glob('*')
linked_targets = {p.resolve(): p for p in paths}

if dev_path in linked_targets:
    return linked_targets[dev_path]

```

48cf: I'd like to keep the behavior of `wwn-*` and `nvme-eui.*` paths being prioritized. Does this code look okay or is there a better way to achieve this?

```py
paths = Path('/dev/disk/by-id').glob('*')
linked_targets = {p.resolve(): p for p in paths}
linked_wwn_targets = {p: unique_targets[p] for p in linked_targets
	if p.name.startswith('wwn-') or p.name.startswith('nvme-eui.')}

if dev_path in linked_wwn_targets:
    return linked_wwn_targets[dev_path]

if dev_path in linked_targets:
    return linked_targets[dev_path]
```

mintsuki (approved):

48cf:
Had a little copy-paste mistake - I forgot to update a variable name from a previous version of the code. Should be all set!

mintsuki (approved):

codefiles:
I like seeing that `_get_microcode()` and `_get_kernel_params()` are put to use here.

mintsuki (approved):

mintsuki (approved):

48cf:
@svartkanin could you please rerun the CI? this time it shouldn't fail as I ran the pre-commit hooks locally

Torxed (approved):

Torxed:
Apologies for the delay, PR looks good and is now merged.
I've begun to wrap up for releasing the next version - however that might take a day or two.

But assuming the ISO build process works as intended this month, this should make it to the next Arch ISO.

Issue #2144, opened 2023-10-03
Hangs on "Waiting for time sync (systemd-timesyncd.service) to complete." during install


peracchi:
As the title says, `archinstall` hangs on _"Waiting for time sync (systemd-timesyncd.service) to complete."_ during install IF NETWORK DO NOT ALLOW NTP TRAFFIC TO GO OUT.

It do not timeout as expected.

The only way I found to do my installation was to open another SSH session and:
```
systemctl stop systemd-timesyncd
nano /etc/systemd/timesyncd.conf
<edit file>
NTP=<my internal NTP server>
<save file>
systemctl start systemd-timesyncd
```

Then, installation proceeds.

Torxed:
Can you confirm that this is happening because network configuration is done manually (like wifi, slow DHCP or even static IP)?

peracchi:
Yes, at work we have a very restrictive firewall, MITM, self-signed certificates and other annoyances.

One or more of these factor prevent NTP communication with outside servers and things only work after I stop NTP service, point it to internal NTP server and then restart NTP service.

Torxed:
I figured something like this was going on ^^
Without even checking, I'm pretty sure this happens:

1. Boot
2. timesync service tries to sync
3. It fails and restart mechanics sucks
4. You configure network
5. Run archinstall
6. It waits for the next timesync to trigger
7. You SSH because the sync never happens
8. You fix it

This is an issue for manual installs to, and I think instead of fixing it here, I'll discuss with some other upstream devs if we can't solve this permanently for all use cases.

peracchi:
I'm installing Arch on a VM on Proxmox using the ISO `archlinux-2023.09.01-x86_64.iso`

There are several other steps I need to do before starting the actual installation.

These steps are:
- boot ISO on Proxmox VM
- change root password to SCP/SSH
- SCP internal self-signed certificates
- `trust anchor <certificates>`
- `pacman-key --init && pacman-key --populate archlinux && pacman -Syy`
- `pacman -S --noconfirm archinstall`
- only then `archinstall`

peracchi:
> We could however add a `--skip-ntp` to the command line.

This would be sufficient.

But the ideal would be to detect the problem with NTP (by timeout, perhaps?), kill the NTP process and proceed with the installation.

alan-barzilay:
> But the ideal would be to detect the problem with NTP (by timeout, perhaps?), kill the NTP process and proceed with the installation.

[...]

PR #2225, opened 2023-11-21
Added a --skip-ntp parameter to deal with #2144

Torxed:
This fix issue:
 * #2144

## PR Description:

Since certain network environments prohibit the default NTP, we might end up in a loop waiting for NTP to happen. Usually this can be resolved while the loop is "hanging", but no information is given about it.

This PR adds both a `warn()` call notifying the user about the potential issue after 5 seconds, but also a flag to `--skip-ntp` which disables the check all together. This can be useful in environments where NTP is not available and time is instead synced manually.


archinstall/lib/installer.py
```diff
@@ -130,13 +130,23 @@ def _verify_service_stop(self):
...
+				if not _notified and time.time() - _started_wait > 5:
+					_notified = True
+					warn("Time syncronization not completing, while you wait - check the docs for workarounds: https://archinstall.readthedocs.io/")
+					
+				time_val = SysCommand('timedatectl show --property=NTPSynchronized --value').decode()
+				if time_val and time_val.strip() == 'yes':
+					break
+				time.sleep(1)
+		else:
+			info('Skipping waiting for automatic time sync (this can cause issues if time it out of sync during installation)')
```
alan-barzilay, line 149: i think this part is worded in a confusing way "if time it out of sync", is it a typo for  if time i**s** out of sync

archinstall/lib/installer.py
```diff
@@ -130,13 +130,23 @@ def _verify_service_stop(self):
...
-			time.sleep(1)
+		if not storage['arguments'].get('skip_ntp', False):
+			info('Waiting for time sync (timedatectl show) to complete.')
+
+			_started_wait = time.time()
+			_notified = False
+			while True:
+				if not _notified and time.time() - _started_wait > 5:
+					_notified = True
+					warn("Time syncronization not completing, while you wait - check the docs for workarounds: https://archinstall.readthedocs.io/")
```
alan-barzilay, line 142: Maybe add something like "(...) completing ,this could me an issue with your network/ntp/environment permissions, while you wait (...)" and also maybe link directly to the relevant section instead of the whole doc?

archinstall/lib/installer.py
```diff
@@ -130,13 +130,23 @@ def _verify_service_stop(self):
...
 		"""
-		info('Waiting for time sync (systemd-timesyncd.service) to complete.')
 
-		while True:
-			time_val = SysCommand('timedatectl show --property=NTPSynchronized --value').decode()
-			if time_val and time_val.strip() == 'yes':
-				break
-			time.sleep(1)
+		if not storage['arguments'].get('skip_ntp', False):
+			info('Waiting for time sync (timedatectl show) to complete.')
```
alan-barzilay, line 135: probably stupid question, why change  from `systemd-timesyncd.service` to` timedatectl show` ?

alan-barzilay (commented):
thank you for the quick response as always! I made some comments, hope they help :)

archinstall/lib/installer.py
Torxed: Ah yes, typo. Also I just realized we should use the translation call `_()` too! I'll push a fix. Thank for checking it out!

PR #2234, opened 2023-11-21
Testing github pages

Torxed:
readthedocs is a bit of a pain to get working with our dependencies.
Not to mention the recommended way of pushing dependencies with `requirements.txt` instead of `pyproject.toml`.

### DIFF
--- a/.github/workflows/github-pages.yml
+++ b/.github/workflows/github-pages.yml
@@ -1,0 +1,27 @@
+name: documentation
+
+on: [push, pull_request, workflow_dispatch]
+
+permissions:
+  contents: write
+
+jobs:
+  docs:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


svartkanin:
@Torxed maybe it's better to keep it all in one place and use github wiki? 
But I remember you mentioned that there's talks about a potential migration to gitlab in the future which might erase that

Torxed:
> @Torxed maybe it's better to keep it all in one place and use github wiki? But I remember you mentioned that there's talks about a potential migration to gitlab in the future which might erase that

Correct, migration will happen so planning for it.
I got github pages working and we'll host it on https://archinstall.archlinux.page/ (just waiting for the certificate). This is where gitlab pages will live too most likely. And it appears to support restructuredtext.

Issue #2214, opened 2023-11-13
Add nvidia-drm.modeset


ryzendew:
Could nvidia-drm.modeset=1 be added to grub please so we can have Wayland for NVidia out of the box?
but only if the NVidia drivers are selected.

Cardiacman13:
Hi, you can add it in modeprobe.d, this way, it will work with all bootloaders.

Torxed:
> Hi, you can add it in modeprobe.d, this way, it will work with all bootloaders.
> 
> In the same way, you could cancel the gdm rule with: 'ln -s /dev/null /etc/udev/rules.d/61-gdm.rules' to give Nvidia users the option to use Wayland on GNOME.

I like this method, easy to maintain. Any dracks? Early KMS might be the only thing missing but that should be solved elsewhere anyway?

Cardiacman13:
> > Hi, you can add it in modeprobe.d, this way, it will work with all bootloaders.
> > In the same way, you could cancel the gdm rule with: 'ln -s /dev/null /etc/udev/rules.d/61-gdm.rules' to give Nvidia users the option to use Wayland on GNOME.
> 
> I like this method, easy to maintain. Any dracks? Early KMS might be the only thing missing but that should be solved elsewhere anyway?

No, normally there shouldn't be any dracks to put it in modeprobe.d except that people are so used to putting it in the boot-loader options that they might open issues saying it's not done XD . 

Also, normally, early KMS isn't necessary, especially since it makes the initramfs huge, which can cause issues for those using multiple kernels.

Torxed:
> > > Hi, you can add it in modeprobe.d, this way, it will work with all bootloaders.
> > > In the same way, you could cancel the gdm rule with: 'ln -s /dev/null /etc/udev/rules.d/61-gdm.rules' to give Nvidia users the option to use Wayland on GNOME.
> > 
> > 
> > I like this method, easy to maintain. Any dracks? Early KMS might be the only thing missing but that should be solved elsewhere anyway?
> 
> No, normally there shouldn't be any dracks to put it in modeprobe.d except that people are so used to putting it in the boot-loader options that they might open issues saying it's not done XD .
> 
> Also, normally, early KMS isn't necessary, especially since it makes the initramfs huge, which can cause issues for those using multiple kernels.

[...]

Issue #2002, opened 2023-08-10
Missing Nvidia Proprietary Driver after archinstall Script with August Arch Linux ISO's 


Cardiacman13:
Hi,

Issue: The August Arch Linux ISO's archinstall script does not install the Nvidia proprietary driver.

Environment:
    Arch Linux ISO: version 2023.08.01 (issue persists despite script update but work with july iso)

0xREAVEN:
Hello, I have exactly the same issue. I noticed after installing KDE and Hyprland that the Nvidia drivers are not installed.

chanhhanh:
I also have this issue. I noticed after installing Arch I tried running apps with DXVK but it doesn't work. So I open my GPU settings and see that I am running nouveau driver (Which I assume that I'm missing proprietary driver).

lutzseverino:
Same issue here. Driver is not present after installing.

dozn:
The September release seems to have fixed it 🥳

Torxed:
We did not release a new version in September, so either it was a package conflict/issue with the `nvidia` package or something is still broken, can any of you who commented before dozn verify a second time that it works?

dozn:
Interesting, I assumed you released every month alongside the Arch ISOs :D Hope it works for others as well!

Cardiacman13:
It's Still broken.

Torxed:
> Interesting, I assumed you released every month alongside the Arch ISOs :D Hope it works for others as well!

We tried, but releases have started getting complicated due to the amount of testing that needs to be done.

> It's Still broken.

Thanks, I suspected as much.
And in the gfx section, I assume you selected `nvidia` when asked during DE selection?
Mind submitting hte logs? https://archinstall.readthedocs.io/help/issues.html#log-files

dozn:
Figured it out, you never install plain ol' `linux-headers` https://github.com/archlinux/archinstall/blob/b1e00687953de168bbc31ea435b2c5bdc7c61e39/archinstall/lib/profile/profiles_handler.py#L217

I was using then Zen kernel, so I Logged into the regular kernel, no drivers, installed `linux-headers`, rebooted, and voila!

Torxed:
Oooh, heh.. ignore my stupidity but are they requires for the `linux` kernel as well?
That feels like a nvidia specific bug or change, because IIRC previously (https://github.com/archlinux/archinstall/issues/585) the headers have only been required for non-standard kernels.

[...]

PR #2241, opened 2023-11-22
Added nvidia-dkms when installing propri driver to 

### LINKED ISSUE #2233: [Solved] broken intramfs, black screen
Broken initramfs after clean install with encryption, kde desktop and propretary nvidia drivers.
1. Loads grub
2. Decrypt drive
3. Loads initramfs
4. Black screen with no output

Torxed:
This should fix #2233, #2214 and #2002

## PR Description:

The issue is not totally clear, but `nvidia-dkms` appears to be missing on some hardwares.
The solution would be to correctly identify when this is needed and only install it then.
But we don't have enough information to do that logic right now, and adding it appears to not harm users who don't need it. It can be safely cleaned up afterwards for now.

### DIFF
--- a/archinstall/lib/hardware.py
+++ b/archinstall/lib/hardware.py
@@ -42,6 +42,7 @@ class GfxPackage(Enum):
 	LibvaMesaDriver = 'libva-mesa-driver'
 	Mesa = "mesa"
 	Nvidia = 'nvidia'
+	NvidiaDKMS = 'nvidia-dkms'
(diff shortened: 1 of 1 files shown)


Torxed:
@svartkanin still wrapping my head around the re-work of the package management around these things. But this is the correct way to install `nvidia-dkms` when `nvidia` (nvidia prop) is selected right?

svartkanin:
Yep that looks all good:)

PR #2261, opened 2023-11-25
Enable systemd-oomd - Update installer.py

ndellagrotte:
Enables systemd-oomd during the install process. As is, archinstall does not enable an OOM killer by default. This is a significant oversight which has undoubtedly resulted in many system freezes and lost work. However, systemd-oomd has been criticized for being too aggressive, which could also result in lost work. Personally, I think it's better than nothing. Also, I have tested the code, and it appears that systemd-oomd is enabled and running after booting into arch.

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -458,6 +458,10 @@ class Installer:
 		# fstrim is owned by util-linux, a dependency of both base and systemd.
 		self.enable_service("fstrim.timer")
 
+	def enable_systemd_oomd(self) -> None:
+		info('Enabling systemd-oomd.service to handle OOM events')
+		self.enable_service('systemd-oomd')
+
 	def enable_service(self, services: Union[str, List[str]]) -> None:
 		if isinstance(services, str):
 			services = [services]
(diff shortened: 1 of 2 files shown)


Torxed:
> This is a significant oversight

I think the choice of words here might be a bit drastic. This is not a significant oversight, if at all an oversight. This is a totally optional thing that users can install and enable post-installation without any affect.

I'll even consider allowing this:
```python
def enable_systemd_oomd(self) -> None:
	info('Enabling systemd-oomd.service to handle OOM events')
	self.enable_service('systemd-oomd')
```

If someone wants to enable it, but then again since it's very user specific they could use this instead directly: https://github.com/archlinux/archinstall/blob/f6c0e6aa84ad48c16021e5e79cba6a8bf64503f1/archinstall/lib/installer.py#L463C6-L463C6

The more I write the more I realize, I'll close this in favor of a more generic PR that enables a configuration option called "Enable services" where the services can be named instead in a list, and we'll call `enable_service()` for those.

I can re-open this if arguments supporting this need can be put forward, especially if we can cite the wiki installation guide on this being needed.

PR #2266, opened 2023-11-26
Fixes #2262 and GFX rewrite

### LINKED ISSUE #2262: [2.7.0- Master GIT]
Tested today : Nivida-dkms is installed but no headers, so no modules was compilated.

Next comment for logs.

Screenshot_20231125_115710

svartkanin:
This fixes #2262 and rewrites the GFX installation logic to something (hopefully) simpler

### DIFF
--- a/archinstall/lib/hardware.py
+++ b/archinstall/lib/hardware.py
@@ -37,18 +37,22 @@ class CpuVendor(Enum):
 
 
 class GfxPackage(Enum):
+	Dkms = 'dkms'
 	IntelMediaDriver = 'intel-media-driver'
 	LibvaIntelDriver = 'libva-intel-driver'
 	LibvaMesaDriver = 'libva-mesa-driver'
 	Mesa = "mesa"
-	Nvidia = 'nvidia'
+	NvidiaDkms = 'nvidia-dkms'
 	NvidiaOpen = 'nvidia-open'
+	NvidiaOpenDkms = 'nvidia-open-dkms'
 	VulkanIntel = 'vulkan-intel'
 	VulkanRadeon = 'vulkan-radeon'
 	Xf86VideoAmdgpu = "xf86-video-amdgpu"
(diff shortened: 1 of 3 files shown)


archinstall/lib/hardware.py
```diff
@@ -37,18 +37,22 @@ def get_ucode(self) -> Optional[Path]:
 
 
 class GfxPackage(Enum):
+	Dkms = 'dkms'
 	IntelMediaDriver = 'intel-media-driver'
 	LibvaIntelDriver = 'libva-intel-driver'
 	LibvaMesaDriver = 'libva-mesa-driver'
 	Mesa = "mesa"
-	Nvidia = 'nvidia'
```
Torxed, line 44: Won't this make it so `nvidia` package is not installed? Which is needed for some platforms? Or we rely solely on the `-dkms` and add the nvidia hooks to build the driver each kernel upgrade?

svartkanin: Maybe I missed something but in the old code Nvidia didn't seem to be installed at all. But please double check that I didn't miss anything

I've actually tried installing nvidia and nvidia-dkms which threw a conflict error

Torxed:
This could solve:
 * https://github.com/archlinux/archinstall/issues/2002
 * https://github.com/archlinux/archinstall/issues/2214 (slightly related)
 
So installing `nvidia-dkms` instead of `nvidia` should work, the downside being that every upgrade will be slower as the kernel modules needs to be compiled during installation (this is done automatically AFAIK).

I'll merge this for now, but we should work on a fix to:
 * Install `nvidia` if __only__ `linux` kernel is selected (And also install `linux-headers`)
 * Install `nvidia-lts` if __only__ `linux-lts` kernel is selected (Also `linux-lts-headers`)
 * Install `nvidia-dkms` if combintions of the above *or* none of the above was selected

That is the supported packaging from the package maintainers. But `-dkms` should work on all of the above - just slower upgrades.

PR #2278, opened 2023-11-30
Fix UKI default value

### LINKED ISSUE #2255: no default value for uki key in configurations.json
(archinstall 2.7.0) 
When using archinstall in a system with no support for uefi, the user is not prompted to choose a value for the uki key. However, when running archinstall with a config file without a uki key defined, it errors out since there is no such key defined with a default value.

At least thats what I understood from the problem looking at PR(https://github.com/archlinux/archinstall/pull/1519).

Adding uki: false in the config file fixed this for me.

by the way, how is uefi supported checked? I though it was impossible to know from OS side

codefiles:
Fixes #2255

### DIFF
--- a/archinstall/scripts/guided.py
+++ b/archinstall/scripts/guided.py
@@ -158,7 +158,7 @@ def perform_installation(mountpoint: Path):
 
 		installation.add_bootloader(
 			archinstall.arguments["bootloader"],
-			archinstall.arguments["uki"]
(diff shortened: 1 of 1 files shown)


Torxed (approved):

Torxed:
Awesome! Thank you!

codefiles:
Excellent job maintaining. I appreciate your dedication and respect your willingness to volunteer.

PR #2282, opened 2023-12-02
downgrade pyparted due to "unrelease"

alan-barzilay:
somehow and for some reason, pyparted 3.13 is no longer available in pypi only in its git repository. To deal with this I downgraded the version requirement and updated to >= instead of the == exact version match

- This fix issue: 

## Tests and Checks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ classifiers = [
 ]
 dependencies = [
     "simple-term-menu==1.6.3",
-    "pyparted==3.13.0",
+    "pyparted>=3.12.0",
 ]
 
 [project.urls]


Torxed:
If I recall correctly, we had issues with older versions of pyparted (libparted), so I'm wondering if this will maybe ensure that pyparted gets installed yes - but breaks calls later?

I'll have to verify this first.
Better odds are probably to get the packager of PyParted's attention and push the release as they should. It's a pretty minor thing as they only need to run ~1 command more or less.

Torxed:
Adding a cross reference to your issue for traceability: https://github.com/dcantrell/pyparted/issues/106

alan-barzilay:
@Torxed , since it has been a few months and nothing happened on the pyparted side, could we move on with the downgrade here? I'm unable to use archinstall in production do to this..
If they release newer versions in the future we can still upgrade the requirement, the downgrade would be just a band-aid to fix the pip installation (which, as far as I understand, is fundamentally broken now) while pyparted does not break their inertia and start moving

alan-barzilay:
@Torxed up haha

Torxed:
I'm not closing this one just yet, but I have submitted a PR that could potentially replace this one. Instead of downgrading, I'm exploring just running around the problem completely. Because.. frustrations are a bit high atm with this dependency :P

Torxed:
Ok review went quick, closing this.
But if we bump into issues we can revisit this one.

Issue #2211, opened 2023-11-11
Superfluous use of `partprobe`


codefiles:
Pertinent `partprobe` manual information: [1]

> **partprobe** is a program that informs the operating system kernel of partition table changes.

From pyparted, `Disk.commit()` also "informs the operating system [kernel] of the changes" to a partition table [2] therefore calling `partprobe` after `DeviceHandler._perform_partitioning()` is unnecessary.

https://github.com/archlinux/archinstall/blob/f6106324eb71905df9b832eb973a08a1f45fa5af/archinstall/lib/disk/device_handler.py#L327-L328

https://github.com/archlinux/archinstall/blob/f6106324eb71905df9b832eb973a08a1f45fa5af/archinstall/lib/disk/device_handler.py#L366-L367

https://github.com/archlinux/archinstall/blob/f6106324eb71905df9b832eb973a08a1f45fa5af/archinstall/lib/disk/device_handler.py#L523-L529

A `partprobe` that is performed after formatting the partitions is unnecessary because formatting the partitions does not change the partition table.

https://github.com/archlinux/archinstall/blob/f6106324eb71905df9b832eb973a08a1f45fa5af/archinstall/lib/disk/device_handler.py#L295-L305

https://github.com/archlinux/archinstall/blob/f6106324eb71905df9b832eb973a08a1f45fa5af/archinstall/lib/disk/device_handler.py#L377-L389

Encrypting a partition does not change the partition table.

https://github.com/archlinux/archinstall/blob/f6106324eb71905df9b832eb973a08a1f45fa5af/archinstall/lib/luks.py#L165

[1] https://man.archlinux.org/man/partprobe.8
[2] https://github.com/dcantrell/pyparted/blob/d5870bb7a5d512c17c92a4e4fe7e9a8e7c8c3528/src/parted/disk.py#L202-L205

svartkanin:
Hmm I remember there was a reason we put the partprobe in place after partitioning, I can't fully remember now but I think there was an issue without it not loading the lsblk info correctly

codefiles:
Regardless of whatever the intent was in the past, no evidence has been provided to justify any of the uses of partprobe outlined in the issue.

qlyoung:
This reintroduced #1759

codefiles:
[...]

PR #2288, opened 2023-12-05
Fix "Unable to determine new partition UUID" for NVMe devices

### LINKED ISSUE #2286: "Unable to determine new partition UUID" on NVMe storage
This problem, first described in https://github.com/archlinux/archinstall/issues/1759, appears to have been reintroduced likely by [this](https://github.com/archlinux/archinstall/pull/2212). I hit it today on 2.7.1.

I was able to work around the issue with this patch (effectively the same as the original patch that fixed #1759):
```diff
diff --git a/archinstall/lib/disk/device_handler.py b/archinstall/lib/disk/device_handler.py
index fcf52013..48689702 100644
--- a/archinstall/lib/disk/device_handler.py
+++ b/archinstall/lib/disk/device_handler.py
@@ -398,6 +398,7 @@ class DeviceHandler(object):

                        if lsblk_info.partn and lsblk_info.partuuid and lsblk_info.uuid:
                                break
+                       self.partprobe()

                if not lsblk_info:
                        debug(f'Unable to get partition information: {path}')
```

cc @Torxed @codefiles @svartkanin

qlyoung:
Fixes #2286

## PR Description:

This patch reintroduces part of a previous [patch](https://github.com/archlinux/archinstall/pull/1770) for this issue; calling partprobe appears to refresh the block device info seen by lsblk sufficiently for the installer to get the partition UUIDs of NVMe storage devices after creating new partitions.

## Tests and Checks

Tested as described [here](https://github.com/archlinux/archinstall/issues/2286#issuecomment-1841394425).

See also:
https://github.com/archlinux/archinstall/issues/1759
https://github.com/archlinux/archinstall/issues/1759#issuecomment-1839950427
https://github.com/archlinux/archinstall/issues/2211
https://github.com/archlinux/archinstall/pull/2212

### DIFF
--- a/archinstall/lib/disk/device_handler.py
+++ b/archinstall/lib/disk/device_handler.py
@@ -399,6 +399,8 @@ class DeviceHandler(object):
 			if lsblk_info.partn and lsblk_info.partuuid and lsblk_info.uuid:
 				break
 
+			self.partprobe()
+
 		if not lsblk_info:
 			debug(f'Unable to get partition information: {path}')
 			raise DiskError(f'Unable to get partition information: {path}')


Torxed:
My apologies, I thought my tested and pyparted solved those issues without having the sleep.

codefiles:
If anyone having this issue (I do not have hardware that triggers the issue) could run the commands I supplied here https://github.com/archlinux/archinstall/issues/2286#issuecomment-1841288880 and report back then it could be determined if a call to `partprobe` is necessary. I suspect that after running those commands any user experiencing this problem will see the PARTUUIDs in the output because libparted does the same thing as partprobe when it commits the partition table. If this is the case then there is an underlying issue in either archinstall or pyparted that should be resolved and not by just sensely adding a partprobe call to archinstall because "it works".

qlyoung:
User [reports](https://github.com/archlinux/archinstall/issues/2286#issuecomment-1845094905) this doesn't fix the issue and has provided the log file @codefiles needs, so I'll close this.

PR #2294, opened 2023-12-08
Fix partition table commit

### LINKED ISSUE #2293: Excessive partition table commits
The logic in the code below will cause the partition table to be written to disk and the operating system kernel to be informed on each iteration of the for loop. Instead this should be done once, after all changes have been made to the partition table layout.

https://github.com/archlinux/archinstall/blob/f107104103b74ac065a56bd94a0878f0d263e177/archinstall/lib/disk/device_handler.py#L534-L538

https://github.com/archlinux/archinstall/blob/f107104103b74ac065a56bd94a0878f0d263e177/archinstall/lib/disk/device_handler.py#L341-L342

https://github.com/archlinux/archinstall/blob/f107104103b74ac065a56bd94a0878f0d263e177/archinstall/lib/disk/device_handler.py#L380-L381

codefiles:
Fixes #2293

### DIFF
--- a/archinstall/lib/disk/device_handler.py
+++ b/archinstall/lib/disk/device_handler.py
@@ -322,7 +322,7 @@ class DeviceHandler(object):
 			part_mod.partuuid = lsblk_info.partuuid
 			part_mod.uuid = lsblk_info.uuid
 
-	def _perform_partitioning(
+	def _setup_partition(
 		self,
 		part_mod: PartitionModification,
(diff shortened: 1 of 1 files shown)


codefiles:
This will also save 3 seconds per partition added to the partition table layout since it removes a sleep of that duration.

svartkanin:
I think that's a great change!

Torxed (approved):

PR #2312, opened 2023-12-24
Improve EFI bootloader installation

HTV04:
## PR Description:

Some improvements to the installation of bootloaders on devices with UEFI:
* GRUB:
  * Adjusted the grub-install parameters
    * Disabled the `--bootloader-id` option, using the default `grub` instead
    * Disabled the `--debug` option, which prints a lot of unnecessary info and may enable a [debug](https://github.com/rhboot/grub2/blob/72b17102c157707acfffb01ae7694de6cf23eb5c/util/grub-install.c#L228) [flag](https://github.com/rhboot/grub2/blob/72b17102c157707acfffb01ae7694de6cf23eb5c/util/grub-install.c#L861)
    * Disabled the `--removable` option, which prevents an EFI entry from being created
      * The other functionality of this option, setting the destination EFI file to the standard `/EFI/BOOT/BOOTxxx.EFI/`, has been implemented manually
* Limine:
  * An EFI entry is now added via `efibootmgr`
* EFISTUB
  * Code clarity tweaks
* All bootloaders:
  * Like GRUB, all bootloaders will now attempt to install again upon failure to assist with buggy EFIs
    * After two failed attempts, the installer will attempt to use a fallback without EFI variables; if this fails, the installer will error out
  * Added a few debug messages for silent errors

## Tests and Checks

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -819,12 +819,29 @@ class Installer:
 			bootctl_options.append(f'--esp-path={efi_partition.mountpoint}')
 			bootctl_options.append(f'--boot-path={boot_partition.mountpoint}')
 
+		command = [
+			'/usr/bin/arch-chroot',
+			'bootctl',
+			'install',
+			*bootctl_options
+		]
+
 		# Install the boot loader
 		try:
-			SysCommand(f"/usr/bin/arch-chroot {self.target} bootctl {' '.join(bootctl_options)} install")
+			SysCommand(command, peek_output=True)
 		except SysCallError:
-			# Fallback, try creating the boot loader without touching the EFI variables
-			SysCommand(f"/usr/bin/arch-chroot {self.target} bootctl --no-variables {' '.join(bootctl_options)} install")
+			# Try normally one more time (helps buggy EFIs)
+			try:
+				SysCommand(command, peek_output=True)
+				debug('Installed systemd-boot after initial error.')
+			except SysCallError:
+				# Fallback, try creating the boot loader without EFI variables
+				command.append('--no-variables')
+				try:
+					SysCommand(command, peek_output=True)
+					debug('Installed systemd-boot without EFI variables.')
+				except SysCallError as err:
+					raise DiskError(f'Failed to install systemd-boot on {boot_partition.dev_path}: {err}')
 
 		# Ensure that the $BOOT/loader/ directory exists before we try to create files in it.
 		#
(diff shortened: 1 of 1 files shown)


vanillaSprinkles:
would it be possible to give the user a prompt, upon ultimate failure, to abort advancing on the entirety of the archinstall scripts, or to continue with a potentially failed grub-install? (or check for an archinstall flag to allow prompt/silent mode or something) - suggested grub-install errors as a my example in https://github.com/archlinux/archinstall/issues/2378

Torxed:
Awesome work, I'd like to marge this.
I'm wondering if we'll bring back this tho: https://github.com/archlinux/archinstall/pull/793

HTV04:
> Awesome work, I'd like to marge this. I'm wondering if we'll bring back this tho: #793

I believe that "fixed" the issue because it prevented grub-install from adding the EFI entry in the first place. Instead, the removable flag treats the destination as a removable drive, only adding a standard `BOOTX64.EFI` file.

Also, this PR might actually fix that issue due to automatically retrying the installation upon such an error.

HTV04:
> would it be possible to give the user a prompt, upon ultimate failure, to abort advancing on the entirety of the archinstall scripts, or to continue with a potentially failed grub-install? (or check for an archinstall flag to allow prompt/silent mode or something) - suggested grub-install errors as a my example in #2378
> 
> edit; i tested this on an ipxe'd bios-install and it has still failed (dropped it on top of the current installer.py, did not check for rebase'ing ect) ( it = https://github.com/archlinux/archinstall/blob/5b9574d3acbe5a106543e16d5cc02e1ed860d7b1/archinstall/lib/installer.py )

I could try, but it would make the installation process less "automatic." This might be in the scope of a future PR to handle common errors during installation.

Torxed:
I could never get a hold of hardware that could trigger the error.
I'm totally fine adding in this PR considering the try/except which should fallback to trying.

And now that we know about the issue, if it comes back we could add back this flag.

vanillaSprinkles:
im wondering now if adding different behaviors for each type of error would be useful? for example for my "fun" hardware the efibootmgr can fail but is not an issue due to having to teak the bios manually anyways; and skipping it by adding `--no-nvram`  to grub-install might be useful (something like that is where a user-prompt maybe okay vs auto-skip (or the never ending add-more-flags/inputs))

Torxed:
Those checks and tweaks would improve reliability if nothing else, and it's a godsend if we could get there. Because this and disk operations have always been a painpoint.

PR #2350, opened 2024-01-28
Fix 2307 - Custom mirrors

### LINKED ISSUE #2307: Need Help: How to Add Custom Mirror Server Url Correctly in Arch Install
I have created a pacman local cache according to [https://wiki.archlinux.org/title/Pacman/Tips_and_tricks#Read-only_cache](url). I use **Nginx** to host the cache. Then I add my custom server url (e.g. `Server = http://192.168.13.xx:8080/` )to the **/etc/pacman.d/mirrorlist** of some other Arch machine in the LAN. It works very well.

But when I try to add the custom mirror following: 
**Mirrors**
**Custom mirrors**
**Add a custom mirror**
**Enter name (leave blank to skip)**
**Enter url**
......

The server url and other info are added to the **pacman.conf**. 
For example:
```
[Home]
SigLevel = Optional TrustAll
Server = http://192.168.13.xx:8080
```

But shouldn't the mirror be added to **/etc/pacman.d/mirrorlist**?

From the code, I find the below function **add_custom_mirrors**:
https://github.com/archlinux/archinstall/blob/f107104103b74ac065a56bd94a0878f0d263e177/archinstall/lib/mirrors.py#L276
It operates on `/etc/pacman.conf`

However, another function **use_mirrors**:
https://github.com/archlinux/archinstall/blob/f107104103b74ac065a56bd94a0878f0d263e177/archinstall/lib/mirrors.py#L290C1-L290C1
operates on `/etc/pacman.d/mirrorlist`

Another **_add_custom_mirror** here:
https://github.com/archlinux/archinstall/blob/f107104103b74ac065a56bd94a0878f0d263e177/archinstall/lib/mirrors.py#L141

The code gives me the feeling that it is trying to add a custom repository, but the name from the menu says **Add a custom mirror**.

It is really confusing.

svartkanin:
This should fix #2307 

The custom mirror was only written to the `pacman.conf` on the target, with this change the server will also be added to the `/etc/pacman.d/mirrorlist`

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -318,27 +318,44 @@ class Installer:
 	def post_install_check(self, *args: str, **kwargs: str) -> List[str]:
 		return [step for step, flag in self.helper_flags.items() if flag is False]
 
-	def set_local_mirrors(self, mirror_config: MirrorConfiguration):
-		local_pacman_conf = Path('/etc/pacman.conf')
+	def set_mirrors(self, mirror_config: MirrorConfiguration, on_target: bool = False):
+		"""
+		Set the mirror configuration for the installation.
 
-		if mirror_config.custom_mirrors:
-			config = mirror_config.custom_mirrors_to_config()
+		:param mirror_config: The mirror configuration to use.
+		:type mirror_config: MirrorConfiguration
 
-			with local_pacman_conf.open('a') as fp:
-				fp.write(config)
+		:on_target: Whether to set the mirrors on the target system or the live system.
+		:param on_target: bool
+		"""
+		debug('Setting mirrors')
 
-	def set_target_mirrors(self, mirror_config: MirrorConfiguration):
 		for plugin in plugins.values():
 			if hasattr(plugin, 'on_mirrors'):
 				if result := plugin.on_mirrors(mirror_config):
 					mirror_config = result
 
-		target_mirror_conf = Path(f'{self.target}/etc/pacman.d/mirrorlist')
-		mirror_list_config = mirror_config.mirror_list_config()
+		if on_target:
+			local_pacman_conf = Path(f'{self.target}/etc/pacman.conf')
+			local_mirrorlist_conf = Path(f'{self.target}/etc/pacman.d/mirrorlist')
+		else:
+			local_pacman_conf = Path('/etc/pacman.conf')
+			local_mirrorlist_conf = Path('/etc/pacman.d/mirrorlist')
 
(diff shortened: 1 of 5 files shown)


Torxed (approved):

wswind:
I just need an option for me to disable reflector since I have already changed the mirrorlist file before using arch-install.
I will have to run `systemctl mask reflector` before using it now

Torxed:
Sadly reflector is part of the ISO and there's nothing the archinstall project can do, other than what we do more or less: https://github.com/archlinux/archinstall/blob/2a33d7cd974b57c0e69b0d4998e9bfe06af86b62/archinstall/lib/installer.py#L162

We wait for the service before we modify mirror definitions.
We could mask perhaps, but that would mean that we would interfere with the expected behavior of the ISO.

wswind:
Thanks for reply. 
In fact if I install arch manully, reflector should not be running automatically. It runs when I want to use it.

Things become weird using archinstall. 
Since I'm not trying to use reflector to refresh the mirrorlist file, I will not set the region option in archinstall mirror section. 
But reflector will be runing automatically even if I'm not touching anything about archinstall mirror option.

Check this post for the problem: https://www.reddit.com/r/archlinux/comments/19a132o/archinstall_stuck_in_waiting_for_automatic_mirror/

Torxed:
> In fact if I install arch manully, reflector should not be running automatically. It runs when I want to use it.

We don't install it to your system, if that's what you're thinking of.
And even if you do a manual install - reflector will be running for you in the ISO.. but usually pretty fast so you won't have time to notice.

So just to be clear, when you select a region in `archinstall` we don't use reflector.
We do `wget 'https://archlinux.org/mirrorlist/?country=<_REGION_>&protocol=https&ip_version=4&ip_version=6&use_mirror_status=on' -O - 2> /dev/null | sed '/#Server/s/^#//g'`.

The "stuck waiting" is usually caused by network issues, where `reflector.service` (which comes default turned ON in every Arch ISO) has not been able to complete a mirror sync.
We'll then wait for it to complete OR for you to turn it off. We don't use it, we just need the service to die or finish so we don't get a race condition.

wswind:
I see, this problem has noting to do with this project. 

Reflector would start on boot in the official iso.  When the network stuck, the mirrorlist file would be override later.
And archinstall prompts waiting for reflector, made me think that archinstall would start reflector service to refresh mirrorlist file, which is not true. 

So it's all my misunderstanding.

Thanks for your time.

PR #2381, opened 2024-03-06
remove plasma-wayland-session

ba-batz:
this package has been deprecated and the kde  install fails due to it

- This fix issue: 

## Tests and Checks

### DIFF
--- a/archinstall/default_profiles/desktops/kde.py
+++ b/archinstall/default_profiles/desktops/kde.py
@@ -19,7 +19,6 @@ class KdeProfile(XorgProfile):
 			"kwrite",
 			"dolphin",
 			"ark",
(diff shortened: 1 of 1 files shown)


pumpkintheproot:
why not change it to `plasma-workspace` instead? that would fix it as plasma-workspace is the replacement to `plasma-wayland-session`

xniksysx:
> why not change it to `plasma-workspace` instead? that would fix it as plasma-workspace is the replacement to `plasma-wayland-session`

plasma-workspace dep in plasma-desktop dep in with plasma-meta

dylanmtaylor:
I second the idea of replacing it with plasma-workspace

PR #2384, opened 2024-03-07
Update microcode handling

fortifiedhill:
## PR Description:

As of mkinitcpio v38, microcode is handled by a hook and inserted into the initrd by default. Therefore, we don't have to add microcode entries to bootloaders anymore. Additionally, the microcode option in .preset files has been deprecated in favor of the new microcode hook.

See: https://archlinux.org/news/mkinitcpio-hook-migration-and-early-microcode/

## Tests and Checks

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -83,7 +83,7 @@ class Installer:
 		# systemd, sd-vconsole and sd-encrypt will be replaced by udev, keymap and encrypt
 		# if HSM is not used to encrypt the root volume. Check mkinitcpio() function for that override.
 		self._hooks: List[str] = [
-			"base", "systemd", "autodetect", "keyboard",
+			"base", "systemd", "autodetect", "microcode", "keyboard",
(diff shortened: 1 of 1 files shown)


Torxed:
This is awesome to see, minimizing code - simpler life :D

Torxed:
We might need to do something about this too: https://github.com/archlinux/archinstall/blob/c549b50622dc652ab58038f6952a337082df6363/archinstall/lib/installer.py#L646-L650

fortifiedhill:
> We might need to do something about this too:
> 
> https://github.com/archlinux/archinstall/blob/c549b50622dc652ab58038f6952a337082df6363/archinstall/lib/installer.py#L646-L650

This code appears to handle adding the respective microcode package to the package install list, depending on the user's CPU. In a VM (after removing [this line](https://github.com/archlinux/archinstall/blob/c549b50622dc652ab58038f6952a337082df6363/archinstall/lib/installer.py#L599)), I've tested this by removing the linked code, and sure enough, the microcode does not install, so it appears that we need that code.

PR #2385, opened 2024-03-07
Update resize root partition ArchWiki link

fortifiedhill:
## PR Description:
This fixes the resize root partition link, so that it jumps to the right spot on the Arch Wiki.


Torxed:
Well spotted!

PR #2404, opened 2024-03-11
Fix Btrfs mount options

### LINKED ISSUE #1571: Bugs: Btrfs compression does not work when subvolumes are selected in the UI, btrfs subvolumes ignore "mount_options" specified in disk layout json file
Hi Torxed,

Btrfs compression does not work when subvolumes are selected in the user interface.  Another bug which is related to this is that btrfs subvolumes ignore "mount_options" specified in the disk layout json file.  

In order to replicate, just start archinstall and do the following:  Select "Wipe all selected drives and use a best-effort default partition layout". Choose yes for btrfs compression and btrfs subvolumes.  Install a minimal system.  After the installation is over, enter the command "cat /proc/mounts".  You will see that the mount options of the subvolumes do not include any compression options.  When you check "/mnt/archinstall/etc/fstab" file you will see that no compression options are defined in it.

There are some other things you can check that might help you to debug this issue and reveal more bugs.  Do the same steps that I mentioned above in the user interface but instead of installing the system, ask the UI to save the disk layouts to "user_disk_layout.json" file.  When you look at the json file, you will see that the file system has **"mount_options": ["compress=zstd"]** specified but the subvolumes have **"compress": false** specified.  Playing with this json file will reveal more bugs.  Edit the file and specify **"compress": true** for all subvolumes.  run "archinstall --disk_layout user_disk_layout.json" and install a minimal system again.  After the installation is over, enter the command "cat /proc/mounts".  You will see that the mount options of the subvolumes include **"compress=zlib:3"**.  When you check fstab you will see that it includes **both "compress=zstd" and "compress=zlib:3"** for all subvolumes.  **Wrong type of compression (zlib:3) was used when installing the system and wrong type of compression was specified in the fstab file (both zstd and zlib are specified at the same time).**

Please bear with me and lets do one more check to reveal another bug.  Edit the "user_disk_layout.json" file one more time and set **"mount_options": ["compress-force=zstd:15"]**.  We will again have **"compress": true** for all subvolumes.  Run archinstall with this new json file that has the custom compression we specified and install a minimum system.  After the installation is over check  "cat /proc/mounts" and fstab one more time.  You will see that the situation is the same.  /proc/mounts will show **"compress=zlib:3"** and fstab will show both "compress=zstd" and "compress=zlib:3".  **Custom mount_option (compression option) specified in the json file was completely ignored.**  

I am not a programmer but I searched the source code to see if I can help you to debug these issues.

It seems like in **line 19 of subvolume.py** file, you have the word "compress" hard coded as a mount option for subvolumes.  Btrfs uses zlib:3 compression by default when "compress" is specified without any options.  If you change that line to "compress=zstd" that will solve part of the problem.  Please also check if the user specified custom compression in the  mount_options field of the json file, otherwise a hard coded "compress=zstd" in that line will override the custom compression specified in the disk layout json file.

**Line 39 of btrfs_helpers.py** file might be responsible for options  "compress=zstd" and "compress=zlib:3" showing up in the fstab file at the same time.  In that line you check if "compress=zstd" is in the mount options and add it if it is not already there.  Since zlib compression was mistakenly used when installing the system (due to line 19 of subvolume.py file), compress=zstd gets added.  Line 39 might cause issues if the user specifies custom compression options such as "compress-force=zstd:15" or "compress=lzo" since they do not match the "compress=zstd:*" pattern.

I could not figure out why mount_options flag from the json file gets completely ignored for subvolumes.  It might be because that flag is completely missing from the subvolume.py file.

I also could not figure out why "compress": false is specified for all subvolumes even if the user selects compression in the user interface.  Something in the code overrides the user selection from true to false.  Currently it is not possible to install a compressed system with subvolumes by using the UI of archinstall.  We need to create a disk layout json file first and modify subvolume "compress" flags to true.  Even then, we have to settle with other bugs that I explained above.  :))

**On a separate note, none of these issues happen if the user does not use subvolumes.**  Compression works fine and defaults to "compress=zstd:3" in the fstab file and /proc/mounts when subvolumes are not specified in the user interface.  Specifying custom compression in the disk layout json file such as "mount_options": ["compress-force=zstd:15"] also work without any problems when subvolumes are not used.  Line 39 of btrfs_helpers.py file does not add an incorrect "compress=zstd" into the fstab file in addition to the "compress-force=zstd:15" that is already in there.  Normally I do not use subvolumes.  That's why I did not notice these bugs in the past.

codefiles:
Fixes #1571

In [btrfs(5)](https://man.archlinux.org/man/core/btrfs-progs/btrfs.5.en) under _MOUNT OPTIONS_ there is this note:

> Most mount options apply to the whole filesystem and only options in the first mounted subvolume will take effect. This is due to lack of implementation and may change in the future. This means that (for example) you can't set per-subvolume nodatacow, nodatasum, or compress using mount options. This should eventually be fixed, but it has proved to be difficult to implement correctly within the Linux VFS framework.

This pull request removes the ability to apply the compress and nodatacow mount options for individual subvolumes for this reason.

Selecting and toggling the nodatacow mount option for the file system has been added. Both the selection and toggle restrict compress and nodatacow from both being present since they conflict.

### DIFF
--- a/archinstall/lib/disk/device_handler.py
+++ b/archinstall/lib/disk/device_handler.py
@@ -437,9 +437,19 @@ class DeviceHandler(object):
 			if not luks_handler.mapper_dev:
 				raise DiskError('Failed to unlock luks device')
 
-			self.mount(luks_handler.mapper_dev, self._TMP_BTRFS_MOUNT, create_target_mountpoint=True)
+			self.mount(
+				luks_handler.mapper_dev,
+				self._TMP_BTRFS_MOUNT,
+				create_target_mountpoint=True,
+				options=part_mod.mount_options
+			)
 		else:
-			self.mount(part_mod.safe_dev_path, self._TMP_BTRFS_MOUNT, create_target_mountpoint=True)
+			self.mount(
+				part_mod.safe_dev_path,
+				self._TMP_BTRFS_MOUNT,
+				create_target_mountpoint=True,
+				options=part_mod.mount_options
+			)
 
 		for sub_vol in part_mod.btrfs_subvols:
 			debug(f'Creating subvolume: {sub_vol.name}')
@@ -450,18 +460,6 @@ class DeviceHandler(object):
 				subvol_path = self._TMP_BTRFS_MOUNT / sub_vol.name
 
 			SysCommand(f"btrfs subvolume create {subvol_path}")
-
-			if sub_vol.nodatacow:
-				try:
-					SysCommand(f'chattr +C {subvol_path}')
-				except SysCallError as err:
-					raise DiskError(f'Could not set nodatacow attribute at {subvol_path}: {err}')
-
-			if sub_vol.compress:
-				try:
-					SysCommand(f'chattr +c {subvol_path}')
-				except SysCallError as err:
-					raise DiskError(f'Could not set compress attribute at {subvol_path}: {err}')
 
 		if luks_handler is not None and luks_handler.mapper_dev is not None:
 			self.umount(luks_handler.mapper_dev)
--- a/archinstall/lib/disk/device_model.py
+++ b/archinstall/lib/disk/device_model.py
@@ -315,6 +315,11 @@ class Size:
 		return self._normalize() >= other._normalize()
 
 
+class BtrfsMountOption(Enum):
+	compress = 'compress=zstd'
+	nodatacow = 'nodatacow'
+
+
 @dataclass
 class _BtrfsSubvolumeInfo:
 	name: Path
@@ -458,8 +463,6 @@ class _DeviceInfo:
 class SubvolumeModification:
 	name: Path
 	mountpoint: Optional[Path] = None
-	compress: bool = False
-	nodatacow: bool = False
 
 	@classmethod
 	def from_existing_subvol_info(cls, info: _BtrfsSubvolumeInfo) -> SubvolumeModification:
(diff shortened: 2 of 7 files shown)


Torxed (approved):

svartkanin:
I just realized, it would be better to keep the mount options as types rather than strings
```
def select_mount_options() -> List[str]:
	prompt = str(_('Would you like to use compression or disable CoW?'))
	options = [str(_('Use compression')), str(_('Disable Copy-on-Write'))]
	choice = Menu(prompt, options, sort=False).run()

	if choice.type_ == MenuSelectionType.Selection:
		if choice.single_value == options[0]:
			return [BtrfsMountOption.compress.value]
		else:
			return [BtrfsMountOption.nodatacow.value]

	return []
```
If  they need to be distinguished later on we have to do string comparison which is unsafe and annoying

codefiles:
It is easier to follow along if you properly link to the code like this:

https://github.com/archlinux/archinstall/blob/df2884085dc06f71f2bb201a316828c23b6299dd/archinstall/lib/interactions/disk_conf.py#L218-L229

> I just realized, it would be better to keep the mount options as types rather than strings

Would making that change interfere with allowing the user to provide mount options in the configuration file or make this more complicated?

```
    "disk_config": {
        "device_modifications": [
            {
                ...
                "partitions": [
                    {
                        ...
                        "mount_options": [example_option],
                        ...
                    }
                ],
            }
```

> If they need to be distinguished later on we have to do string comparison which is unsafe and annoying

Could you provide an example of why they would need to be distinguished later? Seems like everything is already taken care of but I might be missing something.

svartkanin:
I just had another look and I missed that previously we were also storing plain strings, so please disregard my comment. If we wanted to change to this it would need to be something with a bigger picture

PR #2412, opened 2024-03-12
Add udev sync before lsblk that follows formatting

### LINKED ISSUE #2411: Synchronize with udev before lsblk when appropriate
The following is from the _DESCRIPTION_ section of [lsblk(8)](https://man.archlinux.org/man/lsblk.8).

> Note that **lsblk** might be executed in time when **udev** does not have all information about recently added or modified devices yet. In this case it is recommended to use **udevadm settle** before **lsblk** to synchronize with udev.

Accordingly, `udevadm settle` should be used before `lslbk` in the scenario below.

https://github.com/archlinux/archinstall/blob/613bec6562c4d67c4c93d443e6d7e93fdb41ba66/archinstall/lib/disk/device_handler.py#L310-L319

Utilizing synchronization with udev should supplant these retires and sleeps.

https://github.com/archlinux/archinstall/blob/613bec6562c4d67c4c93d443e6d7e93fdb41ba66/archinstall/lib/disk/device_handler.py#L390-L392

https://github.com/archlinux/archinstall/blob/613bec6562c4d67c4c93d443e6d7e93fdb41ba66/archinstall/lib/disk/device_model.py#L1113-L1128

codefiles:
Resolves #2411

### DIFF
--- a/archinstall/lib/disk/device_handler.py
+++ b/archinstall/lib/disk/device_handler.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import json
 import os
-import time
 import logging
 from pathlib import Path
 from typing import List, Dict, Any, Optional, TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


Torxed:
Awesome work, and nice finding.
This might have been the issue I've had issue describing and finding a solution to back when we had issues with partitions not being available in lsblk/partition listing?
Sounds like that's exactly what it's for at least, to let the kernel discover all the nodes properly.

codefiles:
When a device is added or modified udev is updated with information about the device via udev events. If lsblk is used before udev is finished processing the event queue then information may be missing. What `udevadm settle` does is pause until all the udev event are handled.

https://man.archlinux.org/man/udevadm.8

PR #2421, opened 2024-03-16
Rename "Kde" profile to the correct "KDE Plasma" / "Plasma"

C0rn3j:
## PR Description:

Just like shorthand for `Microsoft Office` is `Office`, shorthand for `KDE Plasma` is `Plasma`.

Rename files and code as such, capitalize existing code.

This also removes a dead misspelled profile from schema.json and fixes up the profile description there.

## Tests and Checks

### DIFF
--- a/archinstall/default_profiles/desktop.py
+++ b/archinstall/default_profiles/desktop.py
@@ -15,7 +15,7 @@ class DesktopProfile(Profile):
 		super().__init__(
 			'Desktop',
 			ProfileType.Desktop,
-			description=str(_('Provides a selection of desktop environments and tiling window managers, e.g. gnome, kde, sway')),
+			description=str(_('Provides a selection of desktop environments and tiling window managers, e.g. GNOME, Plasma, Sway')),
 			current_selection=current_selection,
 			support_greeter=True
 		)
--- a/archinstall/default_profiles/desktop.py
+++ b/archinstall/default_profiles/desktop.py
@@ -15,7 +15,7 @@ class DesktopProfile(Profile):
 		super().__init__(
 			'Desktop',
 			ProfileType.Desktop,
-			description=str(_('Provides a selection of desktop environments and tiling window managers, e.g. GNOME, Plasma, Sway')),
+			description=str(('Provides a selection of desktop environments and tiling window managers, e.g. GNOME, KDE Plasma, Sway')),
 			current_selection=current_selection,
 			support_greeter=True
 		)
--- a/archinstall/default_profiles/desktop.py
+++ b/archinstall/default_profiles/desktop.py
@@ -15,7 +15,7 @@ class DesktopProfile(Profile):
 		super().__init__(
 			'Desktop',
 			ProfileType.Desktop,
-			description=str(('Provides a selection of desktop environments and tiling window managers, e.g. GNOME, KDE Plasma, Sway')),
+			description=str(_('Provides a selection of desktop environments and tiling window managers, e.g. GNOME, KDE Plasma, Sway')),
 			current_selection=current_selection,
 			support_greeter=True
 		)
--- a/archinstall/default_profiles/desktops/plasma.py
+++ b/archinstall/default_profiles/desktops/plasma.py
@@ -6,10 +6,30 @@ from archinstall.default_profiles.xorg import XorgProfile
 if TYPE_CHECKING:
 	_: Any
 
-
 class PlasmaProfile(XorgProfile):
 	def __init__(self):
-		super().__init__('Plasma', ProfileType.DesktopEnv, description='')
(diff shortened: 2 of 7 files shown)


svartkanin:
I do understand the idea here, but I'm not sure if that is going to be confusing for people. I think generally people are more familiar with calling it KDE rather than Plasma.

Kubuntu uses both KDE and Pladma terms https://ubuntu.com/desktop/flavours

Fedora calls it KDE Plasma
https://fedoraproject.org/spins/

Can we go the same route here and call it "KDE Plasma"

C0rn3j:
> Kubuntu uses both KDE and Plasma terms https://ubuntu.com/desktop/flavours

They refer to KDE software in general as KDE, and the DE as Plasma, which is its short form.

> Fedora calls it KDE Plasma https://fedoraproject.org/spins/

That's the long form

> Can we go the same route here and call it "KDE Plasma"

Personally I prefer the short form so people stop using a name from 2007 and earlier, but both the "Plasma" the commit uses and "KDE Plasma" that you're suggesting is fine.

---

""description": "Specific profile to be installed based on the 'main' selection; these profiles are present in default_profiles/, use the file name of a profile without the extension to install it (case insensitive)","

Someone please verify that what I wrote here makes sense, I might have misunderstood how the script works. Is it using the file name, or actually what's defined in the file itself?

svartkanin:
Also you will need to test if this is backwards compatible by loading an old configuration with this change

C0rn3j:
> you will need to test if this is backwards compatible

It's not, it renames the profile, and the class.

People using the old wrong naming will need to target the new one.  

Good thing Plasma 6 is just released, so the few people that have custom archinstall configs may not be so upset about having to do so.

svartkanin:
For this to move forward I suggest to change the naming to "KDE Plasma" as it will align with other distributions naming convention. It's probably not the most accurate but less confusing. 
Also I don't think breaking backwards compability  for such a minor update is not a good way to handle it and it doesn't take much to make it compatible.

C0rn3j:
> as it will align with other distributions naming convention

Arch Linux wiki [mostly uses the shorthand in writing](https://wiki.archlinux.org/title/KDE).  
Not even KDE uses the full name on the [Plasma page](https://kde.org/plasma-desktop/).
Either the full name or the short form is fine, as it ceases being "KDE" when Plasma is meant to be referred to.

The profile name has to change, it currently is "Kde" which is wrong AND wrongly capitalized.  
But we do not need to do so in a breaking fashion, I've attempted to keep both right now.

This would keep both names, deprecating the old one, and the old one would be removed in an archinstall release or two.

archinstall/default_profiles/desktop.py
```diff
@@ -15,7 +15,7 @@ def __init__(self, current_selection: List[Profile] = []):
 		super().__init__(
 			'Desktop',
 			ProfileType.Desktop,
-			description=str(_('Provides a selection of desktop environments and tiling window managers, e.g. GNOME, Plasma, Sway')),
```
Torxed: Could we add the translation hook back in?

archinstall/default_profiles/desktops/plasma.py
```diff
@@ -0,0 +1,48 @@
...
+			"plasma-workspace",
+			"egl-wayland"
+		]
+
+	@property
+	def default_greeter_type(self) -> Optional[GreeterType]:
+		return GreeterType.Sddm
+
+# 2024-04-16 TODO deprecated Class with old naming, remove in a future version
+class KdeProfile(XorgProfile):
```
svartkanin, line 30: With backards compability I didn't mean to leave the old profile configuration around but to make sure that reading configurations with the old name will be parsed correctly. Leaving this config around will display 2 profiles now. 

The backwards compability sits in 

    def parse_profile_config(self, profile_config: Dict[str, Any]) -> Optional[Profile]:
        """
        Deserialize JSON configuration for profile
        """
        profile: Optional[Profile] = None


        # the order of these is important, we want to
        # load all the default_profiles from url and custom
        # so that we can then apply whatever was specified
        # in the main/detail sections
        if url_path := profile_config.get('path', None):
            self._url_path = url_path
            local_path = Path(url_path)


            if local_path.is_file():
                profiles = self._process_profile_file(local_path)
                self.remove_custom_profiles(profiles)
                self.add_custom_profiles(profiles)
            else:
                self._import_profile_from_url(url_path)


image

C0rn3j:
@svartkanin I wanted to test it first but got a bit of internet issues, so it was taking a long time to download the premade ISO I wished to use…  
I see you got there before me, thanks!

svartkanin:
It looked good from the code but I'll run through it today, thanks for the change!

C0rn3j:
image

Seems to not be duped anymore at least.

PR #2422, opened 2024-03-16
Locale generator fixes

C0rn3j:
## PR Description:

Unify tab usage instead of mixing tabs and spaces.

Write variables better to not erroneously use a wrong variable, also quote them to handle filenames with whitespace better.

Finally, exit on error instead of ignoring issues. 

Note that this breaks the current git master, as msgfmt has a fatal error in one of the language mutations.  
That's an easy fix, but it's something to fix in the language files, not in the script.

## Tests and Checks

### DIFF
--- a/archinstall/locales/locales_generator.sh
+++ b/archinstall/locales/locales_generator.sh
@@ -1,49 +1,48 @@
-#!/bin/bash
+#!/usr/bin/env bash
+set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


svartkanin:
I wish there was a code style that we can automatically apply 😓

svartkanin (approved):

PR #2439, opened 2024-03-28
Adds dosfstools to the PKGBUILD depends

digital-raven:
One of the tools needed by archinstall is mkfs.fatXX and this wasn't automatically installed.

- I don't know if this issue is present on the github. I'll open one if required; the contributing guideline did not impose a requirement for an existing issue.

## PR Description:
This adds dosfstools, which provides the mkfs.fat binary which is used in device_handler.py

## Tests and Checks
When I run `makepkg -s` on my system with or without this change, I receive an error that my build did not pass the "validity check". I don't know if this has bearing on the integrity of this change, so I left the following checklist for what I believe would be a good test of this change.

- [ ] Using archinstall as installed using `pacman -S archinstall` and try to format any partition using a code path that relies on mkfs.fat, archinstall will error because this tool is not installed.
- [ ] When I ran `makepkg -si`, dosfstools was installed on my system along with archinstall.

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -15,6 +15,7 @@ depends=(
   'btrfs-progs'
   'coreutils'
   'cryptsetup'
+  'dosfstools'
   'e2fsprogs'
   'glibc'
   'kbd'


digital-raven:
Hey guys, I'm still watching this! It's not a big deal for me personally as I was able to just install it and that's a decent workaround, but this is something that I think a newer user would be intimidated by were they to use archinstall on its own. I think it's a short and sweet improvement.

digital-raven:
It's missing when running archinstall on another system. If dosfstools is not already present, then `pacman -S archinstall` will not install it, and it's a necessary tool for archinstall's operation.

svartkanin:
I think in that guess it's reasonable to include it in the configuration

svartkanin (approved):

Torxed:
IT doesn't hurt to add it, if it's used to bootstrap a VM for instance from an existing system.
But we've gotten lucky that the [ISO releng has it shipped automatically](https://gitlab.archlinux.org/archlinux/archiso/-/blob/4b709bcd5fc5e426ba75f72e1b80e59922d7c175/configs/releng/packages.x86_64#L24).

svartkanin:
I assumed it should as it's a requirement for it to work on the iso, but I didn't check to make sure. Thanks @Torxed for confirming

PR #2461, opened 2024-04-15
profiles: Add nftables and Firewalld service

EliasTheGrandMasterOfMistakes:
This include a nftables and firewalld firewall packages for improve security on daily usage

## PR Description:
Add Firewalld service and nftables to provide a firewall.
This fixes unsecure installation without firewall 

## Tests and Checks
i think is not necessary a installation test

### DIFF
--- a/archinstall/default_profiles/desktop.py
+++ b/archinstall/default_profiles/desktop.py
@@ -24,6 +24,8 @@ class DesktopProfile(Profile):
 	def packages(self) -> List[str]:
 		return [
 			'nano',
+			'nftables',
+			'firewalld',
 			'vim',
 			'openssh',
 			'htop',


Torxed:
I think we'd like a separate profile for this, rather than inject it on all desktop profiles (even tho, I agree, this would raise the minimum level security of desktops).

But consider users selecting the `server` profile.
We'd want to increase security there too.

So ideally, what we would like is a separate "security" profile - and allow multi-select on profiles so that people can select:
```
[x] Desktop
[x] Security
[ ] Server
```

(Just as a crude example)

dylanmtaylor:
I think this is too opinionated. There are multiple competing ways to set this up, like ufw for instance. Firewalls could be a post-installation configuration item for the end user.

Torxed:
I agree, for now I'll close this in the current form it's implemented.

PR #2464, opened 2024-04-16
Don't explicitly install plasma-workspace

Samq64:
- This fix issue: none

## PR Description:

Removes `plasma-workspace` from the explicitly installed packages in the Plasma profile.

#2386 replaced `plasma-wayland-session` with `plasma-workspace` but it was already a dependency of multiple packages in the `plasma-meta` group, it just also bundles the Wayland session desktop file now.

## Tests and Checks


Samq64:
I tested in Virtualbox and the plasma-workspace package is installed but the Wayland session hangs on startup.

svartkanin:
> I tested in Virtualbox and the plasma-workspace package is installed but the Wayland session hangs on startup.

So I take it that this package is required to be installed?

PR #2485, opened 2024-04-30
Fix 2479 - remove unnecessary partiton info field

### LINKED ISSUE #2479: Manual Partitioning failed on b8052c5
Manual Partitioning failed on b8052c5.
[install.log](https://github.com/archlinux/archinstall/files/15139144/install.log)

There is no problem on v2.8.0.

svartkanin:
This fixes #2479 

The `PartitionModification` class contained a `part_info` field that is not compatible for `deepcopy` and therefore the error. It seems that this field isn't actually used anywhere and therefore should be save to be removed


svartkanin:
@Torxed could approve this so we can merge it otherwise manual partitioning is broken on master

Torxed (approved):

PR #2495, opened 2024-05-10
Use the current layout instead of the us layout

ngn13:
- This fix issue: none

## PR Description:

By default, the installer selects the "us" layout. With this simple enhancement, it selects the current layout (acquired by parsing `localectl status` output) instead of "us".

So if the user runs `loadkeys` or `localectl set-keymap` before running `archinstall`, they won't have to set their keymap again.

## Tests and Checks
- [x] I have tested the code! (on my local machine, there is no link to the ISO)<br>

### DIFF
--- a/archinstall/lib/locale/locale_menu.py
+++ b/archinstall/lib/locale/locale_menu.py
@@ -1,7 +1,7 @@
 from dataclasses import dataclass
 from typing import Dict, Any, TYPE_CHECKING, Optional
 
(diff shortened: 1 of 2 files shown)


Torxed (approved):
Looks good to me, and I don't see a reason for not setting the default to whatever the user has provided before launching archinstall.

svartkanin:
looks good thanks for that!

PR #2500, opened 2024-05-13
Adding a test "framework" to run real world machine tests

Torxed:
Because we have nothing, I'd like to have something.
And this does something.

It runs `qemu` with test parameters.
It has a test class that can take `qemu` serial output and manage it.

It can cycle through the menu's and handle triggers and monitor things.
It's just one test case, it needs more - and it needs to be able to run multiple tests most importantly.


Torxed:
Working on pushing the .github/.gitlab* runner files for this, which will automatically publish the `.cast` recording of the output.

Torxed:
This has been surpassed by another project, I'll push it to a public repo after I cleaned up some debug variables that might leak a bit of information that doesn't need to be publicly available.

PR #2501, opened 2024-05-14
Patching test-cases to use local runner

Torxed:
Before merging:

- [ ]: It needs to be running isolated on the runner machine's hardware
- [ ]: Limit action to trusted users (!developers on github repo)
- [ ]: Be able to post results back to PR's


Torxed:
Apparently the runner will run under `archlinux/archinstall` when it's a PR targeting that repo url. I need to create a local branch PR first. Apologies.

PR #2508, opened 2024-05-15
Refactor suggest_single_disk_layout()

codefiles:
- Simplify home partition conditional and use less indentation
- Simplify root length conditional by de-duplicating repeated home partition conditional checks
- Only call `process_root_partition_size()` when it is required
- Rename `process_root_partition_size()` parameter `available_space` -> `total_size` for increased clarity
- Consolidate accounting of end alignment buffer

Bonus: -3 lines

### DIFF
--- a/archinstall/lib/interactions/disk_conf.py
+++ b/archinstall/lib/interactions/disk_conf.py
@@ -257,9 +257,9 @@ def select_mount_options() -> List[str]:
 	return []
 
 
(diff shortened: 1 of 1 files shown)


codefiles:
I have more refactors like this for various functions. I am unsure if I should submit pull request for them. Would they possibly be considered too minor and unhelpful?

codefiles:
Great. Thanks.

PR #2512, opened 2024-05-17
Refactor format()

codefiles:


archinstall/lib/disk/device_handler.py
```diff
@@ -198,49 +198,34 @@ def format(
...
+			case FilesystemType.Btrfs | FilesystemType.F2fs | FilesystemType.Xfs:
+				# Force overwrite
+				options.append('-f')
+			case FilesystemType.Ext2 | FilesystemType.Ext3 | FilesystemType.Ext4:
+				# Force create
+				options.append('-F')
+			case FilesystemType.Fat16 | FilesystemType.Fat32:
+				mkfs_type = 'fat'
+				# Set FAT size
+				options.extend(('-F', fs_type.value.removeprefix(mkfs_type)))
```
svartkanin, line 214: Looking at this in some time down the road it will take some time to figure out what this magic is. Maybe keeping them in a simpler (but slightly more verbose) form may be more readable 
```
			case FilesystemType.Fat16:
			    mkfs_type = 'fat'
				options.append('-F16')
			case FilesystemType.Fat32:
			    mkfs_type = 'fat'
				options.append('-F32')
``` 
or maybe 
```
			case FilesystemType.Fat16 | FilesystemType.Fat32:
				mkfs_type = 'fat'
				size = '32' if fs_type == FilesystemType.Fat32 else '16'
				options.append(f'-F{size}')
```

thoughts?

codefiles: There is a comment that says "Set Fat size". If I was not sure what this did I would look to [mkfs.fat.8](https://man.archlinux.org/man/mkfs.fat.8) and see that the `-F` option can take the bit size of the file allocation tables. Then I would look up the value for `FilesystemType.Fat16` and `FilesystemType.Fat32` or use the interpreter.

```
>>> from archinstall.lib.disk.device_model import FilesystemType
>>> FilesystemType.Fat16
<FilesystemType.Fat16: 'fat16'>
>>> FilesystemType.Fat32
<FilesystemType.Fat32: 'fat32'>
>>> def disenchant(fs_type):
... 	mkfs_type = 'fat'
... 	return fs_type.value.removeprefix(mkfs_type)
... 
>>> disenchant(FilesystemType.Fat16)
'16'
>>> disenchant(FilesystemType.Fat32)
'32'
```

Here is what I dislike about the suggested alternatives. The first contains two cases with mostly duplication of code because they are nearly identical cases. The second uses a case with an or to then separate them with a conditional, that feels redundant.

What about another comment and a variable?

```python
			case FilesystemType.Fat16 | FilesystemType.Fat32:
				mkfs_type = 'fat'
				# Determine FAT size of 16 or 32 using the value of fs_type
				fat_size = fs_type.value.removeprefix(mkfs_type)
				# Set FAT size
				options.extend(('-F', fat_size))
```

svartkanin: Let's leave it

svartkanin (approved):

PR #2517, opened 2024-05-21
nvidia-open double fix with nvidia-open-dkms

RageGen:
Fixing package duplication when installing nvidia-open driver, which is unnecessary because one of the packages is already present. In this case, nvidia-open and nvidia-open-dkms duplicate each other, which can lead to misunderstandings among inexperienced users when installing through a script and using a kernel other than generic linux.

### DIFF
--- a/archinstall/lib/hardware.py
+++ b/archinstall/lib/hardware.py
@@ -47,7 +47,6 @@ class GfxPackage(Enum):
 	LibvaMesaDriver = 'libva-mesa-driver'
 	Mesa = "mesa"
 	NvidiaDkms = 'nvidia-dkms'
-	NvidiaOpen = 'nvidia-open'
 	NvidiaOpenDkms = 'nvidia-open-dkms'
 	VulkanIntel = 'vulkan-intel'
 	VulkanRadeon = 'vulkan-radeon'
--- a/archinstall/lib/hardware.py
+++ b/archinstall/lib/hardware.py
@@ -116,7 +116,6 @@ class GfxDriver(Enum):
 				]
 			case GfxDriver.NvidiaOpenKernel:
 				packages += [
-					GfxPackage.NvidiaOpen,
 					GfxPackage.Dkms,
 					GfxPackage.NvidiaOpenDkms
 				]


svartkanin:
This change will break the code as the removed package is still being assigned https://github.com/archlinux/archinstall/pull/2517/files#diff-d39c5771ecacdcd8a8216c41e300b2c0df140c7a871e99f2b29994a85b78cc86L120

According to https://wiki.archlinux.org/title/NVIDIA 
> Alternatively for the [Turing (NV160/TUXXX)](https://nouveau.freedesktop.org/CodeNames.html#NV160) series or newer the [nvidia-open](https://archlinux.org/packages/?name=nvidia-open) package may be installed for open source kernel modules on the [linux](https://archlinux.org/packages/?name=linux) kernel (On other kernels [nvidia-open-dkms](https://archlinux.org/packages/?name=nvidia-open-dkms) must be used).

So this may require some more detailed handling of which one to install on Turing series

RageGen:
@svartkanin 
I didn’t understand why it violates anything, if in fact, by analogy, everything is done in the same way as with the nvidia package, which is proprietary. After all, only nvidia-dkms is used there, regardless of the kernel.

RageGen:
@svartkanin 
I corrected the file, if you meant it. You still haven't justified why in the case of nvidia-open does not work in the same way as in the case of a regular nvidia package. You refer to an article on the wiki, but why? If even when using the script, the pacman itself discards nvidia-open and leaves only its dkms weight.

offensive-vk (approved):

RageGen:
@offensive-vk Thank you for your attention to my request!

Torxed:
So the understanding is:
 * If `linux` kernel is chosen, `nvidia-open` can be used for supported cards but `nvidia-open-dkms` can also be used - but adds unnecessary overhead during updates
 * If `linux-<variation>` kernel is chosen, `nvidia-open-dkms` must be used

So by removing `nvidia-open` we would have to rely on `nvidia-open-dkms`, is this assumption not correct? if it is, then it will add delays and complexity during updates where `nvidia-open` would have been a better fit?

RageGen:
@Torxed So regardless of the kernel, both packages are installed and in the end only dkms remains, I checked. Then what’s the point? And well, even if as you say, then with the usual nvidia driver, it would be worth doing exactly the same.

offensive-vk:
@RageGen, Look briefly:

- Using nvidia-open with the standard Linux kernel is generally faster and simpler for updates.
- Using nvidia-open-dkms is necessary for kernel variations but introduces additional overhead during updates..
- If we remove nvidia-open, you will have to use nvidia-open-dkms, which will add delays and complexity during updates..

_Your assumption is mostly correct , but relying solely on nvidia-open-dkms can complicate and slow down the update process compared to using nvidia-open with the standard kernel._

***

Lets not hurry on changing this, as it might impact on production scale.
so please , Let @Torxed do his work and relax 🤗.  
_Have a nice day._

RageGen:
Okay, thank you very much, sorry. I thought I’d help. @offensive-vk

PR #2532, opened 2024-06-08
Do not overwrite the contents of mkinitcpio.conf

ventureoo:
- This fix issue:   N/A

## PR Description:

After  installation is complete, user may want to edit `/etc/mkinitcpio.conf` at will and may encounter a mismatch between the default configuration examples and what it has after installation since archinstall does not just change some values, but completely overwrites contents of the file removing all comments and other options. This PR fixes this by editing only parameters specified in code like hooks/modules/etc.

## Tests and Checks
- [x] I have tested the code!<br> (did not test install process, but tested the operation of a separate mkinitcpio function on default mkinitcpio.conf file).

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -698,10 +698,11 @@ class Installer:
 				if plugin.on_mkinitcpio(self):
 					return True
 
-		with open(f'{self.target}/etc/mkinitcpio.conf', 'w') as mkinit:
-			mkinit.write(f"MODULES=({' '.join(self._modules)})\n")
-			mkinit.write(f"BINARIES=({' '.join(self._binaries)})\n")
-			mkinit.write(f"FILES=({' '.join(self._files)})\n")
+		with open(f'{self.target}/etc/mkinitcpio.conf', 'r+') as mkinit:
+			content = mkinit.read()
+			content = re.sub("\nMODULES=(.*)", f"\nMODULES=({' '.join(self._modules)})", content)
+			content = re.sub("\nBINARIES=(.*)", f"\nBINARIES=({' '.join(self._binaries)})", content)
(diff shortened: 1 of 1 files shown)


Torxed:
I'm not sure the `re.sub()` is working as intended:
```python
>>> import re
>>> content = '\nMODULES=(vfio_pci vfio vfio_iommu_type1 btrfs nvidia nvidia_modeset nvidia_uvm nvidia_drm)'
>>> _modules = ['testmodule', 'anothermod']
>>> re.sub("\nMODULES=(.*)", f"\nMODULES=({' '.join(_modules)})", content)
'\nMODULES=(testmodule anothermod)'
```

ventureoo:
> I'm not sure the `re.sub()` is working as intended:
> 
> ```python
> >>> import re
> >>> content = '\nMODULES=(vfio_pci vfio vfio_iommu_type1 btrfs nvidia nvidia_modeset nvidia_uvm nvidia_drm)'
> >>> _modules = ['testmodule', 'anothermod']
> >>> re.sub("\nMODULES=(.*)", f"\nMODULES=({' '.join(_modules)})", content)
> '\nMODULES=(testmodule anothermod)'
> ```

Well, yeah, `re.sub()` overwrites the `MODULES`, `FILES`, `BINARIES` lines, but leaves rest of the file intact like the comments and other options, which is the purpose of my PR. Also `MODULES`, `FILES`, `BINARIES` are empty by default in` /etc/mkinitcpio.conf`,  so it's not a big problem as I think.

Torxed:
Apologies, I might have been a bit clear and nitpicking.
My intention was that if we're going to improve the code to be non-invasive we should perhaps include users modification to `MODULES=()` *(assuming they've built their own ISO that has adaptions they need)*.

But coming back to this PR, I realized that it's a step in the right direction and nothing needs to be perfect to be added - as long as it's a step in the right direction.

Issue #2559, opened 2024-07-01
Something happen when i try install arch


ImStoleYourCheese:
i cant undestand

```
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
File "/usr/bin/archinstall", line 5, in <module>
fron archinstall import run_as_a_module
File "/usr/lib/python3.12/site-packages/archinstall/_init_.py", line 11, in <module> from lib import disk
File "/usr/lib/python3.12/site-packages/archinstall/lib/disk/_init__.py", line 1, in <module> fron device handler import device handler, disk_layouts
File "/usr/lib/python3.12/site-packages/archinstall/lib/disk/device_handler.py", line 649, in <module> device handler = DeviceHandler()
File "/usr/lib/python3.12/site-packages/archinstall/lib/disk/device_handler.py", line 38, in _init_
self.load devices() File "/usr/lib/python3.12/site-packages/archinstall/lib/disk/device_handler.py", line 74, in load devices Isbik_info = get_isbik_info(partition.path)
File "/usr/lib/python3.12/site-packages/archinstall/lib/disk/device_nodel.py", line 1136, in get Isblk info if infos: fetch_isbik_info(deu_path):
File "/usr/lib/python3.12/site-packages/archinstall/lib/disk/device nodel.py", line 1121, in _fetch_lsblk_info
raise DiskError(f'Failed to read disk "(deu_path)" with isbik') archinstall.lib.exceptions. DiskError: Failed to read disk "/dev/sdb1" with Isbik
```
![image](https://github.com/archlinux/archinstall/assets/130311040/2a74cd9b-3afe-4f93-a62f-b39acf455b91)

svartkanin:
Can you please provide the log file /var/log/archinstall/install.log

arlsdk:
Could you try to reformat your disks with **cfdisk** and try again? Had the same issue.

codefiles:
Duplicate of #1993

Issue #2536, opened 2024-06-12
archinstall --config $yourconfigfile --creds &yourcredsfile dont work


solisinvictum:
Like title.

To reproduce:

1) download latest iso from archlinux.org
2) start archinstall
3) configure everything from top to bottom
4) save both configs to /root/ for example
5) quit archinstall
6) start archinstall with:  archinstall --config /root/user_configuration.json --creds /root/user_credentials.json
7) see error output:

```
Traceback (most recent call last):
  File "/usr/lib/python3.12/site-packages/archinstall/lib/disk/device_model.py", line 1113, in _fetch_lsblk_info
    result = SysCommand(cmd).decode()
             ^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/archinstall/lib/general.py", line 365, in __init__
    self.create_session()
  File "/usr/lib/python3.12/site-packages/archinstall/lib/general.py", line 414, in create_session
    with SysCommandWorker(
  File "/usr/lib/python3.12/site-packages/archinstall/lib/general.py", line 197, in __exit__
    raise SysCallError(
archinstall.lib.exceptions.SysCallError: ['/usr/bin/lsblk', '--json', '--bytes', '--output', '+name,path,pkname,size,log-sec,pttype,ptuuid,rota,tran,partn,partuuid,parttype,uuid,fstype,fsver,fsavail,fsuse%,type,mountpoint,mountpoints,fsroots', '/dev/sda1'] exited with abnormal exit code [32]: lsblk: /dev/sda1: not a block device
{
   "blockdevices": [

   ]
}


During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/bin/archinstall", line 5, in <module>
    from archinstall import run_as_a_module
  File "/usr/lib/python3.12/site-packages/archinstall/__init__.py", line 11, in <module>
    from .lib import disk
  File "/usr/lib/python3.12/site-packages/archinstall/lib/disk/__init__.py", line 1, in <module>
    from .device_handler import device_handler, disk_layouts
  File "/usr/lib/python3.12/site-packages/archinstall/lib/disk/device_handler.py", line 649, in <module>
    device_handler = DeviceHandler()
                     ^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/archinstall/lib/disk/device_handler.py", line 38, in __init__
    self.load_devices()
[...]

PR #2565, opened 2024-07-06
Fix load_devices() lsblk

### LINKED ISSUE #1993: Archinstall encounters Error when starting - "/dev/sda3: not a block device"
`exited with abnormal exit code [32]: lsblk: /dev/sda3: not a block device`

When I start archinstall, it exists after a few seconds with the mentioned error.
I tried the newest package from the Arch Repos, and I tried the scripts from the master branch, but regardless this error occurrs.
IMG_20230806_222932

Running lsblk shows that sda doesn't have a 3rd partition which puzzles me.
sda isn't even relevant I  what I was going to do. I want to install Arch on a different device. Is there any way to make archinstall ignore this or is temporarily removing the drive the only solution?

codefiles:
Fixes #1993

- Get all lsblk info and search it rather than making several calls to lsblk
- Add udev sync before lsblk call

### DIFF
--- a/archinstall/lib/disk/device_handler.py
+++ b/archinstall/lib/disk/device_handler.py
@@ -18,7 +18,7 @@ from .device_model import (
 	DeviceModification, PartitionModification,
 	BDevice, _DeviceInfo, _PartitionInfo,
 	FilesystemType, Unit, PartitionTable,
-	ModificationStatus, get_lsblk_info, LsblkInfo,
+	ModificationStatus, get_lsblk_info, find_lsblk_info, LsblkInfo,
 	_BtrfsSubvolumeInfo, get_all_lsblk_info, DiskEncryption, LvmVolumeGroup, LvmVolume, Size, LvmGroupInfo,
 	SectorSize, LvmVolumeInfo, LvmPVInfo, SubvolumeModification, BtrfsMountOption
 )
@@ -47,6 +47,7 @@ class DeviceHandler(object):
 	def load_devices(self):
 		block_devices = {}
 
+		all_lsblk_info = get_all_lsblk_info()
 		devices = getAllDevices()
 
 		try:
@@ -58,7 +59,11 @@ class DeviceHandler(object):
 			debug(f'Failed to get loop devices: {err}')
 
 		for device in devices:
-			dev_lsblk_info = get_lsblk_info(device.path)
+			dev_lsblk_info = find_lsblk_info(device.path, all_lsblk_info)
+
+			if not dev_lsblk_info:
+				debug(f'Device lsblk info not found: {device.path}')
+				continue
 
 			if dev_lsblk_info.type == 'rom':
 				continue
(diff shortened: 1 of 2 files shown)


svartkanin (approved):

svartkanin:
Thanks for the fix, @codefiles the PR is in draft, is there anything left?

codefiles:
After submitting I noticed that when a user has an existing btrfs partition and it is not mounted then `get_btrfs_info()` will call `mount()`. I think this patch will just move the bug to the lsblk call in `mount()` given that case.

Maybe using `udevadm settle` to synchronize with udev is needed after `newDisk()` and `freshDisk()`. Take a look at the output of `udevadm monitor --udev` while executing those within the python interpreter. You see udev events, right? I think they could be causing a race condition with the lsblk calls that follow. What do you think?

svartkanin:
I wonder if enforcing a restart after a successful installation or at least preventing to run it multiple installations might be worth exploring?

codefiles:
This does not fix the issue it was intended to but I am fine with it being merged.

@svartkanin, in regards to the linked issue, I do not agree with your conclusions or suggestions but thanks.

codefiles:
I figured out how to reproduce the bug. I added an additional commit and with that the bug should be fixed.

Torxed (approved):

svartkanin (approved):

Torxed:
Nicely done, was it tricky to reproduce the bug?

codefiles:
From https://github.com/archlinux/archinstall/issues/2536#issue-2349753573 the traceback shows this bug happens during an lsblk call in `load_devices()`:

```
  File "/usr/lib/python3.12/site-packages/archinstall/__init__.py", line 11, in <module>
    from .lib import disk
  File "/usr/lib/python3.12/site-packages/archinstall/lib/disk/__init__.py", line 1, in <module>
  File "/usr/lib/python3.12/site-packages/archinstall/lib/disk/device_handler.py", line 649, in <module>
                     ^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/archinstall/lib/disk/device_handler.py", line 38, in __init__
    self.load_devices()
  File "/usr/lib/python3.12/site-packages/archinstall/lib/disk/device_handler.py", line 74, in load_devices
    lsblk_info = get_lsblk_info(partition.path)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/archinstall/lib/disk/device_model.py", line 1136, in get_lsblk_info
    if infos := _fetch_lsblk_info(dev_path):
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/archinstall/lib/disk/device_model.py", line 1121, in _fetch_lsblk_info
    raise DiskError(f'Failed to read disk "{dev_path}" with lsblk')
archinstall.lib.exceptions.DiskError: Failed to read disk "/dev/sda1" with lsblk
```

The function `load_devices()` is called when the application launches and I had seen the bug before from my own testing. I put the function call in an infinite loop.

```patch
diff --git a/archinstall/lib/disk/device_handler.py b/archinstall/lib/disk/device_handler.py
index 8d38dbc8..6db3d94b 100644
--- a/archinstall/lib/disk/device_handler.py
+++ b/archinstall/lib/disk/device_handler.py
@@ -38,7 +38,8 @@ class DeviceHandler(object):

        def __init__(self):
                self._devices: Dict[Path, BDevice] = {}
-               self.load_devices()
+               while True:
+                       self.load_devices()

        @property
        def devices(self) -> List[BDevice]:
```

With this patch I am able to produce the bug with the commit previous to this merge by executing archinstall and letting this loop run until the error is raised. It takes a relatively short amount of time till the error is raised.

I then used that same process to check this pull request with a small addition. I used `tail -f /var/archinstall/log/install.log | grep -E '(Device|Partition) lsblk info not found'` in another tty while the application was running. This was to verify that no devices or partitions were skipped due to missing lsblk infomation since that is a feature of this pull request. Without the last commit to this pull request there were skips but after adding it there were no skips and no error raised.

#2536 and #2559 can be closed as well.

Issue #2539, opened 2024-06-19
/home not encrypted when it should be


onekum:
The past few times I've installed with ArchInstall, I designated my `/home` partition to be encrypted, but the script doesn't appear to follow that instruction. My system boots completely as if I had not designated encryption at all. There's not one prompt for a password to decrypt the system; everything boots just fine but unencrypted. Here is my [install log](https://github.com/user-attachments/files/15894077/install.log).

svartkanin:
According to the logs the encryption was selected for /home and also setup correctly. 
Can you share the output for `lsblk -J -O` from the installed system

onekum:
Upon booting into the desktop, it appears that the `/home` partition is recognized as encrypted through the file explorer, but it is permenantly unlocked, defeating the purpose of the encryption altogether. I rebooted countless times to see if a decryption prompt would ever show up on boot, but it doesn't; despite this, I can access the `/home` directory as if it wasn't encrypted.

Here's the output of `lsblk -J -O`:
```
{
   "blockdevices": [
      {
         "alignment": 0,
         "id-link": "wwn-0x5000c500aa9af812",
         "id": "0x5000c500aa9af812",
         "disc-aln": 0,
         "dax": false,
         "disc-gran": "4K",
         "disk-seq": 1,
         "disc-max": "0B",
         "disc-zero": false,
         "fsavail": null,
         "fsroots": [
             null
         ],
         "fssize": null,
         "fstype": null,
         "fsused": null,
         "fsuse%": null,
         "fsver": null,
         "group": "disk",
         "hctl": "0:0:0:0",
         "hotplug": false,
         "kname": "sda",
         "label": null,
         "log-sec": 512,
         "maj:min": "8:0",
         "maj": "8",
         "min": "0",
         "min-io": 4096,
         "mode": "brw-rw----",
         "model": "ST2000LM007-1R8174",
         "mq": "  1",
         "name": "sda",
         "opt-io": 0,
         "owner": "root",
         "partflags": null,
         "partlabel": null,
         "partn": null,
         "parttype": null,
         "parttypename": null,
         "partuuid": null,
         "path": "/dev/sda",
         "phy-sec": 4096,
         "pkname": null,
         "pttype": "gpt",
         "ptuuid": "273c1279-e312-4b61-9e34-ef41d8e01d4f",
         "ra": 128,
         "rand": true,
         "rev": "SDM2",
         "rm": false,
         "ro": false,
         "rota": true,
[...]

PR #2575, opened 2024-07-12
Handling encryption logic when only one non-root partition is encrypted (such as /home)

Torxed:
- This fix issue: #2539 

## PR Description:

This should make single-non-root-encrypted-partitions be unlocked via `/etc/crypttab` instead of via keyfile, as the keyfile would be unprotected.

## Tests and Checks

### DIFF
--- a/archinstall/lib/disk/device_model.py
+++ b/archinstall/lib/disk/device_model.py
@@ -17,7 +17,7 @@ from parted import Disk, Geometry, Partition
 
 from ..exceptions import DiskError, SysCallError
 from ..general import SysCommand
-from ..output import debug, error
+from ..output import debug, error, warn
 from ..storage import storage
 
 if TYPE_CHECKING:
@@ -1204,10 +1204,21 @@ class DiskEncryption:
 			raise ValueError('LuksOnLvm encryption require LMV volumes to be defined')
 
 	def should_generate_encryption_file(self, dev: PartitionModification | LvmVolume) -> bool:
-		if isinstance(dev, PartitionModification):
+		warn("Deprecation warning: should_generate_encryption_file() has been renamed to should_generate_encryption_keyfile() and should_generate_encryption_file will be removed.")
+		return self.should_generate_encryption_keyfile(dev)
+
+	def should_generate_encryption_keyfile(self, dev: PartitionModification | LvmVolume) -> bool:
+		# If all partitions being encrypted, are non-root file systems
+		# we will not generate key-files as their safety can not be guaranteed.
+		if all([part.mountpoint != Path('/') for part in self.partitions]):
+			return False
+		# If the device mountpoint is anything but root,
+		# we want to create a keyfile for unlocking
+		elif isinstance(dev, PartitionModification):
 			return dev in self.partitions and dev.mountpoint != Path('/')
 		elif isinstance(dev, LvmVolume):
 			return dev in self.lvm_volumes and dev.mountpoint != Path('/')
+
 		return False
 
 	def json(self) -> Dict[str, Any]:
--- a/archinstall/lib/luks.py
+++ b/archinstall/lib/luks.py
@@ -191,6 +191,10 @@ class Luks2:
 
 		self._mapper_dev = None
 
+	def create_crypttab(self, target_path: Path, override: bool = False):
+		crypttab_path = target_path / 'etc/crypttab'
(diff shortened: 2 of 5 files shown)


archinstall/lib/disk/device_model.py
```diff
@@ -1204,10 +1204,21 @@ def __post_init__(self):
...
 
 	def should_generate_encryption_file(self, dev: PartitionModification | LvmVolume) -> bool:
-		if isinstance(dev, PartitionModification):
+		warn("Deprecation warning: should_generate_encryption_file() has been renamed to should_generate_encryption_keyfile() and should_generate_encryption_file will be removed.")
+		return self.should_generate_encryption_keyfile(dev)
+
+	def should_generate_encryption_keyfile(self, dev: PartitionModification | LvmVolume) -> bool:
+		# If all partitions being encrypted, are non-root file systems
+		# we will not generate key-files as their safety can not be guaranteed.
+		if all([part.mountpoint != Path('/') for part in self.partitions]):
```
svartkanin, line 1213: just FYI there's a `part.is_root()` :)

svartkanin: This check doesn't feel right here...on the caller side this is run for each partition now, but we can probably short circuit it 

```
	def generate_keyfile_for_dev(self, dev: PartitionModification | LvmVolume) -> bool:
		# If all partitions being encrypted, are non-root file systems
		# we will not generate key-files as their safety can not be guaranteed.
		if isinstance(dev, PartitionModification):
			return dev in self.partitions and dev.mountpoint != Path('/')
		elif isinstance(dev, LvmVolume):
			return dev in self.lvm_volumes and dev.mountpoint != Path('/')

		return False

	def generate_keyfiles(self) -> bool:
		# If all partitions being encrypted, are non-root file systems
		# we will not generate key-files as their safety can not be guaranteed.
		if all([part.mountpoint != Path('/') for part in self.partitions]):
			return True
		return False
```

And on the caller

```
	def _setup_luks_unlock_partitions(self) -> None:
		if not self._disk_encryption.generate_keyfiles():
			return
		...
```

What do you think?

archinstall/lib/luks.py
```diff
@@ -239,13 +248,35 @@ def _add_key(self, key_file: Path):
...
 			opt = ','.join(options)
 			uuid = self._get_luks_uuid()
-			row = f"{self.mapper_name} UUID={uuid} {key_file} {opt}\n"
-			crypttab.write(row)
+			new_row = f"{self.mapper_name} UUID={uuid} {key_file or 'none'} {opt}\n"
+
+			for index, existing_row in enumerate(existing_data):
+				if uuid in existing_row:
+					if override is False:
+						print(f"Found {uuid} in {existing_row}")
```
svartkanin, line 266: Can we change this  to `debug(...)` rather as it's not really relevant for the user

archinstall/lib/luks.py
```diff
@@ -239,13 +248,35 @@ def _add_key(self, key_file: Path):
...
+						return True
```
svartkanin, line 267: Is this return here right? If override is passed into the function should this terminate ?

archinstall/lib/luks.py
```diff
@@ -239,13 +248,35 @@ def _add_key(self, key_file: Path):
...
+						existing_data.pop(index)
+
+			# Make sure the last entry ends with \n
+			# before inserting our new entry at the end
+			if not existing_data[-1][-1:] == '\n':
+				existing_data.append('\n')
+
+			existing_data.append(new_row)
+
+			print(f"Writing {existing_data}")
```
svartkanin, line 278: This will show up during the installation for the user probably not very useful, `debug(...)` probably better here

archinstall/lib/luks.py
Torxed: Of course, it's still a draft and used print when executing blocks of code in isolated `test.py` outside of archinstall :) but good reminder!

archinstall/lib/luks.py
Torxed: It's checking if we **don't** want to override - and it found that we already have added the data - so it's happy and returns "done" :)

archinstall/lib/disk/device_model.py
Torxed: > just FYI there's a `part.is_root()` :)

Silly me just copied the code from two rows down, but now that you mention it I do remember seeing that function, I'll change!

PR #2588, opened 2024-07-23
Upgrade mypy 1.11

svartkanin:


archinstall/lib/profile/profiles_handler.py
```diff
@@ -101,7 +101,7 @@ def parse_profile_config(self, profile_config: Dict[str, Any]) -> Optional[Profi
 		if not profile:
 			return None
 
-		valid_sub_profiles: List[Profile] = []
+		valid_sub_profiles: List[TProfile] = []  # type: ignore
```
Torxed, line 104: Out of curiosity, why can't we use `Profile` directly?
Is it because MyPy wants a `TypeVar` if it's not a built-in?

svartkanin: @Torxed yeah that change didn't actually fix it therefore the ugly `type: ignore` at the end. 
The original code with `Profile` showed the error 
```
archinstall/lib/profile/profiles_handler.py:125: error: Incompatible types in assignment (expression has type "list[Profile]", variable has type "list[TProfile]")  [assignment]
```

so I updated the type to `TProfile` but that ended up 
```
archinstall/lib/profile/profiles_handler.py:104: error: Type variable "archinstall.default_profiles.profile.TProfile" is unbound  [valid-type]
```

I'm a bit confused about this error as `TProfile` is defined with a bound 

```
TProfile = TypeVar('TProfile', bound='Profile')
```

if anyone has any idea what the problem is that would be great :)

correctmost: Can `TProfile` be removed entirely from the codebase?

mypy didn't emit any errors when I tried replacing all instances of `TProfile` with `Profile`, but I'm not sure if there was another reason for `TProfile` existing.

svartkanin: Nicely spotted, yeah that type is actually redundant! I updated the PR and removed it

Torxed (approved):

Issue #2598, opened 2024-07-30
Bundle an offline copy of mirrorlist


oldherl:
### describe the request

Currently archinstall gets the mirrorlist from https://archlinux.org/mirrorlist/?protocol=https , but in some cases, the user might be blocked by archlinux.org due to attackers in the neighbourhood, and gets an HTTP 403 error for that request.
![image](https://github.com/user-attachments/assets/0d2f0a17-b446-4ad6-bf9e-a81edc63f561)

Note that the user is not an attacker, but unfortunately blocked because of attackers in the same /24 or bigger subnet.

In that case, if archinstall could provide an offline copy of mirrorlist, the user could choose a mirror from that and continue the installation, as long as the chosen mirror doesn't block them.

Torxed:
It would be a frozen mirror list that may or may not be accurate at the time of installation. But for sure, technically this would be an easy thing to implement.

In the meantime, we've adjusted some proxy configs and you should be able to access archlinux.org now :)

oldherl:
Yes it would be a frozen list, but most mirrors in it should continue be usable within a month, which is the release cycle of archiso.

christian-heusel:
Is `pacman-mirrorlist` not coming preinstalled with the ISO? 🤔

Torxed:
True, that package might be there.
We're using the `archlinux.org` call to check for network connectivity to some extent.
And also while filtering/grabbing mirrors based on user selection.

lunelovehearn:
+1 this. I think archinstall should also offer an offline, local repository to install KDE plasma, gnome, xfce and other 3 popular WMs for this case, and when the user has an unstable internet or can't access the internet at all. then, the script and the local repo gets updated monthly.

PR #2599, opened 2024-07-30
Use `/mirrors/status/json/` instead of `/mirrorlist/` to get mirror lists

Torxed:
## PR Description:

This is more of a performance tweak, as the `/json/` endpoint is cached in the backend and machine readable, whereas the ASCII human version of mirror status is not.

 * Improves mirror-list grab latency
 * Introduces `pydantic` as a dependency
 * Creates a MirrorStatus model *(with latency and speed performance checks)*
 * Emulates `reflector.service` behavior after mirror selection *(Sort order: `score` -> `download speed`)*
 * Fixes mirror selection overwriting mirrorlist, instead of appending it after mirrors *(making them never being used essentially)*
 * Implemented a dependency-free `ping()` function. We're currently not using `mirror.latency` but there's a possibility we could in the future if we wanted to.

Future improvements could be that we could use `pacman-mirrorlist` package as source, convert that to JSON that we could consume, solving potential issues such as [#2598](https://github.com/archlinux/archinstall/issues/2598).

## Tests and Checks

### DIFF
--- a/archinstall/lib/models/mirrors.py
+++ b/archinstall/lib/models/mirrors.py
@@ -1,6 +1,5 @@
 import datetime
 import pydantic
-import socket
 import urllib.parse
 import urllib.request
 from typing import (
@@ -23,20 +22,19 @@ class MirrorStatusEntryV3(pydantic.BaseModel):
 	ipv6 :bool
 	details :str
 	delay :int|None = None
-	last_sync :datetime.datetime|None=None
+	last_sync :datetime.datetime|None = None
 	duration_avg :float|None = None
 	duration_stddev :float|None = None
 	completion_pct :float|None = None
 	score :int|None = None
 	_latency :float|None = None
 	_speed :float|None = None
+	_hostname :str|None = None
 
 	@property
 	def speed(self):
 		if self._speed is None:
-			parsed_uri = urllib.parse.urlparse(self.url)
-			hostname, *port = parsed_uri.netloc.split(':', 1)
-			info(f"Checking download speed of {hostname} by getting {self.url}core/os/x86_64/core.db")
+			info(f"Checking download speed of {self._hostname} by getting {self.url}core/os/x86_64/core.db")
 			req = urllib.request.Request(url=f"{self.url}core/os/x86_64/core.db")
 			with urllib.request.urlopen(req, None, 5) as handle, DownloadTimer(timeout=5) as timer:
 				size = len(handle.read())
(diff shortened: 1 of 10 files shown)


svartkanin:
This is great, wgen doing the migration to the new menu I was wondering why the mirror menu takes so long to open

Torxed:
Yeah it's a bit faster, and also we get more useful data from this.
I have one more commit before I'm satisfied, I thought I was done but now realized I have to filter a bit more.

I'm currently sorting the mirrors based on `score` which is effectively how healthy they are. But in my final commit I'll sort on multiple values, and by default it will be:
`speed` -> `score`. The mirror-list should then be as fast as possible, as well as the most healthy it can be :)

Torxed:
I will have to tinker a bit.
Because there's no way for the backend to know what speed the users source of origin will have.

So we have to adopt:
 * Do a quick [rankmirrors](https://gitlab.archlinux.org/pacman/pacman-contrib/-/blob/master/src/rankmirrors.sh.in?ref_type=heads) ping check to get early indications
 * Then do what [reflector](https://xyne.dev/projects/reflector/) does with bandwidth testing *after* we select mirrors to not bandwidth test every mirror possible, only the ones the user needs/wants

This should give us a speedy UI since ping is pretty fast.
While giving the best installation time by performing bandwidth test against the specific selected mirror-list.

And if the user opt-out and just use what `reflector.service` already created on bootup, we'll do nothing of course.

Torxed:
So I have a method of sorting now, I just need to trigger it after region(s) have been selected to avoid:
screenshot

Torxed:
Can't recall if there was a purpose behind doing `a` rather than `w` on the `/etc/pacman.d/mirrorlist`.
But currently when setting mirrors through the menu - they get appended after the default ISO mirrorlist - making the selection a bit redundant.

So I made a small change to #2350 which is `a` got replaced by `w` when saving the mirrorlist.
I left the `/etc/pacman.conf` as `a` still, as it probably only adds custom mirrors? which should go at the end hehe.

Torxed:
I think I'm satisfied, comments would be welcome.

screenshot

archinstall/lib/models/mirrors.py
```diff
@@ -0,0 +1,91 @@
...
+	duration_stddev :float|None = None
+	completion_pct :float|None = None
+	score :int|None = None
+	_latency :float|None = None
+	_speed :float|None = None
+	_hostname :str|None = None
+	_port :int|None = None
+
+	@property
+	def speed(self):
```
svartkanin, line 36: nit: return value type missing

archinstall/lib/models/mirrors.py
```diff
@@ -0,0 +1,91 @@
...
+		if self._speed is None:
+			info(f"Checking download speed of {self._hostname}[{self.score}] by fetching: {self.url}core/os/x86_64/core.db")
```
svartkanin, line 38: `info` won't be shown to the user, so maybe just use `debug`

archinstall/lib/models/mirrors.py
Torxed: Good catch! Fixed :D

archinstall/lib/models/mirrors.py
Torxed: I was thinking if the speed checks are slow, it's nice that they get printed during installation with `info()`:
screenshot

It's shown between selecting the region - and returning to the menu, but just briefly.
But it does disrupt the look and feel of the menu system for sure, so we want to skip it?

svartkanin (approved):

PR #2609, opened 2024-08-06
add xwaylandvideobridge for plasma

alexsch01:
- This fix issue: 

## PR Description:

Fedora is shipping xwaylandvideobridge to KDE Plasma users by default

Debian plans to do the same
- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1077654
- "This is a dependency of Plasma 6 and will be team maintained by the Qt/KDE team."

## Tests and Checks

### DIFF
--- a/archinstall/default_profiles/desktops/plasma.py
+++ b/archinstall/default_profiles/desktops/plasma.py
@@ -19,7 +19,8 @@ class PlasmaProfile(XorgProfile):
 			"dolphin",
 			"ark",
 			"plasma-workspace",
-			"egl-wayland"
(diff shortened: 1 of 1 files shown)


Torxed:
What's the consensus here?
It feels like it's a tool that purposefully breaks the separation design of wayland?
And is perhaps a opt-in feature if you need screen sharing - rather than something we should install by default?

I don't use wayland and I'll probably enter retirement age before X11 stops working, so I don't know much about the inner workings of wayland and what tools users expect. So feel free to update me a bit more unless someone else chimes in with good comments here.

PR #2615, opened 2024-08-17
added cosmic for a desktop

killertofus:
- This fix issue:  #2358
## PR Description:

added cosmic for a desktop
## Tests and Checks
- [x] I have tested the code!<br>

Issue #2358, opened 2024-02-01
Add `COSMIC` as a desktop option


yarons:
https://github.com/pop-os/cosmic-epoch

https://wiki.archlinux.org/title/COSMIC

Written in Rust, looks pretty promising.

omentic:
COSMIS is pre-alpha. Once a stable (or anything but alpha) release is made, it'll get packaged pretty fast and we can add it as an option. For now, it's a non-starter, cause that's probably several years away.

killertofus:
it's out now https://archlinux.org/groups/x86_64/cosmic/

Torxed:
Thank you @killertofus for the PR, I requested some minor changes - after which I can merge and it will be joining the next release :)

Torxed:
> it's out now https://archlinux.org/groups/x86_64/cosmic/

I made the mistake of not checking the fine print on the [`COSMIC` GitHub](https://github.com/pop-os/cosmic-epoch/blob/8ddb4bbe6aa5347e5326cc146dea738111a30c9b/README.md?plain=1#L3) which states: 

> incomplete **alpha**

I will push a patch to hide this DE behind `--advanced` so that we don't cause unintended bug reports during next months ISO build. And I'll put a monitoring on their GitHub for a stable release and remove the advanced flag once it's out.

Torxed:
They released alpha 2 yesterday. Still not out of the woods yet :)

MBA380:
> They released alpha 2 yesterday. Still not out of the woods yet :)

It's in beta now, out of the woods yet? (:

ghost:
@Torxed it's stable now

Torxed:
> [@Torxed](https://github.com/Torxed) it's stable now

Thanks for the ping!

Torxed:
Not sure if they consider it stable, or the [README](https://github.com/pop-os/cosmic-epoch/blob/39c24b0b4eb75f02d37196306b5612770cfe3c04/README.md?plain=1#L3) is just lagging behind:

> Currently in **beta** for Epoch 1 (the first release), which is nearly feature-complete and in the final bug fixing stage.

If this is True, we might be better off waiting still?

bittin:
> Not sure if they consider it stable, or the [README](https://github.com/pop-os/cosmic-epoch/blob/39c24b0b4eb75f02d37196306b5612770cfe3c04/README.md?plain=1#L3) is just lagging behind:
> 
[...]

PR #2619, opened 2024-08-20
Added a advanced=bool flag to Profile() class and enabled it on the cosmic-epoch profile

Torxed:
This is to be able to hide certain profiles behind `--advanced`.
We also enabled it on the `cosmic-epoch` due to it being pre-release alpha still with a note on the upstream repo:

> incomplete **alpha**

This complements issue: #2358 and thus #2615

### DIFF
--- a/archinstall/default_profiles/desktops/cosmic.py
+++ b/archinstall/default_profiles/desktops/cosmic.py
@@ -8,7 +8,7 @@ if TYPE_CHECKING:
 
 class CosmicProfile(XorgProfile):
 	def __init__(self):
-		super().__init__('cosmic-epoch', ProfileType.DesktopEnv, description='')
+		super().__init__('cosmic-epoch', ProfileType.DesktopEnv, description='', advanced=True)
 
 	@property
 	def packages(self) -> List[str]:
--- a/archinstall/default_profiles/profile.py
+++ b/archinstall/default_profiles/profile.py
@@ -34,7 +34,10 @@ class GreeterType(Enum):
 	Sddm = 'sddm'
 	Gdm = 'gdm'
 	Ly = 'ly'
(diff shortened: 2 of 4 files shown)


archinstall/default_profiles/profile.py
```diff
@@ -138,16 +141,16 @@ def is_top_level_profile(self) -> bool:
 		return self.profile_type in top_levels
 
 	def is_desktop_profile(self) -> bool:
-		return self.profile_type == ProfileType.Desktop
+		return self.profile_type == ProfileType.Desktop if self.advanced is False or storage['arguments'].get('advanced', False) is True else False
```
svartkanin, line 144: nit: maybe put this evaluation in a function to make it more readable?

archinstall/default_profiles/desktops/cosmic.py
```diff
@@ -8,7 +8,7 @@
 
 class CosmicProfile(XorgProfile):
 	def __init__(self):
-		super().__init__('cosmic-epoch', ProfileType.DesktopEnv, description='')
+		super().__init__('cosmic-epoch', ProfileType.DesktopEnv, description='', advanced=True)
```
svartkanin, line 11: There was also a new greeter option added in archinstall/default_profiles/profile.py does that also need to be behind the flag?

Torxed: Ah good catch, I think that should be behind the flag too. I'll update the pr with this and a fix for the other comment.

Torxed:
Should be good to go now, tested and verified that greeter isn't visible in other DE's and the desktop profile itself isn't visible either without `--advanced`.

svartkanin (approved):

PR #2623, opened 2024-08-22
Add Catalan translation

Odyssey:
Adds catalan (ca) localization to archinstall script


Torxed (approved):
Thank you for this addition!

PR #2640, opened 2024-08-27
Fix various mypy errors in archinstall/lib/

correctmost:
## PR Description:

This PR fixes various mypy errors in `archinstall/lib/`.  The remaining errors are more complex and can be handled separately.

## Tests and Checks

### DIFF
--- a/archinstall/lib/boot.py
+++ b/archinstall/lib/boot.py
@@ -94,7 +94,7 @@ class Boot:
 
 		return self.session.is_alive()
 
-	def SysCommand(self, cmd: list, *args, **kwargs) -> SysCommand:
+	def SysCommand(self, cmd: list[str], *args, **kwargs) -> SysCommand:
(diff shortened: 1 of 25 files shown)


Torxed (approved):

svartkanin (approved):

svartkanin:
@correctmost do you mind adding any files/directories that are fully fixed to the mypy.ini file as includes, that way they are ensured to remain clean as the check is run on PRs

svartkanin:
@correctmost the mypy.ini that currently exists is an addition to the config in the toml as it's basically strictly checking everything. Not sure the toml can be removed or replaced before all files are compliant with the strict validation

Torxed:
I think we'll need it until the entire project is compliant.
Which might take a while, but we really appreciate the contributions to get close to that goal.
It will speed up error checks a lot once the coverage is complete.

PR #2648, opened 2024-08-28
Add retry for mirror speedtest and confirm the speed not to be None.

suica-me:
- This fix issue: 

## PR Description:

1.  Add 3 times retry for speedtest
```Python
_speedtest_limit:int = 3
_retry = 0
if self._speedtest_limit < 1:
    self._speedtest_limit = 1
while _retry < self._speedtest_limit and self._speed is None:
```
2. If the spped is None, we can not sort the mirror, so if the speedtest failed, set the speed to 0.
```Python
if self._speed is None:
    debug(f"    failed to test speed for {self._speedtest_limit} time(s), set to 0")
    self._speed = 0
```

## Tests and Checks

### DIFF
--- a/archinstall/lib/models/mirrors.py
+++ b/archinstall/lib/models/mirrors.py
@@ -32,28 +32,38 @@ class MirrorStatusEntryV3(pydantic.BaseModel):
 	_speed :float|None = None
 	_hostname :str|None = None
 	_port :int|None = None
+	_speedtest_limit:int = 3
 
 	@property
 	def speed(self) -> float|None:
 		if self._speed is None:
-			info(f"Checking download speed of {self._hostname}[{self.score}] by fetching: {self.url}core/os/x86_64/core.db")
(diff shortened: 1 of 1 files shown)


Torxed:
My only real concern is that this increases testing times quite a bit for marginal benefits.
If we could split this PR into two PR's:

1. Replacing `None` with `0` on the individual exceptions handlers
2. A retry PR that we can improve upon to come to a nice way of retrying things

suica-me:
You are right, retry maybe cost too much time, I will close this and make a new one.
Thank you!

Torxed:
I appreciate that!
I think re-try is a good idea, but we'll have to make it conditional.
For instance "Invalid certificate" should not cause a retry.
But partial read could, because the connection might just have dropped temporarily.

Looking forward to the new PR's (I can push a new release today for the `None -> 0` fix if you have time).

Issue #2646, opened 2024-08-28
Connection reset by peer in mirror test will cause script crash


suica-me:
### Which ISO version are you using?

N/A (archinstall 2.8.4-1)

### The installation log

```json
Hardware model detected: QEMU Standard PC (Q35 + ICH9, 2009); UEFI mode: True
Processor model detected: QEMU Virtual CPU version 2.5+
Memory statistics: 3505344 available out of 3972776 total installed
Virtualization detected: kvm; is VM: True
Graphics devices detected: dict_keys(['Device 1234:1111 (rev 02)'])
Disk states before installing: [
----------------------------------------------------------------------------------
                                 --DLETED--
                                 Some disk info
----------------------------------------------------------------------------------
]
Checking version...
Failed determine pacman version: ['/usr/bin/pacman', '-Qu', 'archinstall'] exited with abnormal exit code [1]: [?25l[?25h
    score: 2
Loaded mirror mirror.aarnet.edu.au with current score of 2
    score: 4
Loaded mirror mirror.aarnet.edu.au with current score of 4
    score: 3
----------------------------------------------------------------------------------
                                 --DLETED--
                                 Load mirror
----------------------------------------------------------------------------------
Loaded mirror mirrors.reitarovskyi.tech with current score of 3
    score: 3
Loaded mirror mirrors.reitarovskyi.tech with current score of 3
    score: 13
Loaded mirror mirror.syxpi.fr with current score of 13
    score: 13
Loaded mirror mirror.syxpi.fr with current score of 13
Checking download speed of mirrors.163.com[9] by fetching: http://mirrors.163.com/archlinux/core/os/x86_64/core.db
    speed: 439726.55680225976 (0.41MiB/s)
Checking download speed of mirrors.ustc.edu.cn[2] by fetching: http://mirrors.ustc.edu.cn/archlinux/core/os/x86_64/core.db
    speed: 2457018.5586694996 (2.34MiB/s)
----------------------------------------------------------------------------------
                                 --DLETED--
                                 Mirror speed test
----------------------------------------------------------------------------------
Checking download speed of mirror.nyist.edu.cn[6] by fetching: https://mirror.nyist.edu.cn/archlinux/core/os/x86_64/core.db
    speed: 1406425.7519304594 (1.34MiB/s)
Checking download speed of mirrors.qlu.edu.cn[11] by fetching: https://mirrors.qlu.edu.cn/archlinux/core/os/x86_64/core.db
[...]

PR #2650, opened 2024-08-28
Use `0` instead of `None` when the speedtest fails.

suica-me:
- This fix issue:
#2646 

## PR Description:
https://github.com/archlinux/archinstall/blob/master/archinstall/lib/mirrors.py#L290
If the speed is None, sorted will throw "TypeError: '<' not supported between instances of 'NoneType' and 'float'"
So use `0` instead of `None` when the speedtest fails.

## Tests and Checks
- [ ] I have tested the code!


Torxed:
Awesome, thanks for the quick submission!

PR #2641, opened 2024-08-27
Fix most mypy errors in archinstall/scripts/, docs/, and the __init__ file

correctmost:
## PR Description:

This PR fixes more of the simple mypy errors in the codebase.  There are still ~150 left.

## Tests and Checks
- [ ] I have tested the code!<br>

PR #2651, opened 2024-08-28
Expand mypy checks to include more files and more checks

correctmost:
## PR Description:

This PR depends on https://github.com/archlinux/archinstall/pull/2641. (I will rebase when 2641 is merged and then move this PR out of draft status.)

This PR makes the mypy checks more strict for more files.  It also moves the `mypy.ini` configuration to `pyproject.toml`.

## Tests and Checks


svartkanin:
I need to have a try with this PR, I'm a bit unsure if this is keeping the existing validations for files

svartkanin:
Nice, looks like it does still validate everything :) thanks for the great update

svartkanin (approved):

PR #1681, opened 2023-03-16
Adding evaluation of Ruff

Torxed:
This is just a small evaluation of a code formatting tool called [Ruff](https://github.com/charliermarsh/ruff)

PR #2654, opened 2024-08-28
Add ruff linting support

correctmost:
## PR Description:

This PR adds ruff to the CI checks and pre-commit checks.

Only linting is enabled, no auto-formatting is performed. (#1681 is related work.)

More rules can be enabled in the future (after the relevant code is fixed):
- https://docs.astral.sh/ruff/rules/

## Tests and Checks

### DIFF
--- a/.github/workflows/ruff.yaml
+++ b/.github/workflows/ruff.yaml
@@ -1,0 +1,12 @@
+on: [ push, pull_request ]
+name: ruff linting
+jobs:
+    ruff:
+        runs-on: ubuntu-latest
+        container:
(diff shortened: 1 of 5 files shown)


svartkanin:
@correctmost is the idea to replace flake8 eventually?

I'm okay with a better (faster) linter and it's a drop in replacement for flake8 as well. I'll let @Torxed have the final say on it :)

Torxed (approved):

Torxed:
Since it's not being destructive and only linting, I say lets go for it :)

Issue #2625, opened 2024-08-23
crash when selecting profile


madfresh:
### Which ISO version are you using?

2024-08-01

### The installation log

```json
Hardware model detected: To Be Filled By O.E.M. To Be Filled By O.E.M.; UEFI mode: True
Processor model detected: Intel(R) Core(TM) i5-4690 CPU @ 3.50GHz
Memory statistics: 14668592 available out of 16051668 total installed
Could not detect virtual system: ['/usr/bin/systemd-detect-virt'] exited with abnormal exit code [1]: none

System is not running in a VM: ['/usr/bin/systemd-detect-virt'] exited with abnormal exit code [1]: none

Virtualization detected: None; is VM: False
Graphics devices detected: dict_keys(['Intel Corporation Xeon E3-1200 v3/4th Gen Core Processor Integrated Graphics Controller (rev 06)'])
Disk states before installing: [
    {
        "children": [],
        "fsavail": 0,
        "fsroots": [
            "."
        ],
        "fstype": "squashfs",
        "fsuse_percentage": "100%",
        "fsver": "4.0",
        "log_sec": 512,
        "mountpoint": "/run/archiso/airootfs",
        "mountpoints": [
            "/run/archiso/airootfs"
        ],
        "name": "loop0",
        "partn": null,
        "parttype": null,
        "partuuid": null,
        "path": "/dev/loop0",
        "pkname": null,
        "pttype": null,
        "ptuuid": null,
        "rota": false,
        "size": "801 MiB",
        "tran": null,
        "type": "loop",
        "uuid": null
    },
    {
        "children": [
            {
                "children": [],
                "fsavail": null,
                "fsroots": [],
                "fstype": "ext2",
                "fsuse_percentage": null,
                "fsver": "1.0",
                "log_sec": 512,
                "mountpoint": null,
                "mountpoints": [],
                "name": "sda1",
                "partn": 1,
                "parttype": "0x83",
                "partuuid": "880c70ca-01",
                "path": "/dev/sda1",
                "pkname": "sda",
                "pttype": "dos",
                "ptuuid": "880c70ca",
                "rota": false,
                "size": "487 MiB",
                "tran": null,
                "type": "part",
                "uuid": "db581bad-d6ca-4b9a-abd0-be2e746587fd"
            },
            {
                "children": [],
                "fsavail": null,
                "fsroots": [],
                "fstype": null,
                "fsuse_percentage": null,
                "fsver": null,
                "log_sec": 512,
[...]

PR #2658, opened 2024-08-29
Add Pylint support and enable Pylint CI checks

correctmost:
## PR Description:

Pylint can help catch issues like #2625:

```
************* Module archinstall.default_profiles.server
archinstall/default_profiles/server.py:34:4: E1101: Instance of 'ServerProfile' has no 'set_current_selection' member (no-member)
```

I didn't add the Pylint checks to the pre-commit config because they might be a bit too slow to run on each commit.

## Tests and Checks

### DIFF
--- a/.github/workflows/pylint.yaml
+++ b/.github/workflows/pylint.yaml
@@ -1,0 +1,22 @@
+on: [ push, pull_request ]
+name: Pylint linting
+jobs:
+    pylint:
+        runs-on: ubuntu-latest
+        container:
+            image: archlinux/archlinux:latest
+        steps:
+            - uses: actions/checkout@v4
+            - name: Prepare arch
+              run: |
+                pacman-key --init
+                pacman --noconfirm -Sy archlinux-keyring
+                pacman --noconfirm -Syyu
+                pacman --noconfirm -Sy python-pip python-pyparted python-simple-term-menu pkgconfig gcc
+            - run: pip install --break-system-packages --upgrade pip
+            - name: Install Pylint and Pylint plug-ins
+              run: pip install --break-system-packages .[dev]
+            - run: python --version
+            - run: pylint --version
+            - name: Lint with Pylint
+              run: pylint .
--- a/archinstall/lib/interactions/general_conf.py
+++ b/archinstall/lib/interactions/general_conf.py
@@ -61,8 +61,8 @@ def ask_for_audio_selection(
 	current: Optional[AudioConfiguration] = None
 ) -> Optional[AudioConfiguration]:
 	choices = [
-		Audio.Pipewire.name,
-		Audio.Pulseaudio.name,
+		Audio.Pipewire.name,  # pylint: disable=no-member
+		Audio.Pulseaudio.name,  # pylint: disable=no-member
 		Audio.no_audio_text()
 	]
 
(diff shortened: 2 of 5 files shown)


svartkanin:
Maybe I'm missing something here but why is there a need for ruff and pylint? Isn't ruff covering everything?

Torxed:
Before we merge this, I have to create a `python-pylint-pydantic` package as we need to be able to source this during [PKGBUILD](https://github.com/archlinux/archinstall/blob/master/PKGBUILD):
```diff
makedepends=(
  'python-setuptools'
  'python-sphinx'
  'python-build'
  'python-installer'
  'python-wheel'
+ 'python-pylint'
+ 'python-pylint-pydantic'
)
```
However that package does not exist currently in [any repo](https://archlinux.org/packages/?sort=&q=pylint&maintainer=&flagged=), not even in [AUR](https://aur.archlinux.org/packages?O=0&K=pylint).

It's not a major showstopper, but I just need to sit down a few minutes and make that happen.

svartkanin:
@Torxed why is it needed for a build? Those packages are part of the dev packages in pyproject.toml so they shouldn't be dependencies for a build?

Torxed:
Because we *should* run [check()](https://wiki.archlinux.org/title/Python_package_guidelines#Check) in [PKGBUILD](https://wiki.archlinux.org/title/PKGBUILD#checkdepends).
I haven't been a model packaging citizen, but now that we're doing good stuff this should ideally be there :)

svartkanin:
Hmm I'm still somewhat worried that mixing the build with linting may cause issues down the line. If these checks run on PRs and only PRs that validate successfully get merged then the master branch should be considered ckean and ready for a build.
I remember @Torxed in the past rare PRs got merged with failing checks coz fixing a bug was higher priority then meeting linting checks. If such a check is now failing and the PR gets merged it'll block a hit fix release completely.

Torxed:
> Hmm I'm still somewhat worried that mixing the build with linting may cause issues down the line. If these checks run on PRs and only PRs that validate successfully get merged then the master branch should be considered ckean and ready for a build. I remember @Torxed in the past rare PRs got merged with failing checks coz fixing a bug was higher priority then meeting linting checks. If such a check is now failing and the PR gets merged it'll block a hit fix release completely.

There's also the odd "arch packaging patches" where any trusted user can go in and submit `-<arch patch>` versions. 

Which is what you see sometimes where, a package didn't change version upstream like `v2.8.6` might be packaged with `v2.8.6-2` only in arch. 

It's a low probability of issues, but best practice in packaging is to have the `check()` hook during packaging that should just run a final test/linting.

And it's no problem for me to make a PKGBUILD for pylint-pydantic as long as it's maintained upstream :)

I just need to find my way to my laptop 😅

svartkanin:
Maybe we should pause the pylint addition for a bit and see how well ruff in combination with the others work, get mypy fully compliant and see if we need pylint in addition?

Torxed:
It's true, we could skip the `check()` in PKGBUILD's for now.
But knowing myself it will probably be forgotten until someone pokes me or when it bites me in the butt.

But the tests that pylint adds sounds nice tbh.

svartkanin:
@correctmost I'm soon (TM) about to raise an integration PR for the new menu system which is going to shuffle things around, so just to prevent any throwaway work being done just ignore any menu related linting

svartkanin:
https://github.com/archlinux/archinstall/pull/2663

Torxed:
I've successfully built the dependencies now. I just need to sign and publish the packages :)

Torxed (approved):

Issue #2762, opened 2024-10-30
Typo?


hugok79:
https://github.com/archlinux/archinstall/blob/59384615a50afd5afcb245ccb91770b11092b618/archinstall/locales/base.pot#L133

"Select which partition to `mask` for formatting"

It shouldn't be `mark` instead of `mask`?

PR #2765, opened 2024-11-01
Fixed Typo

prashasth-nair:
- This fix issue: 

## Tests and Checks
  

Changed "Mask" to "Mark". Issue: #2762


svartkanin:
Just changing the translations will not solve this. You'll need to change it in the code where it's actually uses

Torxed (approved):

Torxed (changes requested):
Change the `_("string")` in the code as well that uses this string.

svartkanin:
As this has been stale for quite a while I'll close this PR for now. 
If you'd like to update the PR we can revisit this in the future

PR #2786, opened 2024-11-08
Replace `Pacman()` syscall logic with libalpm

Torxed:
## PR Description:

This PR will enable us to move away from running `pacman -XYZ` via system calls, and instead rely on [libalpm](https://man.archlinux.org/man/libalpm.3.en) directly, via [pyalpm](https://gitlab.archlinux.org/archlinux/pyalpm).

This has the added benefit that we can do package search efficiently, meaning we can replace: https://github.com/archlinux/archinstall/blob/26cf04655ae1e0ce38cb557a43e5b82f385d1302/archinstall/lib/packages/packages.py#L45-L59

Allowing us to do local searches, and let pacman deal with databases and searching.
This simplifies custom mirrors and allowing us to search against them too.

## Some thoughts

I decided to use a context manager approach, as it allows us to choose between interacting against system-wide pacman data - but during context creation also decide if we want a temporary pacman data structure.

That way we can do:

```python
with Pacman(temporary=True, servers=['http://ftp.lysator.liu.se/pub/archlinux/$repo/os/$arch', 'http://ftp.acc.umu.se/mirror/archlinux/$repo/os/$arch']) as pacman:
	print(pacman.search('nano'))
```
To not collide with system-wide pacman, and perform searches and syncs independently and irregards if existing pacman is executing.

With this we can remove: https://github.com/archlinux/archinstall/blob/26cf04655ae1e0ce38cb557a43e5b82f385d1302/archinstall/lib/pacman/__init__.py#L32-L43

And if we want to interact with pacman system wide, we simply just do:

```python
with Pacman() as pacman:
	pacman.sync()
	pacman.install("vim")
```

I'm sure we could achieve the same thing with using only `__init__` and some if/else.
But for some reason this felt like a logical thing to do, especially since all pacman operations are "transaction based", meaning we need to create a lock file while doing operations - and then remove the lock file once we're done.

And if libalpm/pacman crashes during an operation, we need a way to do easy cleanup and remove said lock file.


Torxed:
Some additional commits came in, will revert those/this PR and clean that up.

PR #2799, opened 2024-11-11
Add Teaching Mode as a CLI Flag

jackdesert:
## PR Description:

Adds a flag through ArgParser for `--teach`, which does the following:

1. Echo each subprocess command to the screen (but only after installation actually begins)
2. Wait for a few seconds before actually running the command so the user has a chance to grok what's going on

## Tests and Checks

### DIFF
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -74,6 +74,8 @@ def define_arguments() -> None:
 	parser.add_argument("-v", "--version", action="version", version="%(prog)s " + __version__)
 	parser.add_argument("--config", nargs="?", help="JSON configuration file or URL")
 	parser.add_argument("--creds", nargs="?", help="JSON credentials configuration file")
+	parser.add_argument("--teach", action="store_true", default=False,
+						help="Teaching mode. Echoes all subprocess commands to the screen with a short pause after each. Useful for understanding what goes on under the hood.")
 	parser.add_argument("--silent", action="store_true",
 						help="WARNING: Disables all prompts for input and confirmation. If no configuration is provided, this is ignored")
 	parser.add_argument("--dry-run", "--dry_run", action="store_true",
@@ -323,6 +325,8 @@ def main() -> None:
 	script = arguments.get('script', None)
 
 	if script is None:
+		# `script` is typically passed in from ArgumentParser,
+		# which comes with a default value.
 		print('No script to run provided')
 
 	mod_name = f'archinstall.scripts.{script}'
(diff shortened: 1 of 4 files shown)


jackdesert:
Screenshots:

svartkanin:
We're already logging every command executed into a file https://github.com/archlinux/archinstall/blob/040a6aba27d20361f0c833a04b75fce09690bb51/docs/help/report_bug.rst#L28 isn't that sufficient?

jackdesert:
I actually built this to educate myself about the arch install process, and to demonstrate to a fellow arch newb the patterns used by the guided installer. 

For me the `--teach` option is about education in the moment. Seeing the play-by-play inline with the narration. The existing narration tells me in friendly words what the installer is about to do. Then seeing the actual commands right after that lets me know how those friendly words are enacted. This lends itself to realizations such as: "Oh look, it takes three commands to make those friendly words happen!" / "Oh look, that's the command where I always wondered how they got it to run without asking for human input!" / "Oh look, that command takes a looong time to run and generates a log of output!"

I also envision the --teach option to be super easy to point newcomers to. For example: "Boot up this .iso, run these three iwctl commands to connect to WiFi, and then run `archinstall --teach` for a guided install that also shows you what commands are running under the hood."

jackdesert:
Here's one more plug for `--teach` mode:

@svartkanin Today I looked at the contents of `/var/log/archinstall/cmd_history.txt` to see if it could serve as a stand-in for the `--teach` mode that I have proposed in this PR. Here is what I observe:

1. `cmd_history.txt` appears to contain the cumulative  logs from all the times I've run `archinstall` during this boot. (This makes it hard to see which commands were used in just one invocation of `archinstall`)

2. `cmd_history.txt` appears to include the commands that were run even before the menu was populated. (This makes it challenging to separate out which commands were run just to set up the menu vs which commands were run to actually install the system. I am most interested in learning the latter.)

3. The timestamp format in `cmd_history.txt` is numerical. (Example: 1731611189.3625734) This makes it hard for my human eyes to determine how many minutes passed from one command to the next.

svartkanin:
`archinstall` is essentially an automation tool for the arch wiki. It tends to implement commands and instructions as they are described in the wiki. 
It adds maintenance cost that doesn't necessarily translate into a larger benefit as I think most folks are looking for an automated installation rather than going through it manually step-by-step. And in the latter case the wiki should be the first point of entry for gathering knowledge.

codefiles:
This idea overlooks that there are operations performed in native python (mkdir, chmod, touch as well as file reading, writing, and manipulation) rather than using a command. For example, there is not something like `echo archinstall > /mnt/etc/hostname` in the code. Also when a command is performed it is possible the output is captured and utilized in a way this will not show. This will leave "gaps" between the commands and therefore fail to achieve the intended purpose of teaching.

PR #2801, opened 2024-11-14
When no options provided, mention that options are available

jackdesert:
## PR Description:

When I first started using archinstall I hadn't thought to check to see if there were options I could pass it on the command line. It turns out these options are needed in order to load a config file from disk. So this PR points out that there are options available. (But it only points them out if no options were passed. This way when people are already using options, they don't see either the notice or get the delay.)

## Tests and Checks
  

See the bottom of this screenshot for what you get with this PR when you call `python -m archinstall` with no options:

### DIFF
--- a/archinstall/scripts/guided.py
+++ b/archinstall/scripts/guided.py
@@ -1,4 +1,6 @@
+import sys
 from pathlib import Path
+from time import sleep
 from typing import Any, TYPE_CHECKING, Optional
 
 import archinstall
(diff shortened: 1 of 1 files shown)


svartkanin:
This means every time someone runs archinstall they will need to wait X seconds for it to proceed, which doesn't make any sense as it will also do that during development.

`archinstall` is a CLI tool like hundreds others, it follows a similar structure and usability and provides the help option `--help`. Considering that most users (if not all) are familiar with CLI tools I think it is fair to assume they know how to look for options.

Torxed:
I'm also not a fan of additional `sleep()` timers. We already have quite a few delays in the code and this one would be super intrusive.

I also think that if anyone ever used a command in linux *(or even other OS:es tbh)*, `man archinstall` or `archinstall --help` will be the first thing you try. And both work.

PR #2803, opened 2024-11-14
Add helpful tips for saving config files in a directory

jackdesert:
## PR Description

The first few times I used archinstall, I struggled with saving config files to a directory. It sounds easy, I know. But at the time I kept getting errors and finally gave up saving any config files at all. 

This PR makes it obvious why tab-completion is a boon in this scenario. It also gives a couple useful working locations, and it clarifies that extra steps are needed if you want your config files to persist through a boot cycle.

Here is a screenshot showing the updated text:

## Tests and Checks

### DIFF
--- a/archinstall/lib/configuration.py
+++ b/archinstall/lib/configuration.py
@@ -152,8 +152,8 @@ def save_config(config: Dict) -> None:
 Into which DIRECTORY should the configuration(s) be saved?
 
 	TIPS
-	1. Tab completion is enabled to help you locate a full directory name.
-	2. To persist a configuration file through a reboot, mount a separate drive before running `archinstall`.
+	1. Tab completion is enabled to help you locate a full directory name
+	2. To persist a configuration file through a reboot, mount a separate drive before running `archinstall`
 	3. If you are not sure what to choose, try "." or "/"
 
 Save Directory:  '''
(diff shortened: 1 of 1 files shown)


archinstall/lib/configuration.py
```diff
@@ -148,7 +148,15 @@ def preview(selection: str) -> Optional[str]:
 		while True:
 			path = input(
 				_(
-					"Enter a directory for the configuration(s) to be saved (tab completion enabled)\nSave directory: "
+					'''\
+Into which DIRECTORY should the configuration(s) be saved?
```
svartkanin, line 152: this is goind to output every time someone presses <tab> which is quite annoying
image

I would be better to show this output only once on the top

svartkanin (changes requested):

Torxed:
Maybe we could mention that the current installation is mounted under `/mnt/archinstall` as well? In case they want to save it to their recently installed system - instead of them having to manually mount something?

Or simply:

```
Do you want to save your configuration choices as a JSON file in your newly installed system?

Yes / No
```

And then put it somewhere safe?

svartkanin:
As this has been stale for quite a while I'll close this PR for now. 
If you'd like to update the PR we can revisit this in the future

PR #2813, opened 2024-11-15
Remove most deprecated typing.Dict usage

correctmost:
## PR Description:

`typing.Dict` is deprecated:
- https://docs.python.org/3/library/typing.html#typing.Dict
- https://docs.astral.sh/ruff/rules/deprecated-import/

This PR doesn't touch some disk-related code to avoid merge conflicts with #2775 .

Follow-up to #2673.

## Tests and Checks

### DIFF
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -7,7 +7,7 @@ import curses
 import traceback
 from argparse import ArgumentParser, Namespace
 from pathlib import Path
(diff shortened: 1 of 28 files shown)


svartkanin (approved):

svartkanin:
@correctmost thanks for cleaning all of these up. Would you know if there's a way to also enforce not introducing them in the future again, maybe with a linting rule?

PR #2824, opened 2024-11-16
Remove redundant simple-term-menu package

svartkanin:
As the new curses menu is merged, the `simple-term-menu` package is no longer needed as a dependency


Torxed (approved):

grazzolini:
We have a python rebuild coming on, I think we should make an archinstall release as soon as possible, so the rebuild doesn't impact December's ISO and also, so we can drop simple-term-menu.

PR #2867, opened 2024-11-18
Reduce the max line length from 220 to 160 by wrapping some lines

correctmost:
## PR Description:
This will make it easier to auto-format import sections.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -3,7 +3,7 @@ count = True
 # Several of the following could be autofixed or improved by running the code through psf/black
 ignore = E722,W191,W503
 max-complexity = 40
-max-line-length = 220
+max-line-length = 160
 show-source = True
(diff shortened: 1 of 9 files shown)


svartkanin:
@correctmost this might be somewhat controversial :) 
I'd rather leave this at the current value as I'm sure there's loads of opinions on this, unless there's a significant benefit on reducing it. 

@Torxed any thoughts?

Torxed:
And there's no way to force multi-imports to use:
```python
from x import (
    a,
    b
)
```
Without this?

svartkanin:
Is 160 a number that will certainly fix it, or fix it for the current state and in the future it may have to be changed again as some imports require an even lower length to be set?

Torxed (approved):

PR #2878, opened 2024-11-19
Remove unused Any instances from TYPE_CHECKING blocks

correctmost:
## PR Description:

This will make it easier to start removing more `Any` instances from the codebase.

Some of the regressions in 3.0.0 are not being caught by mypy because of the prevalence of `Any` in the code.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,9 +86,16 @@ warn_unused_configs = true
 warn_unused_ignores = true
 
 [[tool.mypy.overrides]]
+module = "archinstall.default_profiles.applications.*"
+disallow_any_explicit = true
+
+[[tool.mypy.overrides]]
+module = "archinstall.default_profiles.servers.*"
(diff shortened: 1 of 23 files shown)


pyproject.toml
```diff
@@ -85,10 +85,17 @@ warn_unreachable = true
 warn_unused_configs = true
 warn_unused_ignores = true
 
+[[tool.mypy.overrides]]
+module = "archinstall.default_profiles.applications.*"
+disallow_any_explicit = true
```
svartkanin, line 90: @correctmost will this prevent any future use of the `_: Any`? 
We're using that for translations, so it's likely we'll need it again

svartkanin (approved):

pyproject.toml
Torxed: Great work overall with these quality improvements, it's understated how much we appreciated it I think!

PR #2917, opened 2024-11-23
Use binary units

codefiles:
Address a comment from an ArchWiki [edit summary](https://wiki.archlinux.org/index.php?title=Archinstall&diff=814715&oldid=804271) that mentions archinstall, "strangely shows the partition sizes in [decimal units](https://en.wikipedia.org/wiki/Byte#Units_based_on_powers_of_10)".


Torxed (approved):
Great catch!

PR #2936, opened 2024-11-25
Dataclasses for  args and config

svartkanin:
This PR introduces dataclasses for the command line arguments and for the configurations. 
This is only the implementation and parsing logic, it has not been plugged into the code yet which will be a follow-up PR.


Torxed (approved):
Nice addition with the `pytest`!

PR #2998, opened 2024-12-03
updating plasma.py

zbikkk:
changed kwrite to kate as the two packages merged together in 2022 and removed egl-wayland (why was it there in the first place?) it gets pulled by the nvidia package anyway.


svartkanin (approved):

svartkanin:
Thanks for the improvement

PR #3059, opened 2024-12-30
Add libcamera

EliasTheGrandMasterOfMistakes:
- This fix issue: 

## Tests and Checks

### DIFF
--- a/archinstall/default_profiles/applications/pipewire.py
+++ b/archinstall/default_profiles/applications/pipewire.py
@@ -19,6 +19,7 @@ class PipewireProfile(Profile):
 			'pipewire',
 			'pipewire-alsa',
 			'pipewire-jack',
+			'pipewire-libcamera',
 			'pipewire-pulse',
 			'gst-plugin-pipewire',
 			'libpulse',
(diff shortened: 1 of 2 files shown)


svartkanin:
Why would this package be essential?

EliasTheGrandMasterOfMistakes:
> Why would this package be essential?

libcamera is a library that provides camera stack for webcam and camera devices
like for example intel ipu6, in case of these devices its can be used with libcamera 
and for applications that uses it like gnome-snapshots that iirc uses it with pipewire plugin.

Libcamera-ipa is a dependence and provides image processing algorithms for isps and includes
a softISP version that contains debayer that are necessary for some webcams including for future 
arm devices support.

Fedora already provides libcamera to provide out-of-the-box webcams that uses complex ISPs
working

svartkanin:
Archinatall aims to set up a minimal working installation without any bloat of packages. 
This can easily be installed either via the additional packages or as a post installation

zbikkk:
i agree with @svartkanin, archinstall should aim to be as minimal as possible, if the user wishes for libcamera he may install it himself afterwards.

PR #3071, opened 2025-01-04
disk: rework manual partitioning

### LINKED ISSUE #2444: Manual partitioning is really unfriendly
1. every partition starts the the beginning of the drive, instead of right after the previous partition

image

image

image

Notice how the two partitions overlap. You can spot this if you're paying attention, but you have to go back to figure out where the previous partition ends, then write down the sector number and try again.

2. you can't specify a partition size (relative end). So if you make a:
 - 0 start / 1 GB end
 - 1 GB start / 10 GB end
 - 10 GB start / 50 GB end

You get a 1 GB, a 9 GB and a 40 GB partition. `gdisk` supports both `10G` (ends at 10 GB) and `+10G` (has size of 10 GB).

https://github.com/archlinux/archinstall/issues/1213 suggests this is actually the size, but it doesn't match my experience.

3. the "format" option disappears after toggling it and it's not clear when it's going to format the partition or not
4. this might be intentional, but if you delete an existing partition, it gets marked as deleted; then if you delete it again, it disappears from the list. Now if you try to create a new partition, you can't use the whole disk. I don't have a screenshot unfortumately.

codefiles:
Resolves #2444 and the warning on the [ArchWiki](https://wiki.archlinux.org/title/Archinstall):
> - _archinstall_'s _Manual Partitioning_ does not protect against creating overlapping partitions, [misaligned partitions](https://wiki.archlinux.org/title/Advanced_Format#Partition_alignment) or the last partition overlapping with the backup GPT header.

- Validate disk config
- Display free space segments that can be selected to create new partitions
- Take input for size rather than start and end when creating a new partition
- Fix a bug where an existing partition marked as delete could be deleted again resulting in it being removed from the list

archinstall-manual-part

archinstall-manual-part2

### DIFF
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -274,7 +274,11 @@ def post_process_arguments(args: dict[str, Any]) -> None:
 		path = args['plugin']
 		load_plugin(path)
 
-	load_config()
+	try:
+		load_config()
+	except ValueError as err:
+		warn(str(err))
+		exit(1)
 
 
 define_arguments()
--- a/archinstall/lib/disk/device_model.py
+++ b/archinstall/lib/disk/device_model.py
@@ -907,11 +907,18 @@ class PartitionModification:
 	def is_modify(self) -> bool:
 		return self.status == ModificationStatus.Modify
 
+	def is_delete(self) -> bool:
+		return self.status == ModificationStatus.Delete
+
 	def exists(self) -> bool:
 		return self.status == ModificationStatus.Exist
 
 	def is_exists_or_modify(self) -> bool:
-		return self.status in [ModificationStatus.Exist, ModificationStatus.Modify]
+		return self.status in [
+			ModificationStatus.Exist,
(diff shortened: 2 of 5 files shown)


codefiles:
Excellent! The failed pytest shows the validation is working. It had me worried for a second. :satisfied:

In the test config the root partition starts at `513 MiB` and is `20 GiB`, the home partition starts at `20 GiB`; that has overlap.

https://github.com/archlinux/archinstall/blob/f1f08c811bb28a043479094ec1dcd241ea5ba807/tests/data/test_config.json#L57-L64

https://github.com/archlinux/archinstall/blob/f1f08c811bb28a043479094ec1dcd241ea5ba807/tests/data/test_config.json#L46-L53

https://github.com/archlinux/archinstall/blob/f1f08c811bb28a043479094ec1dcd241ea5ba807/tests/data/test_config.json#L84-L91

svartkanin (approved):

PR #3078, opened 2025-01-05
Remove unused translation check workflow

correctmost:
## PR Description:

The code has been unused since 1e296b263.

This should prevent "Run failed: .github/workflows/translation-check.yaml" emails when creating PRs.


svartkanin:
I implemented this as a validation but actually never fully finished it. Xan we leave it and I'll put this on my todo to actually integrate

PR #3079, opened 2025-01-05
Use float instead of int for pre-validated mirror score fields

correctmost:
## PR Description:
This fixes an unnecessary-round ruff warning in `validate_score`:

```
archinstall/lib/models/mirrors.py:99:12: RUF057 Value being rounded is already an integer
```

It also makes the code match the types seen in https://archlinux.org/mirrors/status/json/.


svartkanin:
@correctmost I'll merge this in, it doesn't solve the problem completely but I'll raise a followup for the other bit. 
@Torxed I think this explains the problem you've seen

svartkanin (approved):

PR #3102, opened 2025-01-11
Replace list with link to contributors page

codefiles:


svartkanin:
Sounds reasonable, but I'll let @Torxed as the owner sign it off

Torxed:
Looks great, the list was nice when we were only a handful.. but it's not maintainable any longer hehe.

PR #3109, opened 2025-01-11
Use removeprefix/removesuffix instead of incorrect lstrip/rstrip calls

correctmost:
## PR Description:

This PR should fix the URL issue mentioned in #3013


svartkanin:
Does this fix the url? I did address it in the mirrorlist parsing PR in the server url function https://github.com/archlinux/archinstall/pull/3104/files#diff-3dfabed340ddd155e3952a2337b2d99f4bf4f47c8c7fd5f30b01b5f9c68c263bR41

svartkanin (approved):

PR #3123, opened 2025-01-15
Fix imports and add checks for BTRFS tools

hugo22dr:
### Description

This pull request fixes imports and adds checks to ensure that the BTRFS tools (btrfs-progs) are available on the system before performing BTRFS-related operations. It also ensures that the mount point is available before mounting the file system.

### Major Changes

1. Fixed imports of the `SysCommand` and `SysCommandWorker` modules.
2. Added `_check_btrfs_tools` and `_ensure_mountpoint_available` functions in the `DeviceHandler` class to check the availability of the BTRFS tools and ensure that the mount point is available.
3. Modified the `create_btrfs_volumes` function in the `FilesystemHandler` class to use the newly added functions.

### Tests

- Verified that the BTRFS tools are available before performing BTRFS operations.
- Ensured that the mount point is available before mounting the filesystem.
- Tested in a local environment to ensure that the changes work as expected.

### Contributor

This is my first contribution to this repository. I appreciate the opportunity to contribute and welcome feedback to improve this pull request.

### DIFF
--- a/archinstall/lib/disk/device_handler.py
+++ b/archinstall/lib/disk/device_handler.py
@@ -850,6 +850,22 @@ class DeviceHandler:
 		except SysCallError as err:
 			debug(f'Failed to synchronize with udev: {err}')
 
+	def _check_btrfs_tools(self) -> bool:
+		"""Verify if btrfs-progs tools are available in the system"""
+		try:
+			SysCommand(['which', 'btrfs'])
+			return True
+		except SysCallError:
+			raise DiskError("btrfs-progs not found. Please install btrfs-progs package.")
+
+	def _ensure_mountpoint_available(self, mountpoint: Path) -> None:
+		"""Ensure the mountpoint is available by unmounting if necessary"""
+		if mountpoint.exists():
+			try:
+				self.umount(mountpoint, recursive=True)
+			except Exception as e:
+				debug(f'Failed to unmount existing mountpoint {mountpoint}: {str(e)}')
+
 
 device_handler = DeviceHandler()
 
--- a/archinstall/lib/disk/filesystem.py
+++ b/archinstall/lib/disk/filesystem.py
@@ -9,7 +9,9 @@ from archinstall.tui import Tui
 from ..hardware import SysInfo
 from ..interactions.general_conf import ask_abort
 from ..luks import Luks2
-from ..output import debug, info
+from ..output import debug, info, error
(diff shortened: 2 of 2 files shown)


svartkanin:
I don't think explicitly checking for btrfs to be available is required to do. Archinstall is meant to run on a live ISO and by defining the relevant dependencies in the pkg build file it is to be assumed that they are present

hugo22dr:
> I don't think explicitly checking for btrfs to be available is required to do. Archinstall is meant to run on a live ISO and by defining the relevant dependencies in the pkg build file it is to be assumed that they are present

Thanks for the feedback. I would like to provide more context on the errors we are experiencing and how the proposed changes aim to address them.

Error Context
We are experiencing errors during the Arch Linux installation with the archinstall parameter. The errors start at __init__.py on lines 337/339 and continue through importlib/__init__.py on line 88. These errors appear to be related to the lack of BTRFS tools in the runtime.

The proposed changes include:

BTRFS Tools Check: We have added a check to ensure that the BTRFS tools (btrfs-progs) are available on the system before performing BTRFS-related operations. This is to avoid errors during the installation when these tools are not present.

Mount Point Availability Guarantee: We have added a function to ensure that the mount point is available before mounting the file system. This helps to avoid mount conflicts that can cause the installation to fail.

We understand that Archinstall is intended to run on a live ISO and that the relevant dependencies must be present. However, in some cases, there may be environments where these tools are not available by default, causing the installation to fail. The proposed changes aim to make the installation process more robust by explicitly checking for the presence of the required tools and ensuring that the mount point is available.

If the explicit check for BTRFS tools is not necessary, we are open to removing this part of the changes. However, we would like to discuss possible solutions to ensure that the installation environment has all the necessary dependencies to avoid errors during installation.

I appreciate the opportunity to contribute and we welcome additional feedback to improve this pull request.

svartkanin:
You mentioned errors, can you provide an error log file from an installation?

When installing archinstall via pacman then all necessary packages and dependencies should also be installed

hugo22dr:
> You mentioned errors, can you provide an error log file from an installation?
> 
> When installing archinstall via pacman then all necessary packages and dependencies should also be installed

Sure! I'll do that right away, sorry for not presenting the problem.

svartkanin:
As there hasn't been an update on this for a while I'll close it for now

PR #3167, opened 2025-02-09
Integrate new arguments data structure

svartkanin:
This is the integration of the new `ArchConfig` data structure in favor of the `archinstall.arguments` structure as that one is merely a `dict[str, Any]` which doesn't allow for any type checking and safety. 

Unfortunately, this PR grow beyond what I originally intended to change here but by introducing the `args.py` as a dependency (which pulls in most data models) there were loads of circular dependencies to be dealt with as the code base isn't well divided hierarchically at the moment so I was forced to move things around to get the interpreter happy. Most notably I moved a lot of model data classes into the `archinstall/lib/models` folder. This aligns with existing models in there as well. 

### Changes
* Replaced all usages of the `archinstall.arguments` dict type and replaced it with the respective config or args types
* Version is no longer explicitly defined in `archinstall.__init__` but will be read in the `args.py` file directly from the `pyproject.toml` @Torxed let me know if that causes issues on release 
* Moved data types from `disk/*` to `models` which resolved the circular dependencies
* Added tests for the output testing of the new arguments structures

As I have only moved the relevant data types into the `models/` module there's still some misalignment in the code base, I'm planning to raise a follow up PR to cleanup the rest of the types as well and move them into the `models` if they fit to avoid any major future dependency nightmares.


codefiles:
@svartkanin, with this commit _Automatic time sync (NTP)_ is set to disabled by default when previously it was enabled. Was that intended?

svartkanin:
No it wasn't, that's a regression, I'll raise a fix.
Thanks for pointing it out

Torxed:
> Version is no longer explicitly defined in archinstall.__init__ but will be read in the args.py file directly from the pyproject.toml @Torxed let me know if that causes issues on release

Forgot one reason why we had `__version__`, and that was packaging, hehe..
https://github.com/archlinux/archinstall/blob/f0159e127891959cb1a7930694f57cc18481869c/PKGBUILD#L59

I'll patch/change so that we rely on the `git tag` instead, which should be more reliable anyway.

svartkanin:
@Torxed could also change the awk to target the `pyproject.toml` instead as that is the only SOT now for the version number

Torxed:
That could work too. But for packaging, `git describe` should do the trick and be the most reliable one :)

PR #3176, opened 2025-02-15
Add support for qemu venus

DriedSnow:
- This fix issue: 

## PR Description:
Adds support for the Venus in qemu to the graphics drivers. Venus is a way to have gpu acceleration in qemu with vulkan. Adds it to the all open source and to the open source vm.

## Tests and Checks

### DIFF
--- a/archinstall/lib/hardware.py
+++ b/archinstall/lib/hardware.py
@@ -105,7 +105,8 @@ class GfxDriver(Enum):
 					GfxPackage.LibvaIntelDriver,
 					GfxPackage.IntelMediaDriver,
 					GfxPackage.VulkanRadeon,
-					GfxPackage.VulkanIntel
+					GfxPackage.VulkanIntel,
+					GfxPackage.VulkanVirtio
 				]
 			case GfxDriver.AmdOpenSource:
 				packages += [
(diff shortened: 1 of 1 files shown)


archinstall/lib/hardware.py
```diff
@@ -69,7 +70,7 @@ class GfxDriver(Enum):
 	NvidiaOpenKernel = 'Nvidia (open kernel module for newer GPUs, Turing+)'
 	NvidiaOpenSource = 'Nvidia (open-source nouveau driver)'
 	NvidiaProprietary = 'Nvidia (proprietary)'
-	VMOpenSource = 'VMware / VirtualBox (open-source)'
+	VMOpenSource = 'Qemu / VMware / VirtualBox (open-source)'
```
svartkanin, line 73: Modifying this will break existing configuration when loading, can we handle this when parsing the configuration to support this new and the old format

DriedSnow: Maybe i should just be a different on for qemu then i don't really know how to fix the parsing.

DriedSnow:
This requires  work that i don't really have the time for because of my health to look though the file for loading config. So i am just marking this a closed.

Torxed:
> This requires work that i don't really have the time for because of my health to look though the file for loading config. So i am just marking this a closed.

Get well, and we totally understand.
If we find time we can pick up where you left off, otherwise we'll wait for a future PR.

Issue #3168, opened 2025-02-10
archinstall crashed while creating partitions


mmahmoudian:
### Which ISO version are you using?

2025.02.01

### The installation log

```json
Hardware model detected: QEMU Standard PC (Q35 + ICH9, 2009); UEFI mode: False
Processor model detected: Intel(R) Core(TM) i9-10900K CPU @ 3.70GHz
Memory statistics: 1660476 available out of 2013620 total installed
Virtualization detected: kvm; is VM: True
Graphics devices detected: dict_keys(['Red Hat, Inc. Virtio 1.0 GPU (rev 01)'])
Disk states before installing:
{
    "blockdevices": [
        {
            "name": "loop0",
            "path": "/dev/loop0",
            "pkname": null,
            "log_sec": 512,
            "size": "824 MiB",
            "pttype": null,
            "ptuuid": null,
            "rota": false,
            "tran": null,
            "partn": null,
            "partuuid": null,
            "parttype": null,
            "uuid": null,
            "fstype": "squashfs",
            "fsver": "4.0",
            "fsavail": 0,
            "fsuse_percentage": "100%",
            "type": "loop",
            "mountpoint": "/run/archiso/airootfs",
            "mountpoints": [
                "/run/archiso/airootfs"
            ],
            "fsroots": [
                "/"
            ],
            "children": []
        },
        {
            "name": "sr0",
            "path": "/dev/sr0",
            "pkname": null,
            "log_sec": 2048,
            "size": "1179 MiB",
            "pttype": "dos",
            "ptuuid": "5f3bc01a",
            "rota": false,
            "tran": "sata",
            "partn": null,
            "partuuid": null,
            "parttype": null,
            "uuid": "2025-02-01-08-29-13-00",
            "fstype": "iso9660",
            "fsver": "Joliet Extension",
            "fsavail": 0,
            "fsuse_percentage": "100%",
            "type": "rom",
            "mountpoint": "/run/archiso/bootmnt",
            "mountpoints": [
                "/run/archiso/bootmnt"
            ],
            "fsroots": [
                "/"
            ],
            "children": []
        },
        {
            "name": "vda",
            "path": "/dev/vda",
            "pkname": null,
            "log_sec": 512,
            "size": "51200 MiB",
            "pttype": null,
            "ptuuid": null,
            "rota": true,
            "tran": "virtio",
            "partn": null,
            "partuuid": null,
            "parttype": null,
            "uuid": null,
            "fstype": null,
[...]

PR #3179, opened 2025-02-16
Add disk sync on partition add to prevent table corruption

Omrifo4:
- This fix issue: #3168 

## PR Description:

The disk is not synced after adding a partition, meaning other writes might corrupt the partition table.

Calling `disk.commit()` afterward solved the problem.

## Tests and Checks
- [x] I have tested the code!

To test this, just use `archinstall` normally, noticing the not-corrupted partition table (with `fdisk`).

### DIFF
--- a/archinstall/lib/disk/device_handler.py
+++ b/archinstall/lib/disk/device_handler.py
@@ -561,6 +561,8 @@ class DeviceHandler:
 
 		try:
 			disk.addPartition(partition=partition, constraint=disk.device.optimalAlignedConstraint)
+
+			disk.commit()
 		except PartitionException as ex:
 			raise DiskError(f'Unable to add partition, most likely due to overlapping sectors: {ex}') from ex
 


Torxed:
I must say it's strange that the partition table would be corrupt after staging changes, and then calling one `commit` here: https://github.com/archlinux/archinstall/blob/01fa51c364c1c1ef1c782b64a76ab970b93dbf9a/archinstall/lib/disk/device_handler.py#L727-L733

Could it potentially be because it otherwise would be an overlap if say the first one gets deleted (but not commited): https://github.com/archlinux/archinstall/blob/01fa51c364c1c1ef1c782b64a76ab970b93dbf9a/archinstall/lib/disk/device_handler.py#L524

And when partition `/dev/..2` comes along, it overlaps with a non-deleted (and commited) partition?

Either way, perhaps doing a commit on each partition isn't a bad idea? It would most likely just introduce a delay back again?

codefiles:
This is not it. Don't merge this.

svartkanin:
Thanks for analysing this @codefiles

Issue #3049, opened 2024-12-25
Please support only EFI partition as FAT32 (allow /boot to be EXT4)


michaeluk01:
### describe the request

The Arch wiki says only /boot/efi or /efi has to be FAT32 and /boot can be EXT4 etc. But archinstall fails if you do that. Can you please add support for this.

abhimangs:
I believe the root partition (/), which usually has the most storage, should be EXT4 for performance and reliability. However, both the boot partition (/boot) and the EFI partition (/boot/efi) should be FAT32 to ensure compatibility with UEFI systems.

It would be helpful if archinstall could enforce this configuration to prevent installation issues and align with UEFI standards. This could simplify the process for users.

michaeluk01:
> I believe the root partition (/), which usually has the most storage, should be EXT4 for performance and reliability. However, both the boot partition (/boot) and the EFI partition (/boot/efi) should be FAT32 to ensure compatibility with UEFI systems.
> 
> It would be helpful if archinstall could enforce this configuration to prevent installation issues and align with UEFI standards. This could simplify the process for users.

Only the partition with EFI files has to be fat32, the boot partition is ext4 by default on Debian, Ubuntu and Fedora, it doesn't contain EFI files but does contain Grub or Systemd-boot files.

g-ramirez:
Hi, 

I added a fix for this here.  Not sure what the devs will think, but you can test it out in your env if you really want this:
https://github.com/archlinux/archinstall/compare/master...g-ramirez:archinstall-grub-boot-efi-fix:master

michaeluk01:
Hi , the arch wiki has removed /boot/efi as a recommended EFI partition but does still recommend /efi .

https://wiki.archlinux.org/title/EFI_system_partition

https://wiki.archlinux.org/title/Partitioning

Having it as /efi means you don't have to have important files like the kernel on a fragile fat32 partition. Something Debian, Ubuntu, Fedora avoids.

But currently the archinstall script fails if you try to use /efi in this way.

g-ramirez:
Come to think of it, this may actually resolve your issue of using /efi as well, as I removed the need to specify the boot directory when using grub in a chroot environment, as this is necessary.  I'll give it a go later, but feel free to test it if you'd like.

dabrown645:
[...]

PR #3181, opened 2025-02-18
Changes to support using /boot/efi for grub

g-ramirez:
- This fix issue:  #3049

## PR Description:

Adds support for using /boot/efi as GRUB's EFI partition similar to the way other distros such as Ubuntu do this.  This does not align with the spec, but it may be useful for users coming from other distros who are familiar with this setup.  This also adds support for mounting partitions to existing subdirectories such as mounting a drive as to /mnt as /mnt/data

## Tests and Checks
  

1) run archinstall
2) In partitioning, create separate /, /boot/, and /boot/efi partitions, with /boot/efi being FAT32 formatted and unmark /boot as bootable and mark /boot/efi as bootable
3) Choose GRUB bootloader
4) Install as usual


michaeluk01:
Hi thanks for replying to my request in the other thread. What exactly does this commit do?


https://wiki.archlinux.org/title/EFI_system_partition

https://wiki.archlinux.org/title/Partitioning

The Arch wiki recommends /boot or /efi as the "EFI System Partition" . Having it as /efi means you don't have to have important files like the kernel on a fragile fat32 partition.

But currently the archinstall script fails if you try to use /efi in this way.

g-ramirez:
> Hi thanks for replying to my request in the other thread. What exactly does this commit do?
> 
> https://wiki.archlinux.org/title/EFI_system_partition
> 
> https://wiki.archlinux.org/title/Partitioning
> 
> The Arch wiki recommends /boot or /efi as the "EFI System Partition" . Having it as /efi means you don't have to have important files like the kernel on a fragile fat32 partition.
> 
> But currently the archinstall script fails if you try to use /efi in this way.

Hi,

It allows you to use /boot/efi as the EFI partition for GRUB.  The kernel and initrd are still stored in /boot, so you have the choice of either creating a separate /boot partition formatted however you'd like, or you can simply format a single / partition but have the fat32 EFI partition mounted in /boot/efi, similar to the way Ubuntu does it

michaeluk01:
> > Hi thanks for replying to my request in the other thread. What exactly does this commit do?
> > https://wiki.archlinux.org/title/EFI_system_partition
> > https://wiki.archlinux.org/title/Partitioning
> > The Arch wiki recommends /boot or /efi as the "EFI System Partition" . Having it as /efi means you don't have to have important files like the kernel on a fragile fat32 partition.
> > But currently the archinstall script fails if you try to use /efi in this way.
> 
> Hi,
> 
> It allows you to use /boot/efi as the EFI partition for GRUB. The kernel and initrd are still stored in /boot, so you have the choice of either creating a separate /boot partition formatted however you'd like, or you can simply format a single / partition but have the fat32 EFI partition mounted in /boot/efi, similar to the way Ubuntu does it

Cool will test it with /boot/efi and /efi . I guess /efi is more important as it's recommended in the wiki.

g-ramirez:
Alright, I gave it a go.  I think all you need is editing a single file:
https://github.com/archlinux/archinstall/pull/3181/files

No need to change the installer.py

michaeluk01:
Does this mean it won't be merged?

PR #3208, opened 2025-02-27
Do not enable UKI at archinstall start if using BIOS

mintsuki:
## PR Description:

As per title.

## Tests and Checks


svartkanin (approved):

codefiles:
There should not have been an issue here since this was addressed in 5db4456 but there was a regression caused by 88b91ae.

PR #3223, opened 2025-03-05
Fix 2379 - Mirror and region definitions

### LINKED ISSUE #2379: bug: changing mirros within prompt/json of archinstall should not change live-cd /etc/pacman.* files
error'ing behavior: 
- a user may want to add a local-mirror (not a repo) to the list of mirrors - and instead a repo is created within the arch-live's /etc/pacman.conf 
- this will then look for the repo's DB file and causes archinstall to fail
- finally upon trying to relaunch the archinstall, it fails immediately saying the repo cannot be found, check your internet (due to arch-live's pacman.conf file having a new-repo added to it, where the DB file can fail to exist)

potential solutions:
- a huge improvement: copy the /etc/pacman.conf and /etc/pacman.d/* files to a well-stated-location, then reference them when setting up the system.

- the add mirrors-prompt currently should read: add custom repo 
  - maybe give a hint to overwrite `core`, `extra`, `multilib`, ect

- another prompt to add mirror (or default mirror replacing /etc/pacman.d/mirrorlist )

svartkanin:
This fixes #2379 
To fix the reported problem and addressing the comment from @Torxed https://github.com/archlinux/archinstall/issues/2379#issuecomment-1985465661 the mirror and region configuration has been reworked. 
The major changes
* There are new configuration options for mirror servers and repositories and the menu entries have been renamed to make it clearer
  * `Select regions` -> Select the mirror regions (Same as currently)
  * `Add custom servers` -> NEW allows to define custom mirror servers 
  * `Add custom repositories` -> Has been renamed to avoid confusion
* The `Additional repositories` menu entry has been moved from the global menu into the `Mirrors and repositories` submenu to keep them all together and also renamed to `Optional repositories`
* The `mirror_config` entry has been adjusted in the `user_configuration.json` to reflect the configuration changes but it is backwards compatible so that existing configurations will not break
```
  "mirror_config": {
    "custom_servers": [
        {
          "url": "https://mymirror.com/$repo/os/$arch"
        }
    ],
    "mirror_regions": {
      "Australia": [
        "http://archlinux.mirror.digitalpacific.com.au/$repo/os/$arch"
      ]
    },
    "optional_repositories": [
      "testing"
    ],
    "custom_repositories": [
      {
        "name": "myrepo",
        "url": "https://myrepo.com/$repo/os/$arch",
        "sign_check": "Required",
        "sign_option": "TrustAll"
      }
    ]
  },
```

The `custom_servers` and `mirror_regions` configurations will only modify the `/etc/pacman.d/mirrorlist` configuration file, whereas the `optional_repositories` and `custom_repositories` will modify the `/etc/pacman.conf` accordingly.


neitsab:
Thanks, this is a great addition! 🙏🏻

Issue #2347, opened 2024-01-23
RTL Support


yarons:
The Arabic, Urdu, and Hebrew are breaking in multiple ways.

### First:
Letter ordering - regardless of font selection.
The sentences appear reversed (as if these were processed with `rev`).
#### For example:
If the string reads **Hello World**, it will appear on screen as **dlroW olleH**.
There's a library called [pyfribidi](https://pypi.org/project/pyfribidi/), which supports most RTL languages pretty well.

### Second:
Since those specific alphabets are written from Right-To-Left (hence RTL), the languages should also appear mirrored, so the menus should align to the right side of the screen instead of left like they are now.
##### RTL
![RTL](https://upload.wikimedia.org/wikipedia/commons/8/8b/VirtualBox_Ubuntu18.04_GNOME_Hebrew.png)

##### LTR (The one used in English)
![LTR](https://upload.wikimedia.org/wikipedia/commons/b/bd/Ubuntu_23.10_Mantic_Minotaur_Desktop_English.png)

This will require fundamental changes to the underlying [simple-term-menu](https://github.com/IngoMeyer441/simple-term-menu).

svartkanin:
The translations are here https://github.com/archlinux/archinstall/blob/master/archinstall/locales/ar/LC_MESSAGES/base.po. Are these correct and displayed in that way?

yarons:
I checked those (Arabic and Urdu) on the native console with a proper `usefont` (out of the box they both appear like squares) and I experienced both the phenomenons that I've mentioned, the text appears in the wrong order and aligned to the left.

BTW I've also filled an issue against [simple-term-menu](https://github.com/IngoMeyer441/simple-term-menu/issues/88).

svartkanin:
This has nothing to fo with the simple term menu and the squares appear because the default font isn't supporting non latin characters.

As I don't speak any of those languages I can't confirm that the display is wrong or not.

Are you able to confirm that the linked translations are correct? And could you share a screenshot of how they are wrongly appearing?

yarons:
I'm setting the font issue aside, this is how Hebrew looks on iTerm (connected to an Archlinux VM):
<img width="634" alt="Screenshot 2024-01-24 at 11 31 06" src="https://github.com/archlinux/archinstall/assets/406826/7ad892bd-d327-4439-8ca7-f2309ed7eab6">

The text itself is reversed ("dlroW olleH" instead of "Hello World").
[...]

PR #3229, opened 2025-03-06
Update Hebrew translation.

yarons:
#2347 is still pending, we also need to set the console font dynamically, otherwise the text will be reversed or won't appear at all.


svartkanin (approved):

yarons:
I have a suggestion.
When it's a non-latin language we can display a small banner allowing the user to press some key combination to a redirection to a charset selection wizard.
The wizard will ask some questions about the language and display several charset options for the desired language, sadly with fbdev it's impossible to display a preview of the charset dynamically one next to the other but it's most definitely possible to provide a way to select and also display some preview text to see that it actually worked.

Issue #3155, opened 2025-01-31
German Keyboard layout Issue


tom-jm69:
### Which ISO version are you using?

"2025-01-01"

#### Description of the issue

When using a German keyboard layout (de, de-latin1, or de-latin1-nodeadkeys) on the Arch Linux install ISO (2025-01-01), special characters like §, ö, ü, and ä do not appear when typed in text fields (e.g., hostname or disk encryption input boxes). The key press is registered in hidden boxes (typing the character § in hidden boxes (such as disk encryption password) shows two * characters), but no character is inserted. This issue occurs with both the built-in and external keyboards.

The key presses do work and are correctly represented when typing on the terminal.

tom-jm69:
Seems to be related with https://github.com/python/cpython/blob/c26bed1160978fe8b1844878b8123778e47870c6/Lib/curses/textpad.py#L100 maybe?

tom-jm69:
https://github.com/python/cpython/issues/50994#issuecomment-1093478338

tom-jm69:
I think we might have to use `get_wch()` instead of `getkey()` in the `EditMenu` and generally for the main window. Maybe there are other windows too that need it.

svartkanin:
I had a go at solving this but it seems rather complex. As pointed out in https://github.com/archlinux/archinstall/issues/3155#issuecomment-2660982459 the library will check at that point for the entered character to be an ascii printable character and only allow such to be printed. 
I've tried creating a custom textbox 
```
class CustomTextbox(Textbox):
	def edit(self, validate=None) -> str:
		while 1:
			ch = self.win.get_wch()
			if validate:
				x = validate(ord(ch))
			if not ch:
				continue
			if not self.do_command(ch):
				break
			self.win.refresh()
		return self.gather()
```

and use `get_wch` instead of `getch` but that will only work in the edit function whereas the ascii check resides in the `do_command` function and is ignored there. 

I'm not sure if this has a solution as curses seems to be limited here.
We can definitely add the restriction to only allow ascii characters to be entered though.

PR #3244, opened 2025-03-11
Only allow ascii chars in input

### LINKED ISSUE #3224: Password Bug when Creating User

2025.03.01

### The installation log

```json
HP250G7
Celerom 4020
4RAM
```

### describe the problem

HP250G7
The numeric keypad shows that you are typing even though it is not activated.
The installation is completed correctly but at the time of logging in the password is incorrect since no numbers were entered, but symbols or something similar.

svartkanin:
As described in https://github.com/archlinux/archinstall/issues/3155#issuecomment-2713413805 non-ASCII characters don't seem to be supported by curses input. 
Fixes #3224 and #3155

* Add ASCII validation for input characters and only allow such
* Display info at the bottom of input fields that only ASCII characters are supported

### DIFF
--- a/archinstall/tui/curses_menu.py
+++ b/archinstall/tui/curses_menu.py
@@ -7,6 +7,7 @@ import signal
 import sys
 from abc import ABCMeta, abstractmethod
 from collections.abc import Callable
+from curses.ascii import isprint
 from curses.textpad import Textbox
 from types import FrameType, TracebackType
 from typing import TYPE_CHECKING, Literal, override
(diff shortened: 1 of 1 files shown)


svartkanin:
@Torxed happy to merge this in?

tom-jm69:
Thank you for taking the time and fixing it :)

Torxed (approved):

Torxed:
Looks good! And after merging a few other things I'll begin testing and see if we can squeeze in a release before the next ISO.

PR #3252, opened 2025-03-12
Use __all__ in __init__.py files to fix unused-import warnings

correctmost:
## PR Description:

This commit allows ruff and flake8 ignores to be removed. It also allows mypy to be run in strict mode by default.

Various mypy directives were removed because they are automatically enabled in strict mode:

```
  --strict                  Strict mode; enables the following flags: --warn-
                            unused-configs, --disallow-any-generics,
                            --disallow-subclassing-any, --disallow-untyped-
                            calls, --disallow-untyped-defs, --disallow-
                            incomplete-defs, --check-untyped-defs, --disallow-
                            untyped-decorators, --warn-redundant-casts,
                            --warn-unused-ignores, --warn-return-any, --no-
                            implicit-reexport, --strict-equality, --extra-
                            checks
```


svartkanin:
There are already breaking changes as I was forced to move datackasses into the models module to avoid issues. 
Given that it might be beneficial to include more breaking changes now and release them all at once.

svartkanin (approved):

PR #3265, opened 2025-03-15
Pin astral-sh/ruff-action to a specific commit

correctmost:
## PR Description:

The GitHub docs recommend pinning third-party actions to specific commits for security hardening purposes:

https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

A recent security issue:
- https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised

### DIFF
--- a/.github/workflows/ruff.yaml
+++ b/.github/workflows/ruff.yaml
@@ -5,4 +5,4 @@ jobs:
         runs-on: ubuntu-latest
         steps:
             - uses: actions/checkout@v4
-            - uses: astral-sh/ruff-action@v3
+            - uses: astral-sh/ruff-action@9828f49eb4cadf267b40eaa330295c412c68c1f9  # v3.2.2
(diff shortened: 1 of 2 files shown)


svartkanin:
Interesting, so this means that we need to keep track of releases and manually update the hash moving forward. So if there was vulnerabilities discovered in previous versions we need to be aware of that and potentially update the hash. 
Generally this seems kinda counter intuitive, the idea of using "latest" IMHO is to not having to monitor things constantly and relay on the most up-to-date version with fixes. Now the recommendation is to move backwards and pin things...

svartkanin:
So would the updates be pushed for every new tag release hash?

svartkanin:
Okay lets try it out

svartkanin (approved):

PR #3267, opened 2025-03-17
Limine: use UUID for accessing boot partition if not same as ESP

mintsuki:


codefiles:
> Which will also happen here:
> 
> https://github.com/archlinux/archinstall/blob/87fb96d249f5660f5dc2095d33dc078028705fdf/archinstall/lib/installer.py#L1066-L1068
> 
> Since the move to `PartitionModification`, all paths will be compared like this:
> 
> ```
> class-instance == class-instance
> ```
> 
> Rather than two paths, making them always `True`.
> 
> I just wanna check in and make sure I'm not missing anything on how we want to use `PartitionModification`, should be safe to patch this right:
> 
> ```diff
> - if efi_partition and boot_partition != efi_partition:
> + if efi_partition and boot_partition.mountpoint != efi_partition.mountpoint:
> ```
> 
> Perhaps we should also add a `__eq__` operator on the classes to avoid confusion, just raising an exception if people to do compare the classes?

@Torxed, I wrote that code (d76f4a0), it is fine.

mintsuki:
> > Which will also happen here:
> > https://github.com/archlinux/archinstall/blob/87fb96d249f5660f5dc2095d33dc078028705fdf/archinstall/lib/installer.py#L1066-L1068
> > 
> > Since the move to `PartitionModification`, all paths will be compared like this:
> > ```
> > class-instance == class-instance
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Rather than two paths, making them always `True`.
> > I just wanna check in and make sure I'm not missing anything on how we want to use `PartitionModification`, should be safe to patch this right:
> > ```diff
> > - if efi_partition and boot_partition != efi_partition:
> > + if efi_partition and boot_partition.mountpoint != efi_partition.mountpoint:
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Perhaps we should also add a `__eq__` operator on the classes to avoid confusion, just raising an exception if people to do compare the classes?
> 
> @Torxed, I wrote that code ([d76f4a0](https://github.com/archlinux/archinstall/commit/d76f4a029604dffe740ef1d44fa5f34ec0b23480)), it is fine.

Can you please explain what is wrong with this PR and #3272 ? Why does this not work? It is true that the comparison is probably fine, but the 2 variables are the exact same, always...

codefiles:
@mintsuki, sorry, I edited my comment. Please see the context that I added.

mintsuki:
> @mintsuki, sorry, I edited my comment. Please see the context that I added.

fwiw it does not work if the ESP is set to `/boot/efi`, or pretty much anything really

codefiles:
@mintsuki, let's discuss this in an issue first. It is a waste of time to try to create a solution before properly understanding/identifying the issue.

mintsuki:
> @mintsuki, let's discuss this in an issue first. It is a waste of time to try to create a solution before properly understanding/identifying the issue.

sure, as you please

mintsuki:
@codefiles as per #3275 , this PR should be fine (despite the fact that I have never gotten it to work with separate `/boot`/ESP (as FAT partitions)). Marked ready for review.

svartkanin (approved):

PR #3268, opened 2025-03-17
More properly support FAT12 and FAT16 ESPs

mintsuki:
FAT12 and FAT16 are also valid filesystems for ESPs, therefore, try to support them


codefiles:
Relevant section of the ArchWiki:

https://wiki.archlinux.org/title/EFI_system_partition#Format_the_partition

Correct and the code looks good enough.

svartkanin (approved):

Issue #3185, opened 2025-02-19
Can someone explain the situation with using /efi as the EFI partition


michaeluk01:
https://wiki.archlinux.org/title/EFI_system_partition

The arch wiki recommends /efi as the EFI system partition because:

```
- mount the ESP to /efi:
- It ensures a separation of concerns between OS- and EFI-related files, which may include other OSes' files better left alone.
- It avoids increasing the size requirement of the ESP by not placing the files installed to /boot in it: only the EFI binaries (the boot loader (and optionally drivers) and/or the unified kernel image) will be stored on the ESP, which saves storage space.
- It allows to preserve Linux-specific filesystem permissions for files residing in /boot, avoiding FAT limitations.
- It allows to mount separately the ESP according to the need, e.g. only when upgrading the boot loader.
- If using system encryption with the appropriate setup, it allows to leave only a few required files unencrypted while /boot remains protected: this can be useful for unified kernel images or boot loaders that have file system drivers capable of accessing the kernel(s) and files that are stored elsewhere.
```

But it seems archinstall will fail using anything except /boot , is that correct?

Torxed:
It should work with `EFI` being mounted anywhere, but the default should be `/efi` *(or at least I think we changed to that a while back)*.

I know it sounds weird that the creator say this, but I'd have to double check because we've moved around the partition logic quite a bit the last year or so.

spsf64:
> https://wiki.archlinux.org/title/EFI_system_partition
> 
> The arch wiki recommends /efi as the EFI system partition because:
> 
> ```
>
Actually the first suggestion is /boot, you can check just above the quoted text you copied.
Also there is a note that says:

Note: If the ESP is not mounted to /boot, make sure to not rely on the [systemd automount mechanism]....

I always use the default /boot from archinstall, it works perfectly.

michaeluk01:
> > https://wiki.archlinux.org/title/EFI_system_partition
> > The arch wiki recommends /efi as the EFI system partition because:
> 
> Actually the first suggestion is /boot, you can check just above the quoted text you copied. Also there is a note that says:
> 
> Note: If the ESP is not mounted to /boot, make sure to not rely on the [systemd automount mechanism]....
> 
> I always use the default /boot from archinstall, it works perfectly.

I would assume they would both work though. There both recommended.

Choosing /boot means the kernel is on a fragile fat32 partition and has no protective permissions something none of the big distributions do. (Fedora, Debian, Ubuntu, Red Hat.)

[...]

PR #3272, opened 2025-03-19
More reliably identify ESP partition

mintsuki:


archinstall/lib/disk/partitioning_menu.py
```diff
@@ -288,8 +289,8 @@ def handle_action(
 					partition.mountpoint = self._prompt_mountpoint()
 					if partition.mountpoint == Path('/boot'):
 						partition.set_flag(PartitionFlag.BOOT)
-						if self._using_gpt:
-							partition.set_flag(PartitionFlag.ESP)
+					if self._using_gpt and partition.type_uuid == PartitionGUID.ESP.bytes:
+						partition.set_flag(PartitionFlag.ESP)
```
codefiles, line 293: Why add a check that the partition is the type that it is being set to? How would that make sense?

archinstall/lib/disk/partitioning_menu.py
```diff
@@ -513,9 +514,9 @@ def _create_new_partition(self, free_space: FreeSpace) -> PartitionModification:
 
 		if partition.mountpoint == Path('/boot'):
 			partition.set_flag(PartitionFlag.BOOT)
-			if self._using_gpt:
-				partition.set_flag(PartitionFlag.ESP)
-		elif partition.is_swap():
+		if self._using_gpt and partition.type_uuid == PartitionGUID.ESP.bytes:
+			partition.set_flag(PartitionFlag.ESP)
```
codefiles, line 518: Same thing as my other comment here.

archinstall/lib/disk/partitioning_menu.py
mintsuki: i believe those flags are internal flags, no?

codefiles: This is code for the setting a partition as an ESP. It could be a partition that has yet to be created and so does not exist.

mintsuki:
@codefiles what i was trying to achieve with the PR (and failed miserably at) is to tell archinstall to detect the ESP based on its filesystem type UUID rather than something arbitrary like mountpoint or filesystem that it is formatted as

mintsuki:
> @codefiles what i was trying to achieve with the PR (and failed miserably at) is to tell archinstall to detect the ESP based on its filesystem type UUID rather than something arbitrary like mountpoint or filesystem that it is formatted as

this being an attempt to get #3267 working

codefiles:
You are off target on the underlying cause. See https://github.com/archlinux/archinstall/issues/3185#issuecomment-2737684538

Issue #1591, opened 2022-12-13
Regular users added to wheel group


mcexit:
I noticed after using archinstall (2.5.2-1) on a recent Arch ISO (2022.11.01) that regular users were also added to the `wheel` group. The `wheel` group is automatically given superuser rights in a few components like `polkit` and is typically reserved for super users, for things like `sudo`.

ajburley:
+1 for this. I just found that after installing polkit my supposedly "non-root" user had root privileges! Such a security risk!

svartkanin:
I started on a fix for this, should be available soonish

Issue #1029, opened 2022-03-10
User credentials should allow specifying hashed password strings in the JSON files.


brettski74:
When creating automated installations, it's undesirable to have plain-text passwords floating around in our configuration files. It would be desirable to be able to specify a hashed password string similar to how the Ansible user module handles passwords. Maybe this could be done by having a different key for specifying the hashed password like "password-hash" and have the script use that preferentially and fall back to "!password" if the password hash was not specified.

brettski74:
I'm new to Arch, so it's possible I'm unaware of the scope of variability, but I don't think we need to solve all use cases at once. If we solve for the default case, that should be a good start. Support for other cases can be added later if there is value, but given that implementing other authentication schemes likely involves other additional work, I'm not sure it's clear what the solution should be for them or if there should be one in the stock setup.

I'm unfamiliar with `with installation.Boot() as session`, but I'm assuming that if there are use cases that need to know the plain-text password, using the fallback option should still allow that. My concern is mainly around setting up root and non-root users on a bunch of bare-metal machines that I want to be able to setup in a repeatable and maintainable way without needing to have plain-text passwords everywhere. The workaround I was planning was to set up all the accounts with dummy passwords and then have an Ansible playbook run as a custom command to update all the users to real passwords using the hashed passwords, but that seems inelegant.

As for sloppy users, there's only so much we can do. We already have plain-text passwords in the output, so the security concern is already there. Anything which provides an opportunity to remove that should be an improvement. I expect that less technical users are going to either go through the guided prompts anyway, or work from a JSON file generated from that. If the guided setup outputs a hashed password string instead of plain-text, then we've got them on the right track at least.  I'm assuming anyone who's rolling their own JSON files should know what they're doing already and understand the security implications and/or complexities of different authentication schemes.

svartkanin:
https://github.com/archlinux/archinstall/pull/3276 provides this now

PR #3276, opened 2025-03-20
Store password as hash instead of plaintext

svartkanin:
# DRAFT

Currently the defined user password is stored in plaintext in the `user_credentials.json` file if the configuration is saved. A security concern was originally reported in https://github.com/archlinux/archinstall/issues/1111 and another related issue to allow users to define hashes instead of passwords https://github.com/archlinux/archinstall/issues/1029. 

This PR is addressing some of the issues, based on this announcement in 2023-09-22 archlinux is using `yescrypt` as by default https://archlinux.org/news/changes-to-default-password-hashing-algorithm-and-umask-settings/.

So far the password for a user account has been set with `chpasswd username:password`, however, the command also allows to pass a yescrypt hash with `chpasswd --encrypted --crypt-method YESCRYPT`. 
We can use `mkpasswd --method=yescrypt --stdin` to create the respective password hash on the fly and set it for the user account. 

* This fixes issue #1111  #1029 
* Also fixes https://github.com/archlinux/archinstall/issues/1591
* Remove deprecated (since 2022)  handler for `!superuser` 
* Convert (with backwards compability) existing password configuration
Previously
```
{
    "!root-password": "super_pwd",
    "!users": [
        {
            "!password": "user_pwd",
            "sudo": true,
            "username": "user_name",
            "groups": [...]
        }
    ]
}
```

New 
```
{
    "root_enc_password": "password_hash",
    "users": [
        {
            "enc_password": "password_hash",
            "sudo": true,
            "username": "user_name",
            "groups": ["wheel"]
        }
    ]
}
```


codefiles:
Would you please create a separate pull request with just the fix for #1591? The fix for that is only a few lines (`-G wheel` conditional on `sudo` parameter). Thank you.

archinstall/lib/profile/__init__.py
```diff
@@ -1,9 +0,0 @@
-from .profile_menu import ProfileMenu, select_greeter, select_profile
-from .profiles_handler import profile_handler
-
-__all__ = [
```
svartkanin, line 4: @correctmost I know you just merged this change but I run into a circular dependency when loading profile configuration from a config file

svartkanin:
@Torxed should be ready for a review

codefiles:
If this is merged all user passwords created with archinstall will use bcrypt instead of the Arch Linux default of yescrypt. If Arch Linux staff decided on yescrypt it should not be up to archinstall to change this. For that reason this gets a no from me.

svartkanin:
I'm aware of the documentation and that the PR deviates from the default behaviour. The reason default values exist is to provide a single set option for a user without having to let them choose between multiple options that would all be viable choices. Same in this case, Arch had to pick a single option to go with as default, but that doesn't mean that other options are necessarily worse. `bcrypt` is widely used as a password hashing algorithm and has been battle proven. 

As stated in the description I couldn't find a way to either use PAM or create `yescrypt` hashes with the provided tools, therefore I opted for a viable alternative, which may violate the "Arch approach" and in that case will require another round of brainstorming and exploring a way forward.

Regarding the issue #1111 at hand, it is correct that this PR does not solve the entire problem with plaintext passwords but only addresses the user account ones as also stated in the description. 

Looking at why archinstall has plaintext passwords stored in a file in the first place the main use case seems to be that users can point the installer to the previously generated configuration files and start the installation without having to manually go through all steps again. 

A potential option to workaround it completely could be: 
* Move non-sensitive config information (username, sudo etc.) into `user_configuration.json`
* Don't write a `user_credentials.json` anymore
  * Manual input -> If a configuration file is provided as argument the user will be prompted to provide a user account password if username is in configuration file and disk encryption password if disk encryption was setup
  * Add argument options for user account passwords and disk encryption passwords (if specified it will not prompt for those manually)
  * Provide alternative hooks to fetch the passwords from a storage location?

Torxed:
I'm having a look! Overall this is a very welcome change.
Let me have a stab at the yescrypt thing.

Torxed:
So I'm not an expert in C, but perhaps this could be a starting point:
```python
import ctypes
import ctypes.util

# Try to locate libxcrypt
lib_path = ctypes.util.find_library("xcrypt") or ctypes.util.find_library("crypt")
if not lib_path:
    raise RuntimeError("libxcrypt not found")

libxcrypt = ctypes.CDLL(lib_path)

libxcrypt.crypt.argtypes = [ctypes.c_char_p, ctypes.c_char_p]
libxcrypt.crypt.restype = ctypes.c_char_p

password = b"some password"
salt = b"$y$j9T$KbYPzGJXFPvMbHGhKSD8x/"

hashed = libxcrypt.crypt(password, salt)
```

Should generate something along the lines of:
```
b'$y$j9T$KbYPzGJXFPvMbHGhKSD8x/$Kip3ZdzbqIHxatfBqYQ1R5t1gHo2w1Z5U6pjNLI5Ed/'
```

Would this help?

svartkanin:
That's some serious black magic :)
I can try to work with that!

Given that ur does tge problem of the cleartext disk encryption password remains. Any thoughts on the option I mentioned to pass passwords via the arguments?

svartkanin:
Or we keep storing all in plaintext passwords and encrypt the credentials file with a master password that needs to be entered on save. And when the credentials file is passed we prompt for it? 
That way only 1 password prompt may appear on launch and no need for a solution for the - disk encryption password

Torxed:
> Or we keep storing all in plaintext passwords and encrypt the credentials file with a master password that needs to be entered on save. And when the credentials file is passed we prompt for it?

I like that approach! There's a overwhelmingly high chance that people will forget it ^^
We could also encrypt the file with a HSM, so there's multiple options here which is great.

The only drawback will be that we'll need to have to have additional handling of the different scenarios. But I like the idea of this :)

Basically a vault, and inside hashed credentials :)

svartkanin:
> I like that approach! There's a overwhelmingly high chance that people will forget it ^^
Indeed, or they will potentially use the same password as they gave their user account ;) 

> The only drawback will be that we'll need to have to have additional handling of the different scenarios. But I like the idea of this :)
I think the friction won't be too much, we will need some form of prompt when a user selects to save the credentials file, and then some handling of decryption when the file gets passed as arg 

> Basically a vault, and inside hashed credentials :)
Will we still need to hash the credentials inside? We can try to hash the user account password with your approach above but the disk encryption one will be in clear text

I don't have access to the ISO atm but looking at https://gitlab.archlinux.org/archlinux/archiso/-/blob/4b709bcd5fc5e426ba75f72e1b80e59922d7c175/configs/releng/packages.x86_64 I can't see `libxcrypt` being explicitly included, @Torxed do you know if it's a transient dependency and available on the ISO? I suspect it must be as `chpasswd` hashes passwords...

Torxed:
> Will we still need to hash the credentials inside? We can try to hash the user account password with your approach above but the disk encryption one will be in clear text

I think we should, in case we want to offer unencrypted JSON files for credentials in the future and while passing to the command-line *(remember that the password if using `--chpasswd --encrypted` will be visible in `cmd_history.txt`, which is fine.. but having it hashed would make things better)*.

Also, the disk password is fine. But we should perhaps do *(a bit of pseudo code)*:
```
with tempfile.TemporaryFile() as tmp_file:
    luks2(pwfile=tmp_file)
```

As it would lower the risk of us forgetting it somewhere.

> I can't see libxcrypt being explicitly included, @Torxed do you know if it's a transient dependency

It is, via: [base](https://archlinux.org/packages/core/any/base/) <- [shadow](https://archlinux.org/packages/core/x86_64/shadow/) <- [libxcrypt](https://archlinux.org/packages/core/x86_64/libxcrypt/)

But just to be sure, for instance in docker containers, we should add it as an explicit dependency. It also allows other packages to explicitly see that we depend on it so any updates to that library will tell them to notify us about any breaking changes :)

svartkanin:
Okay sounds good, in that case I'll rework the PR to switch to the yescrypt hash instead of bcrypt. And in a future PR address the encryption of the file and potential arguments.

svartkanin:
@Torxed Update the PR with the `yescrypt` hash generator

archinstall/lib/models/users.py
```diff
@@ -103,72 +107,150 @@ def _check_password_strength(
...
+		that actually provides the functionality
+		https://github.com/rfc1036/whois/blob/next/mkpasswd.c
+		"""
+		enc_plaintext = plaintext.encode('UTF-8')
+		salt = self._gen_yescrypt_salt()
 
-			if password:
-				return [User(username, password, sudo)]
+		yescrypt_hash = self._libcrypt.crypt(enc_plaintext, salt)
+		return yescrypt_hash.decode('UTF-8')
```
codefiles, line 193: As I have shown from documentation (https://github.com/archlinux/archinstall/pull/3276#issuecomment-2746215504), it is PAM that is responsible for encrypting passwords.

Here is the relevant code:
https://github.com/linux-pam/linux-pam/blob/3da5496be24031a5c806752d9f1892100346d8ff/modules/pam_unix/passverify.c#L494-L514

The rounds variable is handled here:
https://github.com/linux-pam/linux-pam/blob/3da5496be24031a5c806752d9f1892100346d8ff/modules/pam_unix/support.c#L189-L244

So for rounds the [`YESCRYPT_COST_FACTOR` is read in](https://github.com/linux-pam/linux-pam/blob/3da5496be24031a5c806752d9f1892100346d8ff/modules/pam_unix/support.c#L193) from [`/etc/login.defs`](https://github.com/shadow-maint/shadow/blob/b4beda846c4bd2c54d2b246ec18948acc47502c1/etc/login.defs#L374-L377), [set to a default of `5`](https://github.com/linux-pam/linux-pam/blob/3da5496be24031a5c806752d9f1892100346d8ff/modules/pam_unix/support.c#L216-L217) if it is not set, and has [enforcement of a sane value](https://github.com/linux-pam/linux-pam/blob/3da5496be24031a5c806752d9f1892100346d8ff/modules/pam_unix/support.c#L225-L229).

The code you wrote has rounds set to `9`.

This is from just a quick glance over this. If you are going to emulate this, please put in the proper time to understand it so that the implementation is correct.

svartkanin: Thanks for pointing that out, I'll update the rounds accordingly

codefiles: Also recheck the rest of this logic against the relevant PAM code. I was only using rounds to demonstrate that the code does not properly emulate the handling done by PAM.

archinstall/default_profiles/desktops/awesome.py
```diff
@@ -9,7 +9,7 @@
 
 class AwesomeProfile(XorgProfile):
 	def __init__(self) -> None:
-		super().__init__('Awesome', ProfileType.WindowMgr, description='')
+		super().__init__('Awesome', ProfileType.WindowMgr, description='TETESTSKJTLKJLKTJLKJ')
```
Torxed: Poor awesome, living the test life ^^

archinstall/lib/models/users.py
Torxed: To be fair, that's on me as well seeing as I laid the foundation for this code. 

I believe that if we can set sane defaults that work with the login after an install, then that's good enough for a v1.0 of this. 
As the overall benefit from using hashes outweigh 1:1 code logic with PAM or whomever, as long as it works. Separate PR's can be raised for such cases.

codefiles: I assume you got https://github.com/archlinux/archinstall/pull/3276#issuecomment-2747500628 from an LLM chat bot.

codefiles:
If a user interacting with archinstall through the TUI inputs a password for a root or user account, will the password be encrypted using this code and bypass the process in which PAM handles encryption of the password even if the user does not intend to save and reuse credentials?

archinstall/lib/models/users.py
```diff
@@ -103,72 +107,150 @@ def _check_password_strength(
...
+		libc.crypt_gensalt.restype = ctypes.c_char_p
+
+		return libc
+
+	def _gen_yescrypt_salt(self) -> bytes:
+		salt_prefix = b'$y$'
+		rounds = 9
+		entropy = os.urandom(64)
+
+		salt = self._libcrypt.crypt_gensalt(salt_prefix, rounds, entropy, 64)
```
codefiles, line 179: From [`crypt_gensalt(3)`](https://man.archlinux.org/man/crypt_gensalt.3):

> char *
**crypt_gensalt**(_const char *prefix, unsigned long count, const char *rbytes, int nrbytes_);

> If rbytes is a null pointer, an appropriate number of random bytes will be obtained from the operating system, and nrbytes is ignored.

archinstall/lib/models/users.py
Torxed: Not really, why?

It was mostly copy paste from [old garbage](https://github.com/Torxed/Scripts/blob/df89b164f9e91411776fb4948ac805d7d0bf590c/python/check_homeperm.py#L23) I had laying around from 5 years ago plus random googling on this specific hash function parameters.

As I've never worked with yescrypt that was new to me if that's what you're fishing for.

codefiles: I thought it based on the appearance of the first example, the brevity of it and no functions. I took a look at the script you linked and admit I was wrong. Thanks for addressing it.

archinstall/lib/models/users.py
```diff
@@ -103,72 +107,150 @@ def _check_password_strength(
...
-		if len(config_users.keys()) > 0:
-			username = list(config_users.keys())[0]
-			password = config_users[username]['!password']
+		libc.crypt.argtypes = [ctypes.c_char_p, ctypes.c_char_p]
+		libc.crypt.restype = ctypes.c_char_p
+
+		libc.crypt_gensalt.argtypes = [ctypes.c_char_p, ctypes.c_int, ctypes.c_void_p, ctypes.c_size_t]
+		libc.crypt_gensalt.restype = ctypes.c_char_p
+
+		return libc
```
codefiles, line 172: Should this be module-level variables instead of using `@cached_property`?

Here is a use of cytpes in the python source code with module-level variables:
https://github.com/python/cpython/blob/305be5fb1a1ece7f9651ae98053dbe79bf439aa4/Lib/ctypes/util.py#L146-L148

archinstall/lib/models/users.py
codefiles: I recommend moving chunk of code to its own module, something like `crypt.py`, and using this python source code as an example of how the code should be structured:
https://github.com/python/cpython/blob/bfc292abc10091a9298027596d107652ee4fe166/Lib/_pyrepl/_minimal_curses.py

archinstall/lib/models/users.py
codefiles: See also https://github.com/archlinux/archinstall/pull/3276#discussion_r2029840111

svartkanin:
> If a user interacting with archinstall through the TUI inputs a password for a root or user account, will the password be encrypted using this code and bypass the process in which PAM handles encryption of the password even if the user does not intend to save and reuse credentials?

Correct, I opted for consistency to keep the hashing of passwords in a single place. I thought it might confuse users if they specify a password them PAM hashes and sets it in the shadow file and the config contains a different hash generated by archinstall.

codefiles:
The PKGBUILD will need `libcrypt.so` and `libxcrypt` as a dependencies. Both are in the PKGBUILD for PAM:
https://gitlab.archlinux.org/archlinux/packaging/packages/pam/-/blob/ace0ff539c0c8ff8e776757c5be273ffa376d865/PKGBUILD#L12-22

archinstall/lib/models/users.py
codefiles: An implementation of the above recommendation:

`libcrypt.py`

```python
import ctypes


class error(Exception):
    pass


libcrypt = ctypes.CDLL('libcrypt.so')

libcrypt.crypt.argtypes = [ctypes.c_char_p, ctypes.c_char_p]
libcrypt.crypt.restype = ctypes.c_char_p

libcrypt.crypt_gensalt.argtypes = [ctypes.c_char_p, ctypes.c_ulong, ctypes.c_char_p, ctypes.c_int]
libcrypt.crypt_gensalt.restype = ctypes.c_char_p


def crypt(phrase: bytes, setting: bytes) -> bytes:
    result = libcrypt.crypt(phrase, setting)
    if result is None:
        raise error('crypt() returned NULL')
    return result


def crypt_gensalt(prefix: bytes, count: int) -> bytes:
    setting = libcrypt.crypt_gensalt(prefix, ctypes.c_ulong(count), None, 0)
    if setting is None:
        raise error('crypt_gensalt() returned NULL')
    return setting
```

Simple use example:

```python
import sys

import libcrypt


def main() -> int:
    if len(sys.argv) < 2:
        return 2

    password = sys.argv[1]
    prefix = b'$y$'
    rounds = 5

    try:
        setting = libcrypt.crypt_gensalt(prefix, rounds)
        hashed = libcrypt.crypt(password.encode(), setting)
    except libcrypt.error:
        return 1

    print(hashed.decode())
    return 0


if __name__ == '__main__':
    sys.exit(main())

```

archinstall/lib/models/users.py
```diff
@@ -103,72 +107,150 @@ def _check_password_strength(
...
 
-	@classmethod
-	def _parse_backwards_compatible(cls, config_users: dict[str, dict[str, str]], sudo: bool) -> list['User']:
-		if len(config_users.keys()) > 0:
-			username = list(config_users.keys())[0]
-			password = config_users[username]['!password']
+		libc.crypt.argtypes = [ctypes.c_char_p, ctypes.c_char_p]
+		libc.crypt.restype = ctypes.c_char_p
+
+		libc.crypt_gensalt.argtypes = [ctypes.c_char_p, ctypes.c_int, ctypes.c_void_p, ctypes.c_size_t]
```
codefiles, line 169: The first argtype is correct but the rest are wrong. See the first quote from `crypt_gensalt(3)` in https://github.com/archlinux/archinstall/pull/3276#discussion_r2029082231 and https://github.com/archlinux/archinstall/pull/3276#discussion_r2029911717 for the correction.

svartkanin: Arguments have been fixed

archinstall/lib/models/users.py
svartkanin: Updated the call to generate the salt

svartkanin:
@codefiles Update the PR. 
Regarding the linked salt creation PAM attempts to read `/etc/login.defs` as you pointed out https://github.com/linux-pam/linux-pam/blob/3da5496be24031a5c806752d9f1892100346d8ff/modules/pam_unix/support.c#L200 looking at the logic it'll consider `YESCRYPT_COST_FACTOR` if it is not commented out. Looking at the current ISO this value is set to 

```#YESCRYPT_COST_FACTOR=5```

which means it'll be ignore and the default value of 5 applied. I've hardcoded it now to 5 instead of 9 but did not implement the actual reading of the value from `login.defs` (as that might change in the future). If you still consider that necessary I can add that functionality

codefiles:
The ISO just reflects how `/etc/login.defs` is provided by `shadow`. Yes I know, `YESCRYPT_COST_FACTOR` is commented out by default as provided by the shadow repository and Arch Linux does not change it. The point is that it is check for by PAM, that means if it is ever changed, your hard coded `5` will not match what is in `/etc/login.defs`. As it is unlikely to change, I am good with the code as it is now.

I would have preferred you keep `crypt()` and `crypt_gensalt()` as in my example because they were intended to be wrappers of those C functions. Good enough though.

svartkanin:
Thanks for the feedback, I've included attempting to read the value from `/etc/login.defs` with the respective handling of values as in https://github.com/linux-pam/linux-pam/blob/3da5496be24031a5c806752d9f1892100346d8ff/modules/pam_unix/support.c#L225-L229

PR #3283, opened 2025-03-21
Refactor creation of BLS entries

codefiles:


codefiles:
GRUB may get support for BLS entries.

https://lists.gnu.org/archive/html/grub-devel/2025-02/msg00018.html

svartkanin (approved):

PR #3293, opened 2025-03-22
Fix 1591 - add user to wheel only when sudo

### LINKED ISSUE #1591: Regular users added to wheel group

svartkanin:
Fixes #1591


codefiles:
Looks good to go. Draft?

Torxed (approved):

PR #3294, opened 2025-03-22
Fix regression for NTP sync

svartkanin:
Fixes regression found by @codefiles https://github.com/archlinux/archinstall/pull/3167#issuecomment-2741258451


codefiles:
That fixes it.

Torxed (approved):

PR #3302, opened 2025-03-25
fix(hardware): Improve Nvidia driver installation

lisichkavn:
Fixes a few issues related to the installation of Nvidia drivers.

1. No longer install the redundant nvidia-open package as it's provided by the nvidia-open-dkms package.
2. Install vulkan-nouveau when selecting the open-source nouveau driver.
3. Install the libva-nvidia-driver package for hardware accelerated video decoding.

### DIFF
--- a/archinstall/lib/hardware.py
+++ b/archinstall/lib/hardware.py
@@ -48,12 +48,13 @@ class GfxPackage(Enum):
 	IntelMediaDriver = 'intel-media-driver'
 	LibvaIntelDriver = 'libva-intel-driver'
 	LibvaMesaDriver = 'libva-mesa-driver'
+	LibvaNvidiaDriver = 'libva-nvidia-driver'
 	Mesa = "mesa"
 	NvidiaDkms = 'nvidia-dkms'
-	NvidiaOpen = 'nvidia-open'
 	NvidiaOpenDkms = 'nvidia-open-dkms'
 	VulkanIntel = 'vulkan-intel'
 	VulkanRadeon = 'vulkan-radeon'
+	VulkanNouveau = 'vulkan-nouveau'
 	Xf86VideoAmdgpu = "xf86-video-amdgpu"
 	Xf86VideoAti = "xf86-video-ati"
 	Xf86VideoNouveau = 'xf86-video-nouveau'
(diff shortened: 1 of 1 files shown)


Torxed:
If I'm not mistaken, `nvidia-open` is not redundant as it will speed up both the installation and updates for anyone running the `linux` kernel.
The `nvidia-open-dkms` is only for non-"standard" kernels.

lisichkavn:
> If I'm not mistaken, `nvidia-open` is not redundant as it will speed up both the installation and updates for anyone running the `linux` kernel. The `nvidia-open-dkms` is only for non-"standard" kernels.

You're correct in thinking that it'd be nice to only have to use `nvidia-open-dkms` when installing something like `linux-zen`, but since the `nvidia-open-dkms` package provides `nvidia-open`, only one can be installed at a time. Maybe I'm misunderstanding something?

Image of the nvidia-open-dkms package on the Arch Linux website

Torxed:
Yeah, we've been tinkering with that frickin driver for a while heh.
But the packages blocking each other makes sense, as it simply instructs you to choose one or the other.

In our case, we want to use `nvidia-open` for all default scenarios, unless `linux-hardened` is selected for instance.

But yeah, our solution back then was lazy to just say "you know what, lets just do `-dkms` and let it cook for a while". Not proud of it, but that's why.
I guess we could do the same here for now.

Any objections from anyone else? Otherwise we can merge and carry on with hopefully fix this in the future :)

lisichkavn:
> I believe we'd need linux-headers as well tho for this dkms package.

It appears to install fine on my actual machine and a fresh virtual machine without `linux-headers`. The current proprietary driver installation in archinstall doesn't seem to pull in `linux-headers` manually either, so we should be fine here? If it's needed, pacman should pull it down as a dependency for us anyway.

At a quick glance, maybe in the future we could have a parameter in the `gfx_packages` function to indicate if the user has selected an alternative kernel? Would be fairly simple, but that's for another time.

Torxed:
The kernel selection is already there, in the `Installer.packages` (pseudo code), so we should be able to detect it. I just haven't hard the time to put the effort in tbh.

Regarding `linux-headers`, most likely it's down to luck.
They've been pulled in by some other package. And sadly, they also need to be specific to the kernel selected. `linux-headers` for `linux` and [linux-lts-headers](https://archlinux.org/packages/core/x86_64/linux-lts-headers/) for the lts kernel etc.

We could wing it, but there's a chance that the `mkinitcpio -P` step will fail.

Torxed:
> After taking another quick look, it seems `linux-headers` is already pulled in manually by archinstall in another place that I completely missed. See: https://github.com/archlinux/archinstall/blob/master/archinstall/lib/profile/profiles_handler.py#L234
> 
> This should cover our bases regarding the concerns about this patch specifically, though a better long term solution for this issue would be nice.

Funny thing is, I missed it too. I was almost sure we did, but couldn't find it >_O

Let's get this in, and we can test it!

Torxed (approved):

PR #3303, opened 2025-03-26
Detect boot partition regardless of boot flags if mounted on /boot

mintsuki:
## PR Description:

The boot partition is always to be mounted to `/boot` (afaik). This allows skipping potentially misleading checks of partition type to determine whether to use a certain partition as `/boot`, if it is already mounted there.

## Tests and Checks

### DIFF
--- a/archinstall/lib/models/device_model.py
+++ b/archinstall/lib/models/device_model.py
@@ -964,8 +964,9 @@ class PartitionModification:
 	def is_boot(self) -> bool:
 		"""
 		Returns True if any of the boot indicator flags are found in self.flags
+		or if the partition is mounted on /boot
 		"""
-		return any(set(self.flags) & set(self._boot_indicator_flags))
+		return any(set(self.flags) & set(self._boot_indicator_flags)) or Path('/boot') == self.mountpoint
 
 	def is_root(self) -> bool:
 		if self.mountpoint is not None:


mintsuki:
@svartkanin request review, please.

svartkanin:
Looks good from my side, I believe the original code was implemented by @codefiles so I will wait for a potential review otherwise happy to merge in a day or so

codefiles:
@Torxed, not me.

https://github.com/archlinux/archinstall/blame/c4bea107b08615486ea70b9c892eb0f4db8867fd/archinstall/lib/models/device_model.py#L968

Torxed:
```python
if self.mountpoint is not None:
	return Path('/boot') == self.mountpoint
else:
	return any(set(self.flags) & set(self._boot_indicator_flags))
```

I'm wondering if it's wiser to check the flags first? And then use the path as a fallback?

```python
return any(set(self.flags) & set(self._boot_indicator_flags)) or Path('/boot') == self.mountpoint
```

It seems strange that the boot flags used in `_boot_indicator_flags` isn't set correctly, but the mount point would be?

mintsuki:
@Torxed 

> It seems strange that none of the boot flags from _boot_indicator_flags is set correctly in self.flags, but the mount point would be?

But it is a real scenario that does happen when using a pre-mounted configuration, and in general the `_boot_indicator_flags` are quite fragile from my testing. They expect certain partition types to be set for a `/boot` partition to be recognised as such, even when a user is explicitly mounting it on `/boot`...

IMO if there is a partition mounted on `/boot` already, it should be assumed that that is the boot partition, regardless of whether it has the right flags set or not.

Other than that, the point being raised about the order of checks and the issue with the else block not being entered under certain conditions is perfectly valid, and I will update the PR to use the line of code you shared.

Torxed:
Thanks for being accommodating, I agree that it might be fragile.
Curious why the flags aren't detected on mounted volumes, but I don't have the time to debug why currently. So lets get this in :)

Torxed (approved):

Issue #3307, opened 2025-03-29
[master] Cannot list packages or add "Additional packages"


Torxed:
```python
import archinstall.lib.models.packages
import archinstall.lib.packages

print(archinstall.lib.packages.list_available_packages((
    archinstall.lib.models.packages.Repository.Core,
    archinstall.lib.models.Repository.Extra
)))
```

> {}

Built on the latest ISO candidate. Can also be reproduced by just opening the "Additional packages" menu entry in the TUI.

~~Not sure why yet, but opening the issue as it's a blocker for release.~~

Most likely this is because we haven't ran a `pacman -Syy` before trying to list the packages, as that solves the problem manually.

PR #3311, opened 2025-03-30
Made sure to update the package databases before running --info

Torxed:
Fixes: https://github.com/archlinux/archinstall/issues/3307


svartkanin:
@Torxed question on this one, when starting archinstall there's a `pacman -Sy` run unless `--offline` is specified https://github.com/archlinux/archinstall/blob/master/archinstall/__init__.py#L56. Isn't that already making sure we have the latest?

Issue #3308, opened 2025-03-29
[master] Allows for starting installation without `root` or `user` accounts configured


Torxed:
Blocks a release, as it could cause users to install machines without any accounts to use after installation.

PR #3312, opened 2025-03-30
Fixed root_password and users[superuser=True] check before install

Torxed:
This fixes: https://github.com/archlinux/archinstall/issues/3308


svartkanin:
Thanks for fixing this, I missed that

PR #3327, opened 2025-04-02
replace GTK theme package for Budgie build with an active one

EK-92:
- This fix issue:

```error: target not found: arc-gtk-theme```

image

## PR Description:

Ran into an issue with a fresh ISO trying to get Arch Budgie on my machine. A quick search let to a conclusion that the theme package seized to exist, and is only now maintained as a fork on AUR. Materia is but one option of many available, can be a different one as default, just something within `extra` repository, adding a list of currently available below the diff.

```
@override
		return [
			"arc-gtk-theme", => "materia-gtk-theme",
			"budgie",
			"mate-terminal",
			"nemo",
```
```
extra/deepin-gtk-theme 2020.06.10-2 (300.6 KiB 3.5 MiB) [deepin] 
    Deepin GTK Theme
extra/adw-gtk-theme 5.6-1 (101.8 KiB 1.8 MiB) 
    Unofficial GTK 3 port of the libadwaita theme
extra/lxde-icon-theme 0.5.2-1 (4.4 MiB 4.4 MiB) [lxde lxde-gtk3] 
    LXDE default icon theme based on nuoveXT2
extra/adapta-gtk-theme 3.95.0.11-3 (696.4 KiB 17.2 MiB) 
    An adaptive Gtk+ theme based on Material Design Guidelines
extra/materia-gtk-theme 20210322-3 (339.0 KiB 6.0 MiB) 
    A Material Design theme for GNOME/GTK+ based desktop environments
extra/orchis-theme 2024_11_03-4 (330.2 KiB 9.5 MiB) 
    A Material Design theme for GNOME/GTK based desktop environments
extra/pop-gtk-theme 5.5.1.r7.25ea85d9-1 (551.1 KiB 3.3 MiB) 
    System76 Pop GTK+ Theme
extra/breeze-gtk 6.3.3-1 (192.4 KiB 1.2 MiB) [plasma]
    Breeze widget theme for GTK 2 and 3
extra/gtk-theme-elementary 8.2.0-1 (83.2 KiB 10.8 MiB) [pantheon] 
    elementary GTK theme
```

## Tests and Checks

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -41,6 +41,7 @@ makedepends=(
 )
 optdepends=(
   'python-systemd: Adds journald logging'
+  'materia-gtk-theme: theming for GTK-based desktop environments'
 )
 provides=(python-archinstall archinstall)
 conflicts=(python-archinstall archinstall-git)
--- a/archinstall/default_profiles/desktops/budgie.py
+++ b/archinstall/default_profiles/desktops/budgie.py
@@ -12,7 +12,7 @@ class BudgieProfile(XorgProfile):
 	@override
 	def packages(self) -> list[str]:
 		return [
-			"arc-gtk-theme",
+			"materia-gtk-theme",
 			"budgie",
 			"mate-terminal",
 			"nemo",


Torxed (approved):
Awesome, we should add this to the PKGBUILD as a dependencie as well. Mind updating the PR? :)

EK-92:
Added to PKGBUILD as requested, in optdepends.
But now I'm doubtful whether there's a need for it at all since budgie-desktop pulls `gnome-themes-standard` as a dependency so it's more like a good-to-have rather than an actual requirement (unless I'm wrong of course).
An alternative fix would be just not using any theme that isn't explicitly requested by user in custom packages, and relying on gnome themes, unless it doesn't work well with Budgie?

Torxed (approved):

Torxed:
The reason I want it in the PKGBUILD is to create a hard link for packages. So that if someone changes meta or group packages, then we can be notified. Currently all this doesn't work as I would hope, but it's something and allowes us to not rely on group or meta packages.

codefiles:
The package `materia-gtk-theme` is not a dependency or optional dependency of `archinstall` and does not belong in the PKGBUILD. @Torxed, the change here was not related to meta or group packages.

Issue #2585, opened 2024-07-22
Unable to reuse existing /home partition when uid is not the default one (1000)


hublol:
### Which ISO version are you using?

2024-07-01

### The installation log

```json
Installing packages: ['pipewire', 'pipewire-alsa', 'pipewire-jack', 'pipewire-pulse', 'gst-plugin-pipewire', 'libpulse', 'wireplumber']
['/usr/bin/arch-chroot', '/mnt/archinstall', 'su', '-', 'halol', '-c', 'systemctl enable --user pipewire-pulse.service'] exited with abnormal exit code [1]: su: warning: cannot change directory to /home/halol: Permission denied
-bash: /home/halol/.bash_profile: Permission denied
[0;1;31mFailed to enable unit /home/halol/.config/systemd/user/default.target.wants/pipewire-pulse.service: Permission denied[0m

[0;1;31mFailed to enable unit /home/halol/.config/systemd/user/sockets.target.wants/pipewire-pulse.socket: Permission denied[0m
```

### describe the problem

archinstall version 2.8.1

I am running into an error when trying to re-use an existing /home partition without wiping it.

The problem comes from the fact that the user that must be created,
and that is the current owner of the files in /home,
doesn't have the default uid 1000.

Since the uid cannot be specified, archinstall recreate the user with the default uid. This then cause archinstall to fail when setting up pipewire (see log file):

So, it would be handy to be able to specify the uid of the users that are created (can be provided in the user_credentials.json), both because it would be nice to have feature, and because it would help avoid this problem.

By the way, re-using an existing /home partition i
[install.log](https://github.com/user-attachments/files/16337493/install.log)
s not something well documented. It would also be very nice to have a small example of how to do this in the documentation (I am not completely confident that I am doing this right).

Thanks a lot!

kikoucalou:
" By the way, re-using an existing /home partition is not something well documented. It would also be very nice to have a small example of how to do this in the documentation (I am not completely confident that I am doing this right). "

Perhaps, it woud be possible for developpers to add a safe option as "re-using an existing /home partition" in archinstall ?

Torxed:
Nice catch, we'll have to improve in this area. Both in terms of documentation and error handling.

PR #3328, opened 2025-04-02
Force to make pipewire symlinks (#2585)

utuhiro78:
- This fix issue: #2585

## PR Description:

Archinstall causes errors if we use an existing home directory.
We need to change 'ln -s' to 'ln -sf' to make pipewire symlinks.

## Tests and Checks

I tested the code on EndeavourOS.
I could install Arch Linux without problems.


Torxed (approved):

Torxed:
Good catch, merging!

PR #3336, opened 2025-04-03
Improve partition type detection, assignment, and misconfig handling

mintsuki:


archinstall/lib/disk/partitioning_menu.py
```diff
@@ -296,9 +308,13 @@ def handle_action(
...
 					partition.invert_flag(PartitionFlag.BOOT)
-					if self._using_gpt:
-						partition.invert_flag(PartitionFlag.ESP)
+				case 'mark_esp':
+					if PartitionFlag.XBOOTLDR in partition.flags:
+						partition.invert_flag(PartitionFlag.XBOOTLDR)
+					partition.invert_flag(PartitionFlag.ESP)
 				case 'mark_xbootldr':
+					if PartitionFlag.ESP in partition.flags:
+						partition.invert_flag(PartitionFlag.ESP)
```
codefiles, line 317: Could you explain this?

codefiles: Can this instead avoid all invalid combinations? For example, if the SWAP flag was set this would still allow adding the ESP or XBOOTLDR flag without removing the SWAP flag.

mintsuki: Done that.

svartkanin (approved):

PR #3344, opened 2025-04-04
Start the boot partition at 1MiB by default on MBR too

mintsuki:


codefiles:
There is a comment regarding this but it is in the wrong location since it was never moved when code was refactored.

https://github.com/archlinux/archinstall/blob/f5a16307bd78a927fcf216c9d7638da75f81dced/archinstall/lib/interactions/disk_conf.py#L364-L365

Here is the relevant section of the ArchWiki:
https://wiki.archlinux.org/title/GRUB#Master_Boot_Record_(MBR)_specific_instructions

mintsuki:
This PR sets the start of the partition to 1MiB (2048 sectors) which is perfectly fine for use with GRUB. I am not sure why it was done the way it was before, but it seems like a misunderstanding of what the Arch wiki is saying.

I will remove that comment if it is now out of date.

codefiles:
Could you explain the misunderstanding? This is from the section of the ArchWiki I linked above:

> Usually the post-MBR gap (after the 512 byte [MBR](https://wiki.archlinux.org/title/MBR) region and before the start of the first partition) in many MBR partitioned systems is 31 KiB when DOS compatibility cylinder alignment issues are satisfied in the partition table. However a post-MBR gap of about 1 to 2 MiB is recommended to provide sufficient room for embedding GRUB's `core.img` ([FS#24103](https://bugs.archlinux.org/task/24103)).

mintsuki:
As per this PR, the start of the boot partition will be set to 2048 sectors (aka 1MiB), the same as it is done for GPT partition tables, and the same as most partitioning tools will do for either MBR or GPT.

Before this PR the start of the boot partition was set to 3MiB, which is neither the recommendation of the Arch wiki (which is 1 to 2MiB, for some reason, even though GRUB works perfectly fine with a standard start of 2048 sectors), and is quite a weird place to have the first partition start to begin with (which even in archinstall's own partition table editor has this gap end up looking like a "free space" between 2048 sectors and 3MiB).

codefiles:
It says 1 to 2 MiB **gap**, not 1 to 2 MiB **start**, I think you have to account for the the MBR region too. The linked issue on the ArchWiki is about 14 years old so maybe the ArchWiki is out-of-date. I am just bringing to light the reason this was done.

mintsuki:
The MBR itself is 1 sector, the "gap" is 32KiB minus 1 sectors, at least the conventionally guaranteed to be present gap.

The 1MiB alignment for the start of the first partition is common convention, and it works with any reasonable BIOS bootloader. The reason I conflated the gap size with the start of the first partition, above, is that 1 or 2MiB minus 512 bytes (1 sector) is still about 1 or 2MiB.

codefiles:
I thought the post-MBR gap might have been 1 MiB aligned starting at sector 2048. I checked and `core.img` starts at sector 1 directly after the 512 bytes (sector 0) for the MBR. Thanks for helping me understand this and sorry for using up your time with this back and forth.

mintsuki:
No worries at all!

Torxed (approved):

PR #3388, opened 2025-04-12
Split out unicode_ljust and unicode_rjust to break import cycle

correctmost:
## PR Description:

Previously, there was an import cycle between `tui.menu_item` and `lib.output`:

```
archinstall/lib/output.py: error: Cycle detected in import chain
  archinstall/lib/output.py
  archinstall/tui/curses_menu.py
  archinstall/tui/menu_item.py (reportImportCycles)
```


svartkanin (approved):

archinstall/lib/unicode.py
```diff
@@ -0,0 +1,40 @@
+import unicodedata
```
svartkanin, line 1: @correctmost would it make sense to put this into the `utils.py` rather to avoid too many modules with utility functions?

svartkanin: yeah lets move it in there that will do

svartkanin (approved):

PR #3389, opened 2025-04-12
Remove unused dependencies_not parameter from MenuItem

correctmost:
## PR Description:
Just some small clean-up as I try to add stronger typing to `MenuItem`.


svartkanin:
I wouldn't mind leaving this around as it was used in the past but due to some refactoring it wasn't needed anymore for the time being, however it could be useful again in the future

PR #3391, opened 2025-04-13
Add optional file encryption for user credentials configuration

### LINKED ISSUE #1111: insecure password storage

svartkanin:
This is a follow-up PR for https://github.com/archlinux/archinstall/pull/3276 to fix #1111. 

It'll add a option to store the data in the `user_configuration.json` file in encrypted form. While saving the configuration files there's a new optional step to encrypt the content and if chosen provide an encryption key. The newly introduced `cryptography` module will handle the encryption by generating a salt and using Argon2id to encrypt the JSON data with the provided password. 

The stored encryption data will look like
```
$argon2id$<salt>$<encrypted_data>
```

When providing an encrypted `user_credentials.json` file, there are 3 ways to provided the encryption key:
* Via a new argument parameter `--creds-decryption-key` 
* By setting an environment variable `ARCHINSTALL_CREDS_DECRYPTION_KEY`
* If none of the above is specified a prompt on startup will be presented to manually enter the encryption key


archinstall/lib/args.py
```diff
@@ -391,12 +414,58 @@ def _parse_config(self) -> dict[str, Any]:
...
 
 		config = self._cleanup_config(config)
 
 		return config
 
+	def _process_creds_data(self, creds_data: str) -> dict[str, Any] | None:
+		if creds_data.startswith('$'):  # encrypted data
+			if self._args.creds_decryption_key is not None:
+				try:
+					password = Password(plaintext=self._args.creds_decryption_key)
```
codefiles, line 429: The password is initialized with `plaintext` and then is only ever used as plaintext.

archinstall/lib/args.py
```diff
@@ -391,12 +414,58 @@ def _parse_config(self) -> dict[str, Any]:
...
+				with Tui():
+					while True:
+						header = str(_('Incorrect password')) if incorrect_password else None
+
+						decryption_pwd = get_password(
+							text=str(_('Credentials file decryption password')),
+							header=header,
+							allow_skip=False,
+							skip_confirmation=True
+						)
```
codefiles, line 451: `get_password()` initializes `Password` the same way, with `plaintext`.

archinstall/lib/crypt.py
```diff
@@ -69,3 +75,52 @@ def crypt_yescrypt(plaintext: str) -> str:
...
+		iterations=1,
+		lanes=4,
+		memory_cost=64 * 1024,
+		ad=None,
+		secret=None,
+	)
+
+	key = base64.urlsafe_b64encode(
+		kdf.derive(
+			password.plaintext.encode('utf-8')
```
codefiles, line 94: `_get_fernet()`, used by `encrypt()` and `decrypt()`, only uses `Password` for plaintext.

archinstall/lib/models/users.py
```diff
@@ -151,6 +149,10 @@ def __eq__(self, other: object) -> bool:
 
 		return self.enc_password == other.enc_password
 
+	def _encrypt(self, plaintext: str) -> str:
+		from ..crypt import crypt_yescrypt
+		return crypt_yescrypt(plaintext)
+
```
codefiles, line 155: This looks like you are trying to fix the import cycle you have created with use of the `Password` class.

archinstall/lib/configuration.py
```diff
@@ -201,10 +217,33 @@ def preview(item: MenuItem) -> str | None:
...
+	).run()
+
+	password: Password | None = None
+	match result.type_:
+		case ResultType.Selection:
+			if result.item() == MenuItem.yes():
+				password = get_password(
+					text=str(_('Credentials file encryption password')),
+					allow_skip=True
+				)
```
codefiles, line 241: Same as other comment about `get_password()`.

codefiles (commented):
What is the reason for using the `Password` class and not just a `str` throughout? When it is being used in these changes, it is initialized with `plaintext` and then is only ever used as plaintext. This looks like unnecessary complexity.

archinstall/lib/translationhandler.py
```diff
@@ -190,10 +191,7 @@ def __add__(self, other) -> DeferredTranslation:
 	def format(self, *args) -> str:
 		return self.message.format(*args)
 
-	@classmethod
-	def install(cls) -> None:
-		import builtins
-		builtins._ = cls  # type: ignore[attr-defined]
 
+builtins._ = DeferredTranslation  # type: ignore[attr-defined]
```
codefiles, line 195: How is this change relevant to the pull request?

svartkanin: To be able to use the translation `_` in the args.py file it needs to be initialized which is done in the __init__.py file which is an explicit call AFTER all imports are resolved. 
But because the args.py file is loaded before that happens the `_` has not been registered in the builtins yet and therefore it won't be available for the initial decryption prompt.

archinstall/lib/args.py
svartkanin: Correct, I wanted to keep it consistent to use the wrapper data class and now throw around native types. It will make it easier in the future to perform actions on it if necessary and also easier figure out where passwords are used based on the type

codefiles: The wrapper data class is for passwords that are encrypted in a different manner, none of its features are used. It is perfectly fine to use native types. Everything looks complete here. I was not aware that types are intended for improved searching.

svartkanin: Fair point, I'll update it

svartkanin:
@Torxed any concerns with this or can we merge it?

Torxed:
Looks good to me!

Torxed (approved):

PR #3393, opened 2025-04-15
Fix 1747 | Additional post-installation menu options

### LINKED ISSUE #1747: Option to directly reboot after installation
This a suggestion.

Can we create a option on what to do after the installation.
```
- Post-Install
 - Reboot
 - Chroot (Into installed system)
 - Exit (Into the Archiso)
```
So when for example I setup a VM, I can just close the virt-viewer, and access it via ssh.

svartkanin:
Fixed #1747 

The current post-installation menu only asks for chroot, with this change there are more user friendly options displayed 
```
> Exit archinstall
   Reboot system
   chroot into installation for post-installation configurations
```

I removed the printing of the link https://wiki.archlinux.org/index.php/Installation_guide#Post-installation as there isn't really much information available there anyways, happy to put it back if there's a usability for it

### DIFF
--- a/archinstall/lib/interactions/general_conf.py
+++ b/archinstall/lib/interactions/general_conf.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from enum import Enum
 from pathlib import Path
 from typing import TYPE_CHECKING, assert_never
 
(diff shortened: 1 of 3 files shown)


Torxed:
Great addition. This allows us to ask for some (optional) post-installation optimizations as well in the future! Great work and good idea!

Torxed (approved):

PR #3400, opened 2025-04-23
Specify menu return types using generics instead of Any

correctmost:
## PR Description:

This commit reduces the number of `Any` instances in menu code.


Torxed (approved):

svartkanin:
Does the EditMenu need a type? It should always be a string

svartkanin (approved):

PR #3407, opened 2025-04-26
Swapping to python-uv for building and distribution

Torxed:
This will swap to [uv](https://docs.astral.sh/uv/) for building and distribution.
Mainly to evaluate, as `twine` has caused minor issues in the past.


svartkanin:
That's great, I started using it a while back and the speed is fantastic

svartkanin (approved):

Torxed:
I was also impressed with how fast it was, and how little hassle it was to use.

Issue #3412, opened 2025-04-27
Make package selector also look for packages in additional repos


OlegAckbar:
### describe the request

Currently, package selector only looks for packages in `core`, `extra`. If I enable `multilib` repo packages from it won't appear in package selector.

killertofus:
#3417 fixes it

killertofus:
#3417 fixes it

PR #3417, opened 2025-04-28
added multilib as a dep

killertofus:
- This fix issue: 
#3412
## Tests and Checks

### DIFF
--- a/archinstall/lib/interactions/general_conf.py
+++ b/archinstall/lib/interactions/general_conf.py
@@ -172,7 +172,7 @@ def ask_additional_packages_to_install(
 ) -> list[str]:
 	Tui.print(str(_('Loading packages...')), clear_screen=True)
 
-	repositories |= {Repository.Core, Repository.Extra}
+	repositories |= {Repository.Core, Repository.Extra, Repository.Multilib}
 	packages = list_available_packages(tuple(repositories))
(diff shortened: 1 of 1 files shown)


svartkanin:
Hardcoding a repository that is not necessarily specified in the selection is not the way to go. 
This needs to be dynamically passed into the arguments, but it goes beyond that as the multi lib repository needs to be actually enabled in the `/etc/pacman.conf` first and then synced. 
I started looking at this and it's more complicated, I'll try to get something working this week

archinstall/lib/interactions/general_conf.py
```diff
@@ -172,7 +172,7 @@ def ask_additional_packages_to_install(
 ) -> list[str]:
 	Tui.print(str(_('Loading packages...')), clear_screen=True)
 
-	repositories |= {Repository.Core, Repository.Extra}
+	repositories |= {Repository.Core, Repository.Extra, Repository.Multilib}
```
svartkanin, line 175: This will not work, as this is only a filter for listed packages

svartkanin, line 175: This will not work, as this is only a filter for listed packages

svartkanin (changes requested):

PR #3427, opened 2025-05-03
Fix 3412 - Take optional repositories into account

### LINKED ISSUE #3412: Make package selector also look for packages in additional repos

svartkanin:
Fixes #3412 by enabling optional repositories in the local `/etc/pacman.conf` if they are selected


svartkanin:
@Torxed this would fit in 3.0.5 :)

Torxed (approved):
Consider it fit! ^^

Issue #2447, opened 2024-04-06
Add wayfire as a desktop option


heheda123123:
Wayfire is one of the few stacked window managers in Wayland, and many people use it. You may consider adding it.
https://wayfire.org/

erikdubois:
![Image](https://github.com/user-attachments/assets/12d8d942-edfa-46fa-bcad-09698c29956b)

Is it possible something has been forgotten?

I do not believe these package are part of the Arch Linux repos.

Was it the intention to add other repos and was it forgotten or am I seeing it all wrong

Torxed:
Nice catch, I'll revert that addition.
`wayfire` is still in AUR which means we won't be able to support it.

Torxed:
Updated the release notes with:

* Wayfire support by @Mohamed-elg in https://github.com/archlinux/archinstall/pull/3020 (prematurely included in the release, but will be removed in 3.0.3 because it's an AUR package. The arch package archinstall-3.0.2-2 has this profile removed during packaging via a patch in the repositories and thus won't be selectable)

I won't sign a new github release, as I won't have time to do all the steps involved in reverting that specific commit in a clean way.

We can re-visit this once the package gets adopted to `extra`.

PR #3432, opened 2025-05-04
Fixes for release

svartkanin:
* Revert the unneeded opt dependency as pointed out by @codefiles https://github.com/archlinux/archinstall/pull/3327#issuecomment-2772605465
* Remove wayfire from the profiles as it's only available in AUR as pointed out by @Torxed https://github.com/archlinux/archinstall/issues/2447#issuecomment-2609350624


Torxed:
Looks good.
I don't entirely agree with this comment: https://github.com/archlinux/archinstall/pull/3327#issuecomment-2772605465

We **need** a way to link packages we hard code in our profiles, to upstream `PKGBUILD` packages. Otherwise we're on our own in terms of keeping track of package removals/name changes etc.

And having `optdepends` is a hacky way for sure, but it's almost the only way currently to link us to packages, so that packagers can get notified about their changes.

Torxed (approved):

codefiles:
Having them listed as optional dependencies implies that installing them on the host system will add additional features to the archinstall package.

https://wiki.archlinux.org/title/PKGBUILD#optdepends

Torxed:
> Having them listed as optional dependencies implies that installing them on the host system will provide additional features to the archinstall package. That is not the case.
> 
> https://wiki.archlinux.org/title/PKGBUILD#optdepends

I know, but again, due to the limitation of `alpm` and in extent `pacman`, after several discussions back and forth both on the Arch Linux summit as well as mailing lists and chat forums, this is basically our only option to link `archinstall`'s internal package dependencies with actual package hooks that packagers use to keep track of their effect on certain changes.

But if you have a better idea of how to create this weak link, I'm all ears.
Because `depends=()` is too much of a strong link, as it will automatically drag in all packages if you install `archinstall`.

PR #3449, opened 2025-05-08
Remove the separate Pipewire profile type

correctmost:
## PR Description:

This breaks an import cycle between applications/pipewire.py and models/audio_configuration.py:

```
  archinstall/default_profiles/applications/pipewire.py: error: Cycle detected in import chain
    archinstall/default_profiles/applications/pipewire.py
    archinstall/default_profiles/profile.py
    archinstall/lib/installer.py
    archinstall/lib/args.py
    archinstall/lib/models/audio_configuration.py (reportImportCycles)
```


Torxed:
I'm not sure if there was an actual import loop here?
And I personally think it will be easier to maintain if we keep the profiles split into profiles, and not within individual functions.

Torxed:
> Although the cycle doesn't cause runtime issues, it does suggest that there are design flaws.
> 
> The Profile interface has methods like is_graphic_driver_supported() and is_greeter_supported(), which don't seem relevant to Pipewire. I wonder if introducing a new class to handle application installs would be a good compromise between overusing Profile and adding more methods to Installer?

I agree that the design principles here are not ideal. And I do like your suggested approach of splitting up the `Profile` class. Perhaps something similar to:

```mermaid
graph TD;
    Profile-->Desktop;
    Profile-->Application;
```

They share a lot of overlap, but I agree that `is_graphic_driver_supported` doesn't make sense in a PipeWire profile :)

svartkanin:
I like that last suggestion @Torxed , I don't think having it in profiles makes sense as it's fundamentally something else, but having a new category application makes more sense.

This could be a different directory alongside profiles with many an interface class and exposed `def install(installer)` that each application needs to override.

PR #3450, opened 2025-05-08
Disable the loop-iterator-mutation rule in ruff

correctmost:
## PR Description:

This allows `ruff check --preview` to succeed again.

Here's the warning that is now suppressed:

```
archinstall/lib/pacman/config.py:46:6: B909 Mutation to loop iterable `content` during iteration
   |
44 |                 # also uncomment the next line (Include statement) if it exists and is commented
45 |                 if row + 1 < len(content) and content[row + 1].lstrip().startswith('#'):
46 |                     content[row + 1] = re.sub(r'^#\s*', '', content[row + 1])
   |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ B909
47 |
48 |         # Write the modified content back to the file
   |

Found 1 error.
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -216,6 +216,7 @@ ignore = [
     "B008",     # function-call-in-default-argument
     "B904",     # raise-without-from-inside-except
     "B905",     # zip-without-explicit-strict
+    "B909",     # loop-iterator-mutation
     "COM812",   # missing-trailing-comma
     "PLC0415",  # import-outside-top-level
     "PLC1901",  # compare-to-empty-string


codefiles:
Why not fix it rather than suppress it? Would this be preferable?

```patch
diff --git a/archinstall/lib/pacman/config.py b/archinstall/lib/pacman/config.py
index 6ea75d4e..948db2de 100644
--- a/archinstall/lib/pacman/config.py
+++ b/archinstall/lib/pacman/config.py
@@ -33,7 +33,10 @@ class PacmanConfig:

                content = self._config_path.read_text().splitlines(keepends=True)

-               for row, line in enumerate(content):
+               row = 0
+               while row < len(content):
+                       line = content[row]
+
                        # Check if this is a commented repository section that needs to be enabled
                        match = re.match(r'^#\s*\[(.*)\]', line)

@@ -44,6 +47,9 @@ class PacmanConfig:
                                # also uncomment the next line (Include statement) if it exists and is commented
                                if row + 1 < len(content) and content[row + 1].lstrip().startswith('#'):
                                        content[row + 1] = re.sub(r'^#\s*', '', content[row + 1])
+                                       row += 1
+
+                       row += 1

                # Write the modified content back to the file
                with open(self._config_path, 'w') as f:
```

This also enables skipping `content[row + 1]` after substitution has been performed on it.

svartkanin (approved):

PR #3455, opened 2025-05-11
UWSM for Hyprland

TSEPLNK:
Add UWSM.

uwsm is recommended to start hyprland and hyprland with uwsm is the default session.
Currrently, archinstall does not install the package as default.
Including UWSM will make out of the box hyprland experience better.

Note: This code wasn't tested, but its structure aligns with existing profiles.


Torxed (approved):
For reference, this is the recommended and preferred thing from hyprland

Issue #1682, opened 2023-03-16
Evaluate the use of Ruff for code formatting


Torxed:
In light of https://github.com/charliermarsh/ruff/issues/1904#issuecomment-1470946322
I'm intrigued to evaluate Ruff as a code formatter.

It looks like it could combat some of my quarrels with https://github.com/archlinux/archinstall/pull/997 *(Linked with issue https://github.com/archlinux/archinstall/issues/872)*.

Specifically the sentences:

> I wanted a tool that formats things consistently with a set of rules that I can explain in prose, and recognize in the output.

> YAPF caused its results to be sometimes hard to understand. No configuration file we tried saved us from very suboptimal edge cases

> I just wanted a tool that does one thing and does it well.

I ran a super default test and it doesn't appear to be very intrusive.
It can also be found in the WIP PR I opened: https://github.com/archlinux/archinstall/pull/1681
```
~/archinstall (eval-ruff)$ ruff check archinstall/*.py --fix
```

```diff
diff --git a/archinstall/__init__.py b/archinstall/__init__.py
index b22b466..b57b468 100644
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -6,7 +6,6 @@ from .lib.disk import *
 from .lib.exceptions import *
 from .lib.general import *
 from .lib.hardware import *
-from .lib.installer import __packages__, Installer, accessibility_tools_in_use
 from .lib.locale_helpers import *
 from .lib.luks import *
 from .lib.mirrors import *
@@ -14,37 +13,14 @@ from .lib.models.network_configuration import NetworkConfigurationHandler
 from .lib.models.users import User
 from .lib.networking import *
 from .lib.output import *
-from .lib.models.dataclasses import (
-       VersionDef,
-       PackageSearchResult,
-       PackageSearch,
-       LocalPackage
-)
-from .lib.packages.packages import (
-       group_search,
-       package_search,
-       find_package,
-       find_packages,
-       installed_package,
-       validate_package_list,
-)
 from .lib.profiles import *
 from .lib.services import *
 from .lib.storage import *
 from .lib.systemd import *
 from .lib.user_interaction import *
-from .lib.menu import Menu
-from .lib.menu.list_manager import ListManager
-from .lib.menu.text_input import TextInput
-from .lib.menu.global_menu import GlobalMenu
-from .lib.menu.abstract_menu import (
-       Selector,
-       AbstractMenu
-)
[...]

PR #3459, opened 2025-05-12
Ruff formatting

svartkanin:
Related issue: https://github.com/archlinux/archinstall/issues/1682

I know this topic has been up a few times so I thought it's time for another round :) 
Some previous discussion around using YAPF https://github.com/archlinux/archinstall/pull/997

As the code base has been integrated heavily with `ruff` it would make sense to also explore using the `ruff format` extension. 

I run a simple `ruff format archinstall/` to get a feeling of what it might look like and to collect feedback on what works and what looks funky. It seems there are some settings for the formatter but not to an extend of YAPF. 

Documentation: https://docs.astral.sh/ruff/settings/#format

@Torxed any feedback would be great :) 

# Edit
Based on the discussions below, the general consent is to use `ruff format` for now as the main formatting tool. 
This PR therefore 
* Adds ruff formatter definition in `pyproject.toml` 
* Formats all files 
* Adds pre-commit functionality to auto-format on git commit
* Adds a git action to enforce the formatting


svartkanin:
I'm not really a fan of things like 
https://github.com/archlinux/archinstall/pull/3459/files#diff-4d8c19c70790b303e8338fbe439d7203c23bb2247a6b0c08cc995d0ca8271855L54-L58

I find the one item per line much for readable

svartkanin:
I was really hoping there would be more settings but it feels much like `black` in that sense

svartkanin:
@correctmost thanks for that, I pushed an update with that and it seems to have done a bit better. But there are still some things that got moved into a single line such as 

https://github.com/archlinux/archinstall/pull/3459/files#diff-ddf0aac257309838d6f80c1beb71eacc46bd77a91a006332b49096939dd18623R14

https://github.com/archlinux/archinstall/pull/3459/files#diff-4d8c19c70790b303e8338fbe439d7203c23bb2247a6b0c08cc995d0ca8271855R54

Is there something to keep those items per line as well?

svartkanin:
Great that worked! 
I actually looks quite good now with not too many major changes. 

@Torxed what do you think?

Torxed:
It's looking better! I have a question why it's not doing the same here:

```diff
- 	part_mods = [
- 		d.segment for d in data
- 		if isinstance(d.segment, PartitionModification)
- 		and d.segment != entry
- 	]
+ 	part_mods = [d.segment for d in data if isinstance(d.segment, PartitionModification) and d.segment != entry]
```

svartkanin:
I assume that's because the max line length is quite high for a python project with 160. Reducing it I can see that it starts getting broken up, e.g. with 100. 

Having said that, I've noticed some other oddities such as 
```

**Comment by Torxed** [comment]
I really wish these programs weren't so hung up on line length.
I just want the same structured format rather than it happening based on outside parameters.

But at some point, I think this PR might better benefit us, so I'll stand behind it despite the oddities if everyone else like this :)

**Comment by svartkanin** [comment]
I'd be in favour of reducing the line count though, to maybe 120, unless there's a good reason to keep it at 160?

I can try running the formatter with yapf as well to compare the output, that one has lots of options to tweak which might suit better?

**Comment by Torxed** [comment]
I just wish it could keep strings long. I prefer one slightly long string over many smaller strings over multiple lines. Preference things I guess, but try and see what 120 line length would look like. I suspect it will chop up 80% of our function definitions.

**Comment by svartkanin** [comment]
It seems that the `ruff format` collides with flake8 
```
archinstall/lib/disk/device_handler.py:420:2: E704 multiple statements on one line (def)
	def _lvm_info_with_retry(self, cmd: str, info_type: Literal['lv']) -> LvmVolumeInfo | None: ...
	^
archinstall/lib/disk/device_handler.py:423:2: E704 multiple statements on one line (def)
	def _lvm_info_with_retry(self, cmd: str, info_type: Literal['vg']) -> LvmGroupInfo | None: ...
	^
archinstall/lib/disk/device_handler.py:426:2: E704 multiple statements on one line (def)
	def _lvm_info_with_retry(self, cmd: str, info_type: Literal['pvseg']) -> LvmPVInfo | None: ...
	^
archinstall/lib/networking.py:154:76: E203 whitespace before ':'
		checksum += (icmp_packet[i] << 8) + (struct.unpack('B', icmp_packet[i + 1 : i + 2])[0] if len(icmp_packet[i + 1 : i + 2]) else 0)
		                                                                        ^
archinstall/lib/networking.py:154:114: E203 whitespace before ':'
		checksum += (icmp_packet[i] << 8) + (struct.unpack('B', icmp_packet[i + 1 : i + 2])[0] if len(icmp_packet[i + 1 : i + 2]) else 0)
		                                                                                                              ^
archinstall/lib/networking.py:195:59: E203 whitespace before ':'
				if icmp_type == 0 and response[-len(random_identifier) :] == random_identifier:
				                                                     ^
archinstall/tui/menu_item.py:443:12: E203 whitespace before ':'
				items[x : x + self._total_cols],
```

I found some mention in an issue https://github.com/astral-sh/ruff/issues/8642#issuecomment-2615236811 which then later points to what black recommends to ignore them https://ichard26-testblackdocs.readthedocs.io/en/refactor_docs/compatible_configs.html#configuration. 

I pushed another version formatted with `yapf` and the following settings
```
[tool.yapf]
based_on_style = "pep8"
use_tabs = true
column_limit = 160
align_closing_bracket_with_visual_indent = true
blank_line_before_nested_class_or_def = true
coalesce_brackets = true
dedent_closing_brackets = true
continuation_align_style = "VALIGN-RIGHT"
spaces_before_comment = 2
split_before_first_argument = true
split_before_closing_bracket = true
split_all_top_level_comma_separated_values = true
indent_dictionary_value = true
split_arguments_when_comma_terminated = false
```

I had to remove the type variables in classes and functions as yapf couldn't handle them. I raised an issue in yapf for it https://github.com/google/yapf/issues/1277. 

The formatting looks similar to ruff but it allows for more tweaks and doesn't seem to suffer from the same above flake8 deviations.

svartkanin:
Agreed, I think moving forward with ruff is probably the better option for now.

Torxed (approved):
Sounds good, and good job elaborating on this one!

PR #3481, opened 2025-05-19
hyprland updates

DarkXero-dev:
Those are just my thoughts for best experience.

- Thunar is better than having to drag KDE dependencies
- hyprpolkitagent is more tailored for Hyprland
- Rofi-hyprland is way better than the limited wofi


DarkXero-dev:
> Is `Rofi-hyprland` preferred for most Hyprland users? Arch users tend to want minimalistic stuff rather than full featured applications most of the time.
> 
> I don't use hyprland myself, so this is mostly a control question to ask the general community :)

I don't use it either to be honest, but as a Distro maintainer part of my job is to ask around what is preferred over what, so I can confidently answer this by saying yes wofi is extremely limited from what I have been told. Looks great but limited that's why they prefer rofi

bitterhalt:
I vote yes for this. I think Dolphin and it's dependencies are terrible idea in a minimalist environments like Hyprland or any WM in general. Rofi is also way better choice because it's more widely used than Wofi.

Torxed (approved):

DarkXero-dev:
Sorry my bad need to clarify that rofi package is rofi-wayland not rofi-hyprland typo...

archinstall/default_profiles/desktops/hyprland.py
```diff
@@ -24,12 +24,12 @@ def packages(self) -> list[str]:
 			"dunst",
 			"kitty",
 			"uwsm",
-			"dolphin",
-			"wofi",
+			"thunar",
+			"rofi-hyprland",
```
svartkanin, line 28: Needs to change https://archlinux.org/packages/extra-testing/x86_64/rofi-wayland/

svartkanin (changes requested):

PR #3482, opened 2025-05-19
More complete plasma install

DarkXero-dev:
I just added the missing elements for a more complete plasma experience. Feel free to ignore.

### DIFF
--- a/archinstall/default_profiles/desktops/plasma.py
+++ b/archinstall/default_profiles/desktops/plasma.py
@@ -16,7 +16,11 @@ class PlasmaProfile(XorgProfile):
 			"konsole",
 			"kate",
 			"dolphin",
+			"dolphin-plugins",
+			"flatpak-kcm",
+			"kcmutils",
 			"ark",
+			"kf6",
 			"plasma-workspace",
 		]
 


archinstall/default_profiles/desktops/plasma.py
```diff
@@ -16,7 +16,11 @@ def packages(self) -> list[str]:
 			"konsole",
 			"kate",
 			"dolphin",
+			"dolphin-plugins",
+			"flatpak-kcm",
+			"kcmutils",
 			"ark",
+			"kf6",
```
Torxed, line 23: [kf6](https://archlinux.org/groups/x86_64/kf6/) is a pretty massive group of applications. All of which feels optional.

archinstall/default_profiles/desktops/plasma.py
```diff
@@ -16,7 +16,11 @@ def packages(self) -> list[str]:
 			"konsole",
 			"kate",
 			"dolphin",
+			"dolphin-plugins",
+			"flatpak-kcm",
+			"kcmutils",
```
Torxed, line 21: flatpak and especially `kcm` is optional packages and not required for a bare minimum experience that still feels like a decent experience I think.

archinstall/default_profiles/desktops/plasma.py
```diff
@@ -16,7 +16,11 @@ def packages(self) -> list[str]:
 			"konsole",
 			"kate",
 			"dolphin",
+			"dolphin-plugins",
```
Torxed, line 19: I wouldn't mind this, if they just dropped the dropbox integration to be honest. But as this stands, this also feels like it's adding more bells and whistles than necessary for most people.

Torxed (commented):
It will sound harsh, as it always does when you reject something. But I genuinely appreciate the effort to make these profiles better, however these feel a bit too specific for general consumption.

I don't use the plasma profile myself, so I can only guess on behalf of most users, that these are things you can live without.

archinstall/default_profiles/desktops/plasma.py
DarkXero-dev: I understand for flatpak kcm and kf6 bit for dolphin-plugins it's conatains something essential which is the git integration very important

DarkXero-dev: @Torxed 

I mean it will not hurt adding a few kb by default. Not everyone will know where to get them when needed.. It's not too big.. Better. For kf6 they will get installed as dependencies with packages that require them so it's fine. For flatpak kcm not needed unless user adds flathub.. Fine too..

svartkanin:
I would not want to have rhythmbox and git installed by default even uf I'm using the later one daily.
It would also not be shipped with other distros by default

DarkXero-dev:
> I would not want to have rhythmbox and git installed by default even uf I'm using the later one daily. It would also not be shipped with other distros by default

THat's taking minimalism to another level IMO.. Too little might as well not offer profiles leave those up to users then.. If a few kb is too much.. If that's the situation, then kill off the profiles menu.. Better for the ArchInstall project less to maintain..

svartkanin:
Well that argument is just bad. It's not about size of installed packages it's about providing a minimal functional system for a user. If it's too much of an effort to install additional package such as `git` as a post installation by the user than archinstall is probably not what you're looking for in the first place.

DarkXero-dev:
Ok you know what ? In that case just leave ArchInstall profiles as they are. I will tell everyone complaining about them that you guys can be so closed minded opinionated not to bother requesting changes.. Wow I thought I could help but from the looks of it ain't nothing gonna change.. 

Just my 2 cents

DarkXero-dev:
If what I suggested was too much I would agree with you completely, not when I I suggested were literally a couple that make it more functional out the box.. But now I get it ArchInstall profiles only exist as a stop gap between nothing n something. Please in that case do not tell users they can suggest any additions to them just fixes make it clear somewhere since you already made up your minds about what should be included and what shouldn't..

Sorry don't mean to be aggressive just weirded out is all

DarkXero-dev:
Then do not install Plasma.. It's like you saying I love this car I love everything about it but the doors, the side rails the roof.. We all have opinions so this argument will not be won. Why did I even bother coming here ? I will never know..

DarkXero-dev:
> What the fuck are you talking about, fuck, I'm reading complete nonsense, all the functionality of the plasma is out of the box, is that not enough for you? Take your fucking pens and install what you need. No one is going to add all this garbage that is not needed except for one person or a couple

No need to be hostile... Did I suggest to add the entire plasma stack ? All I suggested was dolphin-plugins, flatpak-kcm and one other one.. FOr 3 packages you are acting as if I am asking you to carry the whole world. Relax.. 

If you disagree then close this and let's stop with the toxicity.. I will just stick to my guns and recommend against using ArchInstall profiles.. It's a free world isn't it ?

DarkXero-dev:
Tell that to the countless Arch-based distros out there, CachyOS, EndeavourOS and so on.. You are an Arch purist nothing wrong with that. I have nothing against anyone.. 

We are free to take Arch to the next level aren't we ? So that's what am doing with my Distro.. I wasn't arguing (per say) am just saying I came here at the behest of a few users and I get flack for it.

DarkXero-dev:
You are sneaky aren't you ? LoL.. You have your own Distro ? WoW !

Issue #2587, opened 2024-07-23
Suggestion for cinnamon default_profile


spsf64:
### describe the request

I am a cinnamon enthusiast, was checking other profiles and noticed some interesting packages;
So here are my suggestions:

Remove: 
blueberry (linuxmint cinnamon moved to blueman)

Add:
blueman (to replace blueberry) and bluez-utils
engrampa or xarchive (manage archives)
gnome-screenshot (necessary for printscreen to work)
gvfs-smb (access local windows network)
lightdm-gtk-greeter
xed (gui editor)
xdg-user-dirs-gtk (to create proper home directories)

Thanks for this great project.

svartkanin:
Sounds good to me. 
`lightdm-gtk-greeter` -> the greeter can be specified manually (there's a default set) is that not good?

spsf64:
> Sounds good to me. `lightdm-gtk-greeter` -> the greeter can be specified manually (there's a default set) is that not good?

My bad, I was thinking about `lightdm-gtk-greeter-settings`, but it's more like a personal preference, just ignore this one...
Have a nice day!

spsf64:
@svartkanin: Hi, I was checking about "metacity" and it seems cinnamon >6.x.x does not support it anymore, probably can also be removed from default profile... Thanks

Issue #2507, opened 2024-05-15
[Request] Add packages to ensure proper operation.


Bundy01:
Hi,

I reinstalled Arch with `archinstall` with the DE Cinnamon in VirtualBox initially, it's a real treat. 
However, I had to install two post-install packages to get it right:

* The user folders were not present in `nemo`, so I added `xdg-user-dirs`.
* The Cinnamon translation package was not present. So I installed `cinnamon-translations`.

I didn't try with the other DEs, but it should be the same. For the translation package, it would be good to install it if the user has chosen a language/keyboard other than English.

Regards.

Torxed:
Related discussion from the past: https://github.com/archlinux/archinstall/issues/594

Bundy01:
Hi,

I also saw that the Cinnamon DE choice installed `blueberry`. This package hasn't been updated for almost two years :(
It also has a security problem: the adapter is visible all the time.

It would be nice to replace this package with `blueman` for a `DE Cinnamon` installation.

Regards.

Issue #594, opened 2021-07-07
Missing user directories


Radioactive92177:
So I recently switched from **Manjaro** to **Vanilla Arch** and I couldn't bear the fact that it doesn't make **user directories** automatically.
I had to install **xdg-user-dirs** , and run **xdg-user-update** to make it work.
I believe it should be included in the default packages of the guided installer.

liuhaowen478:
I agree that this should happen by default on desktop profiles.

Torxed:
I'm open to the idea of an optional installation of `xdg-user-dirs` if a desktop environment is selected.
I personally don't like to clutter down the disk with folders that I'll never use.

```
XDG_DESKTOP_DIR="$HOME/Desktop"
XDG_DOCUMENTS_DIR="$HOME/Documents"
XDG_DOWNLOAD_DIR="$HOME/Downloads"
XDG_MUSIC_DIR="$HOME/Music"
XDG_PICTURES_DIR="$HOME/Pictures"
XDG_PUBLICSHARE_DIR="$HOME/Public"
XDG_TEMPLATES_DIR="$HOME/Templates"
XDG_VIDEOS_DIR="$HOME/Videos"
```

Out of all these, I think I would only use `Downloads` because the browsers use it by default.
But if people really use all these and want them (a majority of users), then I'd be open to adding a **optional**  question to create these directories after a desktop profile was chosen. So feel free to submit a PR and I'll go through it and merge it if it looks clean :)

nerdyman:
I realised that the scripts don't create a home directory for the user specified in the config which stumped me for a few hours when trying to get a Pantheon desktop set up working as LightDM immediately logs out if the user's home directory isn't there. It would be useful for the `/home/<user>` directory to be created per user specified in the config (even if the XDG dirs aren't included).

svartkanin:
I believe this is created in the latest version

Torxed:
>  It would be useful for the `/home/<user>` directory to be created

This should definitely be the case, as it's done via `-m` when `useradd` is called: https://github.com/archlinux/archinstall/blob/15e52b77683f266ee843def2bb35ed02cd8e559d/archinstall/lib/installer.py#L1198-L1203

```
-m, --create-home             create the user's home directory
```

svartkanin:
Fantastic thanks for clarifying

PR #3484, opened 2025-05-19
Add missing xdg-user-dirs and Cosmic Greeter

DarkXero-dev:
Installing Cosmic as is results in no greeter so service needs to be started. Also added missing xdg-user-dirs for user directories in cosmic-file-manager..

### DIFF
--- a/archinstall/default_profiles/desktops/cosmic.py
+++ b/archinstall/default_profiles/desktops/cosmic.py
@@ -13,9 +13,10 @@ class CosmicProfile(XorgProfile):
 	def packages(self) -> list[str]:
 		return [
 			"cosmic",
+			"xdg-user-dirs"
 		]
 
 	@property
 	@override
 	def default_greeter_type(self) -> GreeterType:
-		return GreeterType.CosmicSession
+		return GreeterType.CosmicGreeter


Torxed (commented):
Thank you for these fixes.
Is `xdg-user-dirs` strictly necessary for cosmic to function?
I thought that package only created some well known folders, that from previous issues/PR have been requested to be removed IIRC.

DarkXero-dev:
> Thank you for these fixes. Is `xdg-user-dirs` strictly necessary for cosmic to function? I thought that package only created some well known folders, that from previous issues/PR have been requested to be removed IIRC.

Removed ? Why ? This is the norm for DEs except thunar.. It's been the way file managers worked since ever..

svartkanin:
I can't see that package mentioned as requested in the wiki https://wiki.archlinux.org/title/COSMIC also not in the cosmic package group

DarkXero-dev:
> I can't see that package mentioned as requested in the wiki https://wiki.archlinux.org/title/COSMIC also not in the cosmic package group

Remember it's Alpha so could have been missed.. Doesn't mean it shouldn't be there.. Again if a few kb hurt you then in my opinion desktop profiles should be killed off in ArchInstall let users choose what they want.

svartkanin:
The point is that we need to be able to point to something when someone else comes along and complains "why is this package being installed". 
Having a clear source to point to is something valuable, with the wiki being the general source to follow.

 In addition, what is `CosmicGreeter` ? That hasn't been defined at all

DarkXero-dev:
> The point is that we need to be able to point to something when someone else comes along and complains "why is this package being installed". Having a clear source to point to is something valuable, with the wiki being the general source to follow.
> 
> In addition, what is `CosmicGreeter` ? That hasn't been defined at all

I guess you are getting a lot of "bloat" sensitive people complaining.. Well you can just tell them xdg-user-dirs is what enables user folders to show up in file manager side panel without it they will not be there. If they complain again just ignore them way I do coz if they dislike the way things are they can either remove the so called "bloat" and move on or stop using Linux n go back to what suits their need better.. 

What is cosmic greeter ? You must be kidding me right ? It's the included greeter for cosmic it's included if you look close. It's just it's not being enabled via ArchInstall..

svartkanin:
My point was it's not defined in the code.

Torxed:
> But hey ignore me and my "unconventional" suggestions.. You do you, I will keep creating scripts that bypass the ArchInstall profiles section..

This general tone is for me unacceptable.
Part of maintaining something, is to interact with others.
This means interacting socially in a way that is pleasant for everyone!

And this is where I find your "you do you" unacceptable, because what @svartkanin is trying to do, is help you understand that **your code** has a few flaws that needs adjusting before we can merge this without breaking changes.

You either don't understand Python *(judging by looking at your GitHub projects)*, which is fine, or you just don't respond well to criticism. 
Because as @svartkanin pointed out:

> In addition, what is CosmicGreeter ? That hasn't been defined at all

* https://github.com/archlinux/archinstall/blob/5a4773cdd312bf65220af6853c222c01f53a8296/archinstall/default_profiles/profile.py#L30-L39

There's no `CosmicGreeter` type defined in our type cluster. Thus if you try to access `GreeterType.CosmicGreeter` you will get:

screenshot

Again, these are all fixable stuff..
But the way you respond to questions from the community and the developers trying to guide you to a complete PR is tempered and short.

And your recommendations on how to handle community feedback in the manner of:

> If they complain again just ignore them way I do

Is not really what we're about. We're trying to be inclusive, both to you and to others.
That means allowing people to have opinions, as long as they themselves are inclusive.

All that said, which I really shouldn't have to tell another person, because people should understand that being respectful and patient with each other is a basic human thing to do..

Adding [xdg-user-dirs](https://archlinux.org/packages/?name=xdg-user-dirs) has been discussed in the past:

 - https://github.com/archlinux/archinstall/issues/594
 - https://github.com/archlinux/archinstall/issues/2587
 - https://github.com/archlinux/archinstall/issues/2507

And there hasn't been much objections, so I'd say we could add it and see what the response would be like, with the limitation of it being added for desktop profiles *(even for certain desktop profiles since most desktop users differ in their taste.. `awesome` for instance don't need it really IMHO as it's a WM not a DE)*.

But `GreeterType.CosmicGreeter` needs to be reverted, or fully added as it's breaking Python code at the moment.

Until then, refrain from passive aggressive language, or language with finality statements such as "kill profiles in archinstall" as these are counter productive.

DarkXero-dev:
@Torxed 

I do not understand python no, never pretended that I did. And am sorry if I was aggressive, but you understand I was sort of criticized for suggesting what I did. As for the Greeter I always wondered why it was included and not being enabled thank you for explaining it. I was just pointing it out.

User dirs is fine with or without, I came here on the behest of some of my users who, when I mentioned why I skip profiles when I use ArchInstall asked me to report it here for them.. Is all..

Torxed:
Feel free to revert the greeter code, and thus only adding the user dirs.
That way we can merge this. Then if you're open for learning and discussing the greeter fix, we can do that in a separate PR.

Or, we can discuss how to fix the Python code here, that's fine too.
It's up to you really. These are just two viable options.

DarkXero-dev:
@Torxed 

Since I am not a dev, I wouldn't know where to begin that's why I suggested it here. You are the brains of the operation. Although I maintain a full Distro, I do that to the best of my non-dev abilities and it has worked out well for me me for the past 5 years or so. 

Forgive my ignorance when I reported I thought it was just a matter of including an option that enables the service since that was all that I needed to do for it to work post-install but from my understanding now it is way more involved than that.. Hmm.. Well in my Cosmic install I just added sudo systemctl enable cosmic-greeter which works..

Torxed:
> Well in my Cosmic install I just added sudo systemctl enable cosmic-greeter which works..

I believe that should be `systemctl --user enable cosmic-greeter` right?

DarkXero-dev:
@Torxed Well I dunno really haven't looked too deeply into it I just use general one I mentioned and it works since in my mind it should be system wide not just for that user.

archinstall/default_profiles/desktops/cosmic.py
```diff
@@ -13,9 +13,10 @@ def __init__(self) -> None:
...
 		return [
 			"cosmic",
+			"xdg-user-dirs"
 		]
 
 	@property
 	@override
 	def default_greeter_type(self) -> GreeterType:
-		return GreeterType.CosmicSession
+		return GreeterType.CosmicGreeter
```
Torxed: If you can revert this change, we can merge in the `xdg-users-dirs` change.
And we can fix the greeter separately.

Torxed (changes requested):

Torxed (approved):

PR #3502, opened 2025-05-25
Move disk encryption into disk config menu

svartkanin:
This will move the disk encryption menu from the main menu into the disk configuration menu as it makes more sense to group it in there with the other disk related configurations. 

The configuration file has also changed to reflect it and the `disk_encryption` is now inside the `disk_config` block. This change is backwards compatible though, so existing configuration files will still be parsed correctly.

### DIFF
--- a/archinstall/lib/args.py
+++ b/archinstall/lib/args.py
@@ -76,7 +76,6 @@ class ArchConfig:
 
 	# Special fields that should be handle with care due to security implications
 	users: list[User] = field(default_factory=list)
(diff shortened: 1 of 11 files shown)


Torxed:
I will adjust the test framework to reflect this. And I agree, it makes more sense to have it in there :)

Torxed:
> Thanks, the current unit tests are not able to cover the disk configurations as any config is only processed if the disk exists on the host running it. Maybe that could be mocked somehow

Probably can be mocked, but it's also covered in the integration test before release at least

Issue #3530, opened 2025-05-28
load_plugin looks for missing __version__ field in storage


correctmost:
The `load_plugin` function looks for a `__version__` field in storage, but that field was removed in https://github.com/archlinux/archinstall/commit/b57f7f91cf36925e94c16457442702fb592d11b3#diff-9837bf5fe3d807b5c33527591ba23a8c3acd008009f9c0e31d90ede78ae9c43f:

https://github.com/archlinux/archinstall/blob/b2b36a2cefda88e89ada7d19571af4560055c40c/archinstall/lib/plugins.py#L101-L102

PR #3548, opened 2025-05-29
Add type annotations to global storage dictionary

correctmost:
## PR Description:

This will help catch issues like #3530.

## Tests and Checks
- [x] I have tested the code!

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -92,8 +92,6 @@ class Installer:
 
 		self.post_base_install: list[Callable] = []  # type: ignore[type-arg]
 
-		# TODO: Figure out which one of these two we'll use.. But currently we're mixing them..
-		storage['session'] = self
 		storage['installation_session'] = self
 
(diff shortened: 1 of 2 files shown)


archinstall/lib/installer.py
```diff
@@ -92,8 +92,6 @@ def __init__(
 
 		self.post_base_install: list[Callable] = []  # type: ignore[type-arg]
 
-		# TODO: Figure out which one of these two we'll use.. But currently we're mixing them..
-		storage['session'] = self
```
Torxed, line 96: This might be a breaking change for those scripting their own installations (not everyone uses `guided.py` or what's referenced in this repo.

This is mainly a note to self so I don't forget to mention that in the release notes.
We should perhaps have gone out with a breaking change notification in the previous release, but it's an easy `sed -i "s/storage['session']/storage['installation_session']/g' <file>` wherever it's used.

Torxed (approved):

PR #3562, opened 2025-05-31
Remove unused context manager code from SysCommand

correctmost:
## Tests and Checks
- Pyright and mypy didn't detect any errors
- I didn't see `SysCommand` used as a context manager when reviewing the various callers


Torxed:
Same thing here, internally we might not but one point of archinstall is to be able to use it as a library: 
- https://github.com/Torxed/archoffline/blob/0dc2867a1c4ecdb5295d8c32a83a258d52f0b351/archoffline/__init__.py#L229

PR #3566, opened 2025-06-01
Fix return values used with a Textbox.edit callback in curses

correctmost:
## PR Description:
References
- [Typeshed definition](https://github.com/python/typeshed/blob/f318894924afd726f518a9fc454ea544fa12f780/stdlib/curses/textpad.pyi#L9)
- [The edit method in Python](https://github.com/python/cpython/blob/a3a5b4bb232ba29875147baa3b56a5524eece353/Lib/curses/textpad.py#L186-L187)

## Tests and Checks
- [x] I have tested the code!

### DIFF
--- a/archinstall/tui/curses_menu.py
+++ b/archinstall/tui/curses_menu.py
@@ -288,7 +288,7 @@ class EditViewport(AbstractViewport):
 		edit_height: int,
 		x_start: int,
 		y_start: int,
-		process_key: Callable[[int], int | None],
+		process_key: Callable[[int], int],
 		frame: FrameProperties,
 		alignment: Alignment = Alignment.CENTER,
 		hide_input: bool = False,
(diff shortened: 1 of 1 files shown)


archinstall/tui/curses_menu.py
```diff
@@ -606,15 +606,15 @@ def kickoff(self, win: curses.window) -> Result[str]:
...
+	def _process_edit_key(self, key: int) -> int:
 		key_handles = MenuKeys.from_ord(key)
 
 		if self._help_active:
 			if MenuKeys.ESC in key_handles:
 				self._help_active = False
 				self.clear_help_win()
 				return 7
-			return None
+			return 0
```
correctmost, line 617: `0` and `None` are interchangeable because Python only checks for falsiness:

https://github.com/python/cpython/blob/a3a5b4bb232ba29875147baa3b56a5524eece353/Lib/curses/textpad.py#L186-L187

Torxed:
Great work on these!

PR #3571, opened 2025-06-03
Remove unused arguments from internal functions

correctmost:
## PR Description:

This makes it easier to run Pylint's unused-argument checks on the codebase.

## Tests and Checks
- [x] I have tested the code!

### DIFF
--- a/archinstall/lib/models/mirrors.py
+++ b/archinstall/lib/models/mirrors.py
@@ -103,7 +103,7 @@ class MirrorStatusEntryV3(BaseModel):
 		return value
 
 	@model_validator(mode='after')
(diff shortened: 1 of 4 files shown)


archinstall/lib/models/mirrors.py
```diff
@@ -103,7 +103,7 @@ def validate_score(cls, value: float) -> int | None:
 		return value
 
 	@model_validator(mode='after')
-	def debug_output(self, validation_info) -> 'MirrorStatusEntryV3':
```
correctmost, line 106: `validation_info` is an optional parameter: https://docs.pydantic.dev/latest/concepts/validators/#validation-info

PR #3597, opened 2025-06-15
Turkish language current

SerdarSaglam:
It was a very bad and wrong translation, but it is good now.


archinstall/locales/tr/LC_MESSAGES/base.po
```diff
@@ -61,10 +61,10 @@ msgid "Write additional packages to install (space separated, leave blank to ski
 msgstr "Kurulacak ek paketleri yazınız (paketleri boşlukla ayırın, geçmek için boş bırakın): "
 
 msgid "Copy ISO network configuration to installation"
-msgstr "ISO Ağ yapılandırmasını kuruluma kopyala"
+msgstr "ISO ağ yapılandırmasını kuruluma kopyala"
 
 msgid "Use NetworkManager (necessary for configuring internet graphically in GNOME and KDE)"
-msgstr "NetworkManager'ı kullan (GNOME ya da KDE'de interneti grafiksel olarak yapılandırmak için gerekli)"
+msgstr "NetworkManager kullan (GNOME ve KDE ile interneti grafiksel olarak yapılandırmak için gerekli)"
```
Nucceteere, line 67: Instead of `ile`, the usage of `'de` would be better in this context

archinstall/locales/tr/LC_MESSAGES/base.po
```diff
@@ -245,35 +245,34 @@ msgid "Keyboard layout"
...
-msgstr "Yerel ayar dili"
+msgstr "Yerel dil"
 
 msgid "Locale encoding"
-msgstr "Yerel ayar dil kod şeması"
+msgstr "Yerel kodlama"
 
 msgid "Drive(s)"
-msgstr "Disk(ler)"
+msgstr "Diskler"
```
Nucceteere, line 257: The usage of `Disk(ler)` here (as in the old translation) is much more suitable due to the fact that the user may have one disk instead of multiple or vice versa.

archinstall/locales/tr/LC_MESSAGES/base.po
```diff
@@ -245,35 +245,34 @@ msgid "Keyboard layout"
...
 
 msgid "Disk layout"
-msgstr "Disk şeması"
+msgstr "Disk düzeni"
```
Nucceteere, line 260: The word `layout` is usually translated as `şema` (as in the old translation) in technology related uses.

PR #3609, opened 2025-06-18
Update bluetooth.py

JoeArchV:


svartkanin:
Is there a need for this package?

CelestifyX:
Why do we need outdated utilities if Bluetooth works correctly without them?

Issue #3477, opened 2025-05-18
Archinstall crashes when running pacstrap


king-spidey99:
### Which ISO version are you using?

2025-05-01

### The installation log

```json
Go to 0x0.st/8vCM.log the form complained my log is too big.
```

### describe the problem

I was installing arch with archinstall because my install I had done an hour ago nuked itself. I got through all the questions and went to install and it complained with the error in the log provided. I tried fully wiping the drive and trying again but it didn't work. I had removed the default subvolumes and didn't make new ones because I don't know what I'm doing and last time I used the default I deleted my system trying to use snapper. Maybe that's it but then how would I sort out snapper because I can't find anything online or on the arch wiki on how to do that.

svartkanin:
You specified a btrfs filesystem but ther's no subvolumes defined according to the config

svartkanin:
That's a different issue, for the one you posted the problem is that there are no subvolumes defined

king-spidey99:
Im gonna ask for help on the forums you can close this

PR #3615, opened 2025-06-20
Remove btrfs default structure prompt on default disk layout

svartkanin:
There was a previous issue related to an installation with BTRFS but no subvolumes https://github.com/archlinux/archinstall/issues/3477#issuecomment-2889343703. In the default disk setup mode that is possible as there is a question to add the subvolumes or not which doesn't make much sense as choosing "no" will not actually work. 

This removes the question if subolumes should be added in the default disk setup but still keeps the prompt in the manual partitioning menu as users are expected to be more advanced in there

### DIFF
--- a/archinstall/lib/global_menu.py
+++ b/archinstall/lib/global_menu.py
@@ -131,7 +131,7 @@ class GlobalMenu(AbstractMenu[None]):
 				action=self._select_applications,
 				value=[],
 				preview_action=self._prev_applications,
-				key='application_config',
+				key='app_config',
 			),
 			MenuItem(
 				text=tr('Kernels'),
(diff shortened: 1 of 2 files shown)


Torxed (approved):
I still think they should have the option to choose, but if it's broken then we should merge this and circle back later.

svartkanin:
What would be the actual flow though if someone chooses not to use the default structure? There's no other option at that point to add subvolumes

Torxed:
My understanding is that people use btrfs as a normal filesystem (like ext4) if no subvolumes were selected, and they can add some manually later. And if they select subvolumes they'll get some sensible defaults :)

svartkanin:
Okay I tried to reproduce the problem mentioned without specifying subvolumes but everything seems to work fine. I'll decline this for now as it seems to be a working use case

PR #3620, opened 2025-06-22
Added explicit drivers defined by the empty package linux-firmware

Torxed:
As of [linux-firmware 20250613.12fe085f-5](https://archlinux.org/news/linux-firmware-2025061312fe085f-5-upgrade-requires-manual-intervention/), `linux-firmware` will become an empty package with dependencies pulling in different firmwares.

This allows us to have more fine grained control of which firmwares are installed by `archinstall`. This PR will add the default ones as of writing this.

In the future, we can detect and install only the required once perhaps. This adds a slight overhead in order to making sure any future additions/removals of package requirements in `linux-firmware` gets adopted.

### DIFF
--- a/archinstall/lib/interactions/general_conf.py
+++ b/archinstall/lib/interactions/general_conf.py
@@ -161,7 +161,7 @@ def ask_additional_packages_to_install(
 	package_groups = PackageGroup.from_available_packages(packages)
 
 	# Additional packages (with some light weight error handling for invalid package names)
-	header = tr('Only packages such as base, base-devel, linux, linux-firmware, efibootmgr and optional profile packages are installed.') + '\n'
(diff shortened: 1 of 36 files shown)


archinstall/lib/interactions/general_conf.py
```diff
@@ -161,7 +161,24 @@ def ask_additional_packages_to_install(
 	package_groups = PackageGroup.from_available_packages(packages)
 
 	# Additional packages (with some light weight error handling for invalid package names)
-	header = tr('Only packages such as base, base-devel, linux, linux-firmware, efibootmgr and optional profile packages are installed.') + '\n'
+	header = (
+		tr(
+			'Only packages such as '
```
svartkanin, line 166: Shall we remove the package list from the translation string otherwise this will reset every time it changes

Torxed: Sounds like a good idea. I'm not sure it will change very often, but often enough perhaps.

PR #3396, opened 2025-04-16
Update bootctl command for new --variables= option

leebickmtu:
## PR Description:

Starting in systemd-boot v257 a `arch-chroot` environment is detected by `bootctl` as a container environment.
This will cause `bootctl install` to skip EFI variable setup silently. See [systemd issue 36174](https://github.com/systemd/systemd/issues/36174) & [Arch topic](https://bbs.archlinux.org/viewtopic.php?id=301920).

As a result of that issue [systemd is introducing](https://github.com/systemd/systemd/pull/37144) a new `--variables=yes|no` option to `bootctl`. The existing `--no-variables` is left for compatibility and is equivalent to `--variables=no`.

The default variable behavior if not specified will be environment context dependent. So since `arch-chroot` is detected as a container, EFI variables will skip.

This PR explicitly sets `--variables=yes` to over come this. And in the fallback case changes from the compatiblity `--no-variables` to `--variables=no`.

This PR should not be merged until a matching version of systemd is released into Arch packages. I just wanted to preemptively get this PR staged.

## Tests and Checks
- [ ] I have tested the code!<br>

PR #3625, opened 2025-06-23
Added temporary hold on bootctl's --variables=BOOL usage, as it's notin systemd main yet

Torxed:
This will deal with: https://github.com/archlinux/archinstall/pull/3396#issuecomment-2996862019

And once we're satisfied that it works, we can remove the check and revert to the above PR's original code.


svartkanin:
@Torxed shall we get this merged otherwise master is broken?

svartkanin (approved):

Torxed:
Was waiting just to get some feedback in case me fix was crazy :D

svartkanin:
It's temporary, let's not forget to clean it up when it's out

PR #3634, opened 2025-06-27
Add interface to change LUKS iteration time

haouarihk:
## PR Description:
Added a user interface to configure the LUKS encryption iteration time through the interactive menu system.

### Why:
I found myself always changing this value after performing archinstall. 
because i can't bother to wait 10 seconds for my disk to decrypt. 
i would much prefer if its 1-3 seconds.
i know this can be a security issue for some people out there.
so i opted into adding an interface for users to change it instead.

### Changes:
- Added "Iteration time" option in the encryption menu
- Added input validation (100-120000ms range)
- Added translation support for the new interface (just the base.pot)
- Updated encryption configuration to pass iteration time to LUKS

### Affects:
- LUKS encryption on partitions
- LVM on LUKS encryption
- LUKS on LVM encryption

Default: 10000ms (10 seconds). 

**Note**: This PR follows the existing code patterns and maintains consistency with the current architecture. All changes are properly typed, documented, and include appropriate error handling. 

## Tests and Checks
- [x] Tested everything default value(10000ms), seems to slowly decrypt (expected behavior)
- [x] Tested changing iteration time to 100ms, it seems to quickly decrypt

### DIFF
--- a/archinstall/lib/disk/encryption_menu.py
+++ b/archinstall/lib/disk/encryption_menu.py
@@ -17,7 +17,7 @@ from archinstall.tui.result import ResultType
 from archinstall.tui.types import Alignment, FrameProperties
 
 from ..menu.abstract_menu import AbstractSubMenu
-from ..models.device_model import Fido2Device
+from ..models.device_model import DEFAULT_ITER_TIME, Fido2Device
 from ..models.users import Password
 from ..output import FormattedOutput
 from ..utils.util import get_password
@@ -64,6 +64,14 @@ class DiskEncryptionMenu(AbstractSubMenu[DiskEncryption]):
 				dependencies=[self._check_dep_enc_type],
 				preview_action=self._preview,
 				key='encryption_password',
+			),
+			MenuItem(
+				text=tr('Iteration time'),
+				action=lambda x: select_iteration_time(x),
+				value=self._enc_config.iter_time,
+				dependencies=[self._check_dep_enc_type],
+				preview_action=self._preview,
+				key='iter_time',
 			),
 			MenuItem(
 				text=tr('Partitions'),
@@ -120,6 +128,7 @@ class DiskEncryptionMenu(AbstractSubMenu[DiskEncryption]):
 
 		enc_type: EncryptionType | None = self._item_group.find_by_key('encryption_type').value
 		enc_password: Password | None = self._item_group.find_by_key('encryption_password').value
+		iter_time: int | None = self._item_group.find_by_key('iter_time').value
 		enc_partitions = self._item_group.find_by_key('partitions').value
 		enc_lvm_vols = self._item_group.find_by_key('lvm_volumes').value
 
@@ -140,6 +149,7 @@ class DiskEncryptionMenu(AbstractSubMenu[DiskEncryption]):
 				partitions=enc_partitions,
 				lvm_volumes=enc_lvm_vols,
 				hsm_device=self._enc_config.hsm_device,
+				iter_time=iter_time or DEFAULT_ITER_TIME,
 			)
 
 		return None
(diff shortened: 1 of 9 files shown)


archinstall/lib/disk/encryption_menu.py
```diff
@@ -354,3 +375,50 @@ def select_lvm_vols_to_encrypt(
...
+			iter_time = int(value)
+			if iter_time < 100:
+				return tr('Iteration time must be at least 100ms')
+			if iter_time > 120000:
+				return tr('Iteration time must be at most 120000ms')
+			return None
+		except ValueError:
+			return tr('Please enter a valid number')
+
+	try:
```
correctmost: Is this try block necessary?  I don't see similar code in other parts of the codebase.

Some of these imports are already included at the top of the file, too, so I'm not sure why they would fail here.

archinstall/lib/disk/encryption_menu.py
```diff
@@ -65,6 +65,14 @@ def _define_menu_options(self) -> list[MenuItem]:
 				preview_action=self._preview,
 				key='encryption_password',
 			),
+			MenuItem(
+				text=tr('Iteration time'),
+				action=lambda x: select_iteration_time(x),
```
correctmost: It looks like the lambda wrapping can be removed:

```
$ ruff check --preview
archinstall/lib/disk/encryption_menu.py:70:12: PLW0108 Lambda may be unnecessary; consider inlining inner function
```

https://docs.astral.sh/ruff/rules/unnecessary-lambda/

archinstall/lib/models/device_model.py
```diff
@@ -19,7 +19,7 @@
 from ..output import debug
 
 ENC_IDENTIFIER = 'ainst'
-
+DEFAULT_ITER_TIME = 10000
```
correctmost, line 22: It might be good to use the new const in `archinstall/lib/luks.py` too:
https://github.com/archlinux/archinstall/blob/a580da2bbdf810da39530b5a241b432877851ca6/archinstall/lib/luks.py#L79

haouarihk:
Thanks for the feedback.

i haven't tested it locally. because i couldn't get the dev enviroment to work.
i'm just waiting for the iso file here to test it.

Edit: i also ran ruff check and flake8 and fixed the spacing warning

correctmost:
> i haven't tested it locally. because i couldn't get the dev enviroment to work.
> i'm just waiting for the iso file here to test it.

You can try this until the workflows are approved (I don't have approval permissions):
- Download [the latest iso from master](https://github.com/archlinux/archinstall/actions/runs/15924185692/artifacts/3418176123)
- Boot from the iso
- Run these commands after the iso boots:

```
cd /root/
git clone --branch master https://github.com/haouarihk/archinstall.git
cd archinstall

# Repeat these steps whenever you make any code changes locally
rm -rf dist/
uv build --no-build-isolation --wheel
uv pip install dist/*.whl --break-system-packages --system --no-build --no-deps

# Launch your custom-built archinstall
archinstall
```

I still see some warnings when running `ruff check --preview`.  And there's still some bugginess after an invalid value like 10ms is set, so that part needs more testing.

Torxed:
Ah right, new contributors require runners to be approved! They should be running now :)

haouarihk:
@correctmost Thanks soo much. that's very useful. and i got the dev enviroment to work

archinstall/tui/curses_menu.py
haouarihk, line 1: i've been testing for quite a while.
and i found this to solve my issue with input persistancy.

there could be better solutions out there.
but i couldn't get _current_text to persist with the previewed value.

i also tested it in the other inputs and it seems to not effect them.

svartkanin:
The preview of the new iteration is missing on the outer menu 
image

Other than that looks good :+1:

haouarihk:
@Torxed i've tested the compiled iso. and it seems to be working as expected.
whenever you get a time, take a look at it.
i wont be making any more modifications unless i'm told to do so.

Thanks

PR #3647, opened 2025-06-30
Add Pyright CI workflow

correctmost:
## PR Description:

I'm submitting a draft PR to see if there's interest in additional CI checks.

I use Pyright locally because it sometimes finds issues that are missed by mypy, ruff, and Pylint.  Pyright also has some checks that are not present in other tools, like `reportImportCycles`.

**Benefits**:
- Pyright can catch issues missed by other type checkers and static analysis tools.  This seems particularly useful now that AI-generated contributions are becoming more common.
- Pyright can help improve code quality and code architecture by enabling checks like `reportUnnecessaryComparison` and `reportImportCycles`
- Adding the Pyright configuration to `pyproject.toml` provides better on-the-fly IDE checks for developers who use VS Code + Pyright/Pylance
- Adding the Pyright configuration will also help catch regressions in Pyright because archinstall is included in [mypy_primer](https://github.com/hauntsaninja/mypy_primer), which is run on every Pyright PR

**Drawbacks**:
- It can be annoying to appease multiple type checkers.  This may become even harder if there's interest in using [ty](https://github.com/astral-sh/ty) or [Pyrefly](https://github.com/facebook/pyrefly) once they are ready for production use.
- Running Pyright locally and inside pre-commit requires Node.js to be installed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -258,4 +258,3 @@ deprecateTypingAliases = true
 reportImplicitOverride = true
 reportImportCycles = false
 reportPropertyTypeMismatch = true
-reportShadowedImports = true


Torxed:
> Running Pyright locally and inside pre-commit requires Node.js to be installed

I think this was the only thing that made me a bit worried.
If we can exclude it from pre-commit, or make it optional, that would be great.
Personally I wouldn't like to drag Node.js into my machine if I can help it heh.

On the github runners tho, that's fine!

correctmost:
> If we can exclude it from pre-commit, or make it optional, that would be great.
> Personally I wouldn't like to drag Node.js into my machine if I can help it heh.

That's understandable :).

I am okay with leaving Pyright out of pre-commit and seeing how things go with the CI checks.  If the checks become burdensome, we can revisit whether it makes sense to keep them.

svartkanin:
Does this https://github.com/RobertCraigie/pyright-python?tab=readme-ov-file#pre-commit actually need node though? 
I tried out adding it to the `.pre-commit-config.yaml` file 

```
  - repo: https://github.com/RobertCraigie/pyright-python
    rev: v1.1.403
    hooks:
    - id: pyright
```
 
 and it seems to work, despite that I don't have node installed

correctmost:
> Does this https://github.com/RobertCraigie/pyright-python?tab=readme-ov-file#pre-commit actually need node though?

If you don't have node installed, pyright-python will download a pre-built node binary using [nodeenv](https://github.com/ekalinin/nodeenv):

```
$ du -sh ~/.cache/pyright-python/
420M	/home/user/.cache/pyright-python/
```

I'm not sure if people are okay with that happening behind the scenes :).

svartkanin:
Not sure what the concern would be? The downloaded version is only part of the pyright installation and not a system installation no?

It'd get frustrating if GH actions keep failing because the changes weren't validated locally first...

correctmost:
> Not sure what the concern would be? The downloaded version is only part of the pyright installation and not a system installation no?

Yeah, it would just be a local installation that isn't visible to any package managers.

I think some developers avoid Node.js because of ongoing security issues with packages in the npm repository.  Other package managers have similar issues, but the npm ones seem a bit more pervasive.

Aside from that, the various caches and downloads from pre-commit are not automatically cleaned up, so you can end up accumulating files in `~/.cache/pre-commit/` and `~/.cache/pyright-python/`.

> 
> It'd get frustrating if GH actions keep failing because the changes weren't validated locally first...

Agreed.  I'm not sure if there's a good way to make certain pre-commit hooks optional either.

correctmost:
Closing this for now.  I think additional workflows can be reassessed after ty and Pyrefly (and maybe [zuban](https://github.com/zubanls/zuban)?) are stabilized.

Issue #3655, opened 2025-07-07
Install inotify-tools along with grub-btrfs


Salman7236:
### describe the request

grub-btrfs requires inotify tools for the systemd service to work properly.

This applies to btrfs+timeshift snapshot configuration in archinstall. Right now, it only installs timeshift and grub-btrfs, but grub-btrfs requires inotify-tools to work and will throw an error if you try to start the grub-btrfsd systemd service without installing inotify-tools.

PR #3656, opened 2025-07-07
Install inotify-tools along with grub-btrfs when using grub with btrfs config

Salman7236:
- This fixes the first issue mentioned [here](https://github.com/archlinux/archinstall/issues/3655)

## PR Description:

archinstall currently only installs timeshift and grub-btrfs when btrfs+grub configurations are detected. But grub-btrfs require inotify-tools to work and the systemd service fails to start without it. This commit just adds the inotify-tools package.

```
if bootloader and bootloader == Bootloader.Grub:
				self.pacman.strap('grub-btrfs')
                                self.pacman.strap('inotify-tools')
				self.enable_service('grub-btrfsd.service')

```

## Tests and Checks

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -956,6 +956,7 @@ class Installer:
 
 			if bootloader and bootloader == Bootloader.Grub:
 				self.pacman.strap('grub-btrfs')
+				self.pacman.strap('inotify-tools')
 				self.enable_service('grub-btrfsd.service')
 
 	def setup_swap(self, kind: str = 'zram') -> None:


CelestifyX:
as for me, you need to install these packages if Btrfs is selected, because for others it is just an extra garbage package

Salman7236:
> as for me, you need to install these packages if Btrfs is selected, because for others it is just an extra garbage package

Yes, archinstall only installs grub-btrfs if you select both the btrfs option and the GRUB option, but grub-btrfs alone is incomplete without inotify-tools.
As for others, who didn't choose BTRFS (or chose BTRFS but didn't choose GRUB), grub-btrfs doesn't get installed at all. That's why it's under the BTRFS section only in the script:

```
def setup_btrfs_snapshot(
		self,
		snapshot_type: SnapshotType,
		bootloader: Bootloader | None = None,
	) -> None:
		if snapshot_type == SnapshotType.Snapper:
			debug('Setting up Btrfs snapper')
			self.pacman.strap('snapper')

			snapper: dict[str, str] = {
				'root': '/',
				'home': '/home',
			}

			for config_name, mountpoint in snapper.items():
				command = [
					'arch-chroot',
					str(self.target),
					'snapper',
					'--no-dbus',
					'-c',
					config_name,
					'create-config',
					mountpoint,
				]

				try:
					SysCommand(command, peek_output=True)
				except SysCallError as err:
					raise DiskError(f'Could not setup Btrfs snapper: {err}')

			self.enable_service('snapper-timeline.timer')
			self.enable_service('snapper-cleanup.timer')
		elif snapshot_type == SnapshotType.Timeshift:
			debug('Setting up Btrfs timeshift')

			self.pacman.strap('cronie')
			self.pacman.strap('timeshift')

			self.enable_service('cronie.service')

			if bootloader and bootloader == Bootloader.Grub:
				self.pacman.strap('grub-btrfs')
				self.enable_service('grub-btrfsd.service')
```				 

So inotify-tools isn't a garbage package for non-btrfs users because they don't get grub-btrfs and related packages in the first place.

CelestifyX:
why do you need grub-btrfs at all if they use a different bootloader?

Salman7236:
> why do you need grub-btrfs at all if they use a different bootloader?

I think you misunderstood me. What I mean is that if you select any other bootloader during the archinstall setup, grub-btrfs won't get installed. It's only for those who select btrfs in disk partitioning section, and grub in the bootloader section. But the problem is that grub-btrfs doesn't work properly without inotify-tools. So this has nothing to do with those who use a different bootloader. Adding this fix to archinstall won't affect non-grub users in any way.

Current behavior: Only people who who choose btrfs+grub get grub-btrfs, but grub-btrfs requires inotify-tools to work which archinstall doesn't install.

Suggested behavior: Just add inotify-tools next to the grub-btrfs line, so people who choose to get grub-btrfs, actually get a fully functional grub-btrfs.

(That's a lot of GRUBs and BTRFSs lol)

svartkanin:
From the repository installation guide https://github.com/Antynea/grub-btrfs?tab=readme-ov-file#arch-linux it explicitly uses only `grub-btrfs` for arch, so I'm confused why this would be a requirement?

svartkanin:
Right, in the package metadata https://archlinux.org/packages/extra/any/grub-btrfs/ it specifies 

> [inotify-tools](https://archlinux.org/packages/extra/x86_64/inotify-tools/) (optional) - For grub-btrfsd daemon

So I suppose we do need it

svartkanin (approved):

PR #3659, opened 2025-07-08
Fix boot detection to use PartitionFlag.BOOT instead of mountpoint

VENOIRE:
## PR Description:

This PR updates the boot partition detection logic to use `PartitionFlag.BOOT` instead of relying solely on the mountpoint being `/boot`. This resolves an issue introduced in #3336 where users attempting to mount the ESP to `/efi` encountered a "Boot partition not found" error during installation.

I tested the installation script in 3 different situations and everything worked as expected.

### DIFF
--- a/archinstall/lib/models/device_model.py
+++ b/archinstall/lib/models/device_model.py
@@ -967,9 +967,7 @@ class PartitionModification:
 		return PartitionFlag.ESP in self.flags
 
 	def is_boot(self) -> bool:
-		if self.mountpoint is not None:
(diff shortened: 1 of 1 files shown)


Torxed (approved):
Looks good, I'll run some tests to verify that the `BOOT` flag is actually set in all our obscure combination of options.

I appreciate this fix!

Could you merge/rebase so that the changes introduced in https://github.com/archlinux/archinstall/pull/3639 isn't conflicting? :)

PR #3677, opened 2025-07-19
Add --skip-boot to allow for bypassing installation of a bootloader

Torxed:
This allows for workarounds in those semi-rare cases where the bootloader step would normally cause issues, allowing users to manually install one once the other steps are completed.


archinstall/lib/args.py
```diff
@@ -62,7 +63,7 @@ class ArchConfig:
 	profile_config: ProfileConfiguration | None = None
 	mirror_config: MirrorConfiguration | None = None
 	network_config: NetworkConfiguration | None = None
-	bootloader: Bootloader = field(default=Bootloader.get_default())
```
Torxed, line 65: Had to remove this one in order to break a circular dependency from `bootloader.py`'s `get_default` needing to access `args` in order to determine if `None` or default option should be returned.

Torxed:
I still need to check how `--conf` behaves here, and especially if `None` is specified without `--skip-boot` as that should cause a warning or exit. And only be allowed during `--skip-boot` presence.

svartkanin:
Should this maybe just be a "No bootloader" option, maybe only visible via --advanced?

Torxed:
True, maybe it just hides the option all together instead. Good thinking!

archinstall/lib/args.py
svartkanin: We're using a "No audio" option in the audio menu which is basically its own enum type. Would rhat pattern work here as well?

Torxed: That would work indeed, I can tweak that.

Torxed: Fixed now in 70c287e541da150fa151720c4e4e4b85f700323c I believe.

Torxed:
Some minor issues remain.

screenshot

The `NO_BOOTLOADER` option is visible when `--skip-boot` is omitted, which isn't ideal.

screenshot

archinstall/lib/global_menu.py
```diff
@@ -197,6 +197,14 @@ def _get_menu_options(self) -> list[MenuItem]:
 			),
 		]
 
+		if arch_config_handler.args.skip_boot:
```
svartkanin: Do we really want to omit the entire boot menu entry? I was under the impression the `--skip-boot` will enable the "No bootloader" option to be selected?

archinstall/lib/args.py
svartkanin: But as its part of the enum now, can we revert this as its always going to be a `Bootloader` type again

archinstall/lib/global_menu.py
Torxed: Oh, my tired brain read:

> "Should this maybe just be a "No bootloader" option"

As: 

"maybe there shouldn't be a boot loader option"

I think the `Bootloader.NO_BOOTLOADER` option to be selected makes sense, I'll put the menu entry back :)

svartkanin: All good, this should make the change much simpler :)

Torxed: Should be back now, waiting for ISO build to complete so I can test a full install.

Torxed:
Verified with three bootloaders, `grub`, `systemd` and `NO_BOOTLOADER`.

PR #3683, opened 2025-07-24
Update documentation to also mention the --config-url option

giftkugel:
# This fix updates the documentation

Using `archinstall` version `3.0.8`

## PR Description

I wanted to use a remote configuration file with `archinstall --config <url>` as described at https://archinstall.archlinux.page/installing/guided.html#example-usage

Unfortunately it failed.

By accident I took a look at the output of `archinstall --help` and discovered `archinstall --config-url`.

With `archinstall --config-url <url>` it works like expected.

This PR updates the documentation so others don't need to try the wrong option as I did. 😄 

## Tests and Checks
- Did not touch tests as I only updated the documentation

### DIFF
--- a/docs/installing/guided.rst
+++ b/docs/installing/guided.rst
@@ -36,7 +36,12 @@ There are two configuration files, both are optional.
 ``--config``
 ------------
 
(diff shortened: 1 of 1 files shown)


Torxed:
Much appreciated, than you!

PR #3707, opened 2025-08-04
Add support for rEFInd boot manager

### LINKED ISSUE #1951: Add rEFind as an alternative bootloader
GRUB and systemd-boot are popular boot loaders, but I like to use rEFInd instead. There is no option without bootloader to install the one we want.

Thank you for this superb helper.

justbispo:
This fixes #1951.

## PR Description:

This PR simply adds support for rEFInd boot manager. I've checked all the places where the boot managers are called to make sure I didn't miss anything.
I've also followed the [Arch Wiki page about rEFInd](https://wiki.archlinux.org/title/REFInd) to make sure I made the installation function correct.
Before accepting the PR, I'd like an opinion about a small issue I noticed. I didn't know this but per the Arch Wiki, [rEFInd is supposed to use the distro logo](https://wiki.archlinux.org/title/REFInd#Not_using_the_distribution_logo) but after the installation is finished, it uses the Tux logo. After checking the partitions, it's as it says there, `the root partition is of type Linux x86-64 root (/) instead of Linux filesystem`. But the installation guide suggests using `Linux x86-64 root (/)`. The Wiki also shows two other fixes, but I don't know if the label can be changed in the rEFInd installation step and copying the distro image seems easy, tho I don't know if it's the best solution
Is this and issue that can be fixed?

Just a note for the future, I'd also like to include [refind-btrfs](https://github.com/Venom1991/refind-btrfs), like grub includes [grub-btrfs](https://github.com/Antynea/grub-btrfs), to show BTRFS snapshots as boot options, but refind-btrfs isn't in the extra repository.

## Tests and Checks

### DIFF
--- a/archinstall/lib/global_menu.py
+++ b/archinstall/lib/global_menu.py
@@ -456,6 +456,10 @@ class GlobalMenu(AbstractMenu[None]):
 			if boot_partition.fs_type not in [FilesystemType.Fat12, FilesystemType.Fat16, FilesystemType.Fat32]:
 				return 'Limine does not support booting with a non-FAT boot partition'
 
+		elif bootloader == Bootloader.Refind:
+			if not SysInfo.has_uefi():
+				return 'rEFInd can only be used on UEFI systems'
+
 		return None
 
 	def _prev_install_invalid_config(self, item: MenuItem) -> str | None:
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -1530,6 +1530,75 @@ class Installer:
 
 		self._helper_flags['bootloader'] = 'efistub'
 
+	def _add_refind_bootloader(
+		self,
+		boot_partition: PartitionModification,
+		efi_partition: PartitionModification | None,
+		root: PartitionModification | LvmVolume,
+		uki_enabled: bool = False,
+	) -> None:
+		debug('Installing rEFInd bootloader')
+
+		self.pacman.strap('refind')
+
+		if not SysInfo.has_uefi():
+			raise HardwareIncompatibilityError
+
+		if not efi_partition:
+			raise ValueError('Could not detect EFI system partition')
+		elif not efi_partition.mountpoint:
+			raise ValueError('EFI system partition is not mounted')
+
+		info(f'rEFInd EFI partition: {efi_partition.dev_path}')
+
+		try:
+			SysCommand(f'arch-chroot {self.target} refind-install')
+		except SysCallError as err:
+			raise DiskError(f'Could not install rEFInd to {self.target}{efi_partition.mountpoint}: {err}')
+
+		if not boot_partition.mountpoint:
+			raise ValueError("Boot partition is not mounted, cannot write rEFInd config")
+
+		config_path = self.target / boot_partition.mountpoint.relative_to('/') / 'refind_linux.conf'
+		config_contents = []
+
+		kernel_params = ' '.join(self._get_kernel_params(root))
+
+		for kernel in self.kernels:
+			for variant in ('', '-fallback'):
+				if uki_enabled:
+					entry = f'"Arch Linux ({kernel}{variant}) UKI" "{kernel_params}"'
+				else:
+					initrd_path = f'initrd=\\initramfs-{kernel}{variant}.img'
+					entry = f'"Arch Linux ({kernel}{variant})" "{kernel_params} {initrd_path}"'
+
+				config_contents.append(entry)
+
+		config_path.write_text('\n'.join(config_contents) + '\n')
+
+		hook_contents = textwrap.dedent(
+			'''\
+			[Trigger]
+			Operation = Install
+			Operation = Upgrade
+			Type = Package
+			Target = refind
+
+			[Action]
+			Description = Updating rEFInd on ESP
+			When = PostTransaction
+			Exec = /usr/bin/refind-install
+			'''
+		)
+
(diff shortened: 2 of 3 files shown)


Torxed:
Thank you for this contribution, I'll do some tests with this one before merging in.
While I do those tests, could you fix the ruff formatting complaints?

I have never used refind, so I'll have to test and read about the logo situation :)

I could also look into the possibility to package `refind-btrfs` for us into extra instead of AUR, I just need to check the build process and see what it does before we do.

justbispo:
Done, I thought the pre-commit hook would add the formatting changes to the commit.

Torxed (approved):

Torxed:
> Done, I thought the pre-commit hook would add the formatting changes to the commit.

My understanding is that you need to enable them locally, our repo simply contains the hooks that you would need.

Torxed:
I'm still away on a trip and have limited access to proper internet.
So I won't be able to test this (installing packages and what not takes too much data).
This means I don't dare to merge this in for the next minor release that I plan on pushing today.

I still welcome this contribution, just thought I'd let you know why it won't make it in to the next release ~today. But it's coming!

Mattyan89:
I've tried this PR, the system doesn't boot if the esp is mounted to /efi instead of /boot:
Schermata_20250910_154127
Schermata_20250910_153701
The installed system has two refind_linux.conf files, one in /efi and one in /boot: 
Schermata_20250910_155153
To me it looks like it's trying to boot from the iso instead of the installed system, but I'm not sure if it is the only problem as I don't understand the error in the second screen.

justbispo:
I did some testing and noticed that the `boot_partition` and `efi_partition` point to the same `/efi` partition.
Can you try replacing the line 1562 with `config_path = self.target / 'boot' / 'refind_linux.conf'`? It worked for me when I tried to fix your situation.
I've never installed Arch with a separate partition for the ESP, can you confirm that this disk configuration is the correct one, so I know I'm testing it correctly:
```json
{
    "disk_config": {
        "btrfs_options": {
            "snapshot_config": null
        },
        "config_type": "manual_partitioning",
        "device_modifications": [
            {
                "device": "/dev/vda",
                "partitions": [
                    {
                        "btrfs": [],
                        "dev_path": null,
                        "flags": [
                            "boot",
                            "esp"
                        ],
                        "fs_type": "fat32",
                        "mount_options": [],
                        "mountpoint": "/efi",
                        "obj_id": "cbfc8315-b53e-497a-b793-8620731d3b02",
                        "size": {
                            "sector_size": {
                                "unit": "B",
                                "value": 512
                            },
                            "unit": "MiB",
                            "value": 512
                        },
                        "start": {
                            "sector_size": {
                                "unit": "B",
                                "value": 512
                            },
                            "unit": "MiB",
                            "value": 1
                        },
                        "status": "create",
                        "type": "primary"
                    },
                    {
                        "btrfs": [],
                        "dev_path": null,
                        "flags": [
                            "bls_boot"
                        ],
                        "fs_type": "ext4",
                        "mount_options": [],
                        "mountpoint": "/boot",
                        "obj_id": "33ec68f8-2106-4a14-a980-3bdcb63b3a2f",
                        "size": {
                            "sector_size": {
                                "unit": "B",
                                "value": 512
                            },
                            "unit": "GiB",
                            "value": 1
                        },
                        "start": {
                            "sector_size": {
                                "unit": "B",
                                "value": 512
                            },
                            "unit": "B",
                            "value": 537919488
                        },
                        "status": "create",
                        "type": "primary"
                    },
                    {
                        "btrfs": [],
                        "dev_path": null,
                        "flags": [],
                        "fs_type": "ext4",
                        "mount_options": [],
                        "mountpoint": "/",
                        "obj_id": "c9a4c1a3-9306-4451-b6d4-df8f394f6c42",
                        "size": {
                            "sector_size": {
                                "unit": "B",
                                "value": 512
                            },
                            "unit": "B",
                            "value": 19862126592
                        },
                        "start": {
                            "sector_size": {
                                "unit": "B",
                                "value": 512
                            },
                            "unit": "B",
                            "value": 1611661312
                        },
                        "status": "create",
                        "type": "primary"
                    }
                ],
                "wipe": false
            }
        ]
    }
}
```

Mattyan89:
Thank you, I am trying to leave /boot on the root filesystem so the kernel is backed up toghether with the rest of the system when I take a btrfs snapshot. So only /efi is mounted on a separate fat32 partition and should only contain the bootloader. This happens when booting with that line changed:
Schermata_20250912_060749
EDIT: I managed to get it to boot by editing /boot/refind_linux.conf
If the initramfs image is on a btrfs subvolume, the entry has to be: initrd=@\boot\initramfs-linux.img
[https://wiki.archlinux.org/title/REFInd#Auto_detection](https://wiki.archlinux.org/title/REFInd#Auto_detection)

justbispo:
I've investigated further and I think this last commit fixes the cases of different partition layouts.
I've tested with `/`, `/boot` and `/efi` in different partitions, only `/efi` a separate partition and `/boot` inside the `/` partition as BTRFS and ext4. It should cover all cases.

Torxed:
Good job, and thank you for the work you've put in.
I apologies for my radio silence, for personal reasons I haven't been able to dig into this one - but that does not mean this isn't desired or appreciated!

h8d13:
Bump :+1:

svartkanin (approved):

svartkanin:
Given that this is not impacting other things it should be safe to merge and collect feedback if something isn't working as expected

Issue #3705, opened 2025-08-03
pipewire-jack-1:1.4.7-1 and jack2-1.9.22-1 are in conflict


word-sys:
### Which ISO version are you using?

2025-08-1

### The installation log

```json
[2025-08-03 10:31:06] - DEBUG - Installing audio server: pipewire
[2025-08-03 10:31:06] - INFO - Installing packages: ['sof-firmware']
[2025-08-03 10:31:27] - INFO - Installing packages: ['pipewire', 'pipewire-alsa', 'pipewire-jack', 'pipewire-pulse', 'gst-plugin-pipewire', 'libpulse', 'wireplumber']
[2025-08-03 10:31:28] - ERROR - Could not strap in packages: ['/usr/bin/pacstrap', '-C', '/etc/pacman.conf', '-K', '/mnt', 'pipewire', 'pipewire-alsa', 'pipewire-jack', 'pipewire-pulse', 'gst-plugin-pipewire', 'libpulse', 'wireplumber', '--noconfirm'] exited with abnormal exit code [1]: ate[K[2E multilib is up to date[K[1Ewarning: libpulse-17.0+r43+g3e2bb8a1e-1 is up to date -- reinstalling
resolving dependencies...
looking for conflicting packages...
:: pipewire-jack-1:1.4.7-1 and jack2-1.9.22-1 are in conflict (jack). Remove jack2? [y/N] 
error: unresolvable package conflicts detected
error: failed to prepare transaction (conflicting dependencies)
:: pipewire-jack-1:1.4.7-1 and jack2-1.9.22-1 are in conflict
[?25h==> ERROR: Failed to install packages to new root

[2025-08-03 10:34:04] - ERROR - Pacstrap failed. See /var/log/archinstall/install.log or above message for error details: ['/usr/bin/pacstrap', '-C', '/etc/pacman.conf', '-K', '/mnt', 'pipewire', 'pipewire-alsa', 'pipewire-jack', 'pipewire-pulse', 'gst-plugin-pipewire', 'libpulse', 'wireplumber', '--noconfirm'] exited with abnormal exit code [1]: ate[K[2E multilib is up to date[K[1Ewarning: libpulse-17.0+r43+g3e2bb8a1e-1 is up to date -- reinstalling
resolving dependencies...
looking for conflicting packages...
:: pipewire-jack-1:1.4.7-1 and jack2-1.9.22-1 are in conflict (jack). Remove jack2? [y/N] 
[...]

PR #3709, opened 2025-08-04
move additional packages down

EK-92:
- This fix issue: 

https://github.com/archlinux/archinstall/issues/3705

## PR Description:

moved additional packages step in guided script below app_config and profile_config groups

## Tests and Checks


Torxed:
Let's try it, historically what's been working has been:

1. Audio
2. Additional Packages
3. Profiles (desktops etc)

https://github.com/archlinux/archinstall/blob/6653a4e5a9e703a65d0ea2bbc4ff8a0b9e116521/archinstall/scripts/guided.py#L107-L117

But I don't see why this wouldn't work, as long as the audio setup is first, it should be "back to normal".

Issue #3725, opened 2025-08-11
2025.08.01 KeyError: 'installation_session'


DrPsychick:
### Which ISO version are you using?

2025-08-01

### The installation log

```json
[2025-08-11 12:01:49] - INFO - Fetching Arch Linux package database...
[2025-08-11 12:01:49] - INFO - Checking version...
[2025-08-11 12:01:49] - DEBUG - Failed to check for package upgrades: archinstall
[2025-08-11 12:01:49] - DEBUG - No archinstall upgrades found
[2025-08-11 12:01:49] - INFO - Found a tailored profile for this machine called: "52-54-00-12-34-56"
[2025-08-11 12:01:59] - ERROR - Traceback (most recent call last):
  File "/usr/lib/python3.13/site-packages/archinstall/__init__.py", line 104, in run_as_a_module
    rc = main()
  File "/usr/lib/python3.13/site-packages/archinstall/__init__.py", line 94, in main
    importlib.import_module(mod_name)
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/usr/lib/python3.13/importlib/__init__.py", line 88, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 1026, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "/usr/lib/python3.13/site-packages/archinstall/scripts/unattended.py", line 19, in <module>
    install_session = storage['installation_session']
                      ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'installation_session'

[2025-08-11 12:01:59] - WARNING - Archinstall experienced the above error. If you think this is a bug, please report it to
https://github.com/archlinux/archinstall and include the log file "/var/log/archinstall/install.log".

Hint: To extract the log from a live ISO
curl -F'file=@/var/log/archinstall/install.log' https://0x0.st
```

#### Description of the issue

[...]

Issue #3517, opened 2025-05-28
KeyError: 'installation_session'


nwcs-sh:
### Which ISO version are you using?

2025.05.01

### The installation log

```json
See pic...
```

#### Description of the issue

Installing in qemu unattended, tried with 3.0.4 and 3.0.6. Worked fine with the archinstall from 2025.01.01 (unsure on which version this is).

Command is: `archinstall --config-url http://x.x.x.x/config.json --creds-url http://x.x.x.x/creds.json --silent --script unattended`

![Image](https://github.com/user-attachments/assets/25cb64eb-0eca-4fa9-947a-07fda874d4ca)

svartkanin:
Are you using this this script for testing or have you actually used this with your own script? 

Looking at the existing script shipped with archinstall https://github.com/svartkanin/archinstall/blob/836dd21f0910676befc3f3499d059733ccc19c7c/archinstall/default_profiles/tailored.py?plain=1#L19-L22 
This seems to be quite outdated, I'm not sure what the original purpose of it was, besides well automatically running it for a given host with a certain MAC address. Maybe @Torxed has some historical knowledge around it. 

I can't see how this would have worked out of the box, as it requires an installation session to be initialized in the first place, something like https://github.com/svartkanin/archinstall/blob/836dd21f0910676befc3f3499d059733ccc19c7c/archinstall/scripts/guided.py?plain=1#L59-L63.

svartkanin:
That is not the same as running the unattended script, which you seemed to have done according to the error

nwcs-sh:
I decided to retest with latest 3.0.7 and it seems to work fine. Closing.

PR #3733, opened 2025-08-13
Remove unattended script

svartkanin:
To avoid further confusion as in https://github.com/archlinux/archinstall/issues/3725 and https://github.com/archlinux/archinstall/issues/3517

The `unattended.py` script is broken in its current state and has been for a significant time. 

From https://github.com/archlinux/archinstall?tab=readme-ov-file#unattended-installation-based-on-mac-address

> Archinstall comes with an [unattended](https://github.com/archlinux/archinstall/blob/master/examples/mac_address_installation.py) example which will look for a matching profile for the machine it is being run on, based on any local MAC address. For instance, if the machine the code is executed on has the MAC address 52:54:00:12:34:56 it will look for a profile called [52-54-00-12-34-56.py](https://github.com/archlinux/archinstall/blob/master/archinstall/default_profiles/tailored.py). If it's found, the unattended installation will begin and source that profile as its installation procedure.

I'm don't think this still has any value at this point, the profiles were reworked a long time ago and it is now possible to create profile classes that can be implemented and dynamically used. 

@Torxed please tell me if I'm wrong here


Torxed:
I'm inclined to agree, I have used this quite extensively in the past.
However it's probably more reliable and easy to do `archinstall --conf-url https://x.local/$(cat /sys/class/net/enp0s3/address).json` and define the profile that way per MAC address.

This made more sense in the past when archinstall did this automatically from the main script.

PR #3737, opened 2025-08-14
Remove beta label from LVM

svartkanin:
I think this has been soaking long enough to remove the `(Beta)` from the LVM menu entry


Torxed (approved):
It's time to be brave! ^^

PR #3749, opened 2025-08-20
Better Ukrainian v.1.3.3

TSEPLNK:
Now, the translation is complete.
I haven't tested it, but poedit says it is alright, and the translation is ready to be used.


EK-92 (approved):
LGTM

Issue #3770, opened 2025-08-30
Hardware model detection fails due to the lack of virtual DMI device in sysfs


dee-fox:
### Which ISO version are you using?

2025-08-01

### The installation log

```json
n/a!
```

#### Description of the issue

Using archinstall on the following hardware (and firmware!)
Motherboard: ASRock A75M
CPU: AMD A4-3300
GPU: (the CPU above's iGPU)
RAM: 4GB
BIOS: version L2.01 [Beta] (CSM disabled, booting in EFI mode)
results in a crash due to the lack of /sys/devices/virtual/dmi (in which archinstall is trying to find vendor info)
I am as baffled as anyone should be when faced with this fact but even the fully installed system (archinstall worked after commenting out the line that tries to retrieve the system vendor) doesn't have DMI in sysfs.

#### Proposed solution
Gracefully return an "undefined" when DMI doesn't exist in sysfs.

PR #3771, opened 2025-08-30
gracefully return "undefined" if DMI is not in sysfs

dee-fox:
- This (should) fix issue: #3770

## PR Description:

Wraps access to the virtual DMI device in a try/except statement to not crash if it doesn't exist.

## Tests and Checks

### DIFF
--- a/archinstall/lib/hardware.py
+++ b/archinstall/lib/hardware.py
@@ -239,7 +239,7 @@ class SysInfo:
 		return _sys_info.cpu_info.get('model name', None)
 
 	@staticmethod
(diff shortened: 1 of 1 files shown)


svartkanin:
The function shouldn't return some ambiguous string but rather none

dee-fox:
yeah that works, checked on hardware

svartkanin:
You'll need to update the types as well and handle it on the caller side

dee-fox:
type updated, the one place this function is called seems to not really care about nulltypes (other bits of hardware.py are called in the same way and they may also return null)

PR #3780, opened 2025-09-05
Fundamentally improve German translations

veprogames:
## PR Description:

poedit and archinstall now report 100% translation rate for German. I added tr() invocations in some places and adjusted the locale generation script to properly detect them (I didn't work for me before).

## Tests and Checks
- [x] I have tested the TUI roughly<br>
  

But there's a problem: Some translations are not picked up, for example "No audio server". I didn't figure out why, did I overlook something obvious? I ran the locale generation script before committing and the .mo file was changed.

Some translations also had a different amount of {}. I don't know if I broke something without noticing.

Should I regenerate all languages because the .pot changed?


veprogames:
> But there's a problem: Some translations are not picked up, for example "No audio server".

Is it because of this? https://github.com/archlinux/archinstall/blob/9bb7ced10ee9c0116c6d43ac90fd6f7b2f97802f/archinstall/tui/help.py#L33-L36

If so, I'd remove the tr calls from Enum values

archinstall/lib/models/mirrors.py
```diff
@@ -147,9 +149,9 @@ def __eq__(self, other: object) -> bool:
 
 
 class SignCheck(Enum):
-	Never = 'Never'
-	Optional = 'Optional'
-	Required = 'Required'
+	Never = tr('Never')
```
svartkanin: I'm not sure we should translate this, it's the actual value stored in the mirror configuration

svartkanin:
Regarding the audio, the `tr(...)` in the value doesn't make sense it would need a helper function 

```
class Audio(StrEnum):
	NO_AUDIO = auto()
	PIPEWIRE = auto()
	PULSEAUDIO = auto()

	def display_text(self) -> str:
		match self:
			case Audio.NO_AUDIO:
				return tr('No audio server')
			case Audio.PIPEWIRE:
				return 'Pipewire'
			case Audio.PULSEAUDIO:
				return 'Pulseaudio'

```

feel free to ignore that for now

veprogames:
I removed the enum translations and did some final wording changes. Should I squash the commits? Is the change to locale_generator.sh ok? Are the leftover comments in de/LC_MESSAGES/base.po ok?

svartkanin:
Thanks for the changes

PR #3789, opened 2025-09-13
WIP: Add support for systemd-homed

luca020400:


luca020400:
@Torxed Any idea why running inside systemd-nspawn isn't enough to get homectl to create an user? I'm a bit lost at this point.

Reading the trace_log shows the proper stdout from homectl, the logic itself seems _sane_ to me, homectl gets to exec and finishes.

With that said fscrypt should only be allowed on expected filesystems (ext4&f2fs).

I didn't check at all how to handle fido2, and tests/other rest of lib is broken, I'll get to it if I figure out why it doesn't create the user properly.

svartkanin:
This hasn't been updated in a while and needs a lot of rebasing. I'll close this for now and we can re-open an updated version if desired

PR #3794, opened 2025-09-14
Using the 'plasma' package group for the plasma desktop profile

georgemoodier:
This should make the profile more in line with others that also use package groups

also removed `plasma-workspace`, not only is it in the package group but has been a dependency of `plasma-meta` since version 6.4-2


svartkanin:
Thanks for raising this BUT it is somewhat controversial and will definitely steer some opinions. 
It'd would probably be better to raise an issue so this can be discussed there but I'll leave some thoughts here.

Firstly, if this is the way forward then to keep things in line I'd suggest to make all profiles use their dedicated groups if available otherwise there is going to be a lot of inconsistencies. 

Looking at the included packages https://archlinux.org/groups/x86_64/plasma/ this may somewhat overlap and maybe conflict with existing settings, such as `kpipewire`, `bluedevil` being some that will now be installed automatically which makes the specific settings for them counter intuitive as not selecting anything there makes things confusing. 

I'm not against it as it would remove the endless discussions of which packages to install and which are not needed as archinstall just pushes that responsibility then to the packaging maintainers (which will not be pleased as @Torxed pointed out to me in the past :) )

h8d13:
This can also probably be closed using smaller in #4076 and the rest is handled by archinst's other features

georgemoodier:
Thanks for the update on the profile! I’ll close this PR and its related issue

georgemoodier:
I'm reopening this PR as a (possible) alternative to #4336

NOTE: this PR requires a package group dependency backend that addresses package group installation according to users' needs, hence why I've only reopened it as solely a draft (I'm unable to set this backend up as I lack the python proficiency.) If this backend is implemented I'll allow this PR to be un-drafted and ultimately pulled. (see #4392)

svartkanin:
I was addressed with https://github.com/archlinux/archinstall/pull/4389 and a more granular Plasma choice is now available

h8d13:
> I was addressed with #4389 and a more granular Plasma choice is now available

#4449 perhaps would be nice to clarify which option is what (full, mid, small)

Issue #3800, opened 2025-09-18
Failed to enable fstrim.timer error


MythSnipper:
### Which ISO version are you using?

modified version from 2025-04-01

### The installation log

```json
[2025-09-18 13:28:41] - DEBUG - Hardware model detected: LENOVO 2464EB3; UEFI mode: True
[2025-09-18 13:28:41] - DEBUG - Processor model detected: Intel(R) Core(TM) i7-3520M CPU @ 2.90GHz
[2025-09-18 13:28:41] - DEBUG - Memory statistics: 6384092 available out of 7950888 total installed
[2025-09-18 13:28:41] - DEBUG - Could not detect virtual system: ['/usr/bin/systemd-detect-virt'] exited with abnormal exit code [1]: none

[2025-09-18 13:28:41] - DEBUG - System is not running in a VM: ['/usr/bin/systemd-detect-virt'] exited with abnormal exit code [1]: none

[2025-09-18 13:28:41] - DEBUG - Virtualization detected: None; is VM: False
[2025-09-18 13:28:41] - DEBUG - Graphics devices detected: dict_keys(['Intel Corporation 3rd Gen Core processor Graphics Controller (rev 09)', 'NVIDIA Corporation GF108M [NVS 5400M] (rev a1)'])
[2025-09-18 13:28:41] - DEBUG - Disk states before installing:
{
    "blockdevices": [
        {
            "name": "loop0",
            "path": "/dev/loop0",
            "pkname": null,
            "log_sec": 512,
            "size": "988 MiB",
            "pttype": null,
            "ptuuid": null,
            "rota": false,
            "tran": null,
            "partn": null,
            "partuuid": null,
            "parttype": null,
            "uuid": null,
            "fstype": "squashfs",
            "fsver": "4.0",
            "fsavail": 0,
            "fsuse_percentage": "100%",
            "type": "loop",
            "mountpoint": "/run/archiso/airootfs",
            "mountpoints": [
                "/run/archiso/airootfs"
            ],
            "fsroots": [
                "/"
            ],
            "children": []
        },
        {
            "name": "sdb",
            "path": "/dev/sdb",
            "pkname": null,
            "log_sec": 512,
            "size": "953869 MiB",
            "pttype": "gpt",
            "ptuuid": "7e9edf70-bce4-4175-9e82-50869c8a00fb",
            "rota": true,
            "tran": "ata",
            "partn": null,
            "partuuid": null,
            "parttype": null,
            "uuid": null,
[...]

PR #3815, opened 2025-09-18
quick fix to run archinstall and enable systemd services

brunolzrosa:
- This fix issue: #3800

## PR Description:

Just inserted a direct call to SysCommand instead of self.arch_chroot, with the flag --root={self.target}.
This fixed the issue, at least for me.

## Tests and Checks
- [ X ] I have tested the code!<br>

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -667,7 +667,7 @@ class Installer:
 			info(f'Enabling service {service}')
 
 			try:
-				self.arch_chroot(f'systemctl enable {service}')
+				SysCommand(f'systemctl --root={self.target} enable {service}')
 			except SysCallError as err:
 				raise ServiceException(f'Unable to start service {service}: {err}')
 


correctmost:
There was a similar suggestion on the Arch GitLab: https://gitlab.archlinux.org/archlinux/packaging/packages/systemd/-/issues/54#note_322382

btantlinger:
The fix partially works, in that I am able to get Arch installed using archinstall... but still seems like something is broken.  If I arch-chroot into the system after the install and run my post-install script, any systemctl command is failing with:

`Failed to connect to system scope bus via local transport: No data available`

This worked before the systemd update.

correctmost:
> The fix partially works, in that I am able to get Arch installed using archinstall... but still seems like something is broken.

There's an underlying issue in systemd: https://github.com/systemd/systemd/pull/39021

archinstall/lib/installer.py
```diff
@@ -667,7 +667,7 @@ def enable_service(self, services: str | list[str]) -> None:
 			info(f'Enabling service {service}')
 
 			try:
-				self.arch_chroot(f'systemctl enable {service}')
+				SysCommand(f'systemctl --root={self.target} enable {service}')
```
YHNdnzj, line 670: This operates on a chroot in a way native to systemd tools and should be done regardless of arch-chroot working or not.

YHNdnzj (approved):
Note that I've updated the PR description to highlight it's not a systemd bug after all. arch-chroot needs to be fixed and the systemd change just revealed the flaw there.

svartkanin (approved):

archinstall/lib/installer.py
Torxed: We have to remember that we're on Arch Linux and the recommended way *(by our own Wiki)* is/was to configure the system using `arch-chroot`, generally speaking.

It's unfortunate that this issue happens, and we'll try to work around it.
But we wish to use native tooling with default values wherever possible and not use workarounds or too many custom parameters *(again, if possible)*.

Issue #2750, opened 2024-10-22
zstd error 70 write error cannot write block broken pipe on all kernels from archinstall


claudemods:
### Which ISO version are you using?

01/10/2024 "Uk date format"

### The installation log

```json
[
[install.log](https://github.com/user-attachments/files/17484674/install.log)
](url)
```

### describe the problem

I use a vm to test and install mainly i first found a issue with calamares 4 weeks back which led me to try and make a iso with archinstall since i make my own iso creation tools exactly like penguins eggs 

Proceeding to make a system from arch install i found this issue in below photo everything install and when i boot to the newely installed system usbs dont read or harddrives

If i install the kernel on a system with a older kernel preinstalled and make a isos if i use calamares in debug mode it says file format corruption its been 4-5 weeks now with these issues

Im in the uk i use worldwide mirrors no de no swap ext4 no unified kernel and no seperate /home

if i use kernel 6.10.10.zen1-1 or any kernel uploaded on 12th september e.g LTS everything works 

I had to upload photo in comments 
I will post log in photos later also

claudemods:
![Screenshot_20241021_054314_Gallery](https://github.com/user-attachments/assets/1a2e83c5-c9e9-4a6a-9d02-bba2d3a86695)

claudemods:
[install.log](https://github.com/user-attachments/files/17484658/install.log)
this is arch install log with latest linux kernel

ciprianoaljhon:
got the same error when trying to archinstall but with additional error on log file related to swap even tho I configured it correctly. And now I can't install arch :<

RNTUNOOB:
Did you find solution for this ?

Torxed:
I'm actually not sure what the error here is.
The log says "Installation completed without any errors. You may now reboot." and the picture above shows `mkinitcpio` performing a successful image generation?

claudemods:
> . You may now reboot." and the picture above shows `mkinitcpio` performing a successful image generation?

ah my encounter wasnt related to this error and well this was many months ago now when kernel 6.12.4 was released

correctmost:
Here's a related bug report: https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio/-/issues/285

PR #3821, opened 2025-09-19
First contrib

h8d13:
Hi, first of all thanks for so much work put in this project! 

Aims to make the install slightly faster and easier to test, also to save bandwidth which doesn't grow on trees :deciduous_tree:   

See code comments for more info.

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -554,7 +554,9 @@ class Installer:
 		info(f'Updating {fstab_path}')
 
 		try:
-			gen_fstab = SysCommand(f'genfstab {flags} {self.target}').output()
+			# note f flag for filter preventing host system mounts see issue #3599 
+			# this was causing issue to do arch to arch install without using ISO good for devs
+			gen_fstab = SysCommand(f'genfstab {flags} -f {self.target} {self.target}').output()
 		except SysCallError as err:
 			raise RequirementError(f'Could not generate fstab, strapping in packages most likely failed (disk out of space?)\n Error: {err}')
 
@@ -782,7 +784,9 @@ class Installer:
 			mkinit.write(content)
 
 		try:
-			SysCommand(f'arch-chroot {self.target} mkinitcpio {" ".join(flags)}', peek_output=True)
+			# avoid broken pipe errors by not peeking output
+			info('Running mkinitscpio quietly...')
+			SysCommand(f'arch-chroot {self.target} mkinitcpio {" ".join(flags)}')
 			return True
 		except SysCallError as e:
 			if e.worker_log:
--- a/archinstall/lib/pacman/__init__.py
+++ b/archinstall/lib/pacman/__init__.py
@@ -73,12 +73,12 @@ class Pacman:
 					packages = result
 
 		info(f'Installing packages: {packages}')
-
+		# note needed flag to avoid re-installs and save bandwidth for user and mirror providers
 		self.ask(
 			'Could not strap in packages',
 			'Pacstrap failed. See /var/log/archinstall/install.log or above message for error details',
 			SysCommand,
-			f'pacstrap -C /etc/pacman.conf -K {self.target} {" ".join(packages)} --noconfirm',
+			f'pacstrap -C /etc/pacman.conf -K {self.target} {" ".join(packages)} --noconfirm --needed',
 			peek_output=True,
 		)
 


Torxed:
Could you elaborate on this:

> Second: change is to clean up broken pipes logs.

When did broken pipe happen?
As this change will mute all output from `mkinitcpio` which might be handy, both for the user seeing progress but also as it can warn about things silently otherwise.

Other than that little question, thank you for the contribution!

h8d13:
Second change was purely cosmetic: on many installs I did all of them had broken zstd pipes (I'm guessing from the def that is capturing the output of mkinitcpio either that or what is redirecting to cmd_output or cmd_history)  It happens during the output of said section. 

Nothing major on that side. And since it still captures error I thought it could be ran quietly. 

I also made this little thing here that I thought could help for what I was saying about declarative deps + env detection. 

``` 
#!/bin/bash
# ARCHINSTALL SYSTEM SUBDEPS SCRIPT
PACKAGES=(
    # Required (core features)
    "arch-install-scripts"
    "grub"
    "mkinitcpio"
    "pciutils"
    "dosfstools"
    # Optional (depending on options picked in archinstall)
    "btrfs-progs"
    "cryptsetup"
    "libfido2"
    # Core (sanity check)
    "coreutils"
    "e2fsprogs"
    "glibc"
    "shadow"
    "systemd"
    "util-linux"
    "python"
)

MISSING=()

# reliable hacky way of knowing we are using ISO
if [ -d "/run/archiso" ]; then
    echo "Running from Arch Linux ISO / USB Medium"
else
    echo "Running from live system"
    for pkg in "${PACKAGES[@]}"; do
        if pacman -Qi "$pkg" &>/dev/null; then
            version=$(pacman -Qi "$pkg" | grep "^Version" | awk '{print $3}')
            echo "OK $pkg ($version)"
        else
            echo "MISSING $pkg"
            MISSING+=("$pkg")
        fi
    done
fi

# Check python modules
PYTHON_MODULES=(
    "cryptography:python-cryptography"
    "parted:python-pyparted"
    "pydantic:python-pydantic"
)

for module_pkg in "${PYTHON_MODULES[@]}"; do
    module="${module_pkg%%:*}"
    package="${module_pkg##*:}"
    if python -c "import $module" 2>/dev/null; then
        version=$(pacman -Qi "$package" 2>/dev/null | grep "^Version" | awk '{print $3}')
        echo "OK $package (${version:-unknown})"
    else
        echo "MISSING $package ($module)"
        MISSING+=("$package")
    fi
done

[ ${#MISSING[@]} -gt 0 ] && echo -e "\nInstall missing: sudo pacman -S ${MISSING[*]}"
``` 


You would get output something like this or the command for installing them. 

``` 
Running from live system
OK arch-install-scripts (29-1)
OK grub (2:2.12.r359.g19c698d12-1)
OK mkinitcpio (39.2-5)
OK pciutils (3.14.0-1)
OK dosfstools (4.2-5)
OK btrfs-progs (6.16.1-1)
OK cryptsetup (2.8.1-1)
OK libfido2 (1.16.0-2)
OK coreutils (9.7-1)
OK e2fsprogs (1.47.3-1)
OK glibc (2.42+r17+gd7274d718e6f-1)
OK shadow (4.18.0-1)
OK systemd (258-2)
OK util-linux (2.41.1-1)
OK python (3.13.7-1)
OK python-cryptography (46.0.1-1)
OK python-pyparted (3.13.0-5)
OK python-pydantic (2.11.9-1)
```  

Could be useful for debugging stuff around subdeps or when systemd chaos arises...  Also just for devs who want to try help with archinstall. 

Anyways thanks for taking the time to look at my contribution and again thanks for all the hard work in this project.

correctmost:
> Second change was purely cosmetic: on many installs I did all of them had broken zstd pipes

This might be a mkinitcpio bug instead of a `peek_output` bug:
- https://github.com/archlinux/archinstall/issues/2750
- https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio/-/issues/285

h8d13:
Yeah might not be as important to be honest. I can remove it if needed. 

Ooh while i have your attention: I was wondering how can we request for AUR packages to become core or extra ? 

Basically there is this large perl script called ckbcomp that is useful for setting keymap in grub [see here](https://github.com/h8d13/GRUB2_KEYMAPS_FIX/tree/master) especially useful for rescue shell, editing launch lines and also if setting grub passwords.

PR #3830, opened 2025-09-23
Fixed #3828

### LINKED ISSUE #3828: Archinstall installs nvidia-open-dkms instead of nvidia-open, even on default linux kernel.
Arch wiki says that nvidia-open-dkms is a package for custom kernels, and that if using the default kernel you should only install nvidia-open.

Blayung:
- This fix issue: #3828 -->

## PR Description:

I've added the ability to choose between dkms and non-dkms nvidia driver versions.

### DIFF
--- a/archinstall/lib/hardware.py
+++ b/archinstall/lib/hardware.py
@@ -43,7 +43,9 @@ class GfxPackage(Enum):
 	LibvaMesaDriver = 'libva-mesa-driver'
 	LibvaNvidiaDriver = 'libva-nvidia-driver'
 	Mesa = 'mesa'
+	Nvidia = 'nvidia'
 	NvidiaDkms = 'nvidia-dkms'
+	NvidiaOpen = 'nvidia-open'
 	NvidiaOpenDkms = 'nvidia-open-dkms'
 	VulkanIntel = 'vulkan-intel'
 	VulkanRadeon = 'vulkan-radeon'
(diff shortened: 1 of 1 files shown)


dg4l (approved):
good change

svartkanin:
This needs to be made backwards compatible as existing configurations that are loaded will not parse anymore

archinstall/lib/hardware.py
```diff
@@ -59,9 +61,11 @@ class GfxDriver(Enum):
 	AllOpenSource = 'All open-source'
 	AmdOpenSource = 'AMD / ATI (open-source)'
 	IntelOpenSource = 'Intel (open-source)'
-	NvidiaOpenKernel = 'Nvidia (open kernel module for newer GPUs, Turing+)'
+	NvidiaOpenKernel = 'Nvidia (open kernel module for newer GPUs, Turing+, for default kernel)'
```
svartkanin: This will break existing configs and nwwds to be gandled properly

svartkanin (changes requested):

archinstall/lib/hardware.py
dg4l: > This will break existing configs and nwwds to be gandled properly

didn't notice this when reviewing, sorry. qwq

h8d13:
I like this change idea. Faster installs without dkms. 

But then what happens if user picks custom kernel and nvidia driver non dkms

Also the headers would still be added regardless see:

`profiles_handler.py : def install_gfx_driver`

svartkanin:
This was addressed with https://github.com/archlinux/archinstall/pull/4343

PR #3845, opened 2025-09-30
Add no-home-volume layout for LVM installation

recolic:
- This fix issue: 

## Tests and Checks


recolic:
I think `get_default_btrfs_subvols()` might also need a `HomeVolume=True/False` option?

svartkanin:
The description on the top of the configuration still reads 
`Configuration: Default layout (root volume + seperate home volume)` even though no home was selected. 

The saved configuration also doesn't seem to store the correct value 
```
...
        "lvm_config": {
            "config_type": "default",
            "vol_groups": [
                {
...
```

I'd also rename the options to be 
* Default layout
* Root only

Issue #3834, opened 2025-09-24
Pantheon


ProstoJarno:
### describe the request

Pls add pantheon😀

svartkanin:
PR contributions are welcome

PR #3857, opened 2025-10-07
Add Pantheon desktop environment profile

P-rth:
adding pantheon support; addressing #3834 

## PR Description:
this pull request adds support for pantheon desktop by adding pantheon.py in archinstall/default_profiles/desktops/

## Tests and Checks


svartkanin (approved):

svartkanin:
Thanks, could you please check the linting and formatting otherwise it looks fine

P-rth:
really sorry for making a change again and again, Im new to contributing to large projects so still figuring it out, the issue now left is just a single empty line (line 8) right?

svartkanin:
How did you choose these particular packages for the installation? 
Looking at the Arch wiki it recommends the package group `pantheon` to be installed https://archlinux.org/groups/x86_64/pantheon/, which archinstall shouldn't as it should keep things as minimal as possible. 

BUT in your list there are 
* gnome-settings-daemon
* pantheon-print

Which are not even in that group, so I'm curious where this list of packages was derived from?

P-rth:
well while selecting the packages, I mostly used these
https://wiki.archlinux.org/title/Pantheon
https://github.com/jeremybosma/ArchPantheon
gnome-settings-daemon is mentioned in the ArchPantheon install script and I thought that would be crutial for working
pantheon-print is mentioned in the archwiki 

would It be be better if we close this pr and just open a new one

svartkanin:
Have you tried installing it actually with your configuration?

P-rth:
sorry for the delay, but I have tested the changes by making a iso and it seems like pantheon desktop is getting installed, but there are some issues when starting it :-
1) the lightdm login screen has a completely black background
2) when selecting pantheon it just never starts and loops back to the login screen

currently all the packages that are suggested in the archwiki are installed, but I think there are some configuration issues with the greeter, and the official greeter is in the AUR and not official repos

Torxed:
So addressing your two questions,

1. The default greeter in `archinstall` is `lightdm-slick-greeter` when you select this desktop *(because of how this PR is structured, which is okay)*, which by default is assumed to give a black background. You'd need to add a [background=](https://github.com/linuxmint/slick-greeter?tab=readme-ov-file#credit) with a path to the image you'd want as the background. This does not happen by default unless the window manager/desktop environment set it.
2. This is because the default desktop environment that `lightdm-slick-greeter` tries to launch is "classic session", which won't load `pantheon` I'm guessing:
   screenshot
   I don't know much about pantheon, but if you select "Secure session" it starts into this:

   screenshot
   Which looks like a desktop, but I'm not sure if it's the "pantheon you expect"?

Torxed:
I'm also wondering if you're sure the greeter is in AUR?
Because in `extra` we have [lightdm-pantheon-greeter](https://archlinux.org/packages/extra/x86_64/lightdm-pantheon-greeter/) which sounds a bit official?

P-rth:
First of all thanks for all the info and help,
Yeah I think I might have wrongly stated that it was in the aur. Sorry about that.

But for this to be used we would need to add this in the greeters too right?, I did not do that because I was unsure if it was ok to add a new greeter specifically for a single desktop environment

Furthermore yeah the pantheon that opens up right now is kind of broken
Some menus and dock don't work and the on screen keyboard opens up everytime 

This is the expected look

!pantheon ](https://github.com/user-attachments/assets/c4ee528e-ca71-4753-ac61-2d5580e86643)

P-rth:
Ive been trying to get a fix for this, the farthest ive gotten with only the packages in the arch repos is this:
https://www.youtube.com/watch?v=wmOXOfgByFw
(Gnome session , wayland gnome and classic session does not work, please ignore the donny speech module)
I have changed the greeter to pantheon's one and installed all the packages as listed in the archwiki

Installing is manually by following the archwiki step by step olso results in this kindof broken desktop
I have shared some of the crash logs in attached files.

I don't have much experience with pantheon tbh can someone help me out here?
[lightdm-debug.log](https://github.com/user-attachments/files/25053666/lightdm-debug.log)
[gala_crash_details.log](https://github.com/user-attachments/files/25053668/gala_crash_details.log)
[env_config.log](https://github.com/user-attachments/files/25053669/env_config.log)
[explicit_packages.txt](https://github.com/user-attachments/files/25053733/explicit_packages.txt)
[installed_packages.txt](https://github.com/user-attachments/files/25053734/installed_packages.txt)

svartkanin:
This has been outdated and without any updates for some time now, I'll close this and we can reopen it again if there is new interest in it

PR #3865, opened 2025-10-12
Do not install Btrfs module and binary in mkinitcpio

ventureoo:
This is what btrfs hook already does.

- This fix issue: None 

## PR Description:

This is what btrfs hook from btrfs-progs already does: https://gitlab.archlinux.org/archlinux/packaging/packages/btrfs-progs/-/blob/main/initcpio-install-btrfs

## Tests and Checks


svartkanin:
As these functions were only used for btrfs filesystem cases, it would be good to remove them all together.

ventureoo:
> As these functions were only used for btrfs filesystem cases, it would be good to remove them all together.

Done

Issue #808, opened 2021-12-21
mkinitcpio.conf generated incorrectly for AMDGPU.


Galcian79:
As the archwiki installation guide states [https://wiki.archlinux.org/title/AMDGPU#Specify_the_correct_module_order](https://wiki.archlinux.org/title/AMDGPU#Specify_the_correct_module_order), you must ensure that the amdgpu module is loaded before the radeon one: `MODULES=(amdgpu radeon)`
Otherwise the DM will fail to start at boot.

Torxed:
We'll have to implement a package log in: https://github.com/archlinux/archinstall/blob/638a05243c7bab7db7449c24a2ab770e581244d0/archinstall/lib/installer.py#L264

And then in the xorg profile create a similar conditional update on https://github.com/archlinux/archinstall/blob/d2a4572e98044d4af504aabd0203b6c89e9991bf/profiles/xorg.py#L40

Torxed:
Fixed, realized we could detect it based on the graphics driver selected in the prep function. This fix will be out in v2.3.1 :) Thank you for reporting this!

PR #3866, opened 2025-10-12
Do not re-order amdgpu and radeon modules in mkinitcpio

ventureoo:
This is not necessary as kms hook for mkinitcpio already takes care of adding amdgpu and radeon modules.

- This fix issue: None 

## PR Description:

Relevant issue (https://github.com/archlinux/archinstall/issues/808) was created in 2021, but in 2022 mkinitcpio had added `kms` hook (https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio/-/commit/7bfe4861eacb3bf6cb70d9a17a0262542733a8ed), which already takes care of adding amdgpu and radeon modules and it used by default.

## Tests and Checks

### DIFF
--- a/archinstall/lib/profile/profiles_handler.py
+++ b/archinstall/lib/profile/profiles_handler.py
@@ -228,13 +228,6 @@ class ProfileHandler:
 			headers = [f'{kernel}-headers' for kernel in install_session.kernels]
 			# Fixes https://github.com/archlinux/archinstall/issues/585
 			install_session.add_additional_packages(headers)
-		elif driver in [GfxDriver.AllOpenSource, GfxDriver.AmdOpenSource]:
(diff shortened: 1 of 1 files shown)


svartkanin:
According to the wiki https://wiki.archlinux.org/title/AMDGPU#Load_amdgpu_driver

ventureoo:
> According to the wiki https://wiki.archlinux.org/title/AMDGPU#Load_amdgpu_driver

This is written in the context of enabling support for Southern Islands and Sea Islands in amdgpu. This is not enabled by default and archinstall does nothing to enable it.

svartkanin:
Ahh I missed that we don't enable that, then it seems to be good

PR #3867, opened 2025-10-13
Upgrade Pylint to 4.0.0

correctmost:
## PR Description:

This commit also removes the pylint-pydantic plug-in, which is not yet compatible with Pylint v4.

Related Pylint bug report:
- https://github.com/pylint-dev/pylint/issues/10647

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -43,7 +43,6 @@ makedepends=(
   'python-wheel'
   'python-sphinx_rtd_theme'
   'python-pylint'
-  'python-pylint-pydantic'
   'ruff'
 )
 optdepends=(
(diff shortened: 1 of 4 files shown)


PKGBUILD
```diff
@@ -43,7 +43,6 @@ makedepends=(
   'python-wheel'
   'python-sphinx_rtd_theme'
   'python-pylint'
-  'python-pylint-pydantic'
```
correctmost, line 46: @Torxed - The `PKGBUILD` file in https://gitlab.archlinux.org/archlinux/packaging/packages/archinstall might need to be updated too

svartkanin:
Thanks for fixing the upgrade

Issue #1634, opened 2023-01-29
Option to connect to wifi from arch installer


godalming123:

yashmohod:
https://wiki.archlinux.org/title/Iwd#iwctl

godalming123:
Maybe add a tui menu to make this easier or give the user a brief explanation of how to connect to wifi if the installer cannot ping the archlinux mirros EG:
**Failed to connect to network, how to connect to internet:**
- Ethernet: Arch Should detect you hardware automatically
- Wifi - use IWCTL:
  1. Get a list of wifi devices `iwctl device list`
  2. Turn device on: `iwctl device ${device} set-property Powered on`
  3. Scan for WiFi networks available on device and show them: `iwctl station ${device} scan; iwctl station ${device} get-networks`
  4. Finnally connect (if there is a network password the utility will prompt you): `iwtcl station ${device} connect SSID`

PR #3879, opened 2025-10-21
Wifi connection menu with textual

svartkanin:
This introduces a bunch of new things, the primary goal was to introduce a new wifi connection menu to address #1634 . When implementing it it became clear that curses already has challenges with inputs of non ASCII characters which was "okay" for the existing inputs such as username and password configurations. But for existing wifi passwords that most likely contain a special character beyond ASCII it will not work. It could be run in raw mode but it would require more customisations and handling.

Therefore I opted to introduce textual https://textual.textualize.io. It currently lives alongside the curses implementation and is only used for the new Wifi connection menu which is only triggered if archinstall is run in offline mode and wpa_cli can be run/enabled successfully. 

If this is a good way forward and accepted I will migrate existing menus over from curses to textual.


Torxed (approved):
Looks good, and good work on keeping it separated from the other UI stuff so we can swap/test stuff!

Torxed:
I'll make a release to fix all the bugs, and then I'll merge this after that.
That way this can get in to the next ISO but saves me some worries for the smaller patch release hehe.

Torxed:
This is now released! :bomb:

h8d13:
Hey @svartkanin was wondering the menu when pressing 'NO' to connect still says "scanning for networks" is this expected?

Then sometimes click no it just kicks me back to terminal, then pressed 'yes' connected and it brought me back to terminal again saying 'failed' but actually worked. Only then i was able to get into menu :/

Testing on a 2016 MSI if that's relevant

svartkanin:
I made some changes in the open textual migration PR, are you still experiencing it there as well?

h8d13:
Have not tested there yet, was on master

As much as I think setting up WiFi for the user is a good thing, perhaps enforcing is a bit extreme 😅

svartkanin:
There's nothing enforced. If archinstall is launched without internet access it will terminate. The new wifi option at that point gives the opportunity to connect and if the user chooses not to just terminate

h8d13:
But what was weird is that when pressing 'no' it still showed "scanning networks" screen. perhaps again just something you corrected in your future PR but on master

Issue #2776, opened 2024-11-07
Change "Formatting /dev/sdX in ..." message if nothing will actually be formatted


trustytrojan:
I got a scare when I saw the "Formatting <device> in 5..." message even though I setup the partitions such that none of them said "create" or "modify" in the table shown by archinstall. I would definitely prefer that the message only be shown if anything will actually be formatted...

Here's the print statement of the "Formatting" message: https://github.com/archlinux/archinstall/blob/master/archinstall%2Flib%2Fdisk%2Ffilesystem.py#L49

And here's where partitions are filtered for "create" or "modify" partitions:
https://github.com/archlinux/archinstall/blob/master/archinstall%2Flib%2Fdisk%2Ffilesystem.py#L98-L99

This isn't a crucial issue but I feel it could prevent headaches for many users.

andreyrd:
Yes this freaked me out lol

PR #3880, opened 2025-10-21
Fix 3863

### LINKED ISSUE #3863: Final warning message "Formatting /dev/sdX" is wrong and scary
Hello, first time install of arch, I picked a specific partition to install the root fs (not the whole drive), and the final warning it gave me was
`Formatting /dev/nvme0n1 in 5...4..` -> where nvme0n1 is the drive, not the selected partition (which was nvme0n1p6).

It was scary because I have a data partition on this disk I did not want to lose (no backup yolo)

I think this is the code responsible for this line
https://github.com/archlinux/archinstall/blob/b0b87400c275908d9761b4761f0bdb4d9afe36fc/archinstall/lib/disk/filesystem.py#L343

Could this be maybe changed to only include the partitions that will be impacted, rather than the whole device? 
Something like `Formatting selected partitions in /dev/nvme0n1 in 5...4..`

svartkanin:
Fixes #3863 and #2776 

Change the final warning message from 
```Formatting in 1...2...3```

to something more generic as it's difficult to determine what each configuration may entail 

```Starting device modifications in 1...2...3```

### DIFF
--- a/archinstall/lib/disk/filesystem.py
+++ b/archinstall/lib/disk/filesystem.py
@@ -338,7 +338,7 @@ class FilesystemHandler:
 	def _final_warning(self) -> bool:
 		# Issue a final warning before we continue with something un-revertable.
 		# We mention the drive one last time, and count from 5 to 0.
-		out = tr('Starting device operations in ')
+		out = tr('Starting device modifications in ')
 		Tui.print(out, row=0, endl='', clear_screen=True)
 
 		try:

### CODE REVIEW

File: archinstall/lib/disk/filesystem.py [hunk 1]
```diff
@@ -337,10 +335,10 @@ def _lvm_vol_handle_e2scrub(self, vol_gp: LvmVolumeGroup) -> None:
 				Size(256, Unit.MiB, SectorSize.default()),
 			)
 
-	def _final_warning(self, device_paths: str) -> bool:
+	def _final_warning(self) -> bool:
 		# Issue a final warning before we continue with something un-revertable.
 		# We mention the drive one last time, and count from 5 to 0.
-		out = tr(' ! Formatting {} in ').format(device_paths)
+		out = tr('Starting device modifications in ')
```
correctmost, line 341: I think the `We mention the drive one last time, and count from 5 to 0.` comment on line 340 should be updated too

unsignedchar-256:
thanks for the fix! well i thought i almost wiped my entire SSD which contains 9 partitions.

PR #3904, opened 2025-11-06
Update Italian translation

vanmatten:
- Translated new Italian strings
- Run locales_generator


Torxed:
Could you submit this PR without running `locales_generator` as it touches all the other languages too (or at least not do `git add -A .` as I prefer to merge languages separately).

vanmatten:
I amended the commit to include only italian files.

Issue #3160, opened 2025-02-05
Password inputs don't respect the loaded keymap when using diacritics (using pt-latin1, supect others will have the same fate)


BrunoMSantos:
### Which ISO version are you using?

2025/02/01

### The installation log

```json
I ended up not using archinstall in the end for other reasons and I didn't manage to unencrypt the SSD afterwards nor did I remember to check the log before unmounting / rebooting / reinstalling.
```

#### Description of the issue

Trying the guided install for the 1st time, and `loadkeys pt-latin1` before starting, I noticed that SSD and account passwords with diacritic characters would not register correctly. I noticed it only when trying to boot the system afterwards and finding I couldn't.

By characters with diacritics, I mean those one can input directly in one/two strokes with certain layouts: ç, á, ã, â, ä and the like.

At the 2nd attempt (imagining I had mistyped and not being able to guess exactly how the password had been clobbered) I could actually see 2 '**' when inputting a 'single' character with a diacritic, which does look suspicious. Maybe a unicode canonization issue, converting between different byte representations? Didn't look at the code.

Using passwords without diacritics worked fine while still respecting the keyboard layout otherwise.

Beware I was battling some sporadic crashes that I haven't had the time to investigate and had to restart the process several times for each successful attempt. I'll try to update and create issues for the crashes, but realistically, I'll only look at this again in a couple of months or so.

svartkanin:
Looks like a duplicate of https://github.com/archlinux/archinstall/issues/3155

svartkanin:
This patch won't fix the problem per se but should avoid any confusion as the underlying curses does not support any non-ASCII characters https://github.com/archlinux/archinstall/pull/3244

PR #3913, opened 2025-11-09
--needed prevent re-installs & fix #issues/3599

h8d13:
- Fix host-to-target installs using `-f` in genfstab
- Fix overlapping meta package vs drivers installs using `--needed` grep install logs for `re-installing`
- Fix Snapper-Grub integration similar to how `timeshift` is handled
- Fix mkinitcpio v40 `vconsole.conf` hook 

- Added support for hybrid laptops to conditionally add `nvidia-prime` 
- Added VM fix as many apps need software render fallback `vulkan-swrast`

Tested many times using full-disk best effort: on `MSI GP72 6QE` Nvidia 950M + Intel iGPU

This would notably effect positively: Both amd/intel iGPU + NVIDIA dGPU setups:
- Acer Predator/Nitro series 
- Asus ROG/TUF series
- HP Omen/Envy/Zbook
- Most MSI laptops
- ...

Allowing for `prime-run <intensive-app>`

EDIT: Readme fix because I see many logs(mostly on reddit) using outdated version since it's not mentioned you need to update archinstall and then some people recpmmend -Syu archinstall on ISO which isn't great.

I have also added documentation at each step so that if you don't use the full PR it might still help =)

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -554,7 +554,7 @@ class Installer:
 		info(f'Updating {fstab_path}')
 
 		try:
-			gen_fstab = SysCommand(f'genfstab {flags} {self.target}').output()
+			gen_fstab = SysCommand(f'genfstab {flags} -f {self.target} {self.target}').output()
 		except SysCallError as err:
 			raise RequirementError(f'Could not generate fstab, strapping in packages most likely failed (disk out of space?)\n Error: {err}')
 
--- a/archinstall/lib/pacman/__init__.py
+++ b/archinstall/lib/pacman/__init__.py
@@ -78,7 +78,7 @@ class Pacman:
 			'Could not strap in packages',
 			'Pacstrap failed. See /var/log/archinstall/install.log or above message for error details',
 			SysCommand,
-			f'pacstrap -C /etc/pacman.conf -K {self.target} {" ".join(packages)} --noconfirm',
+			f'pacstrap -C /etc/pacman.conf -K {self.target} {" ".join(packages)} --noconfirm --needed',
 			peek_output=True,
 		)
 


svartkanin:
Could you please fix the linting and formatting, you may wanna setup the pre-commit hooks

h8d13:
Hi @svartkanin thanks for tip.  And also, for reproducing the bug in my issue report couple months ago. 

Seems to pass now, sry new to larger code-bases.

h8d13:
Screenshot_20251111_133558
Screenshot_20251111_134321
Screenshot_20251111_134435

h8d13:
**Final note:**  Tested using the patch and official ISO: with layout `be-latin1`

And LUKS2 unlock with grub worked out of the box without requiring ckbcomp files <3

My understanding from: # https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio/-/commit/d0a98218d2444b939cf4932265cc70a6a89a8d22

Is that this call for hook keymap is not needed anymore. And the font is also set there: https://man.archlinux.org/man/vconsole.conf.5

Ln782:

`self._hooks = [hook.replace('systemd', 'udev').replace('sd-vconsole', 'keymap consolefont') for hook in self._hooks]`

also fixes #3155 and #3160

archinstall/lib/installer.py
```diff
@@ -72,8 +72,8 @@ def __init__(
 		"""
 		self._base_packages = base_packages or __packages__[:3]
 		self.kernels = kernels or ['linux']
+		self.config: dict[str, Any] = {}
```
refcar: for what?

h8d13:
Added a def for the vconsole.conf and also `FONT=` option


image

svartkanin:
There are a lot of different changes in this PR and to keep an overview of what is being changed I'd suggest to split them into separate PRs

archinstall/lib/hardware.py
```diff
@@ -134,8 +136,12 @@ def gfx_packages(self) -> list[GfxPackage]:
 			case GfxDriver.VMOpenSource:
 				packages += [
 					GfxPackage.Mesa,
+					GfxPackage.VulkanSwrast,
 				]
 
+		if self.is_nvidia() and (SysInfo.has_intel_graphics() or SysInfo.has_amd_graphics()):
```
svartkanin: shouldn't the second clause contain a `not` ?

archinstall/lib/installer.py
svartkanin: Not needed?

archinstall/lib/installer.py
```diff
@@ -114,6 +114,8 @@ def __init__(
 			'filesystems',
 			'fsck',
 		]
+		# https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio/-/commit/d0a98218d2444b939cf4932265cc70a6a89a8d22
```
svartkanin: Why do we need this comment?

archinstall/lib/installer.py
```diff
@@ -552,9 +554,10 @@ def set_mirrors(
 	def genfstab(self, flags: str = '-pU') -> None:
 		fstab_path = self.target / 'etc' / 'fstab'
 		info(f'Updating {fstab_path}')
-
+		# see GENFSTAB(8) to limit to mountpoint fixes https://github.com/archlinux/archinstall/issues/3599
```
svartkanin: I don't think we need this comment as the PR commit should guide the change to the linked issue

archinstall/lib/installer.py
```diff
@@ -863,6 +866,17 @@ def minimal_installation(
 
 		debug(f'Optional repositories: {optional_repositories}')
 
+		# We need to set vconsole before the first hook in accordance to
+		# https://wiki.archlinux.org/title/Linux_console/Keyboard_configuration#Persistent_configuration
+		# https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio/-/releases/v40
+		# https://www.gnu.org/software/grub/manual/grub/grub.html#Input-terminal
+		locale_cfg = locale_config or LocaleConfiguration.default()
+		self.set_vconsole(locale_cfg)
+		# Also fixes https://github.com/archlinux/archinstall/issues/3160
```
svartkanin: Please raise this fix as a separate PR so it can be properly linked to the issue in GH

svartkanin: Also there's no need to have these issues as comments

archinstall/lib/hardware.py
h8d13: I was thinking it actually needs a is laptop check? 

To safeguard multi gpu setups

We can check this using form factor or battery power. Also no there are many amd iGPU setups too

archinstall/lib/installer.py
h8d13: Yeah the commit hooks were screaming if I didn't add this fir some reason and I'm unfamiliar with how to link to config

archinstall/lib/installer.py
h8d13: Thought it might be useful as hooks might need chnage and wasn't sure about this part :)

archinstall/lib/installer.py
h8d13: Ah was just following how the current code documents issues 

I'll split them and redice the comments

h8d13:
I've split it into 5 clean branches in [repo](https://github.com/h8d13/archinstall-patch)

Master (Pacgenfstab), Hardware, Snapshots, Vconsole & Readme

Removed most of comments kept only ones in vconsole

Hope this is what you needed :+1:

svartkanin (approved):

Issue #3915, opened 2025-11-10
Enabling "testing" repositorys also enables [multilib-testing] regardless if "multilib" is enabled


meatharvester:
### Which ISO version are you using?

archlinux-2025.11.01-x86_64.iso

### The installation log

```json
(installation went fine and I didn't save the log from the usb boot environment)
```

#### Description of the issue

I was Installing using archinstall on a Intel compute stick (STCK1A32WFC) and I installed without multilib (I almost always use testing repos though) to save space as I only have 32gb of emmc storage as it was installing i noticed when it was syncing the mirrors that it was fetching data from [multilib-testing]

After install everything was fine but [multilib-testing] was infact enabled in /etc/pacman.conf

svartkanin:
But wouldn't it make sense to enable that as part of testing?

meatsnails:
> But wouldn't it make sense to enable that as part of testing?

it enables regardless of if multilib is enabled so if you have testing enabled but not multilib it will enable [multilib-testing]

svartkanin:
I'm not sure what people are looking for, if I was to enable "testing" I would sort of expect multilib-testing to be enabled? but maybe I'm wrong with that assumption...

meatsnails:
> I'm not sure what people are looking for, if I was to enable "testing" I would sort of expect multilib-testing to be enabled? but maybe I'm wrong with that assumption...

if i enable testing but not multilib it enables multilib-testing regardless meaning if i don't want multilib im still getting multilib-testing from enabling testing repos

PR #3916, opened 2025-11-10
Added a check to prevent [multilib-testing] from being enabled on non-multilib systems

meatsnails:
A fix to the issue I described in #3915
I added a check to the archinstall/lib/pacman/config.py under the repos_to_enable logic that ensures multilib-testing will only be enabled when multilib and testing are both set

I tested by installing in a VM from the latest iso (archlinux-2025.11.01-x86_64.iso) both with and without multilib + testing and both times (I watched during the install process and checked the installed system's pacman.conf file to ensure the repos we properly set)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ dev = [
     "pre-commit==4.4.0",
     "ruff==0.14.4",
     "pylint==4.0.2",
-    "pytest==8.4.2",
+    "pytest==9.0.0",
 ]
 doc = ["sphinx"]
 


archinstall/lib/pacman/config.py
```diff
@@ -28,7 +28,10 @@ def apply(self) -> None:
 		repos_to_enable = []
 		for repo in self._repositories:
 			if repo == Repository.Testing:
-				repos_to_enable.extend(['core-testing', 'extra-testing', 'multilib-testing'])
+				repos_to_enable.extend(['core-testing', 'extra-testing'])
+			elif repo == Repository.Multilib:
```
correctmost: It looks like this change would prevent `Repository.Multilib.value` from being appended to `repos_to_enable`.

meatsnails: my oversight I think i fixed it and let me test it some more to be sure

meatsnails:
closing this because I made the git log a mess (end my suffering)

PR #3928, opened 2025-11-15
Vconsole.conf KEYMAP= FONT= 
_base mkinitcpio.conf v40 error

h8d13:
We need to set vconsole before the first hook in accordance to
- https://wiki.archlinux.org/title/Linux_console/Keyboard_configuration#Persistent_configuration
- https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio/-/releases/v40
- https://www.gnu.org/software/grub/manual/grub/grub.html#Input-terminal

Also fixes https://github.com/archlinux/archinstall/issues/3160
Since we already validate ASCII only this should not cause any issue or need more changes

- https://github.com/archlinux/archinstall/issues/3155
- https://github.com/archlinux/archinstall/pull/3244


archinstall/lib/installer.py
```diff
@@ -1761,6 +1764,29 @@ def chown(self, owner: str, path: str, options: list[str] = []) -> bool:
...
+		kb_vconsole: str = locale_config.kb_layout
+		# this is the default used in ISO other option for hdpi screens TER16x32
+		# can be checked using
+		# zgrep "CONFIG_FONT" /proc/config.gz
+		# https://wiki.archlinux.org/title/Linux_console#Fonts
+
+		font_vconsole = 'default8x16'
+
+		# Ensure /etc exists
+		vconsole_dir: Path = Path(self.target) / 'etc'
```
svartkanin: `self.target` is already a `Path`

archinstall/lib/installer.py
```diff
@@ -1761,6 +1764,29 @@ def chown(self, owner: str, path: str, options: list[str] = []) -> bool:
...
+		vconsole_dir.mkdir(parents=True, exist_ok=True)
+		vconsole_path: Path = vconsole_dir / 'vconsole.conf'
+
+		# Write both KEYMAP and FONT to vconsole.conf
+		vconsole_content = f'KEYMAP={kb_vconsole}\n'
+		# Corrects another warning
+		vconsole_content += f'FONT={font_vconsole}\n'
+
+		vconsole_path.write_text(vconsole_content)
```
svartkanin, line 1809: Is it certain that this file does not exist at this point? If so then this iwll override any existing content

svartkanin: In addition, shouldn't we use `localectl` to update the file? 
Running a `localectrl set-keymap XXX` will update this file for me, so I don't understand why we would want to do this manually? 
> For convenience, localectl may be used to set the console keymap. It will change the KEYMAP variable in /etc/vconsole.conf and also set the keymap for the current session:

We may have to update the call here https://github.com/svartkanin/archinstall/blob/275259928c34d59cd166055e675938719b2b7276/archinstall/lib/installer.py?plain=1#L1779 to also call `localectl` on target

h8d13: Basically without it the first base hook fails hence why i needed to place it before` _base `

And calling localectl before base well this was more straightforward and uses already existing content of config.

https://www.reddit.com/r/archlinux/comments/1otgoas/issue_with_mkinitcpio/

This also makes encryption layouts possible i guess ? I was using `ckbcomp` with grub manually before to generate layout files for French keyboard from AUR 😩

These need to be latin compatible to my understanding at least with grub which why I had added all the docs comments lmao but since you validate ASCII for passwords and such 

And i think these are different things needs specifically 
`KEYMAP=` and `FONT=`

Maybe we could detect screen size to set font to larger alternative dynamically 

Sorry for bad img from reddit

image

Currently same happens with archinstall but keeps going (non-fatal) with `vconsole.conf` being set:

Screenshot_20251111_134321

h8d13: Also since it's called before it doesn't get overwritten in the subsequent calls



Continuation of #1187 with a *earlier* approach @svartkanin would be great to get some feedback as this could help many users using different kb than US (altho I'm unsure about limitation outside of latin range). Or update instructions: https://github.com/archlinux/archinstall/blob/master/archinstall/locales/README.md

You test on branch: `testing` in https://github.com/h8d13/archinstall-patch

Which also includes my hardware draft for laptops

h8d13:
Hey!

Any ETA/modifications on this @svartkanin I keep seeing reddit logs referecing this error. 

Also again enables latin layouts to work with disk encryption (laptops). Currently you'd get US layout at grub and unable to login.

Perhaps this needs extra logic for non latin layouts

svartkanin:
This issue #3160 and #3155 were issues in the UI, where `curses` was not able to handle non ASCII input characters which was fixed. 

Your PR talks about updating the vconsole.conf configuration. I run an installation from master with the local configuration set to 

```
$ cat user_configuration.json
{
...
    "locale_config": {
        "kb_layout": "de",
        "sys_enc": "UTF-8",
        "sys_lang": "de_DE.UTF-8"
    },
...
}
```

and this is the resulting `vconsole.conf` after the installation: 
```
$ cat /mnt/etc/vconsole.conf
# Written by systemd-localed(8) or systemd-firstboot(1), read by systemd-localed
# and systemd-vconsole-setup(8). Use localectl(1) to update this file.
KEYMAP=de
XKBLAYOUT=de
XKBMODEL=pc105
XKBOPTIONS=terminate:ctrl_alt_bksp
```

The `KEYMAP` as well as the `XKBLAYOUT` have both been set to `de` is that not the expectation?

h8d13:
Hmmm I have to say I'm a bit lost now.

So how did this end up there (from localectl ?) called later?  Also did it pull from `"sys_lang": "de_DE.UTF-8"` (with some parsing) or from the actual asked `kb_layout` (say I leave my sys lang to US but want FR layout). 

My issue is that first `_base` minimal strap would still have screamed an error at you, If I'm not wrong. (vconsole.conf no found) Not a great look when you start an install to see error image generation was not successful. It also let's us set highDPI terminal font if needed (we could add this to detection somehow?)

h8d13:
@Torxed any thought about this? I discussed it a bit with you over on reddit but I couldn't find more definitive answers from anyone really...

And first _base strap still screaming :(

Torxed:
> @Torxed any thought about this? I discussed it a bit with you over on reddit but I couldn't find more definitive answers from anyone really...
> 
> And first _base strap still screaming :(

I'll have a look after this weekend, where in the log does it screan before _base? Just to speed up my looking ^^

h8d13:
Very first _base mkinitcpio call would say:

```
Vconsole.conf not found
...
Warning errors were encountered issues during the build. The image may not be complete. 
```

Non-fatal since its set later on but not a good look if/when you start an install.

svartkanin:
The initial description of the issue is somewhat cryptic and referencing the issues that aren't really related to this PR weren't helpful either. 

This is the output during the installation 

```
...
==> Starting build: '6.17.9-arch1-1'
  -> Running build hook: [base]
  -> Running build hook: [systemd]
  -> Running build hook: [autodetect]
  -> Running build hook: [microcode]
  -> Running build hook: [modconf]
  -> Running build hook: [kms]
  -> Running build hook: [keyboard]
  -> Running build hook: [keymap]
  -> Running build hook: [sd-vconsole]
==> ERROR: file not found: '/etc/vconsole.conf'
  -> Running build hook: [block]
  -> Running build hook: [filesystems]
  -> Running build hook: [fsck]
==> Generating module dependencies
==> Creating zstd-compressed initcpio image: '/boot/initramfs-linux.img'
==> WARNING: errors were encountered during the build. The image may not be complete.
error: command failed to execute correctly
(13/15) Reloading system bus configuration...
  Skipped: Running in chroot.
(14/15) Checking for old perl modules...
(15/15) Updating the info directory file...


Enabling periodic TRIM
Enabling service fstrim.timer
Setting keyboard language to us
Running as unit: run-p23380-i23381.service
Press ^] three times within 1s to disconnect TTY.
Keyboard language for this installation is now set to: us
Running as unit: arch-chroot-23560_mnt.service
Press ^] three times within 1s to disconnect TTY.
...
```

BUT despite this showing the error the actual `vconsole.conf` file still is created correctly with the correct entries. This is most likely due to the call to `localectl` here https://github.com/svartkanin/archinstall/blob/601e03318854a26c454871296b2e70ed85ce2231/archinstall/lib/installer.py?plain=1#L1802, which happens after 
```
...
Keyboard language for this installation is now set to: us
...
```

This means that at the point of bootstrapping https://github.com/svartkanin/archinstall/blob/601e03318854a26c454871296b2e70ed85ce2231/archinstall/lib/installer.py?plain=1#L871 the `vconsole.conf` needs to be setup otherwise the error in the output is shown. 

The questions I have around this are: 
* If a user was to go through a manual installation process, does the wiki indicate that vconsole.conf needs to be setup? 
* Is archinstall doing something during the bootstrap it shouldn't be doing?

h8d13:
Thanks for the follow-up @svartkanin : The main thing I found in wiki: https://wiki.archlinux.org/title/Linux_console/Keyboard_configuration#Persistent_configuration

I believe that `localectl` just isn't available yet (since it's installed in first _base strap) Which is why I thought we can just pull it from config. Notice order is just before the first pacstrap.  

I've also observed when doing host-to-target installs that that would change the layout of host (which isn't great considering I could be installing for another laptop, which is a bit annoying to change again).

Why I mentioned the issues is personal experience (Grub Luks2 password needing to be typed like if my keyboard was US, when it's FR layout).

Hope this clarifies a little, and of course there might be caveats I haven't been able to test like layouts outside latin range.

svartkanin:
> https://wiki.archlinux.org/title/Linux_console/Keyboard_configuration#Persistent_configuration

Yes I get the intention of this PR to do that, but as the wiki states 

_For convenience, localectl may be used to set the console keymap. It will change the KEYMAP variable in /etc/vconsole.conf and also set the keymap for the current session_
_```# localectl set-keymap --no-convert keymap```_

> I believe that localectl just isn't available yet (since it's installed in first _base strap) Which is why I thought we can just pull it from config. Notice order is just before the first pacstrap.

That is probably correct and I don't disagree with you on that and I can see how the PR is intended to solve this. My questions from above however still stand, why is this error happening in the first place and is either 

a) a user also seeing this error simply doing a manual install via the manual wiki steps
b) archinstall is not doing something correctly *before* the bootstrap *or* with the bootstrap

> I've also observed when doing host-to-target installs that that would change the layout of host (which isn't great considering I could be installing for another laptop, which is a bit annoying to change again).

Could you elaborate on this one?

h8d13:
The error happens **because** of order of operations, you're doing it correctly just not in time. So my solution was hacky in that sense.

When wizards changed the hooks in v40.x keymap/consolefont are now considered essential (again makes sense for encryption)(they might also have changed other things? is why I had added the commit info from mkinitcpio [changes](https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio/-/commit/d0a98218d2444b939cf4932265cc70a6a89a8d22) and wiki/grub docs).

It used to be only the warning and a successful image build. Yes this would also happen on manual but since we have the layout info for us it should be easier to fix. 

> Could you elaborate on this one? 

Say I'm installing from my Desktop that uses US layout, but my target needs FR (laptop). Archinstall will over-write my current keymap (host polluted). Because we do not have a safeguard of difference between a from ISO (where the env is temporary doesn't matter, versus from an actual install, one I don't want any changes to happen on).

For this I would add an entry def that checks for `/run/arch-iso` then I believe we can set it for the terminal session only instead of globally. 

This would also **allow us to check for subdeps** needed for archinstall, if running from a live system (since the ISO has many tools that a regular arch installation might be missing and will cause FATAL errors).

The official readme says: _(Usually from a live medium)._ But I think it's a good tool to not have to use ISOs. For that we probably need some work

svartkanin:
Here's a related discussion around the same https://bbs.archlinux.org/viewtopic.php?id=310236. 
I think your approach makes sense to avoid the error, but maybe https://bbs.archlinux.org/viewtopic.php?pid=2273010#p2273010 sounds reasonable too 

_The easy way to avoid this is probably to have an empty vconsole.conf dummy._

If this is caused by a new change I'm still convinced that this is not the behaviour that should happen from `mkinitcpio`. But for the sake of addressing it for now I think it's fine

PR #3929, opened 2025-11-15
Readme Tweak

h8d13:
Simply because I see many logs on reddit using outdated version of archinstall

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,10 +14,8 @@ The installer also doubles as a python library to install Arch Linux and manage 
 * archinstall [documentation](https://archinstall.archlinux.page/)
 
 # Installation & Usage
-
-> In the ISO you are root by default. Use sudo if running from an existing system. 
-
-Make sure to use the latest version...
+> [!TIP]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,8 +14,15 @@ The installer also doubles as a python library to install Arch Linux and manage
 * archinstall [documentation](https://archinstall.archlinux.page/)
 
 # Installation & Usage
+
+> In the ISO you are root by default. Use sudo if running from an existing system. 
+
+Make sure to use the latest version...
```
svartkanin: This can be removed

README.md
```diff
@@ -14,8 +14,15 @@ The installer also doubles as a python library to install Arch Linux and manage
 * archinstall [documentation](https://archinstall.archlinux.page/)
 
 # Installation & Usage
+
+> In the ISO you are root by default. Use sudo if running from an existing system.
```
svartkanin: Not sure why this is a quote, but I'd suggest to use a `note` or `tip` instead 

> [!TIP]
> In the ISO you are root by default. Use sudo if running from an existing system

PR #3932, opened 2025-11-16
Add dialog to install EFI bootloader to removable location

Mintsuki:
This is just for GRUB and Limine for now.

### DIFF
--- a/archinstall/lib/bootloader/__init__.py
+++ b/archinstall/lib/bootloader/__init__.py
@@ -1,3 +1,0 @@
-from .bootloader_menu import BootloaderMenu
-
-__all__ = ['BootloaderMenu']
--- a/archinstall/lib/global_menu.py
+++ b/archinstall/lib/global_menu.py
@@ -12,7 +12,7 @@ from archinstall.tui.menu_item import MenuItem, MenuItemGroup
 from .applications.application_menu import ApplicationMenu
 from .args import ArchConfig
 from .authentication.authentication_menu import AuthenticationMenu
-from .bootloader import BootloaderMenu
+from .bootloader.bootloader_menu import BootloaderMenu
 from .configuration import save_config
 from .hardware import SysInfo
 from .interactions.general_conf import (


svartkanin:
i think this should go under a new Bootloader sub-menu instead of a global namespace setting. 
The current bootloader selection should go in there together with this setting

archinstall/lib/interactions/system_conf.py
```diff
@@ -90,6 +90,42 @@ def ask_for_bootloader(preset: Bootloader | None) -> Bootloader | None:
 			raise ValueError('Unhandled result type')
 
 
+def ask_for_bootloader_removable(preset: bool = False) -> bool:
+	prompt = (
+		tr('Would you like to install the bootloader to the default removable media search location?\n')
```
svartkanin: do put any characters such as \n \r or bullet points into the translation as that will make it difficult to reuse them

h8d13: https://wiki.archlinux.org/title/EFI_system_partition#Typical_mount_points

Isn't referring to the second section here ? I think that would be a lot more work to integrate end-to-end kinda works in my fork but need a lot of specific code for systemd-boot

Mintsuki: > https://wiki.archlinux.org/title/EFI_system_partition#Typical_mount_points
> 
> Isn't referring to the second section here ? I think that would be a lot more work to integrate end-to-end kinda works in my fork but need a lot of specific code for systemd-boot

@h8d13 it is not related to that. This is about the location of the bootloader within the ESP itself, not about where the ESP is mounted.

Mintsuki: This should be resolved.

Mintsuki:
The PR should be ready for review now.

Please, beware that I have used LLMs for most of these changes due to my unfamiliarity with the codebase, and while I have reviewed and tested the code, I would strongly recommend someone familiar with the codebase to thoroughly review the changes.

Mintsuki:
@svartkanin alright, sorry, I added a bunch more commits, but I should be done now, at least for what this PR is concerned.

archinstall/lib/bootloader/bootloader_menu.py
```diff
@@ -0,0 +1,236 @@
...
+			MenuItem(
+				text=tr('Bootloader'),
+				action=self._select_bootloader,
+				value=self._bootloader_conf.bootloader,
+				preview_action=self._prev_bootloader,
+				mandatory=True,
+				key='bootloader',
+			),
+			MenuItem(
+				text=tr('Unified kernel images'),
```
svartkanin, line 55: why was this moved here?

Mintsuki: UKIs are inherently a bootloader thing, are they not? they are basically a fancy version of EFISTUB that gets chainloaded, it logically made sense to have it in this menu, and it made sense from a code perspective.

svartkanin: Right, maybe it does make sense to have it here, I got confused by the name

Mintsuki: I'll mark it as resolved, then.

Mintsuki:
Anything else I should change before approval?

archinstall/lib/bootloader/bootloader_menu.py
```diff
@@ -0,0 +1,236 @@
...
+			if not SysInfo.has_uefi() or not bootloader.has_removable_support():
+				removable_item.enabled = False
+				removable_item.value = False
+				self._bootloader_conf.removable = False
+			else:
+				removable_item.enabled = True
+
+		return bootloader
+
+	def _select_uki(self, preset: bool) -> bool:
```
svartkanin, line 121: there's no need to wrap these around the ask* functions, just call them directly

archinstall/lib/bootloader/bootloader_menu.py
```diff
@@ -0,0 +1,236 @@
...
+
+	@override
+	def run(
+		self,
+		additional_title: str | None = None,
+	) -> BootloaderConfiguration:
+		super().run(additional_title=additional_title)
+		return self._bootloader_conf
+
+	def _select_bootloader(self, preset: Bootloader | None) -> Bootloader | None:
```
svartkanin, line 97: can we move this outside the class as a standalone function similar to the ask* functions. Reason being they can be reused if someone wants to build their own menu and just use these selections

archinstall/lib/global_menu.py
```diff
@@ -402,9 +387,10 @@ def _prev_kernel(self, item: MenuItem) -> str | None:
 			return f'{tr("Kernel")}: {kernel}'
 		return None
 
-	def _prev_bootloader(self, item: MenuItem) -> str | None:
-		if item.value is not None:
-			return f'{tr("Bootloader")}: {item.value.value}'
+	def _prev_bootloader_config(self, item: MenuItem) -> str | None:
```
svartkanin, line 390: This should provide a bit of a better output 
image
similar to other submenus with 
```
Bootloader: ...
UKI: ...
Removable: ...
```

archinstall/lib/bootloader/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .bootloader_menu import BootloaderMenu
```
svartkanin: lets remove thsi import, we had multiple issues in the past with circular dependencies as loading the booloader/ module starts cascading the import downstream

svartkanin:
Sorry I was too busy, I tested the code and it seems to work, great job in making the config backwards compatible. 
I left some small comments if you can address those then we can merge the PR

archinstall/lib/bootloader/bootloader_menu.py
Mintsuki: but that function depends on `self` because its purpose is to ensure the submenu's UKI/removable options are dynamically updated depending on the chosen bootloader.

archinstall/lib/bootloader/__init__.py
Mintsuki: what is the suggested change?

svartkanin: just remove the content of the __init__.py and leave it empty

archinstall/lib/bootloader/bootloader_menu.py
svartkanin: Right, I missed that there's already `ask_for_bootloader` outside then all good

archinstall/lib/bootloader/__init__.py
Mintsuki: I did try that and CI won't pass.

Mintsuki: I figured it out, my bad. Issue is resolved now!

PR #3947, opened 2025-11-25
Default `partuuid` parameter of `_get_kernel_params_partition()` to false

Mintsuki:
There isn't a reason I could think of for preferring PARTUUIDs to UUIDs, since as far as I know that should work fine with all supported filesystems (and GRUB already forces this to false). This is especially important when using MBR instead of GPT, as that doesn't have proper UUIDs, and it will instead use the kernel's own conversion that is based on the MBR ID and more prone to collisions.

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -1115,7 +1115,7 @@ class Installer:
 		self,
 		root: PartitionModification | LvmVolume,
 		id_root: bool = True,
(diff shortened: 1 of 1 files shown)


svartkanin:
Couold you provide a description of why this is needed?

Mintsuki:
> Couold you provide a description of why this is needed?

I have updated the original description.

Issue #3945, opened 2025-11-23
Fallback entry not working on systemd-boot


Mattyan89:
ISO: 2025-11-01 / Archinstall 3.0.13
When using systemd-boot (not sure about other boot loaders, I haven't tried), a fallback menu entry is created, but it is not functional because of the recent update to mkinitcpio, which disables fallback image creation for new installs.

svartkanin:
Arch wiki https://wiki.archlinux.org/title/Mkinitcpio#Fallback_initramfs_generation

h8d13:
Is there a world where we can make this configurable ?

PR #3949, opened 2025-11-25
Do not create BLS and Limine entries for fallback initramfs

Mintsuki:
Fallback initramfs seem to no longer be built by default.

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -1160,19 +1160,18 @@ class Installer:
 			f"""\
 			# Created by: archinstall
 			# Created on: {self.init_time}
-			title   Arch Linux ({{kernel}}{{variant}})
+			title   Arch Linux ({{kernel}})
 			linux   /vmlinuz-{{kernel}}
-			initrd  /initramfs-{{kernel}}{{variant}}.img
+			initrd  /initramfs-{{kernel}}.img
 			options {' '.join(self._get_kernel_params(root))}
 			""",
 		)
 
 		for kernel in self.kernels:
-			for variant in ('',):
-				# Setup the loader entry
-				name = entry_name.format(kernel=kernel, variant=variant)
-				entry_conf = entries_dir / name
-				entry_conf.write_text(entry_template.format(kernel=kernel, variant=variant))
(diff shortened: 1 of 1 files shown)


archinstall/lib/installer.py
```diff
@@ -1168,7 +1168,7 @@ def _create_bls_entries(
 		)
 
 		for kernel in self.kernels:
-			for variant in ('', '-fallback'):
+			for variant in '':
```
svartkanin: This doesn't make any sense, it's not going to execute this for loop

Mintsuki: I apologise, I didn't actually test this one as I went to bed shortly thereafter; my first approach was to do `for variant in ('')`, but that triggered a ruff style check thingy, and this was the suggested solution, so I trusted it.

What do you recommend? Doing `for variant in ('')` and suppressing the style warning, or something else?

Mintsuki: Okay, I think the issue should be resolved now: I did not know a trailing comma was mandatory for single element tuples. Now I used `('',)`.

Mintsuki:
This PR is ready for review.

svartkanin:
Would you have any reference to this claim, the wiki still lists fallback as an entry example https://wiki.archlinux.org/title/Systemd-boot#Adding_loaders

svartkanin:
This is relevant for issue #3945 

This is due to mkinitcpio change and a workaround is mentioned on the wiki here  https://wiki.archlinux.org/title/Mkinitcpio#Fallback_initramfs_generation

> Change PRESETS=('default') line to PRESETS=('default' 'fallback') in the respective .preset files in /etc/mkinitcpio.d/.

Mintsuki:
Yeah, but it is disabled by default, so I reckon we should probably not create the entry? That is unless we decide to reenable it.

svartkanin:
Yeah that makes sense, going against defaults isn't wise...

Can you remove the loop completely in that case as it's not needed. You can remove the `variant` from the template as that would get replaced with '' anyways

Mintsuki:
Should be good to go :)

svartkanin (approved):

PR #3950, opened 2025-11-26
Fix GRUB fallback-from-removable logic introduced in c095eb56d8

Mintsuki:
Commit c095eb56d878b51a80d49cdef3c82950bcfcdecd was supposed to introduce logic
such that if the `grub-install` command failed with a `--removable` flag, then
another attempt would be made with such flag removed.

This was broken because the `--removable` flag was kept in both cases (likely a
copy-paste mistake). This has been an issue since, in all future iterations of
the code.

What this commit does is fix this logic, but also invert the cases tested:
first test without `--removable`, then add it should that case fail, as this is
the most sensible thing to do.

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -1321,11 +1321,15 @@ class Installer:
 
 			try:
 				SysCommand(command, peek_output=True)
-			except SysCallError:
-				try:
-					SysCommand(command, peek_output=True)
-				except SysCallError as err:
-					raise DiskError(f'Could not install GRUB to {self.target}{efi_partition.mountpoint}: {err}')
+			except SysCallError as err:
+				if not bootloader_removable:
+					command.append('--removable')
+					try:
+						SysCommand(command, peek_output=True)
+					except SysCallError:
+						pass
+
+				raise DiskError(f'Could not install GRUB to {self.target}{efi_partition.mountpoint}: {err}') from err
 		else:
 			info(f'GRUB boot partition: {boot_partition.dev_path}')
 


svartkanin:
This seems like a hack and a "lets try until it works" thing. I'd rather this to be an actual logic rather than abusive errors as if-else cases

Mintsuki:
Well, I wouldn't necessarily call this a hack per se, and in any case the original logic was broken regardless.

I would also be okay with removing this fallback altogether and having users manually select the removable option from the menu (the one I added in a previous PR).

h8d13:
Because --removable is always more reliable in both cases at least from personal experience

Mintsuki:
> Because --removable is always more reliable in both cases at least from personal experience

Yes, but that is not a good argument for unconditionally using the `--removable` flag. If that was the intended usage of `grub-install`, it'd be the default.

The UEFI specification intends the removable path (on x86-64: `/EFI/BOOT/BOOTX64.EFI`) to only be used for, well, removable media, where the firmware cannot know ahead of time what the name of the EFI application to load is.

For non-removable media, the intended way is to let the firmware know what bootloaders/EFI applications are available with NVRAM entries (as managed by `efibootmgr` on Linux, for example).

Now, it is true that on *some* motherboards `efibootmgr` may fail, or NVRAM entries may be ignored, and on these motherboards, yes, one should use the removable location, short of other workarounds being available, but this shouldn't be what the installer defaults to. It should be a fallback.

The way Debian's installer sorts this out is to simply do both, but let the user choose if they also want to use the removable location, with a short explanation of why they may want to do that.

My opinion is this: the code as-is is broken, and it never made sense. We should either remove that fallback path entirely and only decide based on whether the user says they want to use the removable path or not, or we should do as this PR does currently, where the non-removable code path is tried first.

h8d13:
From the Grub Wiki Page:

`--removable`
```
If you are installing GRUB on a [Mac](https://wiki.archlinux.org/title/Mac), you will have to use this option. 
Some desktop motherboards will only look for an EFI executable in this location, making this option mandatory, in particular with MSI boards.
If you execute a UEFI update, this update might delete the existing UEFI boot entries. 
Therefore, it is a potential fallback strategy to have the "removable" boot entry enabled.
```

It is a better default I think, generally speaking.

Mintsuki:
It is not a good default, it's a fallback. At the very least the user should be prompted about it like Debian does. Recently merged PR #3932 adds a screen that prompts the user whether they want to install to the removable location or not.

svartkanin:
For the sake of fixing this for now I'll merge the PR. The thing that I don't particularly like about it is that if a user does not set the removable setting now archinstall will simply go ahead and add it anyways on the failure. 
It'd be good if there could be a check before running an installation if the selected setup is a viable or now and may lead to the error beforehand.

svartkanin (approved):

archinstall/lib/installer.py
```diff
@@ -1321,11 +1321,15 @@ def _add_grub_bootloader(
...
-					raise DiskError(f'Could not install GRUB to {self.target}{efi_partition.mountpoint}: {err}')
+			except SysCallError as err:
+				if not bootloader_removable:
+					command.append('--removable')
+					try:
+						SysCommand(command, peek_output=True)
+					except SysCallError:
+						pass
+
+				raise DiskError(f'Could not install GRUB to {self.target}{efi_partition.mountpoint}: {err}') from err
```
correctmost, line 1332: It looks like the fallback branch with `--removable` will always fall through and raise an exception here.  Am I overlooking something?

Issue #3948, opened 2025-11-25
Better search feature in additional package


Nurysso:
### describe the request

The `ask_additional_packages_to_install` function currently sorts searched results alphabetically, which can sometimes lead to less relevant packages appearing before the most obvious match.

For example, searching for `firefox` shows `browserpass-firefox` before the browser itself, and searching for `git` lists unrelated packages like `auth-tarball-from-git` higher up.

The ordering rules of the search functionality could be improved to prefer exact matches and closer package names in order to more quickly and easily provide the user with the packages they are looking for.

I would like to improve the search algorithm and ranking of the results regarding this issue.

Additionally, I'd appreciate any specific guidelines I should follow regarding code compatibility and performance, particularly given that archinstall runs in a wide variety of hardware and often within an ISO environment.

h8d13:
I say comma seperated list (with check) :D

Nurysso:
well if you mean buy typing git,firefox this wont be able to find anything

<img width="950" height="527" alt="Image" src="https://github.com/user-attachments/assets/b1a604b9-2e95-4b5a-ba60-e4657b37a5c2" />

h8d13:
I meant just making the selection a `git, firefox, micro` > Check

Nurysso:
Correct me if I’m wrong, but didnt older versions of archinstall allow typing additional packages separated by spaces? The newer versions introduced a searchable list to reduce typos(i guess). While it’s slower than before, I think it’s a better approach overall. Shouldn’t improvements aim to move us forward rather than backward?

PR #3967, opened 2025-12-02
Menu now filters items and sorts using priority, improving UX.

okayGravity:
- This fixes issue: #3948 
_Note: While packages are the common use case for search, this change also effects searching all menu items._

## PR Description:
Previously, using search ('/') captured user input and filtered menu items using a lambda function with .contains(). While this removed non-matching items, it retained the original alphabetical sort. This resulted in _less relevant_ matches appearing first.

For example, a user searching 'git' in packages will be resulted 'auth-tarball-from-git' because it appears first alphabetically than 'git'. (see below)

_example of previous results using filter "git"_
521531278-1e433869-6648-4cb9-8a03-6d6f18e33cce

## Fix:
This change adds the _items_score() function as a helper. This function scores, and gives priority to items that start with the user given filter. In menu_item.py, the items() function now uses sorted() with this scoring key (replacing the previous list() conversion) to return results in the desired relevance order.

_example of new results using filter "git"_
521532631-6493c17e-1a1f-4600-8e59-f2abfc90bd27

## Tests and Checks
- [x] I have tested the code locally!<br>


okayGravity:
I noticed there was some interest in this issue from @Nurysso last week, but since it hadn't been assigned or updated, I went ahead and pushed a fix. Happy to make changes if needed!

Nurysso:
Wait we can just fix it with just using                 l_items = sorted(items, key=self._items_score)

I wrote like 30 some lines to do the same 😭. Anyways thanks for adding this, I was going to test and send pr of my code tomorrow but yours looks better mine

Torxed:
I appreciate you taking the time to improve based on others suggestions! :D Looks good to me.

okayGravity:
@Nurysso I didn't mean to snipe you! I also thought of the issue around the same time as well, but didn't notice you asked to work on it. Maybe we can collaborate on some other issues.

Nurysso:
That's no issue your was a simpler approach, and would love to work on something with you in future

okayGravity:
> I appreciate you taking the time to improve based on others suggestions! :D Looks good to me.

Of course! I'll be around, I'm still pretty fresh but I'm happy to be contributing!

h8d13:
You can also use python built in Sequence matcher for better fuzzy search ;) 

```
from difflib import SequenceMatcher
##
similarity = SequenceMatcher(None, search_lower, name_lower).ratio()
## then sort using a threshold
```

okayGravity:
> You can also use python built in Sequence matcher for better fuzzy search ;) 
> 
> ```
> from difflib import SequenceMatcher
> ##
> similarity = SequenceMatcher(None, search_lower, name_lower).ratio()
> ## then sort using a threshold
> ```
> 

I did know about this but didn't try it to see how it felt, I'm working on an issue right now with the way focused items in the items list works when filtering, I'll see if I can improve the search with this after I finish with that issue. Thanks for laying this out for me, I saw that a common threshold was about 60%, would you say this is good? I'm unsure exactly how it works as I haven't looked at it too much.

okayGravity:
I added a change to the way items are focused and view is changed. Originally, the user's focused item would stay focused even when adding, removing or changing filter. This also caused an issue where when moving through items in menu, sometimes the view would not update when moving _up_ the list, as seen below.

search_old

This was caused by a check with a previous solution that reloaded the focused item when filter changes were made. 

This fix sets the focus item to be focus_first(), which always focuses the first item in the list when using a filter which now better directs a user to their desired selection since filter logic will always display the most likely result first. 

The issue involving the menu failing to update when moving up a list was fixed by removing a check when displaying menu items in item_group.has_filter() which would set the view to the first result in the list. This was removed and replaced with a check if focus_item == 0, since our searched item would always be the first result, and now does not conflict when the user removes the filter and wants to navigate the list. 

search_fix

h8d13:
I used something like .35 and it worked pretty great

mintsuki (approved):

svartkanin (approved):

PR #3973, opened 2025-12-07
Explicitly allow selecting any additional repository

### LINKED ISSUE #3915: Enabling "testing" repositorys also enables [multilib-testing] regardless if "multilib" is enabled

svartkanin:
Fixes #3915

### DIFF
--- a/archinstall/lib/mirrors.py
+++ b/archinstall/lib/mirrors.py
@@ -354,9 +354,14 @@ def select_optional_repositories(preset: list[Repository]) -> list[Repository]:
 	:rtype: Repository
 	"""
 
-	repositories = [Repository.Multilib, Repository.Testing]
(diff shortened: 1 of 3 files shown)


h8d13:
Bump :+1:

Issue #3599, opened 2025-06-15
/etc/fstab boot-hang


h8d13:
Hi. I was testing on dell latitude I got from workplace. Newer ssd, entreprise type.

I wanted to test something other than Ext4 (which I never had any issues with).
So tried both BTRFS and XFS. But both were hanging on first boot (for about a minute or so) due to zram0 line in /etc/fstab. Then say skipped UUID or failed UUID. And I can still get in to correct: When commenting out the zram0 line. Boot works fast again. 

Was just wondering if I made a mistake or if it's something to do with installer ? Not entirely sure how it all works just wanted to try give some feedback. 

zram0 253:0 0 4G 0 disk [SWAP]
nvme0n1 259:0 0 238.5G 0 disk
├─nvme0n1p1 259:1 0 1G 0 part /boot
└─nvme0n1p2 259:2 0 237.5G 0 part /

svartkanin:
Can you provide the installation log

h8d13:
``` 
   "archinstall-language": "English",
    "audio_config": {
        "audio": "pipewire"
    },
    "bootloader": "Grub",
    "custom_commands": [],
    "disk_config": {
        "btrfs_options": {
            "snapshot_config": null
        },
        "config_type": "default_layout",
        "device_modifications": [
            {
                "device": "/dev/nvme0n1",
                "partitions": [
                    {
                        "btrfs": [],
                        "dev_path": null,
                        "flags": [
                            "boot",
                            "esp"
                        ],
                        "fs_type": "fat32",
                        "mount_options": [],
                        "mountpoint": "/boot",
                        "obj_id": "bbcfc292-5ce8-4b4f-a31a-60ca314ea461",
                        "size": {
                            "sector_size": {
                                "unit": "B",
                                "value": 512
                            },
                            "unit": "GiB",
                            "value": 1
                        },
                        "start": {
                            "sector_size": {
                                "unit": "B",
                                "value": 512
                            },
                            "unit": "MiB",
                            "value": 1
                        },
                        "status": "create",
                        "type": "primary"
                    },
                    {
                        "btrfs": [],
                        "dev_path": null,
                        "flags": [],
                        "fs_type": "xfs",
                        "mount_options": [],
                        "mountpoint": "/",
                        "obj_id": "80f89869-58a4-4a29-9cd5-6dae2540fe12",
                        "size": {
                            "sector_size": {
                                "unit": "B",
                                "value": 512
                            },
                            "unit": "B",
[...]

PR #3978, opened 2025-12-08
H2T (Host-to-target) installs and prevent host pollution

h8d13:
This is a follow up to my last [PR#3928](https://github.com/archlinux/archinstall/pull/3928) and original [ISSUE#3599](https://github.com/archlinux/archinstall/issues/3599) 

What we achieve with such a patch:
- Prevent **polluting** the host by having a hacky `running_from_host` definition 
    - Skip `set_kb_layout`  (will be set on target only) via `installer.set_keyboard_language()` and `set_vconsole()`
- This is added to main `__init__.py` to print: EXAMPLE:

```
Fetching Arch Linux package database...
Checking version...
Running from host (H2T Mode)...
```
Or `Running from ISO (Live Mode)`

And will be present in logs to have a trace too (useful if outdated system deps or straight up missing deps). 

**End result:** Makes it so that if you are installing say from a US layout computer and target uses a different layout (ex: laptop), we _do not modify the host system_ and can be built-upon for other host-checks if needed (say for example: extra deps checks or other host isolation related needs).

- Readme clarification Host to target installs without an ISO! 
- Pyright type annotations fixes (was giving me errors but can revert it if needed)

### DIFF
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -113,9 +113,9 @@ def main() -> int:
 
 	if running_from_host():
 		# log which mode we are using
-		info('Running from host (H2T Mode)...')
+		debug('Running from Host (H2T Mode)...')
 	else:
-		info('Running from ISO (Live Mode)...')
+		debug('Running from ISO (Live Mode)...')
 
 	script = arch_config_handler.get_script()
 


svartkanin:
I understand the intention of the PR but I'm not a big fan of deviating behavior depending on where archinstall is run as it will make things more complex. 

Having said that I do agree that modifying the host system is not a good idea and I hope to find an alternative option that behaves the same as well, from a quick look:

* `localectl` seems good for the console as permanent changes are short lived there anyways
* `setxkbmap` only seems to work on X11 sessions

So I guess the options would be

1) Don't modify anything anywhere and the user has to set it up manually outside of archinstall
2) Find a solution that sets it temporary but that requires host inspection and different tools for different systems
3) Accept the different behavior BUT display the current keyboard layout in all inputs so there less confusion 

@Torxed any thoughts?

h8d13:
Thanks for the feedback @svartkanin 

It might seem a bit too hacky but makes sense with how it works (usually from live medium but _why not from an existing install_). 

We just need to be able to differentiate maybe you have a better solution or better spot for it (I thought early is better) and make it abstract bool so that it can be re-used. 

It doesn't really affect much of the rest is more just of guardrails for H2T.  The key here: is just the **return early** here and makes **a lot of sense.**  And still sets for target !

(Please feel free to modify the rest!) ,because again it would just make it so that we do not touch the host's settings, if in live-install scenario vs being on an live-medium (where we don't care).  

If anything I don't think it adds much complexity but clearly states what archinstall should behave like in different scenarios. I've had run manually countless times before I found this hack and I think is one of the cooler parts of archinstall is not needing an ISO at all and being able to use your current install to create others :D 

About your other solutions I don't think it addresses the root cause: Unless we use `loadkeys` which is only for the current session/tty.  But again still wouldn't make sense in ISO vs on running system

archinstall/lib/locale/utils.py
```diff
@@ -79,6 +79,11 @@ def get_kb_layout() -> str:
 
 
 def set_kb_layout(locale: str) -> bool:
+	if running_from_host():
+		# Skip when running from host - no need to change host keymap
+		# The target installation keymap is set via installer.set_keyboard_language()
```
h8d13, line 84: As an alternative to return
```
		try:
			SysCommand(f'loadkeys {locale}')
		    return True
		except SysCallError as err:
			raise ServiceException(f"Unable to set locale '{locale}' for console: {err}")
```

svartkanin:
So the assumption after this will be 
* If on ISO a user has probably not setup the desired keymap yet and archinstall will do it automatically for them whenever a new layout is choosen
* If on host, a user will use their preferred keymap already and most likely the target will have the same (in which case there's no need to set it) 

To avoid confusion I'd still like to have the current keymap displayed in any input (can be done in a future PR as I have some major migration to `textual` upcoming)

archinstall/__init__.py
```diff
@@ -110,6 +111,12 @@ def main() -> int:
 				info(new_version)
 				time.sleep(3)
 
+	if running_from_host():
+		# log which mode we are using
+		info('Running from host (H2T Mode)...')
```
svartkanin: Can we have this as `debug` only, no need to display it to the user

archinstall/lib/locale/utils.py
svartkanin: Lets not use `loadkeys` as it will not work on wayland which will have taken over most systems in the near future

h8d13:
> * If on host, a user will use their preferred keymap already and most likely the target will have the same (in which case there's no need to set it) 

This is where the issue is, if on host and select a keymap (different from host) I've polluted my host but only wanted this in target (and assumptions that they are the same is just wrong). 

I've had this many times installing for laptops (forced into a different keymap) from my main setup (us).

I think the loadkeys (set in session only) + `running from host` def seems like a good solution (see comment on code, could be re-used for other things).


Again thanks for follow up ! And hopefully this makes more sense now. Sorry if I was unclear, it's a weird case I get a lot with French/Belgian layouts.

svartkanin:
> I think the loadkeys (set in session only) + running from host def seems like a good solution (see comment on code, could be re-used for other things).

I'm not too keen on using a tool that is not supported on Wayland as most system will be on Wayland by now. So this will add another layer of behaviour that needs to be understand as just looking at the code path is not enough. 

If you can change the `info` to `debug` I think the PR is fine

svartkanin (approved):

Issue #3986, opened 2025-12-12
Error Installing ly Greeter


pneumus:
### Which ISO version are you using?

2025-DEC-01

### The installation log

```json
Unfortunately I can not provide the log, but it is really easy to reproduce this, based on the description.
```

#### Can not install ly greeter with KDE Plasma

When trying to install Arch Linux with KDE Plasma on an Alienware X14 Laptop, 
and the ly Greeter ( instead of the default SDDM ): 
There is an error in the Python Script.

<img width="1238" height="676" alt="Image" src="https://github.com/user-attachments/assets/af104c2a-28c8-49c2-bbb0-de2aaf9e8285" />

imolinuevo:
I have the same error when trying to install the ly greeter with Hyprland, for the moment I`m using GDM that works fine.

pneumus:
@imolinuevo I am using SDDM. Also works fine.
But I would prefer ly and I can not use it because the archlinux python script is obviously buggy.

h8d13:
Ly devs changed something that broke any install seen both on reddit, bbs and their git issues.  `ly.service to ly@ttyX.service`

Not archinstall issue :) please follow their git sources instead.

pneumus:
@h8d13 the URL does not work.

h8d13:
Sorry they probably removed it ? See pull request #4006

But package has to also be updated I believe in repos still
https://archlinux.org/packages/extra/x86_64/ly/

Cryptic7505:
Same issue in my case 
logs: https://pastebin.com/d7vzzEeS

h8d13:
Thanks for reporting @Cryptic7505 unfortunately we need to wait package to be rebuilt in repos. Issue is fixed in archinstall see pr above

svartkanin:
As all fixes are done from our end I'll close this

PR #3988, opened 2025-12-13
Fix installing ly

Oleg-Chan:
- This fix issue: #3986

Since 1.3.0 ly "Added support for running Ly on multiple TTYs on systemd" therefore changing the system service and requiring to specify the tty + disabling getty service at that tty
## PR Description:

Changed the name of the service. repurposed enable_service function to add disable_service function

## Tests and Checks
Tested on the vm, worked fine.

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -675,11 +675,11 @@ class Installer:
 				if hasattr(plugin, 'on_service'):
 					plugin.on_service(service)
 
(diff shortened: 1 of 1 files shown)


svartkanin (approved):

svartkanin:
There are `mypy` failures can you please address those

svartkanin:
Thanks for the PR, as it wasn't updated to resolve the actions #4006 was merged instead to address it.

PR #3997, opened 2025-12-15
Migrate UI to textual

svartkanin:
This PR migrates all menus to the `textual` framework introduced with https://github.com/archlinux/archinstall/pull/3879. 

The current implementation is creating a new textual app and destroying it with every menu/submenu which is not how it is supposed to be. 
The final version will use a single app instance and switch between screens smoothly but to keep things (somewhat) manageable, that change will be done in a follow-up PR in which the code will be moved to an async model (see the wifi screen that is already doing that https://github.com/svartkanin/archinstall/blob/fdb002fa31471d86e1d5e99bc6ec4c21ba1a3b15/archinstall/lib/network/wifi_handler.py?plain=1#L1) 

I have tested everything somewhat but not extensively yet. I have opened the PR regardles for now in case someone wants to have a go at it and provide some feedback. 

[X] Test everything thoroughly

### DIFF
--- a/archinstall/tui/ui/menu_item.py
+++ b/archinstall/tui/ui/menu_item.py
@@ -25,8 +25,8 @@ class MenuItem:
 
 	_id: str = ''
 
-	_yes: ClassVar[MenuItem | None] = None
-	_no: ClassVar[MenuItem | None] = None
+	_yes: ClassVar[Self | None] = None
+	_no: ClassVar[Self | None] = None
 
 	def __post_init__(self) -> None:
 		if self.key is not None:
@@ -46,14 +46,14 @@ class MenuItem:
 		return self.value
 
 	@classmethod
-	def yes(cls, action: Callable[[Any], Any] | None = None) -> 'MenuItem':
+	def yes(cls, action: Callable[[Any], Any] | None = None) -> Self:
 		if cls._yes is None:
 			cls._yes = cls(tr('Yes'), value=True, key='yes', action=action)
 
 		return cls._yes
 
 	@classmethod
-	def no(cls, action: Callable[[Any], Any] | None = None) -> 'MenuItem':
+	def no(cls, action: Callable[[Any], Any] | None = None) -> Self:
 		if cls._no is None:
 			cls._no = cls(tr('No'), value=False, key='no', action=action)
 
@@ -138,21 +138,22 @@ class MenuItemGroup:
 	def get_enabled_items(self) -> list[MenuItem]:
 		return [it for it in self.items if self.is_enabled(it)]
 
-	@staticmethod
-	def yes_no() -> 'MenuItemGroup':
-		return MenuItemGroup(
+	@classmethod
+	def yes_no(cls) -> Self:
+		return cls(
 			[MenuItem.yes(), MenuItem.no()],
 			sort_items=True,
 		)
 
-	@staticmethod
+	@classmethod
 	def from_enum(
+		cls,
 		enum_cls: type[Enum],
 		sort_items: bool = False,
 		preset: Enum | None = None,
-	) -> 'MenuItemGroup':
+	) -> Self:
 		items = [MenuItem(elem.value, value=elem) for elem in enum_cls]
-		group = MenuItemGroup(items, sort_items=sort_items)
+		group = cls(items, sort_items=sort_items)
 
 		if preset is not None:
 			group.set_selected_by_value(preset)


h8d13:
Looks cool @svartkanin :D 

Tested a little bit, I came across was f1 help screen crashed and user setup screen was missing `should user be sudo` text and preview for this has 2 screens

The rest seems like it's already fully functional (considering what you said about the flicker), like how you can use cursor, very bamby friendly :computer_mouse: 

Would be cool to add globally visible hints for `esc go back` `ctrl+c clear a field` `enter : confirm` `space/tab : select multiple` `/ : search` 

Then more specific go in help screen ?

svartkanin:
Thanks @h8d13 for testing! 

* Fixed the select sudo header
* Can you share a screenshot or error that you get for the f1 help screen crash? I cannot reproduce that
* Not sure what you mean with `preview for this has 2 screens`

h8d13:
My bad was missing `pacman -S python-linkify-it-py` for f1 crash. 

For my other point sorry was not clear I meant when you are in the submenu 
image

It shows you the overview again but the Esc key doesn't register to go back to main menu, same when you are here: (Need to use arrows menu 2x instead of pressing Esc to go back like in the rest, applies to other submenus too, would be nice to have kb handling)

image

Let me know if you have anything specific, you need tested :) Or need help with something !

svartkanin:
Most submenus (The ones with a `Back` option) don't have a support for `Esc` as that's reserved for leaf selection menus. That behavior exists already

h8d13:
Yeah I thought it would be a good thing if it also worked in these menus reduce keys needed to get around

Had another question, I saw that runners were going to be more restrictive (and paid) for github and always thought the ISO one takes a while (seems unnecessary for smaller changes) . 

Is there anyway we can disable it unless on major releases/needed ?

Torxed:
I like this, and I think there's only one thing that we might need to address first *(and then I'll leave some general things i noticed while trying that we don't need to fix now)*

## Issue

- The espeekup doesn't work, when you select a menu entry it doesn't read the entry, it just repeats the titlebar "archinstall"

  https://github.com/user-attachments/assets/9dac8b23-1398-4630-972e-05d890280ea0



## For later

- Selecting "Use best effort partitioning" and then highlight an entry and pressing enter, doesn't select+continue, it does nothing. You have to use space to select an entry before hitting enter.
- I can't go into partitioning and de-select a partition if I want to un-do my chosen setup (lets say I changed my mind, and pre-mounted things, and now want to un-select the partition i selected)

There's also a small anomaly that I don't think is related to this PR at all, but I'm noting it down because it kept happening:

- When selecting a mirror region, it gets stuck on loading the region list (with the new cool loader animation). And it feels like a timeout on the retrieval of the list and the menu system doesn't detect the timeout/error.

Torxed:
We might need to change the header color combo too.
I suspect the contrast for color blind people might be frustrating on the title bar:

image

This might be nitpicking tho :)

archinstall/tui/ui/components.py
```diff
@@ -158,30 +557,57 @@ def __init__(
...
-		self.update_selection()
+		self._update_selection()
+
+	def action_focus_right(self) -> None:
+		if self._is_btn_focus():
+			self._group.focus_next()
+			self._update_selection()
 
-	def update_selection(self) -> None:
+	def action_focus_left(self) -> None:
```
Torxed, line 657: I wonder if here, we can somehow use [.notify()](https://textual.textualize.io/api/app/#textual.app.App.notify) and perhaps hide the `Toast` with CSS? As espeakup appears to read those texts.

If you don't have a more clever solution to making espeakup understand the menu changes ofc.

svartkanin:
> The espeekup doesn't work, when you select a menu entry it doesn't read the entry, it just repeats the titlebar "archinstall"

I'll have a look at this one. 

> Selecting "Use best effort partitioning" and then highlight an entry and pressing enter, doesn't select+continue, it does nothing. You have to use space to select an entry before hitting enter.

this was a deliberate decision as the textual examples themselves came with this behavior and IMHO it does make some sense 
* Space is for selecting
* Enter is for accepting the current selections 

But I understand that it's deviating from the current behavior and if we think it will cause too much friction I can add that back in. 

> I can't go into partitioning and de-select a partition if I want to un-do my chosen setup (lets say I changed my mind, and pre-mounted things, and now want to un-select the partition i selected)

Good catch, I'll have a look at it.

Torxed:
> But I understand that it's deviating from the current behavior and if we think it will cause too much friction I can add that back in.

We can deviate from the current, it's one more button to press but it makes it more explicit, as I wasn't always convinced that I selected anything in the old menu so.. this might actually be a good thing hehe.

svartkanin:
> We might need to change the header color combo too.
> I suspect the contrast for color blind people might be frustrating on the title bar:
> 
> image
> 
> This might be nitpicking tho :)

Fair point, and gives my bad eyes I already had issues myself.

I think I picked the official arch blue colir hex, but I can make it darker and hopefully that improves it

Torxed:
> I think I picked the official arch blue colir hex, but I can make it darker and hopefully that improves it

Ah, the color is official but the terminal in the ISO doesn't render it correctly (it doesn't map the escape codes perfectly to the CSS I think).

archinstall/tui/ui/components.py
svartkanin: Not sure I'm following on this one, which toast are you referring to?

svartkanin:
> When selecting a mirror region, it gets stuck on loading the region list (with the new cool loader animation). And it feels like a timeout on the retrieval of the list and the menu system doesn't detect the timeout/error.

This may be fixed with https://github.com/archlinux/archinstall/pull/4031

svartkanin:
> I can't go into partitioning and de-select a partition if I want to un-do my chosen setup (lets say I changed my mind, and pre-mounted things, and now want to un-select the partition i selected)

Hmm not sure I fully grasped this one, so when I do 

`Disk configuration` -> `Partitioning` -> `Default` -> `Select partitions` -> `Enter` 

and then 

`Disk configuration` -> `Partitioning` -> `Pre-mounted` 

It should be correct no? 

What do you mean with 
> and now want to un-select the partition

svartkanin:
Okay I had a bold assumption that accessibility would be available considering the maturity of the framework, but it seems I was wrong

https://github.com/Textualize/textual/issues/2425
https://textual.textualize.io/roadmap/

Maybe I can hack around it by setting the hardware cursor...

Torxed:
> > I can't go into partitioning and de-select a partition if I want to un-do my chosen setup (lets say I changed my mind, and pre-mounted things, and now want to un-select the partition i selected)
> 
> Hmm not sure I fully grasped this one, so when I do
> 
> `Disk configuration` -> `Partitioning` -> `Default` -> `Select partitions` -> `Enter`
> 
> and then
> 
> `Disk configuration` -> `Partitioning` -> `Pre-mounted`
> 
> It should be correct no?
> 
> What do you mean with
> 
> > and now want to un-select the partition

I guess that works fine, as I don't know why I was expecting to be able to un-select an option and "leave it blank for now" (aka, un-do my choice):

rec

Torxed:
> Okay I had a bold assumption that accessibility would be available considering the maturity of the framework, but it seems I was wrong
> 
> [Textualize/textual#2425](https://github.com/Textualize/textual/issues/2425) https://textual.textualize.io/roadmap/
> 
> Maybe I can hack around it by setting the hardware cursor...

Oh damn, I actually also just assumed it would be there :sweat_smile:
Yeah try with the cursor, and worst case maybe the popups from `notify()` can work, and just hide them via CSS.

h8d13:
Just an idea for the actual installation screen: Is there any way we could display "major steps" 

What I mean is that they see all the process which is great but have a kind of "Summary" steps:

Perhaps like in bottom right "Formatting disks (1/14)". 
This would be good because of logs being very fast/hard to read. And make the install process more transparent to user. To know what's going on in the full logs. Something I've been wanting to do in my fork but never got around to.

svartkanin:
That's something for the future and unrelated to this PR

codefiles:
@svartkanin, if you are going to remove `examples/mac_address_installation.py` would you please create a separate pull request for it. With that removed it will soon be possible to remove `archinstall/lib/storage.py`. Thank you.

svartkanin:
Sure, I removed it because it was breaking imports but I can work around it

PR #4002, opened 2025-12-16
Hwclock/Tz - disclaimer and display

h8d13:
Didn't find too much about this in the common issues/docs and have seen this countless times on reddit/other forums...

Why I thought it was good idea to have it in FAQ (with some more useful info) and also a visual in the TUI.

**Tests:** 
Performed a few basics test to show concept: 
- Set BIOS to 2002 -> Offline install would fail (Shows RTC time as 2002)
- Set wrong tz -> Online install fails -> Set to `None` + `mandatory=True` 
> Forces to make a considerate choice instead of using `UTC` default 
- Set any tz -> Hides RTC time and shows picked (like before)

Screenshot_20251216_152507
Screenshot_20251216_152530

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ For a quick fix the below command will install the latest keyrings
 
 ```pacman -Sy archlinux-keyring```
 
-## GPG verification errors pacstrap
+## PGP verification errors pacstrap
 
 This can occur if the **timezone selected** is incorrect and/or hardware clock in motherboard settings is not accurate (fails during verification of packages at first base-strap). Ie. when CMOS battery is dead and power cut-off happens this can reset other settings such as Clock, Sata Mode, Fast Boot, etc...
 


archinstall/lib/args.py
```diff
@@ -73,7 +73,7 @@ class ArchConfig:
 	packages: list[str] = field(default_factory=list)
 	parallel_downloads: int = 0
 	swap: bool = True
-	timezone: str = 'UTC'
+	timezone: str | None = None
```
svartkanin, line 76: why not leaving it here as default with all the others?

archinstall/lib/global_menu.py
```diff
@@ -330,7 +331,12 @@ def _prev_applications(self, item: MenuItem) -> str | None:
 
 	def _prev_tz(self, item: MenuItem) -> str | None:
 		if item.value:
-			return f'{tr("Timezone")}: {item.value}'
+			output = f'{tr("Timezone")}: {item.value}'
+			return output
+		else:
+			# Show RTC time when no timezone is selected yet
```
svartkanin, line 337: When would that happen? Isn't `UTC` always used as a default?

h8d13: Basically I thought it would avoid a lot of errors just to set to `None` + `mandatory=True` because it forces user not to skip this step. Do let me know if I over-edited something (first time making changes to json cfg template). 

I also thought that RTC time is relevant here for people doing offline installs using archiso fully cached

Issue #4001, opened 2025-12-16
Bigger root partition size for LVM on LUKS


samu-developments:
### describe the request

AFAIK it is not possible to change the partition size for the root partition when using LVM on LUKS with archinstall. On my install the root is 21 GB, which quickly fills up. Would be great if archinstall supported an increased default size to eg. 32 GB, or ideally make it configurable. Arch docs recommend 23-32 GB https://wiki.archlinux.org/title/Installation_guide#Example_layouts

h8d13:
There are lots of areas where configurable sizes would be nice and even more options (like `/efi`) but this has to take in consideration a lot of things (bootloaders capabilities, FS types, etc). 

This also why there are the manual and pre-mounted options (which I haven't used much either to be totally honest with you). 

This is for both LVM with encryption or without. Cheers

PR #4005, opened 2025-12-17
Change LVM /root def to adapt dynamically

h8d13:
For our friend in #4001

Also from [wiki](https://wiki.archlinux.org/title/Installation_guide#Example_layouts)
> At least 23–32 GiB.

### DIFF
--- a/archinstall/lib/interactions/disk_conf.py
+++ b/archinstall/lib/interactions/disk_conf.py
@@ -602,7 +602,7 @@ def suggest_lvm_layout(
 		[p.length for p in other_part],
 		Size(0, Unit.B, SectorSize.default()),
 	)
-	root_vol_size = Size(32, Unit.GiB, SectorSize.default())
+	root_vol_size = process_root_partition_size(total_vol_available, SectorSize.default())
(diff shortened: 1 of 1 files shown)


archinstall/lib/interactions/disk_conf.py
```diff
@@ -602,7 +602,7 @@ def suggest_lvm_layout(
 		[p.length for p in other_part],
 		Size(0, Unit.B, SectorSize.default()),
 	)
-	root_vol_size = Size(20, Unit.GiB, SectorSize.default())
+	root_vol_size = Size(32, Unit.GiB, SectorSize.default())
```
svartkanin: Instead of hardcoding it shall we use the same function to calcutlate the root for LVM as for regular disk config https://github.com/svartkanin/archinstall/blob/c8e414160a83c8f8bb3bfacb3f1cd0aa60c8e680/archinstall/lib/interactions/disk_conf.py?plain=1#L304-L316

h8d13:
image

Thanks for recommendation, it does look much better without the hardcoded value: tested display only on a 256GiB disk ^^

PR #4009, opened 2025-12-18
Update general QEMU instructions in main README

h8d13:
Since we already have a good Building and Testing section, I thought it would be better to have general QEMU/Other virt section and simplify a little (also updated a bit for extra possibilities, since this docs was from 2021).

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,13 +175,27 @@ qemu-system-x86 -enable-kvm -m 4096 -cpu host -smp 4 \
 
 More advanced use cases can use `virglrenderer qemu-hw-display-virtio-gpu-pci qemu-hw-display-virtio-gpu qemu-ui-sdl qemu-ui-gtk qemu-audio-pipewire`. And full list of subjects can be found [here](https://wiki.archlinux.org/title/Category:Virtualization)
 
-Possibly get hardware acceleration/sound to work via `-device virtio-gpu-pci`, `-vga none -display sdl,gl=on` and `-audiodev pipewire,id=snd0 -device intel-hda -device hda-duplex,audiodev=snd0`. Illustrative examples.
+Possibly get hardware acceleration/sound to work via `-device virtio-gpu-pci`, `-vga none -display sdl,gl=on` and `-audiodev pipewire,id=snd0 -device intel-hda -device hda-duplex,audiodev=snd0`. Illustrative examples. The guest system my also require drivers such as `vulkan-swrast` or `vulkan-virtio`.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -135,45 +135,58 @@ The profiles' definitions and the packages they will install can be directly vie
...
+
+More advanced use cases can use `virglrenderer qemu-hw-display-virtio-gpu-pci qemu-hw-display-virtio-gpu qemu-ui-sdl qemu-ui-gtk qemu-audio-pipewire`. And full list of subjects can be found [here](https://wiki.archlinux.org/title/Category:Virtualization)
+
+Possibly get hardware acceleration/sound to work via `-device virtio-gpu-pci`, `-vga none -display sdl,gl=on` and `-audiodev pipewire,id=snd0 -device intel-hda -device hda-duplex,audiodev=snd0`. Illustrative examples.
+
+Or use any of the GUI managers which should handle parts of this for you: [VirtManager](https://wiki.archlinux.org/title/Virt-manager), [VirtualBox](https://wiki.archlinux.org/title/VirtualBox), [VMWare](https://wiki.archlinux.org/title/VMware), RemoteBox, etc.
+
+## Without a Live ISO Image
 
-    # truncate -s 20G testimage.img
```
svartkanin, line 164: can we leave this in as it is very useful for running archinstall locally during dev and use the loopback device

README.md
```diff
@@ -135,45 +135,58 @@ The profiles' definitions and the packages they will install can be directly vie
 
 ## Using a Live ISO Image
 
-If you want to test a commit, branch, or bleeding edge release from the repository using the standard Arch Linux Live ISO image,
-replace the archinstall version with a newer one and execute the subsequent steps defined below.
-
-*Note: When booting from a live USB, the space on the ramdisk is limited and may not be sufficient to allow
```
svartkanin, line 141: I think this is still relevant or not?

README.md
h8d13: I tried them locally and didn't really feel it was clear enough (second command didn't seem to work) maybe I'm missing something here and we can clarify.

README.md
h8d13: I added `git clone -b testing-branch https://github.com/archlinux/archinstall` As I think that clearer way to put it and a link to forks which I thought is cleaner

README.md
svartkanin: They work fine for me. 
I think this one 
```
# losetup --partscan --show --find ./testimage.img
```
has changed recently as the `--find` doesn't seem to work anymore, but removing that will make it work

PR #4013, opened 2025-12-20
feat(applications): add CUPS installation support

ghost:
This pull request adds support for installing [CUPS](https://openprinting.github.io/cups/) via the applications menu which is required if you'd like to print documents.

I figured this would be worth adding and wouldn't cross the very arbitrary line of what is and isn't considered bloat and user opinion since I'd classify this as hardware enablement and not just some random application we're forcing on people. 

It could also be considered expected behavior by possibly somewhat new users who don't know that you need to install a separate package & enable it's service to be able to print documents, and all that implies they know what the package/service is called anyway. This should make it more clear and much easier for those newer users that may install their favorite desktop environment and wonder why the printer settings in their desktop's respective control panel are missing or just don't work at all.

### DIFF
--- a/tests/test_args.py
+++ b/tests/test_args.py
@@ -132,7 +132,7 @@ def test_config_file_parsing(
 		app_config=ApplicationConfiguration(
 			bluetooth_config=BluetoothConfiguration(enabled=True),
 			audio_config=AudioConfiguration(audio=Audio.PIPEWIRE),
-			print_service_config=PrintServiceConfiguration(enabled=True)
+			print_service_config=PrintServiceConfiguration(enabled=True),
 		),
 		auth_config=AuthenticationConfiguration(
 			root_enc_password=Password(enc_password='password_hash'),


archinstall/lib/applications/application_menu.py
```diff
@@ -94,6 +109,34 @@ def select_bluetooth(preset: BluetoothConfiguration | None) -> BluetoothConfigur
 			raise ValueError('Unhandled result type')
 
 
+def select_printer(preset: PrinterConfiguration | None) -> PrinterConfiguration | None:
```
svartkanin: this sounds misleading as it might indicate to select "a printer", can we rename it to `select_print_service`

archinstall/applications/printer.py
```diff
@@ -0,0 +1,23 @@
+from typing import TYPE_CHECKING
+
+from archinstall.lib.output import debug
+
+if TYPE_CHECKING:
+	from archinstall.lib.installer import Installer
+
+
+class PrinterApp:
```
svartkanin: Given its a print service, lets rename this to `PrintServiceApp`

svartkanin:
This sounds like a good addition, can you rename the 2 mentioned things then it's good to go in

archinstall/applications/printer.py
ghost: Fixed in [2fe0fd4](https://github.com/archlinux/archinstall/commit/2fe0fd406e04bee9a3d50ee2cbcde2066e714743)

archinstall/lib/applications/application_menu.py
ghost: Fixed in [2fe0fd4](https://github.com/archlinux/archinstall/commit/2fe0fd406e04bee9a3d50ee2cbcde2066e714743)

ghost:
Should be good to go!

svartkanin:
Formatting isn't pasing

ghost:
Eek! My formatter removes those useless commas, it should be happy now.

PR #4014, opened 2025-12-20
fix: fix translation for bluetooth preview_action

ghost:
This pull requests fixes one of the bluetooth menus not respecting the user's language.


svartkanin:
I was under the impression that that's the name used in any language but maybe not?

ghost:
It probably is, but doesn't hurt to make it able to be translated just in case.

Issue #2557, opened 2024-06-30
power-profiles-daemon package should be installed by default


emir814:
https://archlinux.org/packages/extra/x86_64/power-profiles-daemon/ I think it's a necessary thing, of course it's up to the developers.

Torxed:
I've been thinking for years, how this would be a nice thing if for instance there was a way to detect if we're on a laptop 

And if we could put it under a "optimizations" menu that is strictly optional. 

Would that be something? 
I don't want to make people think we're trying to push things on them - but I do a lot of laptop installs and things like these would free up a lot of work.

arlsdk:
> I've been thinking for years, how this would be a nice thing if for instance there was a way to detect if we're on a laptop
> 
> And if we could put it under a "optimizations" menu that is strictly optional.
> 
> Would that be something? I don't want to make people think we're trying to push things on them - but I do a lot of laptop installs and things like these would free up a lot of work.
> 
> Is there a better way than checking `/sys/class/power/BAT*`?

Maybe checking the lid?
`/proc/acpi/button/lid` [Original answer](https://stackoverflow.com/a/68676700)

ymcx:
I'd say that going with [tuned](https://archlinux.org/packages/extra/any/tuned/) instead of PPD would be a better choice. Fedora did the migration a while back for the reasons listed [here](https://fedoraproject.org/wiki/Changes/TunedAsTheDefaultPowerProfileManagementDaemon).

PR #4015, opened 2025-12-20
feat(applications): add support for power-profiles-daemon/tuned as a power management daemon

ghost:
This pull request adds the power-profiles-daemon package to the desktop profile by default.

I have a feeling that most users probably end up installing this (or something similar such as [tuned](https://tuned-project.org/)) after installation or by selecting it in the additional packages menu anyway, especially people doing installations on laptops or other mobile devices, so it may be worth it to just install this by default for people selecting the desktop profile. 

This also has the benefit of improving battery life on laptops and other mobile devices for users who may not know they need to install this package to get basic power management functionality.

### DIFF
--- a/archinstall/lib/global_menu.py
+++ b/archinstall/lib/global_menu.py
@@ -329,6 +329,11 @@ class GlobalMenu(AbstractMenu[None]):
 				output += tr('Enabled') if app_config.print_service_config.enabled else tr('Disabled')
 				output += '\n'
 
+			if app_config.power_management_config:
+				power_management_config = app_config.power_management_config
+				output += f'{tr("Power management")}: {power_management_config.power_management.value}'
+				output += '\n'
+
 			return output
 
 		return None


h8d13:
You'd need to enable the service too and not all DEs package an integration to this, I have a similar patch [here](https://github.com/archlinux/archinstall/compare/master...h8d13:archinstall-patch:dot-hw)

ghost:
> You'd need to enable the service

power-profiles-daemon and tuned (via tuned-ppd) are both started by D-Bus automatically, this could possibly be an issue for people who end up installing a window manager and not one of the big desktop environments, but we're also shipping pipewire which is started by D-Bus too and no one has complained yet (that I can see) so it should be fine?

> not all DEs package an integration to this

All the full desktop environments that currently have profiles support it (i.e: GNOME, Plasma, Cinnamon, COSMIC, Xfce, etc.) out of the box without any additional packages as long as something is providing the expected D-Bus interface. I guess it could be considered 'bloat' for some people who end up installing a window manager, but I suspect even laptop users who install a window manager will end up installing either this or tuned.

h8d13:
It's made available for control over d-bus but I think it still needs to be explicitly enabled through `systemctl` have you tested ?

ghost:
I was pretty sure this was enough since this is how I've always installed it on my personal machines, but I decided to spin up a virtual machine just to double check, and yes I can confirm that it is. A fresh installation using the desktop profile with both KDE Plasma and GNOME (didn't bother testing the others due to slow internet, if someone wants to test this patch further, be my guest!) seem to work just fine.

Here is a screenshot from the KDE Plasma install showing power-profiles-daemon running and working even though the service is disabled:
image

The power profiles not being "supported" is just due to it being a virtual machine, ignore that.

h8d13:
My bad then. Huh I think it's a good change, not sure I'd do it by default for desktops, hence why I wanted a new `is_laptop` def in my patch. Other than that I also thought would be nice QOL

svartkanin:
There is an open issue for this https://github.com/archlinux/archinstall/issues/2557 which a discussion running there and I can see people suggesting `tuned` over `power-profile-daemon`. So I think this will definitely cause some contention by just adding it as an installed package. 

I think this will potentially need at least a selection between the 2 and maybe even only be available if the host device is a laptop/battery?

ghost:
I personally don't see the benefit of tuned over power-profiles-daemon as both it and power-profiles-daemon essentially do the exact same thing, but I'm not completely against it either, it is arguably more customizable to the end user (which is really the only point brought up in the Fedora proposal that is linked in the discussion). If anyone has any arguments in favor of tuned, feel free to attempt to persuade me!

I'm just personally not a fan of the long startup time of tuned (~400ms or so on my older laptop) since well, it's written in Python, and I think presenting users with the option to pick between the two could potentially confuse newer users who may not know the differences between the two, or even what their purpose is.

I definitely agree with the laptop/mobile device check though, I'll experiment with some things and work on implementing that.

ghost:
I added a check to see if a battery is present before installing the package in [a5573ce](https://github.com/archlinux/archinstall/pull/4015/commits/a5573ce414e45e05d8f0661095368e8ee5730a1e). It's very naïve and could potentially have false positives with weird hardware since it simply checks if a battery is present on the machine and nothing else, but it works on my test machines (desktop and multiple laptops) without any problems so it should be alright. 

I also wasn't really too sure if this is the ideal location to put this snippet, even after looking around for a bit, but I figured since it was profile specific it belonged there in ``desktop.py``. If it doesn't, or there's a more ideal location you'd like this, let me know.

h8d13:
Hey since you're ln a roll you mind reading over this:

https://archlinux.org/news/nvidia-590-driver-drops-pascal-support-main-packages-switch-to-open-kernel-modules/

Does this imply changes for `hardware.py` file ?

Brain doesn't work properly have some fever lmao

Also saw an older pull requests about this to let users pick non-dkms (and not install headers)

ghost:
Yes it would, I have a branch over in my working repository for changes related to this, just haven't gotten around to finishing them and I think it's time for me to clock out for the day :) I'll probably open a pull request sometime soon (in a couple of days?) as I'm hoping to land them before the January monthly .iso release, discussion can continue there when that happens as to not fill this thread with unrelated discussions :+1:

archinstall/lib/hardware.py
```diff
@@ -198,6 +198,19 @@ def loaded_modules(self) -> list[str]:
 
 
 class SysInfo:
+	@staticmethod
+	def has_battery():
+		for device in os.listdir('/sys/class/power_supply/'):
```
svartkanin: Can be simplified with 
```
Path('/sys/class/power_supply/').glob('*/type')
```

svartkanin:
As there are alternatives to PPD available https://wiki.archlinux.org/title/CPU_frequency_scaling#Userspace_tools I'd prefer if this is a new option in the `Application` sub-menu instead

archinstall/lib/hardware.py
ghost: :+1:

ghost:
This should be ready for review again.

archinstall/lib/applications/application_menu.py
```diff
@@ -81,6 +103,29 @@ def _prev_print_service(self, item: MenuItem) -> str | None:
 		return None
 
 
+def select_power_management(preset: PowerManagementConfiguration | None = None) -> PowerManagementConfiguration | None:
+	items = [MenuItem(a.value, value=a) for a in PowerManagement]
```
svartkanin: just an FYI there's a helper for enums
```
MenuItemGroup.from_enum
```

archinstall/lib/hardware.py
```diff
@@ -210,6 +210,18 @@ def graphics_devices(self) -> dict[str, str]:
 
 
 class SysInfo:
+	@staticmethod
+	def has_battery() -> bool:
```
svartkanin, line 226: For a future PR, we can move this into _SysInfo and cache the return value so it doesn't get calculated multiple times

archinstall/lib/applications/application_menu.py
```diff
@@ -54,8 +63,21 @@ def _define_menu_options(self) -> list[MenuItem]:
 				preview_action=self._prev_print_service,
 				key='print_service_config',
 			),
+			MenuItem(
+				text=tr('Power management daemon'),
+				action=select_power_management,
```
svartkanin, line 68: This should just be "Power management"

archinstall/lib/models/application.py
```diff
@@ -3,6 +3,16 @@
 from typing import Any, NotRequired, TypedDict
 
 
+class PowerManagement(StrEnum):
+	NO_POWER_MANAGEMENT = 'No power management daemon'
```
svartkanin: We don't need this one, after a selection it can be reset with ctrl+c

archinstall/lib/applications/application_menu.py
```diff
@@ -81,6 +103,29 @@ def _prev_print_service(self, item: MenuItem) -> str | None:
...
 
 
+def select_power_management(preset: PowerManagementConfiguration | None = None) -> PowerManagementConfiguration | None:
+	items = [MenuItem(a.value, value=a) for a in PowerManagement]
+	group = MenuItemGroup(items)
+
+	if preset:
+		group.set_focus_by_value(preset.power_management)
+
+	result = SelectMenu[PowerManagement](
```
svartkanin, line 112: add 
```
allow_reset=True
```

archinstall/lib/models/application.py
ghost: This doesn't seem very intuitive to me, I actually didn't even know this was possible since it doesn't seem to be labeled and or documented anywhere, though maybe this is common sense and I was just unaware.

I'll probably save this change for another pull request since the audio configuration also has a ``NO_AUDIO`` value, feel like those changes should be lumped into one for book keeping sake. If you're alright with that.

ghost:
This should be ready for a final review.

archinstall/lib/models/application.py
svartkanin: It is visible in the help menu

archinstall/lib/applications/application_menu.py
```diff
@@ -81,6 +103,29 @@ def _prev_print_service(self, item: MenuItem) -> str | None:
...
+		frame=FrameProperties.min(tr('Power management')),
+	).run()
+
+	match result.type_:
+		case ResultType.Skip:
+			return preset
+		case ResultType.Selection:
+			return PowerManagementConfiguration(power_management=result.get_value())
+		case ResultType.Reset:
+			raise ValueError('Unhandled result type')
```
svartkanin: As you're now allowing a reset with `allow_reset=true` this needs to be handled here and return the default which is `None`

archinstall/lib/models/application.py
```diff
@@ -3,6 +3,16 @@
 from typing import Any, NotRequired, TypedDict
 
 
+class PowerManagement(StrEnum):
+	NO_POWER_MANAGEMENT = 'No power management'
+	POWER_PROFILES_DAEMON = 'power-profiles-daemon'
+	TUNED = auto()
```
svartkanin: Use `tuned` not `auto()`

ghost:
That should be good to go then. I'm still not sure about the whole resetting thing, I messed around with it in my test environment and it seems very unintuitive. If I select a power management daemon and then go right back into the applications menu and reset it, all is fine and dandy and it goes back to being none (previously 'no power management'), but if I select an option, exit the menu, then go back in and select the other option then try to reset it, it goes right back to the previously selected value even though I'm returning ``None`` for the reset result type? Maybe this is intended behavior and it's supposed to just act as a 'previous value' key, but this doesn't seem right?

svartkanin:
I tested it and it seems to work fine. 
* Select PM -> *any*
* Select PM -> Ctrl+c (to resett) -> "yes"
* No PM config

Something missing is the preview of the PM in the application of the main menu https://github.com/archlinux/archinstall/blob/6f768ea87c6266108323f0dd444dcb8aeefdaff3/archinstall/lib/global_menu.py?plain=1#L331

ghost:
Completely missed that, thanks!

PR #4019, opened 2025-12-21
fix(hardware): multiple nvidia driver installation fixes

ghost:
This pull request fixes a couple of issues related to the installation of the NVIDIA driver:

1. The DKMS variant of the driver was being used even when it wasn't necessary. It will now only be installed if you've selected a non-standard kernel package such as ``linux-zen`` or ``linux-lts``, which should help speed up the installation process for a majority of users.
2. It brings the installation process in line with [the packaging changes announced yesterday](https://archlinux.org/news/nvidia-590-driver-drops-pascal-support-main-packages-switch-to-open-kernel-modules/)

### DIFF
--- a/archinstall/lib/hardware.py
+++ b/archinstall/lib/hardware.py
@@ -2,6 +2,7 @@ import os
 from enum import Enum
 from functools import cached_property
 from pathlib import Path
+from typing import Any
 
 from .exceptions import SysCallError
 from .general import SysCommand
@@ -64,7 +65,7 @@ class GfxDriver(Enum):
 	MesaVirtualized = ('mesa_vm', 'VirtualBox (open-source)')
 
 	@classmethod
-	def from_key(cls, key: str):
+	def from_key(cls, key: str) -> Any:
 		for member in cls:
 			if member.value[0] == key:
 				return member
--- a/archinstall/lib/models/profile.py
+++ b/archinstall/lib/models/profile.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from multiprocessing import Value
 from typing import TYPE_CHECKING, TypedDict
 
 from archinstall.default_profiles.profile import GreeterType, Profile
@@ -18,7 +17,7 @@ class _ProfileConfigurationSerialization(TypedDict):
 	greeter: str | None
 
 
-def _parse_gfx_driver(value: str):
+def _parse_gfx_driver(value: str) -> GfxDriver:
 	# A really ugly hack to deal with older configs that use values
 	# instead of keys for their config.
 	match value:
--- a/archinstall/lib/profile/profile_menu.py
+++ b/archinstall/lib/profile/profile_menu.py
@@ -2,9 +2,7 @@ from __future__ import annotations
 
 from typing import override
 
-from archinstall import warn
 from archinstall.default_profiles.profile import GreeterType, Profile
-from archinstall.lib.output import info
 from archinstall.lib.translationhandler import tr
 from archinstall.tui.curses_menu import SelectMenu
 from archinstall.tui.menu_item import MenuItem, MenuItemGroup


h8d13:
I'll just ping these here again: #3828 #3830

If I understand correctly, there is also an option that doesn't need dkms or headers correct (but would **only work for the mainline** kernel?).  EDIT I see you already working on that (::

We might want to actually merge this early as proprietary is already a dead option (seen 2-3 reports of error logs already on reddit). Then, work on enhancements, just to avoid issues incoming 🤔 
    
About Sway disclaimer, I always thought it was unnecessary (I've used sway on Nvidia countless times?)  

**Also we should move kernel selection before profile (graphics drivers).** perhaps after bootloader would make for better logic.

ghost:
This should be enough for now, I tested it briefly on my machines and it seems to work fine, so I'm marking this ready for review. I had (and still have) other changes planned, but I'm quite busy due to the holiday season rapidly approaching so I'm putting those on hold for now so we can at least get something working in time for the monthly .iso release next week.

I'm also not too sure about the locale changes I made in [416235a](https://github.com/archlinux/archinstall/commit/416235a87cb15ba90f6e6f858936f048ed03a650), I'm assuming they are fine and it's up to the translators to update that for their respective language? If that's not how this works, feel free to let me know and I can revert that commit or bring it in line with how things are supposed to be. :)

svartkanin:
I need to look into this more as it will take quite some testing to make sure this doesn't break anything. 

In any case though the changes are not backwards compatible. Unfortunately, we're currently saving the selection value to the config 
image

which isn't ideal when parsing the config. It would be better to have that changed to the key instead which is some more work but needs to be taken care off either in this change or a future PR.

ghost:
Ouch, didn't think about that... that's quite an annoying problem to deal with because it seems no matter how we go about this, there will be at least some breakage.

I'll have to think about this more in depth after the holidays, but in the moment I feel like this is one of those situations where breakage is possibly warranted and better in the long run. If we make the config start using keys instead of the selection's label and just accept the breakage, it would make dealing with potential problems like this much easier.

ghost:
Alrighty @svartkanin, this should be good for another review.

It's not my proudest work, and there's definitely some code smells in here (Python is not my favorite language, but I make do!), but it's a simple solution to get things working in time for the monthly .iso release while also laying the ground work for the future refactoring I have in mind.

This should also be fully backwards compatible with existing config files as well thanks to the hack that is ``_parse_gfx_driver`` in ``models/profile.py``. I went ahead and made the decision to fall back to the nouveau driver for people that had the old, and now deprecated, NVIDIA proprietary driver selected in their config since if we end up installing the new open kernel module for those people and they happen to be running pre-Turing era hardware, installing the open kernel module will leave them with a broken system (they will just end up with an unresponsive black screen). I figured it was better to install something that will at least get them working graphics and they can correct everything post install (and hopefully update their archinstall config too!)

svartkanin:
There was no need to do all these changes, with the `key` I meant that we can just change this line https://github.com/archlinux/archinstall/blob/601e03318854a26c454871296b2e70ed85ce2231/archinstall/lib/models/profile.py?plain=1#L31
to 

```
'gfx_driver': self.gfx_driver.key if self.gfx_driver else None,
```

and then this line https://github.com/archlinux/archinstall/blob/601e03318854a26c454871296b2e70ed85ce2231/archinstall/lib/models/profile.py?plain=1#L45

needs to handle both the old
```
GfxDriver(gfx_driver) if gfx_driver else None,
```

and new one
```
GfxDriver[gfx_driver] if gfx_driver else None,
```

so maybe 

```
if gfx_driver:
    try:
        GfxDriver(gfx_driver)
     catch ...:
         GfxDriver[gfx_driver] 
```

ghost:
I went with the approach I went with because I personally think it is more robust since if we ever have to update the 'description' (value) of a driver, we're going to be back to the whole backwards compatibility issue for people who have older configs.

If this isn't desirable, I'm happy with just going with a simple solution like the one you proposed, but that's just my two cents.

PR #4022, opened 2025-12-21
Do not install base-devel by default

h8d13:
This a follow up to [discussion](https://github.com/archlinux/archinstall/discussions/159) I guess this is something that might have been a to-do but never got ticked. Discussion dates back from **2021**, but is a generally valid  point IMO. 

> We could remove it I suppose. I don't recall a specific reason for it being there other than this was one of my default packages back in ~2014. It does pull in `sudo` and and `which`, as well as being an optional pacman dependency.
> 
> But in reality, submitting a PR with a suggestion to remove it is welcome and we'll see the community feedback (if any) :)

From @Torxed originally ^^ Comment from June **2025**

Screenshot_20251221_201225

Basically allows us to make `base-devel` something that should explicitly declared by user.  As it shouldn't be assumed for non dev use-case. 

Screenshot_20251221_203126

This also aligns more with arch philosophy and makes it so that build-tools/other explicitly have dep `-devel`. 

I added some disclaimers of this change but am unsure to be 100% honest if that's how you edit them :D And since the first line of most AUR helpers is `sudo pacman -S --needed git base-devel`

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -54,11 +54,6 @@ from .storage import storage
 # Any package that the Installer() is responsible for (optional and the default ones)
 __packages__ = ['base', 'sudo', 'linux-firmware', 'linux', 'linux-lts', 'linux-zen', 'linux-hardened']
 
-# base-devel is actually more a developer's thing and isn't strictly required in arch systems
-# it also overlaps the following pacman, grep, sed, gawk, findutils, file, gettext, gzip, archlinux-keyring (already in base)
-# not needed for a common user gcc, make, autoconf, automake, binutils, flex, bison, etc.
-# see https://github.com/archlinux/archinstall/discussions/159
-
 # Additional packages that are installed if the user is running the Live ISO with accessibility tools enabled
 __accessibility_packages__ = ['brltty', 'espeakup', 'alsa-utils']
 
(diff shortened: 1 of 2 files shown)


archinstall/lib/installer.py
```diff
@@ -52,7 +52,12 @@
 from .storage import storage
 
 # Any package that the Installer() is responsible for (optional and the default ones)
-__packages__ = ['base', 'base-devel', 'linux-firmware', 'linux', 'linux-lts', 'linux-zen', 'linux-hardened']
+__packages__ = ['base', 'sudo', 'linux-firmware', 'linux', 'linux-lts', 'linux-zen', 'linux-hardened']
+
+# base-devel is actually more a developer's thing and isn't strictly required in arch systems
```
svartkanin: I don't think we need this comment here, it's enough if it's on the PR and we can find it through the commit

archinstall/locales/base.pot
```diff
@@ -46,10 +46,13 @@ msgid "Choose an audio server"
 msgstr ""
 
 msgid ""
-"Only packages such as base, base-devel, linux, linux-firmware, efibootmgr "
+"Only packages such as base, sudo, linux, linux-firmware, efibootmgr "
```
svartkanin: Please don't update these translations in here manually as that will mess things up

PR #4024, opened 2025-12-22
Fix LVM creation/ info 

### LINKED ISSUE #3424: Installation is stuck on VirtualBox virtual machine "Setting up LVM config..."

2025-05-01 (latest)

### The installation log

```json
http://0x0.st/84qf.log
```

### describe the problem

I'm using best-effort partition split with LVM enabled. All the partitions are set to ext4. All the settings are pretty much the default ones.

A similar issue:
https://github.com/archlinux/archinstall/issues/2924

h8d13:
Aims to fix #3424

Can be tested on branch:

```
git clone -b fix-lvm-hang https://github.com/h8d13/archinstall-patch
```

I believe both `while True` and missing sync/polls was causing race condition

### DIFF
--- a/archinstall/lib/disk/device_handler.py
+++ b/archinstall/lib/disk/device_handler.py
@@ -422,8 +422,8 @@ class DeviceHandler:
 		cmd: str,
 		info_type: Literal['lv', 'vg', 'pvseg'],
 	) -> LvmVolumeInfo | LvmGroupInfo | LvmPVInfo | None:
-		# Retry for up to 30 seconds (10 attempts * 3 seconds)
-		max_retries = 10
+		# Retry for up to 5 mins
+		max_retries = 100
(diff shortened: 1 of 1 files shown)


archinstall/lib/disk/device_handler.py
```diff
@@ -422,11 +422,18 @@ def _lvm_info_with_retry(
 		cmd: str,
 		info_type: Literal['lv', 'vg', 'pvseg'],
 	) -> LvmVolumeInfo | LvmGroupInfo | LvmPVInfo | None:
-		while True:
+		# Retry for up to 30 seconds (10 attempts * 3 seconds)
```
svartkanin: If it's a large disk this may not be enough, it should be higher than that maybe 5min?

svartkanin (approved):

PR #4025, opened 2025-12-22
Enable IWD to be used as back-end in network selection

h8d13:
This PR should solve a lot of issues for Intel users among other legacy hardware.

Issue originally from user: Coda_Phi : On a surface pro 7 (Notoriously Intel contributes to IWD dev, and has bad support with default wpa_supplicant default backend) This notably results in "No networks" in desktops.

image

So I tested a little on a `Intel AX211` similar to user's `AX201`

It follows https://wiki.archlinux.org/title/NetworkManager#Using_iwd_as_the_Wi-Fi_backend
And https://wiki.archlinux.org/title/NetworkManager#iwd_support_for_captive_portal_support_on_legacy_hardware

By setting enable services to `False` then enabling only `NetworkManager` and writing the necessary config file to use IWD as back-end. It also still copies from ISO re-using existing definitions too.

Simply makes it so that you configure in ISO env before or during archinstall => Still works in DE

### DIFF
--- a/archinstall/lib/models/network.py
+++ b/archinstall/lib/models/network.py
@@ -164,15 +164,15 @@ class NetworkConfiguration:
 					if profile_config.profile.is_desktop_profile():
 						installation.add_additional_packages('network-manager-applet')
 
-				if self.type == NicType.NM:
-					# Default just NM service
-					installation.enable_service('NetworkManager.service')
-				else:  # NM_IWD special handling
+				installation.enable_service('NetworkManager.service')
+				if self.type == NicType.NM_IWD:
+					# NM_IWD special handling
 					installation.configure_nm_iwd()
+					installation.disable_service('iwd.service')
+
 			case NicType.MANUAL:
 				for nic in self.nics:
 					installation.configure_nic(nic)
-
 				installation.enable_service('systemd-networkd')
 				installation.enable_service('systemd-resolved')
 
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -776,14 +776,6 @@ class Installer:
 		iwd_backend_conf = nm_conf_dir / 'wifi_backend.conf'
 		iwd_backend_conf.write_text('[device]\nwifi.backend=iwd\n')
 
-		# Disable standalone iwd service
-		# and enable NetworkManager
-		def post_install_configure_nm_iwd(*args: str, **kwargs: str) -> None:
-			self.disable_service('iwd.service')
-			self.enable_service('NetworkManager.service')
-
-		self.post_base_install.append(post_install_configure_nm_iwd)
-
 	def mkinitcpio(self, flags: list[str]) -> bool:
 		for plugin in plugins.values():
 			if hasattr(plugin, 'on_mkinitcpio'):


archinstall/lib/models/network.py
```diff
@@ -25,6 +26,8 @@ def display_msg(self) -> str:
 				return tr('Copy ISO network configuration to installation')
 			case NicType.NM:
 				return tr('Use NetworkManager (necessary to configure internet graphically in GNOME and KDE Plasma)')
+			case NicType.NM_IWD:
+				return tr('Use NetworkManager with iwd backend (copies ISO config, enables graphical tools)')
```
svartkanin: I'd be very happy if we can rename the whole thing to 
```
Network Manager (default backend)
Network Manager (iwd backend)
```

svartkanin:
The `wpa_supplication` package is installed by default https://github.com/archlinux/archinstall/blob/601e03318854a26c454871296b2e70ed85ce2231/archinstall/default_profiles/desktop.py?plain=1#L35, I think this will need to be only added if the default NM is selected.

archinstall/lib/models/network.py
```diff
@@ -148,11 +151,20 @@ def install_network_config(
...
 				if profile_config and profile_config.profile:
 					if profile_config.profile.is_desktop_profile():
 						installation.add_additional_packages(['network-manager-applet'])
 				installation.enable_service('NetworkManager.service')
+			case NicType.NM_IWD:
+				# Copy ISO config first for initial connectivity
+				installation.copy_iso_network_config(enable_services=False)
+				# Configure NetworkManager with iwd backend
+				installation.add_additional_packages('networkmanager')
+				if profile_config and profile_config.profile:
```
svartkanin, line 163: should we also install the applet on default NM?

h8d13: I think its still needed for both desktops, I could maybe simplify 

But it checks that its a desktop profile 

`if profile_config.profile.is_desktop_profile():` For both `NM` and `NM_IWD` 
See bellow

h8d13: Idek if this looks better makes more sense in OoO

h8d13:
Maybe we hold on to this for a little testing on appropriate hardware

archinstall/lib/models/network.py
```diff
@@ -147,12 +150,31 @@ def install_network_config(
 				installation.copy_iso_network_config(
 					enable_services=True,  # Sources the ISO network configuration to the install medium.
 				)
-			case NicType.NM:
-				installation.add_additional_packages(['networkmanager'])
+			case NicType.NM | NicType.NM_IWD:
+				# Common setup for both NetworkManager configurations
+				if self.type == NicType.NM_IWD:
+					# Copy ISO config first for initial connectivity
+					installation.copy_iso_network_config(enable_services=False)
```
svartkanin: Why are now doing the iso config copy for NM setup?

archinstall/lib/models/network.py
```diff
@@ -147,12 +150,31 @@ def install_network_config(
 				installation.copy_iso_network_config(
 					enable_services=True,  # Sources the ISO network configuration to the install medium.
 				)
-			case NicType.NM:
-				installation.add_additional_packages(['networkmanager'])
+			case NicType.NM | NicType.NM_IWD:
+				# Common setup for both NetworkManager configurations
```
svartkanin: I guess you're using an LLM, can we remove these redundant comments

archinstall/lib/models/network.py
h8d13: This is how i had done it in the first place. I didnt see the downside since makes it so it works + keeps the ISO's set-up (usually already through iwctl) I can remove the copy / comments ?

archinstall/lib/models/network.py
h8d13: Don't really appreciate the comment tho, and always write these myself + test extensively

archinstall/lib/models/network.py
Torxed: The issue is that it will probably cause a conflict between `systemd-networkd` and `NetworkManager`, I'm on a phone so can't verify but I suspect we also enable the systemd-networkd service in there too, and that's the colliding factor if that's the case.

h8d13: Ah i thought that `(enable_services=False)` was handling that. I can just remove it to be honest with you, do let me know :)

Torxed:
I think this looks good to go in now.
Future idea: maybe by selecting NM we can have a sub-menu or something to "configure" it, when it's selected a submenu could be to "select backend".. or just a "popup" asking :)

But that can be a separate/later PR when the new TUI is in.

svartkanin (approved):

svartkanin:
I think the "Copy ISO config" option is somewhat misleading altogether as it won't copy any manually setup wifi networking which would be ideal as I imagine laptop users will mostly connect to it

PR #4027, opened 2025-12-24
Set up Zram dynamically based on best practice

h8d13:
This patch follows [Wiki:ZRAM1.3](https://wiki.archlinux.org/title/Zram#Using_zram-generator)
But adapted in python logic :

Half of the total RAM available written back to the conf file (with minimum 4GiB regardless of size).


CelestifyX:
In my opinion, it would be better to give the user the opportunity to choose the size themselves, or use the default value.

svartkanin (approved):

Issue #4026, opened 2025-12-24
Regression: Grub install on MSI/Macs boards


h8d13:
I had already commented this didn't look good in #3932
https://wiki.archlinux.org/title/GRUB first tip is to use `--removable`

I don't see what value it provides either to have this only for 2 boot-loaders. And also commits were messy, and get pushed in faster than minor changes...

 _Originally posted by @h8d13 in [eb815d8](https://github.com/archlinux/archinstall/commit/eb815d817f3eacf0f83f1f678ecef1233e486bb6#r171157375)_

This leads to kernel panic/unable to find kernel/initrd files, on MSI boards (if not others) and can be easily reproduced: 
`git checkout 840c1c0` will work just fine.  The same is true in QEMU or using os-prober with grub. 

The prompt for removable bootloader option also Defaults to "no" and so hides a sane default behind a toggle.

h8d13:
@svartkanin could you please check this out when you have time before next month's release since it affects a lot of hardware (for a month now), thanks in advance and happy holidays, Hade

Mintsuki:
Happy holidays. I opened #4030 to address this issue.

Issue #950, opened 2022-02-07
`--removable` in `grub-install` fails on some hardware


Torxed:
The full thread can be found here: https://discord.com/channels/726808923662712894/726809631342592081/937766587044364321
But here are some exerpts:
 * Log: https://0x0.st/oHEp.log

We'll need to detect the need for `--removable` or retry if it fails, here's the relevant code: https://github.com/archlinux/archinstall/blob/1515fb9fdeeb807b1301bc803d625dfebca8aaca/archinstall/lib/installer.py#L770

timmypidashev:
I feel a good idea would be to enable the --debug flag just so people can get an idea of what went wrong with grub in the future, but its also just nice to see it compile. Makes me calm at least. This is just my opinion, won't fix the bug aforementioned above.

Torxed:
We can in a pretty simplistic way add `peak_output=True` to any command we run, which will render it to stdout. We can do that for the grub commands.

I’m in the process of refactoring `SysCommand` so that the forking doesn’t break. Currently we’re getting quite lucky that it doesn’t run commands twice and stuff. So no one will know, but that needs to be addressed before people know :)

timmypidashev:
Awesome, I look forward to that.

Torxed:
I've added `peak_output` to the `grub-install` commands, which should give a more detailed although "spammy" output of what's going on. I've also added `--debug` to show in greater detail where it might fail.

This will for sure help us debug issues down the road, but at some point we'll have to remove this or pipe it somewhere.

This will fix the above mentioned issue, but not other grub issues.
But it will help to diagnose :)

PR #4030, opened 2025-12-25
Make removable location the default for bootloader installation

Mintsuki:
Also update the option's description to make it clear that it being enabled is the sane default.

### DIFF
--- a/archinstall/lib/args.py
+++ b/archinstall/lib/args.py
@@ -185,7 +185,7 @@ class ArchConfig:
 			uki = args_config.get('uki', False)
 			if uki and not bootloader.has_uki_support():
 				uki = False
-			arch_config.bootloader_config = BootloaderConfiguration(bootloader=bootloader, uki=uki, removable=False)
+			arch_config.bootloader_config = BootloaderConfiguration(bootloader=bootloader, uki=uki, removable=True)
 
 		# deprecated: backwards compatibility
 		audio_config_args = args_config.get('audio_config', None)
--- a/archinstall/lib/bootloader/bootloader_menu.py
+++ b/archinstall/lib/bootloader/bootloader_menu.py
@@ -83,8 +83,8 @@ class BootloaderMenu(AbstractSubMenu[BootloaderConfiguration]):
 
 	def _prev_removable(self, item: MenuItem) -> str | None:
 		if item.value:
-			return tr('Will install to /EFI/BOOT/ (removable location)')
-		return tr('Will install to standard location with NVRAM entry')
+			return tr('Will install to /EFI/BOOT/ (removable location, safe default)')
+		return tr('Will install to custom location with NVRAM entry')
 
 	@override
 	def run(
@@ -114,6 +114,9 @@ class BootloaderMenu(AbstractSubMenu[BootloaderConfiguration]):
 				removable_item.value = False
 				self._bootloader_conf.removable = False
 			else:
+				if not removable_item.enabled:
+					removable_item.value = True
+					self._bootloader_conf.removable = True
 				removable_item.enabled = True
 
 		return bootloader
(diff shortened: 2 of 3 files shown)


Mintsuki:
Addresses #4026

h8d13:
Bump 👍🙏

Torxed:
@svartkanin IIRC we made this an option in the partitioning/grub menu, didn't we? to make it optional instead a forced default for everyone?

h8d13:
Yes the problem is hardware support and text was miss-leading to press "No" 
On Macs/MSI just causes black/blue-screen and a nice QR code

Torxed:
I'm perfectly fine with updating the help text and descriptions.
But I don't think we should change the actual default behavior? As this is hardware specific right now.

Mintsuki:
This reverts the behaviour to what it was before we added the dialog, in the sense that it defaults to removable just like it was before, when the dialog was not there to begin with.

Torxed:
I can only find [one old issue](https://github.com/archlinux/archinstall/issues/950) on the topic so maybe it's just my brain being biased that we used to have issues with the old `--removable`.

h8d13:
https://www.reddit.com/r/arch/s/BGuxPuKNx8 (yesterday) 

I've seen quite a few in (since this has been the case the whole month) 

I also reproduced on quite recent msi board

Torxed:
Ah I'm not debating that users are experiencing issues when `--removable` is absent.
I just have a vague memory that when we added `--removable` in as a default, it caused more issues on the other side of the hardware spectrum.. I just can't find any issues linked to it.

So I'm worried that by adding back the old way of defaulting to `--removable` we'll be back in the old boat :sweat_smile:


tl;dr: I'm just nervous this brings back old bugs ^^

h8d13:
Ah makes sense. 

Perhaps just the help text being better is enough (its also documented on the grub page of wiki). 

And can default to False

Torxed:
I would appreciate that, as it would make users less confused - but make me sleep better at night :sweat_smile:

svartkanin:
It seems both may cause issues for some users if not set correctly for their hardware. 
In a perfect world we would be able to detect the hardware and pick the proper one but I think that may always lead to uncovered edge cases.

IMHO we should choose the default that represents the larger group of users and for whoever runs into the problem andd a FAQ entry in the README covering it.

archinstall/lib/args.py
```diff
@@ -185,7 +185,7 @@ def from_config(cls, args_config: dict[str, Any], args: Arguments) -> 'ArchConfi
 			uki = args_config.get('uki', False)
 			if uki and not bootloader.has_uki_support():
 				uki = False
-			arch_config.bootloader_config = BootloaderConfiguration(bootloader=bootloader, uki=uki, removable=False)
```
Torxed, line 188: Ok lets change this back to `False` for now.

archinstall/lib/bootloader/bootloader_menu.py
```diff
@@ -83,8 +83,8 @@ def _prev_uki(self, item: MenuItem) -> str | None:
 
 	def _prev_removable(self, item: MenuItem) -> str | None:
 		if item.value:
-			return tr('Will install to /EFI/BOOT/ (removable location)')
-		return tr('Will install to standard location with NVRAM entry')
+			return tr('Will install to /EFI/BOOT/ (removable location, safe default)')
```
Torxed, line 86: And we can keep the hints that help users choose `removable=True`!

archinstall/lib/models/bootloader.py
```diff
@@ -65,7 +65,7 @@ def from_arg(cls, bootloader: str, skip_boot: bool) -> Bootloader:
 class BootloaderConfiguration:
 	bootloader: Bootloader
 	uki: bool = False
-	removable: bool = False
```
Torxed, line 68: And keep `False` here.

archinstall/lib/models/bootloader.py
```diff
@@ -74,12 +74,14 @@ def json(self) -> dict[str, Any]:
 	def parse_arg(cls, config: dict[str, Any], skip_boot: bool) -> BootloaderConfiguration:
 		bootloader = Bootloader.from_arg(config.get('bootloader', ''), skip_boot)
 		uki = config.get('uki', False)
-		removable = config.get('removable', False)
```
Torxed, line 77: And keep `False` here.

Torxed (changes requested):
After these changes, let's get it in and we'll push a new minor release before the new year :)

Mintsuki:
I personally think that keeping it defaulting to true (as this PR already does) is probably for the best, unless there is an actual pop-up screen like Debian does that explains what the choice entails. Because whether one likes it or not, people are going to skip over a bunch of stuff and then land in support forums like the past month has shown.

Generally speaking, there *shouldn't really* be an issue with going back to the default that archinstall already used, and using the removable location *shouldn't really* cause issues, or if it does, it should be in a smaller proportion than the reverse.

Torxed:
If you think so, I'm happy to merge and see what happens ^^

PR #3974, opened 2025-12-07
Optimized the speed of retrieving mirrorlist

xhdndmm:
- This fix issue: 

## PR Description:
The waiting time and request timeout were limited based on the original code.
```python
	try:
        #Added timeout period
		mirrorlist = fetch_data_from_url(url, timeout=1)
		self._status_mappings = self._parse_remote_mirror_list(mirrorlist)
		return True
	except Exception as e:
		debug(f'Error while fetching mirror list: {e}')
		#The original wait was deleted.
		time.sleep(0.5)
```

## Tests and Checks
- [ ] I have tested the code!<br>

PR #4031, opened 2025-12-25
Fix mirrors hang when /status endpoint is down

### LINKED ISSUE #4029: Arch install stuck in mirror
So i installed arch using archinstall like I always do and it always end up normal, but this time it's stuck in the mirror and repositories when I enter. My laptop keeps showing Loading mirror regions and it didn't go away or show any mirrors.

h8d13:
Aims to fix #4029 or start a discussion about it (I've had this when the endpoint is down).

By assigning a bool to the def `load_remote_mirrors` in the parent (which was naïve about the state). 

This can be tested on the branch above and using something similar to block URL (temporarily):

`echo "127.0.0.1 archlinux.org" | sudo tee -a /etc/hosts` or when it's down like currently (15:00 UTC+1).

Second thing I'm wondering about: 

```python
  def get_status_by_region(self, region: str, speed_sort: bool) -> list[MirrorStatusEntryV3]:
      mappings = self._mappings()
      region_list = mappings[region]

      # Only sort if we have remote mirror data with score/speed info
      # Local mirrors lack this data and can be modified manually before-hand
      # Or reflector potentially ran already
      if self._fetched_remote and speed_sort:
          # original return
          return sorted(region_list, key=lambda mirror: (mirror.score, mirror.speed))
      # just return as-is without sorting?
      return region_list
``` 

Another day another DDoS ? I'll ping thsi for the funny lore too #3974

### DIFF
--- a/archinstall/lib/models/mirrors.py
+++ b/archinstall/lib/models/mirrors.py
@@ -105,10 +105,13 @@ class MirrorStatusEntryV3(BaseModel):
 
 	@model_validator(mode='after')
 	def debug_output(self) -> 'MirrorStatusEntryV3':
+		from ..args import arch_config_handler
+
 		self._hostname, *port = urllib.parse.urlparse(self.url).netloc.split(':', 1)
 		self._port = int(port[0]) if port and len(port) >= 1 else None
 
-		debug(f'Loaded mirror {self._hostname}' + (f' with current score of {self.score}' if self.score else ''))
+		if arch_config_handler.args.verbose:
+			debug(f'Loaded mirror {self._hostname}' + (f' with current score of {self.score}' if self.score else ''))
 		return self
 
 


h8d13:
Also had fun on another branch where we can use mirror endpoint from archlinux.de (or other models?) 

https://github.com/archlinux/archinstall/compare/master...h8d13:archinstall-patch:refs/heads/dot-mirrors-de

Torxed:
We could also fall back to `/etc/pacman.d/mirrorlist` as it should contain a new enough mirror list when we can't use the online lookup.

And the regions are "there" in the shape of comments, so we'd just need to parse "which region are we currently on" and all subsequent mirror entries (even if commented out) belong to that region.

h8d13:
It is already the fallback behavior problem was that timeout didn't work and the parent logic didn't know about the child status

The extra options are totally optional and can be cherry picked if you don't want them (today's commits)

Torxed:
Ah, I think I'm slowly waking up and getting the hang of this.

So would the `--offline` flag be sufficient instead of the following, or do we need the fine tuning of arguments and code logic: https://github.com/archlinux/archinstall/blob/421892bade1be654ce0380ee8501cceccf51b308/archinstall/lib/args.py#L397-L402

I'm not against it, just wondering if `--offline` could serve well here, as it would help with other online issues too.

archinstall/lib/networking.py
```diff
@@ -121,7 +121,7 @@ def enrich_iface_types(interfaces: list[str]) -> dict[str, str]:
 	return result
 
 
-def fetch_data_from_url(url: str, params: dict[str, str] | None = None) -> str:
+def fetch_data_from_url(url: str, params: dict[str, str] | None = None, timeout: int = 3) -> str:
```
Torxed: The default timeout can probably be increased to `15` or even `20`.
As I know some regions struggle with latency and speeds.

Torxed:
> Ah, I think I'm slowly waking up and getting the hang of this.
> 
> So would the `--offline` flag be sufficient instead of the following, or do we need the fine tuning of arguments and code logic:
> 
> https://github.com/archlinux/archinstall/blob/421892bade1be654ce0380ee8501cceccf51b308/archinstall/lib/args.py#L397-L402
> 
> I'm not against it, just wondering if `--offline` could serve well here, as it would help with other online issues too.

Also regarding this, the ISO should already be running `reflector.service` as a pre-cursor that `archinstall` waits for, so we shouldn't need to run it again?

But we could perhaps re-name the `--local-mirr` to `--skip-reflector`/`--skip-mirrors`? And keep it, but make it so this: https://github.com/archlinux/archinstall/blob/5fcea379b9cd1f65343056dcfd70a441dc0ab744/archinstall/lib/installer.py#L197-L199

Behaves like this: https://github.com/archlinux/archinstall/blob/5fcea379b9cd1f65343056dcfd70a441dc0ab744/archinstall/lib/installer.py#L180-L181

archinstall/lib/mirrors.py
```diff
@@ -404,12 +405,18 @@ def get_mirror_regions(self) -> list[MirrorRegion]:
...
 
 	def load_mirrors(self) -> None:
+		# Skip if already loaded
+		if self._status_mappings is not None:
+			return
+
 		from .args import arch_config_handler
 
-		if arch_config_handler.args.offline:
+		if arch_config_handler.args.local_mirr or arch_config_handler.args.offline:
```
h8d13: @Torxed We check here `--offline`

archinstall/lib/networking.py
```diff
@@ -121,7 +121,7 @@ def enrich_iface_types(interfaces: list[str]) -> dict[str, str]:
 	return result
 
 
-def fetch_data_from_url(url: str, params: dict[str, str] | None = None) -> str:
+def fetch_data_from_url(url: str, params: dict[str, str] | None = None, timeout: int = 20) -> str:
```
h8d13: The issue with 20 is that it's x3 + incremental if i understand correctly 

From def `load_remote_mirrors`

archinstall/lib/installer.py
```diff
@@ -194,9 +194,12 @@ def _verify_service_stop(self) -> None:
 		else:
 			info(tr('Skipping waiting for automatic time sync (this can cause issues if time is out of sync during installation)'))
 
-		info('Waiting for automatic mirror selection (reflector) to complete.')
-		while self._service_state('reflector') not in ('dead', 'failed', 'exited'):
-			time.sleep(1)
+		if not arch_config_handler.args.offline:
```
h8d13, line 197: As so ? @Torxed  or should this also check for `--skip-mirrors`

archinstall/lib/networking.py
Torxed: That is true, but currently in a spot in Germany with heavy load I will not be able to access the mirror list in 3 seconds even with 3 attempts. And I know the same will be true in a lot of south american places.

(aside from the DDoS going stronk heh).

So It's true that we'll be doing 3 attempts: https://github.com/archlinux/archinstall/blob/5fcea379b9cd1f65343056dcfd70a441dc0ab744/archinstall/lib/mirrors.py#L417-L421

But all those three attempts will need pretty generous timeouts :) So the 20 sec looks good! You've learned the code base pretty well I must say :)

h8d13: Oh btw I also made a funny patch since `.de` is unaffected by DDoS 

https://github.com/archlinux/archinstall/compare/master...h8d13:archinstall-patch:refs/heads/dot-mirrors-de

and thanks, i enjoy trying to fix things for users here it's fun

archinstall/lib/installer.py
Torxed: I like this, but we could perhaps create and use a `--skip-mirrors` after-all, as it will make here like with the others right next to it?

The `--offline` should ideally block anything and everything that tries to touch the outside world (wkd sync, mirrors, new version check etc) now that I think about it.

Sorry for going back and fourth on this, it just clicked when I saw the option next to the others, that it should probably be called `--skip-<something>`.

h8d13: Ah yeah already did the renames for both user facing arg and `skip_mirrors: Bool`

h8d13:
Imma test an install just for sanity check :d 

Both tests pass with and without the new arg, and expected behaviors (one hanging a bit then going through, and the other skipping instantly both issues). 🕺

archinstall/lib/networking.py
Torxed: Because this is an official project, I don't think we can use third party sites without their consent/approval, and should generally try to avoid going outside `archlinux.org` for anything as it will look weird ^^

And it will inevitably lead to the DDoS spreading there when they find out we use that site, and then they're in trouble too heh.

archinstall/lib/mirrors.py
```diff
@@ -404,12 +405,18 @@ def get_mirror_regions(self) -> list[MirrorRegion]:
 		return available_mirrors
 
 	def load_mirrors(self) -> None:
+		# Skip if already loaded
+		if self._status_mappings is not None:
```
svartkanin: I don't think this function should terminate, rather the calling side should conditionally call it so that any consumers can decide on their own. 
In this case only this call should be guarded https://github.com/archlinux/archinstall/blob/1faac77c0d476165a9dee1616274c1a2749c49db/archinstall/lib/mirrors.py?plain=1#L307

archinstall/lib/args.py
```diff
@@ -392,6 +393,12 @@ def _define_arguments(self) -> ArgumentParser:
 			default=False,
 			help='Disabled online upstream services such as package search and key-ring auto update.',
 		)
+		parser.add_argument(
```
svartkanin: I'm somewhat against introducing a new flag to specifically handle the mirror fetching failures, if this continues we'll have a flag for every menu entry which isn't sustainable and very confusing for users. 

I'd rather recommend to handle this gracefully in the menu and provide the users with options if possible or just fall back to the local config.

archinstall/lib/installer.py
```diff
@@ -194,9 +194,12 @@ def _verify_service_stop(self) -> None:
 		else:
 			info(tr('Skipping waiting for automatic time sync (this can cause issues if time is out of sync during installation)'))
 
-		info('Waiting for automatic mirror selection (reflector) to complete.')
-		while self._service_state('reflector') not in ('dead', 'failed', 'exited'):
-			time.sleep(1)
+		if not arch_config_handler.args.offline or arch_config_handler.args.skip_mirrors:
+			info('Waiting for automatic mirror selection (reflector) to complete.')
+			while self._service_state('reflector') not in ('dead', 'failed', 'exited'):
+				time.sleep(1)
```
svartkanin, line 202: Maybe this should terminate at some point to avoid infinite waiting

archinstall/lib/mirrors.py
```diff
@@ -404,12 +405,18 @@ def get_mirror_regions(self) -> list[MirrorRegion]:
...
 		from .args import arch_config_handler
 
-		if arch_config_handler.args.offline:
+		if arch_config_handler.args.skip_mirrors or arch_config_handler.args.offline:
+			self._fetched_remote = False
 			self.load_local_mirrors()
 		else:
-			if not self.load_remote_mirrors():
+			self._fetched_remote = self.load_remote_mirrors()
+			if not self._fetched_remote:
```
svartkanin, line 416: add a debug here to track this

archinstall/lib/networking.py
svartkanin: 20 seems a bit odd, shall we do 30 seconds as that's more like a standard timeout?

archinstall/lib/args.py
Torxed: That works too. As long as we have a way to deal with these DDoS issues, otherwise archinstall is "useless" as it won't be able to pass the mirror stage. 

So currently these `--skip-???` are needed for environments where NTP is blocked, GPG is blocked or mirror generators are unaccessible (reflector + our /status logic, both relies on the same DDoS'ed endpoint).

After that I don't think we have more `--skip-???` flags needed as those are the online resources currently blocking progress.

And the menu would need to internally handle and det a skip feature to circumvent the issues.

archinstall/lib/installer.py
Torxed: Good idea!

archinstall/lib/networking.py
Torxed: Sounds reasonable ^^

archinstall/lib/installer.py
h8d13: I set it to 60, lmk if that's too short
And your other feedbacks ! 
Removed the new flag (reverted to 2 days ago) + changes

archinstall/lib/args.py
svartkanin: Then instead of having the flag would it make sense to have a "failed to fetch mirrors, fallback to locale?" Yes/no?

h8d13: @svartkanin i removed it... see files changed based on both of your feedbacks

h8d13:
Its been happening one or twice per month now, so I'd consider fixing this more priority, you can sleep better knowing the installer works regardless of initial fetch. (along with the NVIDIA changes). If there is still anything with the code that needs change lmk

svartkanin:
Seems to work fine

Issue #4035, opened 2025-12-26
Ask on startup to resume/reuse "old" config as seed for menu


Torxed:
### describe the request

It might be a good idea if the menu system detects `/var/log/archinstall/user_configuration.json`, and asks if the user wants to continue/reuse that configuration or start over.

For instance in a crash where the user made a mistake somewhere, they wouldn't need to re-do everything.

It's a small thing, but perhaps that would make more use of the automatically created user config.

h8d13:
Can be tested `git clone -b dot-resume https://github.com/h8d13/archinstall-patch/`

svartkanin:
I think this is probably something for developers rather than users, so if we're happy with it I'd suggest to put this behind the `--debug` flag. 

Also it's not self-contained as the credentials file will not be saved (which I hope we won't for security reasons :)).

h8d13:
This should be ready for review again and in a separate file now :)

PR #4036, opened 2025-12-26
Resume/abort

h8d13:
#4035 draft (comes from my dev fork)

This would be especially good for testing ;)

### DIFF
--- a/archinstall/lib/configuration.py
+++ b/archinstall/lib/configuration.py
@@ -2,6 +2,7 @@ import json
 import readline
 import stat
 from pathlib import Path
+from typing import Any
 
 from archinstall.lib.translationhandler import tr
 from archinstall.tui.curses_menu import SelectMenu, Tui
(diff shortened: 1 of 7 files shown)


Torxed:
Well holy crap that was fast :sweat_smile:

h8d13:
Comes from a fork so this is passage 2. Also my logic was bad your pre-commit checks help a lot
```
  ⎿  c848f947 (h8d13 2025-11-14 09:44:17 +0100 292) def load_saved_config() -> dict | None:                     
     c848f947 (h8d13 2025-11-14 09:44:17 +0100 293)     """Load saved config and credentials from hade_box folder"""
```

h8d13:
I'll mark this as ready as the logic looks sound and feels correct on testing branch

Torxed:
Thoughts on this one @svartkanin? :)

*(Aside from having to move it to the new TUI in #3997)*

archinstall/lib/configuration.py
```diff
@@ -242,3 +243,63 @@ def preview(item: MenuItem) -> str | None:
 			config_output.save_user_creds(dest_path, password=enc_password)
 		case 'all':
 			config_output.save(dest_path, creds=True, password=enc_password)
+
+
+def has_saved_config() -> bool:
+	"""Check if there's a saved config in /var/log/archinstall"""
+	config_file = logger.directory / 'user_configuration.json'
```
svartkanin: the file names are defined here https://github.com/archlinux/archinstall/blob/1faac77c0d476165a9dee1616274c1a2749c49db/archinstall/lib/configuration.py?plain=1#L32, to make sure any future changes are easy we should define them once and reuse a variable instead of c/p

archinstall/scripts/guided.py
```diff
@@ -18,11 +18,59 @@
...
 from archinstall.lib.profile.profiles_handler import profile_handler
 from archinstall.lib.translationhandler import tr
 from archinstall.tui import Tui
+from archinstall.tui.curses_menu import SelectMenu
+from archinstall.tui.menu_item import MenuItem, MenuItemGroup
+from archinstall.tui.result import ResultType
+from archinstall.tui.types import Alignment
+
+
+def _check_for_saved_config() -> None:
```
svartkanin: If we want this feature it should be usable in all the provided scripts that ship, so this should go into a generic file so it can be re-used by any script.

archinstall/lib/configuration.py
```diff
@@ -242,3 +243,63 @@ def preview(item: MenuItem) -> str | None:
 			config_output.save_user_creds(dest_path, password=enc_password)
 		case 'all':
 			config_output.save(dest_path, creds=True, password=enc_password)
+
+
+def has_saved_config() -> bool:
```
svartkanin: Can we move these into the `ConfigurationOutput` class and Maybe rename that thing to `ConfigurationHandler` it will make future refactoring much easier

svartkanin:
In addition to the comments, I can see some benefits for it but also it's not fully self-contained for regular users as the `user_credentials.json` is not saved and so a full "automatic" re-installation cannot be done. 
That does not mean that I'm suggesting to store the credentials file as that would be quite bad from a security perspective. 

If the intention with this is to "be easier during dev" I'd suggest to put this behind the `--debug` flag which we have already

Torxed:
> In addition to the comments, I can see some benefits for it but also it's not fully self-contained for regular users as the `user_credentials.json` is not saved and so a full "automatic" re-installation cannot be done. That does not mean that I'm suggesting to store the credentials file as that would be quite bad from a security perspective.
> 
> If the intention with this is to "be easier during dev" I'd suggest to put this behind the `--debug` flag which we have already

I agree, skipping credentials is good and if that's the only step when re-running it still save time. 

And putting it behind `--debug` makes sense for now if nothing else :)

archinstall/scripts/guided.py
h8d13: I think it only makes sense for guided ? I integrated it to the Class and renamed as you asked
that already makes it re-usable. There are also safeguards for `silent` and check that this only happens in `--debug`

Altho it does change the exit behavior slightly with an extra `Abort` / `Cancel` which isn't necessarily a bad thing

h8d13 (commented):
Address feedback rounds from torxed/svart

Do lmk I'd more changes are needed

h8d13:
It's now hidden behind `--debug` and changes based on feedback. (In a seperate lib file) 

Also tested a little, and is really helpful to be able to test stuff faster.

archinstall/scripts/guided.py
svartkanin: I think the functionality makes sense to expose via an API in a lib file rather than chucking this into the script layer. 

The script should be as lightweight as possible to call lib APIs to perform things

h8d13: @svartkanin made a new file `resumehandler.py` and move all the logic there

h8d13:
This conversation would still be nice to re-open for dev :)

PR #4037, opened 2025-12-26
Add rescue mode script

h8d13:
This morning I deleted `sudo` by accident playing around with `--noconfirm` scripts. Specifically wanted to build a parser that would support same format as `-r requirements.txt` but for `pacman` and it went wrong in the `-Rns` category lol. 

Anyways I thought it would be good to have a utility that can save your ass from ISO or other install. Obviously isn't perfect but I tried to re-use as much as possible from existing menus. And it showed my root partition selected it ran `pacman -S sudo`, rebooted and was done.

### DIFF
--- a/archinstall/scripts/rescue.py
+++ b/archinstall/scripts/rescue.py
@@ -69,31 +69,6 @@ def mount_partition(partition: LsblkInfo, mount_point: Path) -> bool:
 		return False
 
 
-def mount_special_filesystems(mount_point: Path) -> bool:
-	"""Mount special filesystems needed for chroot (/dev, /proc, /sys, /run)."""
-	special_mounts = [
-		('/dev', 'dev', ['--bind']),
-		('/proc', 'proc', ['--bind']),
-		('/sys', 'sys', ['--bind']),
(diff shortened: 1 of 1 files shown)


svartkanin:
We had some form of "helper" scripts before with `swiss.py` which was supposed to be more for dev. It is extremely painful to maintain and keep up to date and I'm not to keen on getting back into that as it's extremely time sonsuming with limited testing options. 

This seems more like a personal thing to hack around and I'm not sure there's much value including it into the repo itself

h8d13:
I think it would be invaluable. The amount of times i see people on forums saying "just re-install". Just because they do not have a mount/chroot script. This would explicitly makes that available.

I do see the issue with maintaining, but if we re-use from the code-base as much as possible and simply use `arch-chroot` it kinda updates itself. It's also why you made this structure if I'm not mistaken scripts are supposed to help also manage existing installations.  

Just wanted to show a PoC that actually helped me fix my stuff, so yes personal, but hell why not?

h8d13:
Imma close this. I created a repo here https://github.com/h8d13/archinstall-contrib to inject it at runtime.

PR #4042, opened 2025-12-28
Feat: Zram algorithm config

h8d13:
Continuation to PR #4027 

Would allow to set the algo of zram directly. Available ones can be checked using 

`cat /sys/block/zram0/comp_algorithm`

We could however extend this to support re-compression as seen on the [wiki](https://wiki.archlinux.org/title/Zram)

```
/etc/systemd/zram-generator.conf

[zram0]
compression-algorithm = lzo-rle zstd(level=3) (type=idle)
``` 

But for now already giving an option looks better.

https://github.com/user-attachments/assets/17441320-f9fa-4114-ab2c-12c8dbdba75d

This also fixes a quirk where the default highlighted value was "no" causing an extra interaction for sane default.

### DIFF
--- a/archinstall/lib/global_menu.py
+++ b/archinstall/lib/global_menu.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from typing import override
 
 from archinstall.lib.disk.disk_menu import DiskLayoutConfigurationMenu
-from archinstall.lib.models.application import ApplicationConfiguration
+from archinstall.lib.models.application import ApplicationConfiguration, ZramConfiguration
 from archinstall.lib.models.authentication import AuthenticationConfiguration
 from archinstall.lib.models.device import DiskLayoutConfiguration, DiskLayoutType, EncryptionType, FilesystemType, PartitionModification
 from archinstall.lib.packages import list_available_packages
@@ -80,7 +80,7 @@ class GlobalMenu(AbstractMenu[None]):
 			),
 			MenuItem(
 				text=tr('Swap'),
-				value=True,
+				value=ZramConfiguration(enabled=True),
 				action=ask_for_swap,
 				preview_action=self._prev_swap,
 				key='swap',


archinstall/lib/args.py
```diff
@@ -72,7 +72,7 @@ class ArchConfig:
 	ntp: bool = True
 	packages: list[str] = field(default_factory=list)
 	parallel_downloads: int = 0
-	swap: bool = True
+	swap: bool | dict[str, Any] = True
```
Torxed: Perhaps we could swap `Any` here if possible.

h8d13: `dict[str, str]` ?

Torxed: Yepp that looks correct *(I say just looking at the code, haven't tested it)*

h8d13: Will also test later tonight :+1: 

 I like this (would) make the option in menu more useful than yes/no :D

svartkanin: Lets use a type instead of the entire dict. Ideally also temove the bool and when parsing the config auto transform old bool vales into the type with defaults

archinstall/lib/args.py
```diff
@@ -72,7 +72,7 @@ class ArchConfig:
 	ntp: bool = True
 	packages: list[str] = field(default_factory=list)
 	parallel_downloads: int = 0
-	swap: bool = True
+	swap: bool | dict[str, str] = True
```
svartkanin: This should be a actual type not dict as we moved away from raw types completely and they are not type safe. 
The bool should also be removed and during the config parsing transformed into the new type with the default values, similar to the audio config parsing https://github.com/svartkanin/archinstall/blob/c0b3abd88fba6fbc407916394a20c94abb9e25bd/archinstall/lib/args.py?plain=1#L189-L194

svartkanin (changes requested):

archinstall/lib/global_menu.py
```diff
@@ -372,7 +372,12 @@ def _prev_disk_config(self, item: MenuItem) -> str | None:
 	def _prev_swap(self, item: MenuItem) -> str | None:
 		if item.value is not None:
 			output = f'{tr("Swap on zram")}: '
-			output += tr('Enabled') if item.value else tr('Disabled')
+			if isinstance(item.value, dict):
```
svartkanin: with the mentioned new type this will not be necessary anymore

archinstall/lib/interactions/system_conf.py
```diff
@@ -113,6 +118,37 @@ def ask_for_swap(preset: bool = True) -> bool:
...
+			enabled = result.item() == MenuItem.yes()
+			if not enabled:
+				return False
+
+			# Ask for compression algorithm
+			algo_preset = 'zstd'
+			if isinstance(preset, dict):
+				algo_preset = preset.get('algo', 'zstd')
+
+			algorithms = ['zstd', 'lzo-rle', 'lzo', 'lz4', 'lz4hc']
```
svartkanin: These algorithms should be put in an Enum and used in the new zram type

archinstall/lib/args.py
```diff
@@ -72,7 +72,7 @@ class ArchConfig:
 	ntp: bool = True
 	packages: list[str] = field(default_factory=list)
 	parallel_downloads: int = 0
-	swap: bool = True
+	swap: ZramConfiguration = field(default_factory=lambda: ZramConfiguration(enabled=True))
```
svartkanin: No need to use `field` here as that was only for the `dict`, you can assign the `ZramConfiguration(true)` directly

archinstall/lib/args.py
```diff
@@ -211,7 +211,15 @@ def from_config(cls, args_config: dict[str, Any], args: Arguments) -> 'ArchConfi
 		if parallel_downloads := args_config.get('parallel_downloads', 0):
 			arch_config.parallel_downloads = parallel_downloads
 
-		arch_config.swap = args_config.get('swap', True)
+		# Parse swap config - transform bool/dict into ZramConfiguration
+		swap_arg = args_config.get('swap', True)
```
svartkanin: Lets change the fallback for the `get` to the new type
```
swap_arg = args_config.get('swap', ZramConfituration(enabled=true))
```

archinstall/lib/args.py
```diff
@@ -211,7 +211,15 @@ def from_config(cls, args_config: dict[str, Any], args: Arguments) -> 'ArchConfi
...
+		if isinstance(swap_arg, bool):
+			arch_config.swap = ZramConfiguration(enabled=swap_arg)
+		elif isinstance(swap_arg, dict):
+			algo = swap_arg.get('algo', ZramAlgorithm.ZSTD.value)
```
svartkanin: Can we move this into a `parse_args` function like all the others, example https://github.com/archlinux/archinstall/blob/5fcea379b9cd1f65343056dcfd70a441dc0ab744/archinstall/lib/models/application.py?plain=1#L62

archinstall/lib/installer.py
```diff
@@ -990,7 +990,7 @@ def setup_btrfs_snapshot(
 			self._configure_grub_btrfsd(snapshot_type)
 			self.enable_service('grub-btrfsd.service')
 
-	def setup_swap(self, kind: str = 'zram') -> None:
+	def setup_swap(self, kind: str = 'zram', algo: str = 'zstd') -> None:
```
svartkanin: `algo` should be type `ZramAlgorithm`

archinstall/lib/interactions/system_conf.py
```diff
@@ -113,6 +110,32 @@ def ask_for_swap(preset: bool = True) -> bool:
...
 			return preset
 		case ResultType.Selection:
-			return result.item() == MenuItem.yes()
+			enabled = result.item() == MenuItem.yes()
+			if not enabled:
+				return ZramConfiguration(enabled=False)
+
+			# Ask for compression algorithm
+			algo_items = [MenuItem(a.value, value=a) for a in ZramAlgorithm]
+			algo_group = MenuItemGroup(algo_items, sort_items=False)
```
svartkanin: You can use
```
group = MenuItemGroup.from_enum(ZramAlgorithm)
```

archinstall/lib/interactions/system_conf.py
```diff
@@ -113,6 +110,32 @@ def ask_for_swap(preset: bool = True) -> bool:
...
+				allow_skip=True,
+			).run()
+
+			algo: ZramAlgorithm
+			match algo_result.type_:
+				case ResultType.Skip:
+					algo = preset.algorithm
+				case ResultType.Selection:
+					algo = algo_result.get_value()
+				case ResultType.Reset:
```
svartkanin: Reset is not enabled in the `SelectMenu` (`allow_reset` is default to false) so this is unnecessary

examples/config-sample.json
```diff
@@ -146,7 +146,7 @@
   },
   "script": "guided",
   "silent": false,
-  "swap": true,
+  "swap": {"algo": "zstd"},
```
svartkanin: This needs to be updated with the right output

h8d13:
Sorry about the messy commits, learning about the full structure still. 

Really do appreciate all the feedback from both of you. I will do more testing today. If you have the chance to look at the LVM branch I'd be really grateful too.

Torxed:
> Sorry about the messy commits, learning about the full structure still.

No worries, we squish commits to one right before merging - so it'll look like one commit in the end :)

tests/test_args.py
```diff
@@ -225,7 +232,7 @@ def test_config_file_parsing(
 		ntp=True,
 		packages=['firefox'],
 		parallel_downloads=66,
-		swap=False,
+		swap=ZramConfiguration(enabled=False),
```
svartkanin, line 235: This should probably be test the `true` case with a proper algorithm

Issue #4046, opened 2025-12-28
Mirrors sort


h8d13:
```
[2025-12-28 16:12:49] - DEBUG - Checking download speed of mirror.moson.org[0.10946855478612] by fetching: http://mirror.moson.org/arch/core/os/x86_64/core.db
[2025-12-28 16:12:50] - DEBUG -     speed: 75406.25530018455 (0.07MiB/s)
[2025-12-28 16:12:50] - DEBUG - Checking download speed of mirror.moson.org[0.16033788096610865] by fetching: https://mirror.moson.org/arch/core/os/x86_64/core.db
[2025-12-28 16:12:52] - DEBUG -     speed: 96603.7381084319 (0.09MiB/s)
```

```
## Germany
Server = http://mirror.moson.org/arch/$repo/os/$arch
Server = http://arch.jensgutermuth.de/$repo/os/$arch
Server = http://mirror.ubrco.de/archlinux/$repo/os/$arch
Server = https://berlin.mirror.pkgbuild.com/$repo/os/$arch
Server = https://mirror.moson.org/arch/$repo/os/$arch
```

Huh bit confused about this one did it pick the slowest mirrors?

Issue #4045, opened 2025-12-28
lvm + btrfs doesn't mount correctly before pacstrap


Torxed:
A few users have reported issues with btrfs on lvm.
And I managed to reproduce it by:

1. Partitioning -> use best effort -> select disk
2. Partitioning -> LVM -> btrfs -> yes to subvolumes -> compression

<img width="1280" height="747" alt="Image" src="https://github.com/user-attachments/assets/a258323d-36c2-46cb-b70e-9ba238892429" />

<img width="1280" height="747" alt="Image" src="https://github.com/user-attachments/assets/3beeb5f1-6eb3-4386-9bff-b3139a41cb9a" />

So we're "running out of disk space" because stuff isn't mounted under `/mnt`, but the disk setup was created.

h8d13:
Can you do me a favor and test with encryption activated ?

https://github.com/h8d13/archinstall-patch/commit/42af1fdb29c47e839d22e8754bc663fc97e4b45e

(Whoops committed a bunch of other stuff) but going to take a break for today

Torxed:
Yepp, with disk encryption it appears to be working *(luks + lvm + btrfs)*.

h8d13:
So that should fix it ^^ https://github.com/h8d13/archinstall-patch/commit/42af1fdb29c47e839d22e8754bc663fc97e4b45e

PR #4047, opened 2025-12-28
Lvm2/LUKS fixes/Mirror Logic

h8d13:
Aims to address #4045  #4046

I then did sanity check with 3rd attempt without encryption same settings LVM BTRFS, and got "signature exists" errors (this why i tried the `wipefs` call).

(That being said/done) For real clean-ups we should have kind of like traps in bash (especially test this running from a host, since you'd need to check the devices are not still there `/dev/ArchinstallVg/root`  (which i believe they are, from earlier testing my main rig) And should really happen at the very end (even after synchronizing disks).

image

Tested LVM against both of these and without encryption

### DIFF
--- a/archinstall/lib/mirrors.py
+++ b/archinstall/lib/mirrors.py
@@ -450,8 +450,8 @@ class MirrorListHandler:
 		# Local mirrors lack this data and can be modified manually before-hand
 		# Or reflector potentially ran already
 		if self._fetched_remote and speed_sort:
-			# Sort by speed descending (higher is better) - fastest mirrors first
-			return sorted(region_list, key=lambda mirror: -mirror.speed)
+			# original return
+			return sorted(region_list, key=lambda mirror: (mirror.score, mirror.speed))
 		# just return as-is without sorting?
 		return region_list
 


h8d13:
I might actually have introduced the new bug when using `while`. When trying to address #3424 , so I apologize for the mistake. So for this one I did empirical testing: Testing all 3: BTRFS/LVM/LUKS2(on/off).  

Please feel free to test as-well `
git clone -b dot-hotfix-lvm https://github.com/h8d13/archinstall-patch
`  

I'll mark the PR as open/ready for review. But obviously this should be addressed further.

archinstall/lib/mirrors.py
```diff
@@ -450,8 +450,8 @@ def get_status_by_region(self, region: str, speed_sort: bool) -> list[MirrorStat
 		# Local mirrors lack this data and can be modified manually before-hand
 		# Or reflector potentially ran already
 		if self._fetched_remote and speed_sort:
-			# original return
-			return sorted(region_list, key=lambda mirror: (mirror.score, mirror.speed))
+			# Sort by speed descending (higher is better) - fastest mirrors first
+			return sorted(region_list, key=lambda mirror: -mirror.speed)
```
Torxed: Could we move this to a separate PR? Just to not fix two things in one, in case one needs to be reverted :)

h8d13: Sry bout that was frustrated at mirrors when testing

Torxed: 10/10 understand :sweat_smile:

PR #4048, opened 2025-12-29
Modify archinstall language display to be handled like other sections.

h8d13:
Screenshot_20251229_113905 NEW FORMAT

Screenshot_20251229_113941 OLD FORMAT

Basically the value of `lang.display_name` was breaking the menu when too long. So just changed it to be handled like other sections.

### DIFF
--- a/archinstall/tui/types.py
+++ b/archinstall/tui/types.py
@@ -115,7 +115,7 @@ class Chars:
 	Block = '█'
 	Triangle_up = '▲'
 	Triangle_down = '▼'
-	Check = '✓'
+	Check = '+'
 	Cross = 'x'
 	Right_arrow = '←'
 


h8d13:
@svartkanin any thoughts on this one, I know these are fixed in textual but still in the meantime not a bad idea to make some UI enhancements

svartkanin:
I absolutely hate Githubs mobile app it's abysmal as it doesn't seem to be able to even add comments anymore.

So I'll put it here:

This utf character isn't supported in the ISO


On another note, you need to be a bit patient with getting reviews, there's no ETA when they will happen as it's an open source project. We'll get to it when we get to it :) this may be a day, a week or even longer.

h8d13:
No worries sorry about that I'm quite hyped at finally sending in all my stuff :D Also I agree lmao (about github mobile)

svartkanin (approved):

PR #4049, opened 2025-12-29
Make tz selection deliberate

h8d13:
Instead of over complex like I tried in #4002

This is a much simpler approach we set the fallback in `args.py` and make it mandatory in the menu. 
Again this to address lots of logs that get PGP verif errors because TZ was not set.

### DIFF
--- a/archinstall/lib/args.py
+++ b/archinstall/lib/args.py
@@ -213,7 +213,8 @@ class ArchConfig:
 
 		arch_config.swap = args_config.get('swap', True)
 
-		arch_config.timezone = args_config.get('timezone', 'UTC')
+		if 'timezone' in args_config:
+			arch_config.timezone = args_config.get('timezone', 'UTC')
 
 		if services := args_config.get('services', []):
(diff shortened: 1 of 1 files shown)


Torxed:
I'm not sure I understand what the original problem was?
When I'm lazy I usually just stick with the default UTC and it works fine, unless the hardware clock is way off - but that's more an issue of it being off than what time zone I picked?

h8d13:
Just because I see of lot of error logs than yeah its hardware clock (which also gets reset on dead CMOS), so i thought making it explicit choice + showing RTC clock originally (previous PR #4002)

Torxed:
hmm, well `NTP` should make stuff "just work", as it should happen before archinstall starts strapping in packages. And a timezone shouldn't be needed?

h8d13:
My thought was that I could be dead (the service, happens sometimes) or disconnected for offline installs with cached files, so time being wrong there would cause pgp errors.

But yeah I do see how it's technically a non-problem for regular usage.

PR #4052, opened 2025-12-29
lib/network: adds symlink to configure systemd-resovled properly

wahrwolf:
When "Copy network config from ISO" is selected, it sets up `systemd-resolved` and `systemd-networkd`, however, `systemd-resolved` requires `resolve.conf` to be a [stub](https://wiki.archlinux.org/title/Systemd-resolved#DNS) in order to work properly.

This PR addresses the issue that it's currently not a stub, but a core filesystem file meant for other types of resolvers.

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -754,8 +754,13 @@ class Installer:
 				for psk in psk_files:
 					shutil.copy2(psk, f'{self.target}/var/lib/iwd/{os.path.basename(psk)}')
 
-		# Add symlink to enable systemd-resolved
-		os.symlink('/run/systemd/resolve/stub-resolv.conf', f'{self.target}/etc/resolv.conf')
+		# Enable systemd-resolved by (forcefully) setting a symlink
+		# For further details see  https://wiki.archlinux.org/title/Systemd-resolved#DNS
+		resolv_config_path = Path(f'{self.target}/etc/resolv.conf')
+		if resolv_config_path.exists():
+			os.unlink(resolv_config_path)
+		os.symlink('/run/systemd/resolve/stub-resolv.conf', resolv_config_path)
+
 		# Copy (if any) systemd-networkd config files
 		if netconfigurations := glob.glob('/etc/systemd/network/*'):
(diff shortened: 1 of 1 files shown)


svartkanin:
Can you provide a description of the PR and the problem

Torxed:
> Can you provide a description of the PR and the problem

I filled in the blanks as I forced poor @wahrwolf to submit this PR :sweat_smile: Sorry for the confusion!

wahrwolf:
I tested the change and after the first reboot ```resolvectl status``` confirms the stub-resolver

Here is the weird part:
From the iso live system the symlink looks fine.
However if you chroot into the system after the archinstall script finished, the symlinks is not correctly displayed.
Both file and ls detect it as a local ascii file.
Tbh I do not understand it and it hints to something weird with os.symlink, the arch-chroot script, ext4 or all together.

Everything looks normal after the reboot, so I assume this can be safely ignored

archinstall/lib/installer.py
```diff
@@ -754,6 +754,13 @@ def post_install_enable_iwd_service(*args: str, **kwargs: str) -> None:
 				for psk in psk_files:
 					shutil.copy2(psk, f'{self.target}/var/lib/iwd/{os.path.basename(psk)}')
 
+		# Enable systemd-resolved by (forcefully) setting a symlink
+		# For further details see  https://wiki.archlinux.org/title/Systemd-resolved#DNS
+		resolv_config_path = Path(f'{self.target}/etc/resolv.conf')
+		if resolv_config_path.exists():
+			os.unlink(resolv_config_path)
+		os.symlink('/run/systemd/resolve/stub-resolv.conf', resolv_config_path)
```
gdamjan, line 762: this is actually unnecessary, since the first boot would've created the file anyway (when it doesn't exist).

h8d13: Im guessing it should only be handled if chrooting into the (pre)install then ?

Torxed: > this is actually unnecessary, since the first boot would've created the file anyway (when it doesn't exist).

The problem is that the core package/filesystem contains the file, so it won't be created on the first boot. And it's better to be explicit here I think.

PR #4053, opened 2025-12-30
fix(hardware): bring nvidia driver installation in line with packaging changes

ghost:
Title.

I ended up going with the simpler solution proposed in the earlier pull request (#4019)

### DIFF
--- a/archinstall/lib/models/profile.py
+++ b/archinstall/lib/models/profile.py
@@ -42,13 +42,20 @@ class ProfileConfiguration:
 
 		_gfx_driver: GfxDriver | None = None
 		if gfx_driver:
-			try:
-				_gfx_driver = GfxDriver(gfx_driver)
-			except Exception:
-				_gfx_driver = GfxDriver[gfx_driver]
+			# Note: This is for backwards compatability with older configs.
+			# We fall back to the open-source nouveau driver here because if
+			# we end up installing the open kernel modules on a machine with pre-Turing
+			# hardware, the user will end up with a broken install (unresponsive black screen).
+			if gfx_driver == 'Nvidia (proprietary)':
+				_gfx_driver = GfxDriver.NvidiaOpenSource
+			else:
+				try:
+					_gfx_driver = GfxDriver(gfx_driver)
+				except Exception:
+					_gfx_driver = GfxDriver[gfx_driver]
 
 		return ProfileConfiguration(
 			profile,
-			_gfx_driver if _gfx_driver else None,
+			_gfx_driver,
 			GreeterType(greeter) if greeter else None,
 		)


archinstall/lib/models/profile.py
```diff
@@ -40,8 +40,22 @@ def parse_arg(cls, arg: _ProfileConfigurationSerialization) -> 'ProfileConfigura
...
 		gfx_driver = arg.get('gfx_driver', None)
 
+		_gfx_driver: GfxDriver | None = None
+		if gfx_driver:
+			# Note: This is for backwards compatability with older configs.
+			# We fall back to the open-source nouveau driver here because if
+			# we end up installing the open kernel modules on a machine with pre-Turing
+			# hardware, the user will end up with a broken install (unresponsive black screen).
+			if gfx_driver == 'Nvidia (proprietary)':
+				_gfx_driver = GfxDriver.NvidiaOpenSource
```
Torxed, line 50: hmm, I understand the "why".
But I'm not sure it's a great idea to let the user be un-informed when they chose "proprietary" but got "nouveau" :/

ghost: What do you propose then? I'm not sure how we could clearly communicate this unless there is some pop up when the installer is first ran telling them what is going to happen and that they should probably update their config (and read the announcement regarding packaging changes)

Torxed: I think one option is to not show the nvidia prop driver at all, if it's not supported?
But not sure how to do that based on hardware ID's and stuff

ghost: This pull request removes the proprietary driver from the menus, but if you read the previous pull request (#4019) @svartkanin was worried about backwards compatibility with older configs, hence why this is here.

As of driver release [590.48.01](https://www.nvidia.com/en-us/drivers/details/259267/), they are no longer supporting the proprietary driver hence it was removed from the repositories which only leaves the open kernel module available and it only supports Turing+ hardware (GTX 1650/RTX 2000 and newer)

If we fall back to the open kernel module for these older configs and the user happens to have pre-Turing hardware, they're going to be met with a black screen when they boot after installation, which I find undesirable in my opinion. I'd rather someone end up getting the 'incorrect' but working graphics driver that they can fix themselves after the installation than them be met with a broken install that may leave a new user confused with a system they can't use.

ghost:
Bumping this as I'd like to hopefully get this merged in time for the February .iso release since it didn't make it in time for the January one.

Torxed:
I think the issue of https://github.com/archlinux/archinstall/pull/4053#discussion_r2653880733 is valid tho and needs to be addressed.

ghost:
I'm open to suggestions because I'm just not too sure what would be the best way to handle such a thing. I really dislike the idea of installing ``nvidia-open`` if they have the legacy ``nvidia`` package selected in their config because if we go that route and the user has pre-Turing hardware they will be left with an unusable system after installation since the new driver release does not support hardware that is pre-Turing like I addressed in the previous comments.

The way this pull request technically handles things could be seen as "misleading" since we're installing the open-source Mesa stack instead for people that have the legacy ``nvidia`` package selected in their configs, but at least this way they'd have a functioning system that they could at least get to the desktop with.

Another option is to display a warning message when they run the installer with a config that has the old driver selected that warns them their config is out of date and will that it will install a graphics driver that may not be compatible with their hardware, and if they ignore the warning, I guess that falls on them?

PR #4056, opened 2025-12-30
Add sponsors button

h8d13:
Just thought since this is one of archlinux's most popular project, perhaps this would be useful.

Just redirects to either archlinux github sponsors or /donate link. 

For obvious reasons I cannot test this on my private repo, has to be added to settings (if we can also add a social preview in the repo settings that would be cool too :)

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,4 +1,4 @@
 # These are supported funding model platforms
 
 github: [archlinux]
-custom: [https://archlinux.org/donate/]
+custom: ['https://archlinux.org/donate/']


svartkanin:
Ill let @Torxed decide if this is desired

Torxed:
I'll just swing it by the other staff members too aside from me and @svartkanin, I suppose it's fine but I just want to make sure we don't appear greedy or something else. We're all volunteers after all so I want to make sure we're all on the same page :)

h8d13:
The thing is with such a codebase, we need to move with the time and it's difficult to do so generally speaking
Especially when the more it grows, the less it feels like making a change is something people even dare to do scared of touching something magical, but that's also how a project loses momentum. 

Anyways, my thought is that as it is, it's complicated to submit code, not feel like some things should be taken care of or are taken seriously on review, or on how to actually fix something we see often, be able to test reproduce etc. Then other things linger for no reason or are over-discussed for simple change, when a simple edit and post changes would work. 

Anyways I'm saying it takes a lot of time to test, code and is not possible for me to keep sending work when I just want to keep moving forward, in a place where it basically needs to move quite fast I believe, and where doing this all out in the public without any real interaction, isn't something on my list of want to do this everyday.

On the other side actually making a project like this feel like my own is incredibly rewarding as my friends have even gotten interested and show me support to be have passion for such a subject. 

Would benefit of lot of people and also ask is anyone actively testing, gathering info as to why this or that might change or reports, etc... Notably when a lot of reports occur like with imports currently or other withstanding to-dos. 

We are all volunteers (as much as I wish this was my job) but dedicating a lot of time and effort, to such a project is something to me more valuable than anything, is freedom of choice (and competition between standards) gain of time (for the user, not the wizard). It also means making "anything" possible, unlike many installers.

Make more time for this than for my actual jobs lmao

So yeah I kinda gave up and instead put it all inside a single branch of simplifications over hard-coding or complex. I really want to help with this project but I'm not sure I can help much more than trying my best at using/modifying as much I can :d

Anyways code-named it `archinstoo` in a fit of rage and really just tried to remove as many things for the 2nd time lmao, let's just say last time I had cut a bit too much [meme](https://www.reddit.com/r/programmingmemes/comments/1o5oykp/programming_memes/) this time I removed less, but still considerable amount. (using `vulture` to find least used) but this time adding other kind of choice (notably `--script` additions, resume handler, all the `Applications`, `doas` alternative, `Custom commands`) 

Maybe all along the key is/was just issues/discussions that once sparked here, and another time there, ... 
But the issue is also that to move fast sometimes a dev goes fast, makes mistakes, needs testing, or even collaborating with other people who enjoy coding this stuff lmao

The more configurable archinstall is the better, to me this means giving the user the choice of everything but in the simplest of ways. I do not pretend to code well, to me its more like legos that make for the death star. But also wanted to make it a folder without the 10000 files, skip flags, etc, just because I guess I live there now, and somebody who would actually want to modify this probably would just do the same and check out the code, so I don't think such a large structure is needed for core features where it gets confusing the more it grows. 

Anyways I'm posting it all on that repo and perhaps you can diff it on a Sunday with some tea (altho I've changed/removed about 100 files) some important things you would say, so a lot of tea or maybe whiskey depending on the mood and perhaps enjoy the effort for the rest of the changes

Sorry for writing so much, i'll leave the pr open, i do not think anyone is being greedy, I just think archlinux should promote archlinux projects and vice-versa

PR #4058, opened 2025-12-31
Bumping version to: 3.0.15

Torxed:
I also added a `info()` to let users know why it's "hanging" after waiting for the wkd sync, as it wasn't totally clear that we were doing operations (that some times takes a lot of time if the connection is slow or there's a lot of mirrors).

### DIFF
--- a/archinstall/lib/mirrors.py
+++ b/archinstall/lib/mirrors.py
@@ -450,7 +450,7 @@ class MirrorListHandler:
 		# Local mirrors lack this data and can be modified manually before-hand
 		# Or reflector potentially ran already
 		if self._fetched_remote and speed_sort:
-			info("Sorting your selected mirror list based on the speed between you and the individual mirrors (this might take a while)")
+			info('Sorting your selected mirror list based on the speed between you and the individual mirrors (this might take a while)')
 			# Sort by speed descending (higher is better in bitrate form core.db download)
 			return sorted(region_list, key=lambda mirror: -mirror.speed)
 		# just return as-is without sorting?


h8d13:
Happy new year in advance friends :1st_place_medal: 

Was just wondering should we not include the Nvidia changes ?

Torxed:
> Happy new year in advance friends 🥇
> 
> Was just wondering should we not include the Nvidia changes ?

I have limited connectivity to test various changes, and I won't be able to test those enough before a reelease. And the next ISO is being generated tomorrow so I'd like to get a release in before that..

Unless the nvidia thing is a HUGE issue, if that's the case then maybe.. but i feel getting the current `master` out there as-is even, is a good idea.

Happy new year in advance as well! :D

h8d13:
I'm wondering if it's a critical when you pick proprietary in menu 

Can you extend a few hours I can do some testing and maybe @sigreturn can pitch in

Nevermind seems like there is a redirect of some kind

Torxed:
> I'm wondering if it's a critical when you pick proprietary in menu

Not sure I'm following. I believe fixing graphics drivers in general can be done post-install if needed, so it shouldn't be too critical.. if we're talking about users picking the wrong driver for the wrong hardware?

I think getting the boot loader fixes in, lvm fixes, mirror status fix (so we can get around the DDoS issues) are more important for instance :) And I sadly don't have hours, as this needs to be pushed in order for the mirrors to sync the packages before the ISO is being built (and it's new years soon, so I'll be AFK until 2:d) :)

If the nvidia thing is urgent, and a installation blocker - please point out what that is as I'm sure I might have missed something if that's the case :)

Ps. I appreciate your offer to help testing - but I also feel a responsibility to do testing myself before releasing so I know all the variations and combinations have been tested extensively.

h8d13:
No it's actually fine I thought that picking prop would just crash instant. But there is a redirect in place

Torxed:
Goodie :D Onwards and upwards with the release hehe.
Sucks to do this over a "56k connection", so it's already been delayed for 2 days :sweat:

h8d13:
Good luck with release notes this month: @h8d13 a lot...

Issue #4057, opened 2025-12-30
Request: Install x11 libs conditionally


h8d13:
### describe the request

Currently in [`hardware.py`](https://github.com/archlinux/archinstall/blob/master/archinstall/lib/hardware.py#L83-L84) we automatically add `xorg-` libs even for profiles that do not need them (sway, and the likes (Hyprland, Niri, River, Labwc)). 

For the lore: #251 introduced a file called `wayland.py` in April 2021 was later removed in commit `d510ad5b` (the very next day). If you currently go the the Graphics section it will always show `xorg-server` and `xorg-xinit` in the list to be installed. 

Anyways, I've been using sway a lot and think it's a discussion we should re-open for sure would only need to be modified at the `Profile`  level instead of `Graphics Drivers`

I'm also hoping someone can beat me to it since I'm a bit tired 🥲 Do see also: [`xorg.py`](https://github.com/archlinux/archinstall/blob/master/archinstall/default_profiles/xorg.py)

JotaRandom:
The actual situation has flipped, now most stuff is Wayland or Wayland-only when the exception is Xorg only like iceWM or Xfce4 but while the former are tinhat'd against Wayland the later (Xfce4) IS doing steps to be Wayland compatible showing is just the minority who aren't traveling to Wayland.

I think it is time to revisit how we do the whole display thingies (X11 vs Wayland).

h8d13:
I've mostly finished the changes in my fork, and its about 130 pkgs less

Also added shell choices based on your feedback and doas alternative @JotaRandom 😀

Issue #3899, opened 2025-11-03
Add wayland just like Xorg


JotaRandom:
### describe the request

In the same way that the option to install Xorg is included in the settings, there should also be an option to install only Wayland, or to install both.

The desktop environments should also specify whether they work with only Xorg, with Wayland (full or experimental), or both. This would help those who want a system without Xorg and vice versa.

svartkanin:
Similar to https://github.com/archlinux/archinstall/issues/3616

The DE/WM installs whatever they ship with as their primary

PR #4066, opened 2026-01-01
fix: HW changes / X11 #4057 #3899 

h8d13:
Happy new year everyone :))

Resources used:

https://wiki.archlinux.org/title/QEMU - 

https://archlinux.org/news/plasma-640-will-need-manual-intervention-if-you-are-on-x11/
https://www.phoronix.com/news/GNOME-Mutter-Drops-X11
#4057 #3899 

VM changes is something I noticed by running `Adwaita/GTK` apps that simply wouldn't load. Using `vulkan-icd-loader` provides automatic fallback to `vulkan-swrast` or `vulkan-virtio` this notably allows you do run VMs with acceleration or software render fallback. 

Screenshot_20260101_142943

Removed the sway warning because sway works perfectly fine under NVIDIA (have been using it for months on bare-metal). 

This also notably aligns more with newer practices of Wayland vs x11 by setting an override for certain profiles:
And cuts down the pkgs installed by 137.

- Sway (TESTED)
- Hyprland
- Niri
- River
- Labwc

Also the same is true for `GNOME` (TESTED) and `KDE Plasma` (altho plasma stills pulls in `xorg-server` for `SDDM` which is bound to change soon) https://invent.kde.org/plasma/plasma-login-manager and is a choice in `Greeter`

Would be great if someone can help test other profiles too :)))
Wanted to open the discussion and reviews.

### DIFF
--- a/archinstall/default_profiles/desktops/niri.py
+++ b/archinstall/default_profiles/desktops/niri.py
@@ -2,11 +2,6 @@ from typing import override
 
 from archinstall.default_profiles.profile import GreeterType, ProfileType
 from archinstall.default_profiles.wayland import WaylandProfile
-from archinstall.lib.translationhandler import tr
-from archinstall.tui.curses_menu import SelectMenu
-from archinstall.tui.menu_item import MenuItem, MenuItemGroup
-from archinstall.tui.result import ResultType
-from archinstall.tui.types import Alignment, FrameProperties
 
 
 class NiriProfile(WaylandProfile):


CelestifyX:
Happy new year @h8d13 !!

h8d13:
To you too @CelestifyX !! Wanna test some other DEs on the patch ? :>

archinstall/default_profiles/custom.py
```diff
@@ -203,7 +203,6 @@
 # 			ProfileType.CustomType,
 # 			packages=packages,
 # 			services=services,
-# 			support_gfx_driver=True
```
svartkanin, line 206: this is dead code, if you don't mind could you delete this entire commented out block

archinstall/lib/interactions/system_conf.py
```diff
@@ -45,7 +50,7 @@ def select_kernel(preset: list[str] = []) -> list[str]:
 			return result.get_values()
 
 
-def select_driver(options: list[GfxDriver] = [], preset: GfxDriver | None = None) -> GfxDriver | None:
+def select_driver(options: list[GfxDriver] = [], preset: GfxDriver | None = None, profile: 'Profile | None' = None) -> GfxDriver | None:
```
svartkanin: I think the annotation should be `'Profile` | None`

archinstall/lib/hardware.py
```diff
@@ -2,13 +2,17 @@
...
 from pathlib import Path
+from typing import TYPE_CHECKING
 
 from .exceptions import SysCallError
 from .general import SysCommand
 from .networking import enrich_iface_types, list_interfaces
 from .output import debug
 from .translationhandler import tr
 
+if TYPE_CHECKING:
```
svartkanin: doesn't need to be behind TYPE_CHECKING

archinstall/lib/hardware.py
```diff
@@ -71,17 +76,30 @@ def is_nvidia(self) -> bool:
 			case _:
 				return False
 
-	def packages_text(self) -> str:
-		pkg_names = [p.value for p in self.gfx_packages()]
+	def packages_text(self, servers: set['DisplayServer'] | None = None) -> str:
```
svartkanin: Type can be without quotes

archinstall/lib/interactions/system_conf.py
```diff
@@ -56,7 +61,9 @@ def select_driver(options: list[GfxDriver] = [], preset: GfxDriver | None = None
 	if not options:
 		options = [driver for driver in GfxDriver]
 
-	items = [MenuItem(o.value, value=o, preview_action=lambda x: x.value.packages_text()) for o in options]
+	# Get display server requirements from profile if available
```
svartkanin: comment unnecessary

archinstall/default_profiles/desktops/cosmic.py
```diff
@@ -20,3 +20,7 @@ def packages(self) -> list[str]:
 	@override
 	def default_greeter_type(self) -> GreeterType:
 		return GreeterType.CosmicSession
+
+	@override
+	def display_servers(self) -> set[DisplayServer]:
```
svartkanin: would any profile ever return multiple values? If no, then this should not be a set

archinstall/default_profiles/profile.py
```diff
@@ -164,17 +169,23 @@ def is_tailored(self) -> bool:
...
-			return self._support_gfx_driver
-		else:
-			if any([p._support_gfx_driver for p in self.current_selection]):
-				return True
-			return False
-
 	def is_greeter_supported(self) -> bool:
 		return self._support_greeter
 
+	def display_servers(self) -> set[DisplayServer]:
```
svartkanin, line 174: lets rename this to `get_all_display_servers` to not get confused with the per profile function

archinstall/lib/hardware.py
```diff
@@ -71,17 +76,30 @@ def is_nvidia(self) -> bool:
...
 			text += f'\t- {p}\n'
 
 		return text
 
-	def gfx_packages(self) -> list[GfxPackage]:
-		packages = [GfxPackage.XorgServer, GfxPackage.XorgXinit]
+	def gfx_packages(self, servers: set['DisplayServer'] | None = None) -> list[GfxPackage]:
+		from archinstall.default_profiles.profile import DisplayServer
+
+		# empty list
```
svartkanin: all these comments are unnecessary

archinstall/lib/hardware.py
```diff
@@ -71,17 +76,30 @@ def is_nvidia(self) -> bool:
...
+		packages = []
+
+		# Only add X11 packages if required by said profile
+		if servers is None:
```
svartkanin: Can be simplified to 

```
		if servers is None or DisplayServer.X11 in servers:
			packages = [GfxPackage.XorgServer, GfxPackage.XorgXinit]
```

archinstall/lib/hardware.py
```diff
@@ -108,7 +126,7 @@ def gfx_packages(self) -> list[GfxPackage]:
 			case GfxDriver.IntelOpenSource:
 				packages += [
 					GfxPackage.Mesa,
-					GfxPackage.LibvaIntelDriver,
+					# GfxPackage.LibvaIntelDriver, #ISSUE#3005 (legacy)
```
svartkanin: should be removed if not needed, however the linked issue states that this driver is for older hardware which may be the case for some users. I cannot see a downside in leaving this in?

archinstall/lib/interactions/system_conf.py
```diff
@@ -9,6 +11,9 @@
 
 from ..hardware import GfxDriver, SysInfo
 
+if TYPE_CHECKING:
```
svartkanin: Not needed to put behind TYPE_CHECKING

archinstall/default_profiles/desktops/cosmic.py
h8d13: yes if user selects several DEs ;)

archinstall/lib/hardware.py
h8d13: My guess was conflict concerns from user, but you're right for broad compat

archinstall/default_profiles/desktops/cosmic.py
svartkanin: This is the config per profile, not for the overall profile handler, why would a single profile have multiple values?

h8d13: They do? 

You could make KDE: Xorg and Wayland

Would just be added: `plasma-x11-session` and `kwin11` and boom you support more hardware.  That has actually been requested many times. 

If you have some better way of doing this please me know other than that I've addressed all your other comments

svartkanin: KDE has dropped X11 support and ships with wayland as default. We should adhere to that and not make changes to default behaviour. Given the config you crated I cannot see any profile configs returning multiple values, so my question was if there would be a case that support is needed

h8d13: Well yeah for KDE it has been requested many times. Some hardware/use case prefers x11, they didn't drop support at all, they separated concerns

Also xorg still gets daily fixes as seen here: https://gitlab.freedesktop.org/xorg/xserver/-/commits/master?ref_type=HEADS

Contrary to Linux stereotypes we should aim to handle both, hence why I prepared this that way. And perhaps in the future even more display protocols.

Think it would make a lot of happy folks to show that Plasma works on both.

archinstall/default_profiles/desktops/gnome.py
```diff
@@ -20,3 +20,7 @@ def packages(self) -> list[str]:
 	@override
 	def default_greeter_type(self) -> GreeterType:
 		return GreeterType.Gdm
+
+	@override
+	def get_all_display_servers(self) -> set[DisplayServer]:
```
svartkanin: My apologies I misunderstood this part here, I thought it'd be an independent function from the `Profile` class but it's inherited, could you please switch the name back to `display_servers`

archinstall/default_profiles/profile.py
```diff
@@ -164,17 +169,23 @@ def is_tailored(self) -> bool:
...
-			return self._support_gfx_driver
-		else:
-			if any([p._support_gfx_driver for p in self.current_selection]):
-				return True
-			return False
-
 	def is_greeter_supported(self) -> bool:
 		return self._support_greeter
 
+	def get_all_display_servers(self) -> set[DisplayServer]:
```
svartkanin: this should be moved to the https://github.com/archlinux/archinstall/blob/5fcea379b9cd1f65343056dcfd70a441dc0ab744/archinstall/lib/profile/profiles_handler.py?plain=1#L32

h8d13:
Im guessing this is what was required, altho there are still some conventions I'm not too sure to understand like the specifc `xorg.py` and `is_xorg_type_profile` when we used to just add it in `hardware.py` regardless :d

Anyways perhaps this would require a bit more re-thinking, will be testing this afternoon.

h8d13:
I've done a little more work on this.

Added an actual Class for `WaylandProfile` which does not show up in menu
Fixed `river` to also ask for `seat_access`

Also makes it cleaner than overrides by following the same pattern as the current

Will be testing more today

svartkanin:
Can we keep PRs simple and not try to squeeze too many things in them at once. It will make reviewing them easier and quicker. 

I would suggest to revert that new WaylandProfile and keep that as a attribute per profile. In general these profiles are way too complex and should be reworked to not even inherit from the Xorg one, but rather store things as attributes.

PR #4069, opened 2026-01-01
improves additional packages loading

h8d13:
Basically this patch makes loading the `Additional Packages` section instant.

It uses `pacman -Sl` instead of the full `-S --info`

`pacman -Sl  0.14s user 0.03s system 99% cpu 0.171 total`

And builds a `multilib lib32-v4l-utils 1.32.0-1` `repo | name | version` stub.

Then updates the `Package info` just for the current selection using `-Si <pkg>`. 

This would notably help even more in regions where loading the full info is slow (even more so than for me)

### DIFF
--- a/archinstall/lib/interactions/general_conf.py
+++ b/archinstall/lib/interactions/general_conf.py
@@ -195,25 +195,31 @@ def ask_additional_packages_to_install(
 	menu_group = MenuItemGroup(items, sort_items=True)
 	menu_group.set_selected_by_value(preset_packages)
 
-	# Helper to prefetch next packages in background
+	# Helper to prefetch packages in both directions in background
 	def _prefetch_packages(group: MenuItemGroup, current_item: MenuItem) -> None:
 		try:
 			filtered_items = group.items
 			current_idx = filtered_items.index(current_item)
 
-			# Collect next 50 packages
+			# Collect next 50 packages (forward)
 			prefetch = []
 			for i in range(current_idx + 1, min(current_idx + 51, len(filtered_items))):
 				next_pkg = filtered_items[i].value
 				if isinstance(next_pkg, AvailablePackage):
 					prefetch.append(next_pkg)
 
+			# Collect previous 50 packages (backward)
+			for i in range(max(0, current_idx - 50), current_idx):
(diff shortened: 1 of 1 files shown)


svartkanin:
This makes navigating the menu extremely slow, if you hold the down/up arrows it is lazy loading every entry which isn't a very nice experience.

h8d13:
> This makes navigating the menu extremely slow, if you hold the down/up arrows it is lazy loading every entry which isn't a very nice experience.

My bad was using `/` search to test. This should do it tho

svartkanin:
It's still slow, if you're on the first entry and press up (to get to the last one) it will still be lazy loading.

I'm not sure what you're trying to solve here, is it really that bad to wait for a couple of seconds until the list is loaded?

h8d13:
Well on slow connections this takes way more time than for you or me. I like testing on horrible hardware.

It's also way more efficient to use `-Sl` takes about `0.15`s (and already contains 1/5 of the needed info). 

But good spot didn't think of the opposite way (now we do ± 50).

svartkanin:
I'm not sure that the complexity of this logic is warranted, I don't think requiring users to wait even 10 seconds for this to load is an issue. 

In the new menu there will be an option to jump by page with PGDN and PGUP, how will that be handled? There are way too many edge cases to safely start pre-loading stuff

h8d13:
I understand, it could also load 100 +/- allowing for this

Never hurts to expirement QOL

svartkanin:
That's a very hacky and flaky solution. 

The approach you took will fail again at some point which makes it unreliable and doesn't solve the problem to a full capacity. 

To give you an idea of an alternative approach, maybe we should build a loading mechanism in the background when archinstall starts. By that I mean to kick off a asynchronous job (preferrably this should be a generic framework to register multiple jobs in the future) that loads the packages. When the user finally reaches the additional packages section they will be loaded already via that job.

It'd be better to raise an issue first were we can discuss these things instead of doing it in PRs

h8d13:
Yeah I reverted it to a draft for that reason. Will open the issue :)

It seems to work pretty good from my testing tho

Issue #2474, opened 2024-04-24
FS type bcachefs is Unknown


kezyr:
Hi, when I select Pre-mounted configuration for Disk configuration where rootfs is bcachefs then it shows FS type as Unknown. 
![image](https://github.com/archlinux/archinstall/assets/62434319/b15852b8-3180-42ac-afd7-ed2a00dc87cc)

I use Hyper-V with 1 disk, latest version of archinstall.
I formated sda2 as bcachefs and mounted to /mnt, and sda1 as vfat and mounted to /mnt/efi

Then the installer craches because of fstype not set

svartkanin:
Bcachefs isn't something parted supports which is being used for partitioning under the hood.

codefiles:
> Bcachefs isn't something parted supports which is being used for partitioning under the hood.

That is not accurate for two reasons:

1. The user is attempting an install with a pre-mounted configuration and therefore has already partitioned and formatted.
2. [parted(8)](https://man.archlinux.org/man/extra/parted/parted.8.en) states that providing a file system when creating a partition is optional. This is true of pyparted as well, give it a try.

I will submit a pull request to fix this since it should work.

svartkanin:
Apologies I misread the issue, I thought it was for manual partitioning. In that case yes, pre-mounted should work and it's a bug.
However, the pre-mounted feature is in general somewhat limited still as it doesn't support many detections, btrfs and lvm being some of them.

codefiles:
> However, the pre-mounted feature is in general somewhat limited still as it doesn't support many detections, btrfs and lvm being some of them.

Btrfs works with the pre-mounted feature in the latest version.

dathide:
This issue is still occurring for me

Torxed:
Even after the release yesterday?

kezyr:
yes the issue is still here (checked using archlinux-2024.09.01-x86_64.iso)
During install there is an error message "File system type not set".

avnyu:
https://github.com/archlinux/archinstall/blob/957fa06ade3e6ba144b96a1ff2796039e6821396/archinstall/lib/installer.py#L1042
removing this line fix the issue for me, and without root fs type my desktop still boot normally

Issue #2376, opened 2024-02-28
pls add bcachefs to atomatic format options


ba-batz:
"Bcachefs is an advanced new filesystem for Linux, with an emphasis on reliability and robustness and the complete set of features one would expect from a modern filesystem." - https://bcachefs.org/

codefiles:
Duplicate of #2201

Issue #2201, opened 2023-11-06
add bcachefs option (task for ~January release)


kojq:
This can be done next year since that is around when we will have kernel `6.7`.

- We will need to consider the TRIM aspect, as the fs currently needs a `--discard` flag for support to be enabled. There is no scrub feature at the moment.

- There is an encryption option built in.

- I suggest `zstd` compression as in `btrfs` (flags are different).

Torxed:
If we adopt [bcachefs](https://wiki.archlinux.org/title/Bcachefs) *(which I think we should)*, it would be nice if we from the start go in with the mindset of supporting multiple disks joined in as one using [multiple drives](https://wiki.archlinux.org/title/Bcachefs#Multiple_drives) configuration.

This is something we don't have today in btrfs, mainly because the code was already a bit complex and handling that logic would scare me a bit.. definitely more likely to get that support today but from the start that was nearly impossible.

And we should aim for supporting [snapshots](https://bcachefs.org/Snapshots/) out of the gate too. I did a quick search as I haven't even looked at this before properly.. and I don't see much tooling around this yet. So perhaps we have to rely on users using CLI for accessing snapshots until it gets popular.

Cool idea, I like it :)

Plarpoon:
Same really, would be nice to have this for the Feb release, if possible of-course

armandothefirst:
I second this, hyped for a bcachefs option.

codefiles:
Here is a basic patch I created for the [single drive](https://wiki.archlinux.org/title/Bcachefs#Single_drive) setup.

https://github.com/archlinux/archinstall/commit/edc0767f5e4d3df45ddb679e3a5c309f8ee29ab1

Plarpoon:
Nice! Awesome!

c3Ls1US:
This would be great. 

However, currently, there's an open issue that results in [large files reliably being corrupted](https://github.com/koverstreet/bcachefs/issues/619). This seems serious and it's suspected to be compression related so until that is resolved, I think it would be wise not to apply compression yet.

Also, replicating the default btrfs subvolume layout in Archinstall would be a problem because currently bcachefs subvolumes cannot be mounted via `mount`. So if users wanted to manage a specific subvolume offline, targeting it (e.g. @) from a device wouldn't be possible. As a result, the subvolume layout would need to be flat for `chroot` to work and there'd just be two entries in fstab--the root and ESP. 

@codefiles Have you explored this?

codefiles:
[...]

PR #4072, opened 2026-01-02
feat: bcachefs support #2201

h8d13:
[Lore](https://www.phoronix.com/forums/forum/software/general-linux-open-source/1572618-linus-torvalds-marks-bcachefs-as-now-externally-maintained/page19) 1.38M commits, 5K+ contributors and dropped out in `Linux-6.18` [repo](https://github.com/koverstreet/bcachefs)

From our issues: #2201 and more dupes
#2474 #2376

So I thought let's settle it once and for all (and any re-adoption would make us just remove code).

Credits to @codefiles for the original implementation (back when it was in Kernel). And the fact that vs ZFS (licensing) this might be the actual future alternative with features similar to `btrfs`.

I'm posting this from a `bcachefs` install done through this patch/branch (currently hidden behind `--advanced`) since it's single-disk only. But now this only worked from a host (vs from an ISO) because `cow_spacesize` is too limited for the `_verify_bcachefs_support` block.


h8d13:
To bypass the problem I made an ISO with with 1GB of `cow_space` [here](https://github.com/h8d13/Int-ISO/releases/tag/ArchInt-2026.01.02) 

Screenshot 2026-01-02 16-04-14

Not sure if these are related to me being on a VM or misconfiguration + would need to test on real-hardware and see shutdown messages too (which I'm not too sure how to do).

Issue #2193, opened 2023-10-28
GRUB menu entries fail to boot when using UKIs


codefiles:
When `grub-mkconfig` is executed it will create menu entries for kernels when it detects the `vmlinuz-*` kernel files via `10_linux`. [1] When using unified kernel images these menu entries are created but will fail to boot (probably because it can not detect a corresponding `initramfs-*.img` file since those are removed by the installer when using UKIs as they are redundant [2]). The Gentoo wiki states GRUB has support for UKIs since 2.12 and will dynamically detect them if they are in the _Linux_ directory on the EFI System Partition. [3] I have had no success with dynamic detection but chainloading as mentioned in the ArchWiki works. [4]

[1] https://git.savannah.gnu.org/cgit/grub.git/tree/util/grub.d/10_linux.in?id=4e10213de66aae53c9292038ff6b8b6ae3d1ed6a
[2] https://wiki.archlinux.org/title/Unified_kernel_image#.preset_file
[3] https://wiki.gentoo.org/wiki/Unified_kernel_image#Boot_Loaders
[4] https://wiki.archlinux.org/title/GRUB#Chainloading_a_unified_kernel_image

codefiles:
I have written code that generates menu entries following the chainloading example and adds them to the `/etc/grub.d/40_custom` (could also decide to write them to `/boot/grub/custom.cfg` which does not require re-running `grub-mkconfig` when it is changed). [1] The invalid boot entries will also need to either be prevented from being generated or be negated after they are generated. This should be the course of action if it is determined that dynamic detection is not available in GRUB.

[1] https://wiki.archlinux.org/title/GRUB#Generate_the_main_configuration_file

codefiles:
The GRUB boot loader option is now restricted from the unified kernel images option. Waiting to see what happens with https://github.com/rhboot/grub2/pull/119

PR #4073, opened 2026-01-03
fix: bootloader changes

h8d13:
This patch makes it so that Grub supports UKI. Also makes it so that rEFInd is handled consistently like other boot-loaders (no fallback entry).

Finally it enables UKI by default if the system has UEFI and a supported boot-loader. Thanks to u/Synthetic451 u/Hermocrates and u/plasmamax1 on Reddit [post](https://www.reddit.com/r/archlinux/comments/1q0c31x/comment/nx2et2j/) and is only since mkinitcpio v40.

Making this default fixes systemdboot that needs a `XTLDRBOOT` part to function for `/efi`, IF UKI is not selected (something we do not have in our disk handling). 

Again safer default so user does not get not boot-able system. 

This is fixed in `def get_default(cls)` in this [line](https://github.com/archlinux/archinstall/compare/master...h8d13:dot-grubrefind?expand=1#diff-131e6745520c95c354c92cefa0144a03e0c042431f7633e14abc332eecab4633R85)

### DIFF
--- a/archinstall/lib/models/bootloader.py
+++ b/archinstall/lib/models/bootloader.py
@@ -20,7 +20,7 @@ class Bootloader(Enum):
 	Refind = 'Refind'
 
 	def has_uki_support(self) -> bool:
-      	return self != Bootloader.NO_BOOTLOADER
+		return self != Bootloader.NO_BOOTLOADER
 
 	def has_removable_support(self) -> bool:
 		match self:


codefiles:
Based on the tag [grub-2.14-rc1](https://cgit.git.savannah.gnu.org/cgit/grub.git/tag/?h=grub-2.14-rc1) (see https://cgit.git.savannah.gnu.org/cgit/grub.git/commit/?id=14c2966c725774aa58a811dada3277c7095cbcb3), GRUB will support UKI <strike>without chainloading as is being done in this pull request</strike>. I have been waiting for 2.14 to release to submit a patch (see #2193).

Edit: I was wrong about the chainloading.

```diff
+  src = grub_xasprintf ("%schainloader (%s)%s/%s%s%s\n",
+			blsuki_save_default ? "savedefault\n" : "",
+			entry->devid, entry->dirname,
+			entry->filename,
+			(options != NULL) ? " " : "",
+			(options != NULL) ? options : "");
```

h8d13:
Also thanks for the info :)) oh btw see the bcachefs draft ? Hehe 

Rest looks good to you ? Always scared when making changes to stuff like this. Altho did a lot of testing today

codefiles:
I have seen probably most of what you have submitted. Pretty good stuff.

h8d13:
Thanks !! Starting to run out of ideas 😅

PR #4074, opened 2026-01-03
feat: firewalls

h8d13:
Inspired by the @sigreturn changes on ppd, I thought this is also something that users might be(should be) doing either-way. Didn't implement another simply because I'm only familiar with this one. But think it would be a good addition.

### DIFF
--- a/archinstall/applications/firewall.py
+++ b/archinstall/applications/firewall.py
@@ -14,6 +14,12 @@ class FirewallApp:
 			'ufw',
 		]
 
+	@property
+	def ufw_services(self) -> list[str]:
+		return [
+			'ufw.service',
+		]
+
 	def install(
 		self,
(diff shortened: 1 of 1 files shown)


archinstall/applications/firewall.py
```diff
@@ -0,0 +1,26 @@
+from typing import TYPE_CHECKING
+
+from archinstall.lib.models.application import Firewall, FirewallConfiguration
+from archinstall.lib.output import debug
+
+if TYPE_CHECKING:
+	from archinstall.lib.installer import Installer
+
+
+class FirewallApp:
```
svartkanin, line 10: Shouldn't this also enable the service `ufw.service` ?

h8d13: Good call mb

PR #4076, opened 2026-01-04
Change plasma-meta -> -desktop.

h8d13:
Second time I open up this subject, instead of creating a separate profile we should use this option directly:

First of all it installs `flatpak` related libs which should be up to the user as to which containerization tech he wants to use. 

Second of all it also pulls in `sddm-kcm` which isn't needed if user selected another login manager through `archinstall`. 

This still results in a fully functional system and aligns closer to arch philosophy (plus 1GiB saved in bandwidth)

### DIFF
--- a/archinstall/default_profiles/desktops/plasma.py
+++ b/archinstall/default_profiles/desktops/plasma.py
@@ -12,7 +12,7 @@ class PlasmaProfile(XorgProfile):
 	@override
 	def packages(self) -> list[str]:
 		return [
-			'plasma-meta',
+			'plasma-desktop',
 			'konsole',
 			'kate',
 			'dolphin',


svartkanin:
I think that's fine for the "minimal" approach

ghost:
> Second time I open up this subject, instead of creating a separate profile we should use this option directly:
> 
> First of all it installs `flatpak` related libs which should be up to the user as to which containerization tech he wants to use.
> 
> Second of all it also pulls in `sddm-kcm` which isn't needed if user selected another login manager through `archinstall`.
> 
> This still results in a fully functional system and aligns closer to arch philosophy (plus 1GiB saved in bandwidth)

Doesn't result in a functional system. Tested it in April Arch ISO and it sucks. Can only see a mouse on a black desktop.

ghost:
> I think that's fine for the "minimal" approach

Did you test it? It's completely black, only the mouse appears and move.

PR #4079, opened 2026-01-04
feat: explicit lock root option

h8d13:
Inspired by Fedora installer which has a toggle button for this. 

I do realize that leaving root pw empty would achieve the same but making it an explicit option in menu, even if user has already created both root pw and sudo user seems like a good security/awareness practice.

### DIFF
--- a/examples/config-sample.json
+++ b/examples/config-sample.json
@@ -1,5 +1,8 @@
 {
   "archinstall-language": "English",
+  "auth_config": {
+    "lock_root_account": false
+  },
   "audio_config": {
     "audio": "pipewire"
   },


svartkanin:
I'm confused about all your draft PRs, are you still working on them?

h8d13:
I think its fair say im actively working on all of them ? Some might be a draft for a good reason: usually needs testing, which takes more time than coding itself 

Also thanks all for all the valuable feedback on other PRs

svartkanin:
Okay then I will not review any PRs in draft as I will consider them incomplete and wait for the review request

h8d13:
That's fair enough, maybe having a direct communication channel with you would be nice too, for things I do not want to share in public where I get sassy comments from randos. Perhaps even enable collaborating on larger changes/branches

svartkanin:
Raising issues is always the preferred way to start an idea and conversation. It allows others who have interest to chime in as well. 

It's the internet so not everyone is nice and especially not in open source where opinions are plenty, so it takes some getting used to it, but assuming the best intention of people is a good start

h8d13:
Morning (: since I would like to contribute to this repo in the long run and actively open both issues and describe my PRs when they are not related to one, I'm not sure its necessary every time

About open comments I will just block users that I deemed to not want notifications for

Issue #3990, opened 2025-12-14
Issue installing omarchy on Toshiba laptop


bmadhusu:
### Which ISO version are you using?

Omarchy-3.2.2.iso

### The installation log

```json
[2025-12-13 20:30:49] - DEBUG - Creating yescrypt hash with rounds 5
[2025-12-13 20:30:56] - DEBUG - Hardware model detected: TOSHIBA Satellite C55-B; UEFI mode: True
[2025-12-13 20:30:56] - DEBUG - Processor model detected: Intel(R) Celeron(R) CPU  N2840  @ 2.16GHz
[2025-12-13 20:30:56] - DEBUG - Memory statistics: 3409568 available out of 3888172 total installed
[2025-12-13 20:30:56] - DEBUG - Could not detect virtual system: ['/usr/bin/systemd-detect-virt'] exited with abnormal exit code [1]: none

[2025-12-13 20:30:56] - DEBUG - System is not running in a VM: ['/usr/bin/systemd-detect-virt'] exited with abnormal exit code [1]: none

[2025-12-13 20:30:56] - DEBUG - Virtualization detected: None; is VM: False
[2025-12-13 20:30:56] - DEBUG - Graphics devices detected: dict_keys(['Intel Corporation Atom Processor Z36xxx/Z37xxx Series Graphics & Display (rev 0e)'])
[2025-12-13 20:30:56] - DEBUG - Disk states before installing:
{
    "blockdevices": [
        {
            "name": "loop0",
            "path": "/dev/loop0",
            "pkname": null,
            "log_sec": 512,
            "size": "6023 MiB",
            "pttype": null,
            "ptuuid": null,
            "rota": true,
            "tran": null,
            "partn": null,
            "partuuid": null,
            "parttype": null,
            "uuid": null,
            "fstype": "squashfs",
            "fsver": "4.0",
            "fsavail": 0,
            "fsuse_percentage": "100%",
            "type": "loop",
            "mountpoint": "/run/archiso/airootfs",
            "mountpoints": [
                "/run/archiso/airootfs"
            ],
            "fsroots": [
                "/"
            ],
            "children": []
        },
        {
            "name": "sda",
            "path": "/dev/sda",
            "pkname": null,
            "log_sec": 512,
            "size": "476940 MiB",
            "pttype": "gpt",
            "ptuuid": "940d6706-fd28-4c59-bb01-592a4098e1c8",
            "rota": true,
            "tran": "sata",
            "partn": null,
[...]

Issue #3975, opened 2025-12-07
Error installing on btrfs partition with subvolumes and LUKS encryption


alvarosainzpardo:
### Which ISO version are you using?

2025-12-01

### The installation log

```json
[2025-12-07 16:20:00] - DEBUG - Hardware model detected: QEMU Standard PC (Q35 + ICH9, 2009); UEFI mode: True
[2025-12-07 16:20:00] - DEBUG - Processor model detected: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
[2025-12-07 16:20:00] - DEBUG - Memory statistics: 7644604 available out of 8106552 total installed
[2025-12-07 16:20:00] - DEBUG - Virtualization detected: kvm; is VM: True
[2025-12-07 16:20:01] - DEBUG - Graphics devices detected: dict_keys(['Red Hat, Inc. Virtio 1.0 GPU (rev 01)'])
[2025-12-07 16:20:01] - DEBUG - Disk states before installing:
{
    "blockdevices": [
        {
            "name": "loop0",
            "path": "/dev/loop0",
            "pkname": null,
            "log_sec": 512,
            "size": "971 MiB",
            "pttype": null,
            "ptuuid": null,
            "rota": true,
            "tran": null,
            "partn": null,
            "partuuid": null,
            "parttype": null,
            "uuid": null,
            "fstype": "squashfs",
            "fsver": "4.0",
            "fsavail": 0,
            "fsuse_percentage": "100%",
            "type": "loop",
            "mountpoint": "/run/archiso/airootfs",
            "mountpoints": [
                "/run/archiso/airootfs"
            ],
            "fsroots": [
                "/"
            ],
            "children": []
        },
        {
            "name": "sr0",
            "path": "/dev/sr0",
            "pkname": null,
            "log_sec": 2048,
            "size": "1467 MiB",
            "pttype": "dos",
            "ptuuid": "16da8f38",
            "rota": true,
            "tran": "sata",
            "partn": null,
            "partuuid": null,
            "parttype": null,
            "uuid": "2025-12-01-16-13-59-00",
            "fstype": "iso9660",
            "fsver": "Joliet Extension",
            "fsavail": 0,
            "fsuse_percentage": "100%",
            "type": "rom",
            "mountpoint": "/run/archiso/bootmnt",
            "mountpoints": [
                "/run/archiso/bootmnt"
            ],
            "fsroots": [
                "/"
            ],
            "children": []
        },
        {
[...]

PR #4080, opened 2026-01-04
feat: make bug report URL configurable

h8d13:
This would notably allow for forks to provide their own URL for bug reports (especially since they might be frozen behind or use very modified configs). 

Less false-positives for us and more critical bug reports for everyone (:

### DIFF
--- a/archinstall/__init__.py
+++ b/archinstall/__init__.py
@@ -143,9 +143,9 @@ def run_as_a_module() -> None:
 			error(err)
 
 			text = (
-				'Archinstall experienced the above error. If you think this is a bug, please report it to\n'
-				'https://github.com/archlinux/archinstall and include the log file "/var/log/archinstall/install.log".\n\n'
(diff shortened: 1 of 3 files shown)


CelestifyX:
nonsense

h8d13:
sry but who tf are you ? as far as i can telly you've never contributed anything here, but make sassy comments
please refrain unless you've got something of value to add :) #3990 #3975 #3781

h8d13:
"interlocutor" would assume:
1) your feedback was constructive which "nonsense" is not 
2) you wouldn't comment on every post 

Now please again refrain from useless comments, and don't waste my time and pollute PR conversations :)

svartkanin:
I don't think this is something to cater for. Forks can do whatever they want, why would archinstall need to care about forks being able to more easily add bug report urls

h8d13:
Because archinstall should set the standard for forks. 

They should have their own respective issues based on their setups, ISOs, and lag-behind time, etc. #3990 #3975 #3781

PR #4081, opened 2026-01-05
post screen instructions

h8d13:
Something I see frequently in the Discord support thread
Just a small reminder to eject media/change boot prios

### DIFF
--- a/archinstall/lib/interactions/general_conf.py
+++ b/archinstall/lib/interactions/general_conf.py
@@ -19,9 +19,9 @@ from ..translationhandler import Language
 
 
 class PostInstallationAction(Enum):
-	EXIT = tr('Exit archinstall')
-	REBOOT = tr('Reboot system')
-	CHROOT = tr('Chroot into installation for post-installation configurations')
+	EXIT = tr('exit archinstall')
+	REBOOT = tr('reboot system')
+	CHROOT = tr('chroot into installation for post-installation configurations')
 
 
 def ask_ntp(preset: bool = True) -> bool:


svartkanin:
Can you link to the actual problem people are reporting or raise an issue with some more context.

h8d13:
Seen in the [Discord#help](https://discord.gg/RYNPVfFJ) (latest messages in help thread) and countless other times on other forums

> its my first time installing arch and everytime i finish an install if i reboot after using the archinstall it doesn't boot but goes to the install medium menu sry to bother

svartkanin:
I'd suggest to move the `Reboot` option to the top so it's the selected one, and reformulate the message to 

`After reboot, remove the installation medium`

When the medium is removed changes to the boot order shouldn't be required, I think that may also mentioning that would also confuse users more

h8d13:
That sounds great will do right now

archinstall/lib/interactions/general_conf.py
```diff
@@ -19,8 +19,8 @@
 
 
 class PostInstallationAction(Enum):
+	REBOOT = tr('Reboot system (and remove install media)')
```
svartkanin: Sorry maybe my initial suggestion wasn't clear, lets not change this text as it's fine but instead add that header you had and change it to 
```
After reboot, remove the installation medium
```

Reason being, that header message applies also to cases when a user just exists archinstall and at some point later reboots.

archinstall/lib/interactions/general_conf.py
```diff
@@ -21,7 +21,7 @@
 class PostInstallationAction(Enum):
 	EXIT = tr('Exit archinstall')
 	REBOOT = tr('Reboot system')
-	CHROOT = tr('chroot into installation for post-installation configurations')
+	CHROOT = tr('Chroot into installation for post-installation configurations')
```
svartkanin: we should leave this lowercase as it's referring tot he command

h8d13: Should I set them all to lowercase since the two others also are commands then ?

svartkanin: I see. At one hand `Reboot` and `Exit` are verbs indicating the next step, but `chroot` could be interpreted the same way. Maybe I was wrong and we should make them all title case `Reboot`, `Exit` and `Chroot`

PR #4088, opened 2026-01-06
Fix an invalid-type-form error reported by ty

correctmost:
## PR Description:

The 'bytes' annotation for the 'bytes' property was actually a forward reference to the property instead of builtins.bytes: https://github.com/astral-sh/ty/issues/1747#issuecomment-3609042917

ty was reporting this error:
```
archinstall/lib/models/device.py:775:21: error[invalid-type-form] Variable of type `property` is not allowed in a type expression
```

### DIFF
--- a/archinstall/lib/models/device.py
+++ b/archinstall/lib/models/device.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import builtins
 import math
 import uuid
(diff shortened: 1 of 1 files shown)


svartkanin:
I'm looking forward to `ty` being mature enough, if it's as good as `uv` it should be a blast

PR #4090, opened 2026-01-06
h2t: protect host cfg

h8d13:
Still on my spree of testing directly from one install to another.

This a simple patch to not write into a host `pacman.conf` prevents from accidentally modifying the host's `/etc` when using `Custom Repositories` section in the menu but persists on target regardless

### DIFF
--- a/archinstall/lib/pacman/config.py
+++ b/archinstall/lib/pacman/config.py
@@ -1,3 +1,4 @@
+import atexit
 import re
 from pathlib import Path
 from shutil import copy2
@@ -47,10 +48,12 @@ class PacmanConfig:
 				if row + 1 < len(content) and content[row + 1].lstrip().startswith('#'):
 					content[row + 1] = re.sub(r'^#\s*', '', content[row + 1])
 
-		# Modify file only in ISO env
+		# Apply temp using backup then revert on exit handler
 		if running_from_host():
-			pass
-		else:
+			temp_copy = self._config_path.with_suffix('.bak')
+			copy2(self._config_path, temp_copy)
+			atexit.register(lambda: copy2(temp_copy, self._config_path))
+		else:  # Apply directly
 			with open(self._config_path, 'w') as f:
 				f.writelines(content)
 


archinstall/lib/pacman/config.py
```diff
@@ -46,9 +47,12 @@ def apply(self) -> None:
 				if row + 1 < len(content) and content[row + 1].lstrip().startswith('#'):
 					content[row + 1] = re.sub(r'^#\s*', '', content[row + 1])
 
-		# Write the modified content back to the file
-		with open(self._config_path, 'w') as f:
-			f.writelines(content)
+		# Modify file only in ISO env
+		if running_from_host():
```
svartkanin, line 52: Hmm that would mean that the setting is irrelevant altogether when running on a host which is somewhat misleading. If someone sets this in the menu they'd expect it to be applied. Instead we should probably lock down the menu with a message for this config saying not available on the host. But again I'm not sure that's wanted, some users may want to actually add a custom repo here for an install which is then fully disabled

svartkanin (changes requested):

archinstall/lib/pacman/config.py
h8d13: Maybe we should set it temporarily and revert at exit ?

And it's not really irrelevant since it still uses the host's existing cfg and then persists on target

archinstall/lib/pacman/config.py
```diff
@@ -46,9 +48,14 @@ def apply(self) -> None:
 				if row + 1 < len(content) and content[row + 1].lstrip().startswith('#'):
 					content[row + 1] = re.sub(r'^#\s*', '', content[row + 1])
 
-		# Write the modified content back to the file
-		with open(self._config_path, 'w') as f:
-			f.writelines(content)
+		# Apply temp using backup then revert on exit handler
+		if running_from_host():
+			temp_copy = self._config_path.with_suffix('.bak')
+			copy2(self._config_path, temp_copy)
```
svartkanin, line 54: Every time you're running this apply function you will copy the pacman.conf file, which means if you run it 2 times, it will override the first backup file.

PR #4100, opened 2026-01-09
hotfix: Firewall

h8d13:
Basically needed for `ufw status` to return `active` when inside the env as enabling the service only adds it to bootlevel. And inherits from default rules properly then, IIUC

### DIFF
--- a/archinstall/lib/global_menu.py
+++ b/archinstall/lib/global_menu.py
@@ -341,6 +341,11 @@ class GlobalMenu(AbstractMenu[None]):
 				output += f'{tr("Power management")}: {power_management_config.power_management.value}'
 				output += '\n'
 
+			if app_config.firewall_config:
+				firewall_config = app_config.firewall_config
+				output += f'{tr("Firewall")}: {firewall_config.firewall.value}'
+				output += '\n'
+
 			return output
 
 		return None


archinstall/applications/firewall.py
```diff
@@ -31,3 +31,4 @@ def install(
 			case Firewall.UFW:
 				install_session.add_additional_packages(self.ufw_packages)
 				install_session.enable_service(self.ufw_services)
+				install_session.arch_chroot("sed -i 's/ENABLED=no/ENABLED=yes/' /etc/ufw/ufw.conf")
```
codefiles: `Installer` has the path `target`:
https://github.com/archlinux/archinstall/blob/62edc56a52879846d16bbd9544fcd8642f5bf666/archinstall/lib/installer.py#L65

Why would you use sed and not Python here, does this not work?

```suggestion
				ufw_conf = install_session.target / 'etc/ufw/ufw.conf'
				ufw_conf.write_text(ufw_conf.read_text().replace('ENABLED=no', 'ENABLED=yes'))
```

Edit: Fix that the code was not rendering for the given link.
The latest commit when I linked to the above excerpt of code was https://github.com/archlinux/archinstall/commit/62edc56a52879846d16bbd9544fcd8642f5bf666 or `62edc56a52879846d16bbd9544fcd8642f5bf666`, ending in `666`, perfection.

h8d13: Sorry my bad was trying `ufw enable` before then just defaulted to thinking in bash

PR #4102, opened 2026-01-09
feat: editors

h8d13:
This patch removes `vim` and `nano` from default in `desktop` and gives a menu option for `Editor` inside of `Applications`. This also properly enables it to be used as default through `/etc/environment`

Fixes the preview in main menu too and makes it more in line so that if you perform a minimal install you might still want to be able to edit files. This also enables `$EDITOR` global key value pair to be used consistently

Guys help I'm stuck in VIM

### DIFF
--- a/archinstall/applications/editor.py
+++ b/archinstall/applications/editor.py
@@ -22,9 +22,11 @@ class EditorApp:
 		install_session.add_additional_packages(editor_config.editor.value)
 
 		editor_binary = self.get_editor_binary(editor_config.editor)
+		# special handling for shorter names needs to happen after pkg has been added above as 'neovim'
+		editor_binary_short = {'neovim': 'nvim'}.get(editor_binary, editor_binary)
 		environment_path = install_session.target / 'etc' / 'environment'
 
-		debug(f'Setting EDITOR={editor_binary} in {environment_path}')
+		debug(f'Setting EDITOR={editor_binary_short} in {environment_path}')
 
 		with open(environment_path, 'a') as f:
-			f.write(f'EDITOR={editor_binary}\n')
+			f.write(f'EDITOR={editor_binary_short}\n')


archinstall/lib/models/application.py
```diff
@@ -38,6 +38,19 @@ class FirewallConfigSerialization(TypedDict):
 	firewall: str
 
 
+class Editor(StrEnum):
+	NANO = 'nano'
+	MICRO = 'micro'
+	VI = 'vi'
+	VIM = 'vim'
+	NEOVIM = 'neovim'
+	EMACS = 'emacs'
```
codefiles:
```suggestion
class Editor(StrEnum):
	NANO = auto()
	MICRO = auto()
	VI = auto()
	VIM = auto()
	NEOVIM = auto()
	EMACS = auto()
```

https://docs.python.org/3/library/enum.html#enum.StrEnum

archinstall/applications/editor.py
```diff
@@ -0,0 +1,79 @@
...
+			case Editor.MICRO:
+				return 'micro'
+			case Editor.VI:
+				return 'vi'
+			case Editor.VIM:
+				return 'vim'
+			case Editor.NEOVIM:
+				return 'nvim'
+			case Editor.EMACS:
+				return 'emacs'
```
codefiles:
```suggestion
		return 'nvim' if editor == editor.NEOVIM else editor.value
```

archinstall/applications/editor.py
```diff
@@ -0,0 +1,79 @@
...
+	def vim_package(self) -> list[str]:
+		return ['vim']
+
+	@property
+	def neovim_package(self) -> list[str]:
+		return ['neovim']
+
+	@property
+	def emacs_package(self) -> list[str]:
+		return ['emacs']
```
codefiles:
```suggestion
```

This can be removed or as an alternative could be `_get_editor_package()` like `_get_editor_binary()`. See why in the next suggestion.

archinstall/applications/editor.py
```diff
@@ -0,0 +1,79 @@
...
+			case Editor.MICRO:
+				install_session.add_additional_packages(self.micro_package)
+			case Editor.VI:
+				install_session.add_additional_packages(self.vi_package)
+			case Editor.VIM:
+				install_session.add_additional_packages(self.vim_package)
+			case Editor.NEOVIM:
+				install_session.add_additional_packages(self.neovim_package)
+			case Editor.EMACS:
+				install_session.add_additional_packages(self.emacs_package)
```
codefiles:
```suggestion
		install_session.add_additional_packages(editor_config.editor.value)
```

See suggestion on <code><i>editor</i>_package</code> properties.

h8d13:
You the goat of reviewing :D Cheers

archinstall/applications/editor.py
```diff
@@ -0,0 +1,30 @@
...
+from archinstall.lib.models.application import Editor, EditorConfiguration
+from archinstall.lib.output import debug
+
+if TYPE_CHECKING:
+	from archinstall.lib.installer import Installer
+
+
+class EditorApp:
+	def _get_editor_binary(self, editor: Editor) -> str:
+		# codefiles hack since its the only name that differs
```
codefiles:
```suggestion
```

If you feel it is a hack then do it the "right way". Please do not use my pseudonym in a code comment.

h8d13: I meant it in a good way sry

archinstall/applications/editor.py
```diff
@@ -0,0 +1,30 @@
...
+		return 'nvim' if editor == Editor.NEOVIM else editor.value
+
+	def install(
+		self,
+		install_session: 'Installer',
+		editor_config: EditorConfiguration,
+	) -> None:
+		debug(f'Installing editor: {editor_config.editor.value}')
+
+		install_session.add_additional_packages([editor_config.editor.value])
```
codefiles: In the last suggestion of the first review I showed the argument to this call not being a list intentionally since that will be handled by `Pacman.strap` later anyway.

https://github.com/archlinux/archinstall/blob/5fbe1deea4816120580a9a017a79ffa5d23d1f84/archinstall/lib/installer.py#L1834-L1835
https://github.com/archlinux/archinstall/blob/5fbe1deea4816120580a9a017a79ffa5d23d1f84/archinstall/lib/pacman/__init__.py#L67-L68

archinstall/applications/editor.py
```diff
@@ -0,0 +1,30 @@
...
+
+		install_session.add_additional_packages(editor_config.editor.value)
+
+		editor_binary = self._get_editor_binary(editor_config.editor)
+		environment_path = install_session.target / 'etc' / 'environment'
+
+		debug(f'Setting EDITOR={editor_binary} in {environment_path}')
+
+		with open(environment_path, 'a') as f:
+			f.write(f'EDITOR={editor_binary}\n')
```
codefiles: Should the menu have a message stating that selecting an editor will set that editor globally?

https://wiki.archlinux.org/title/Environment_variables#Using_pam_env

archinstall/lib/global_menu.py
```diff
@@ -341,6 +341,16 @@ def _prev_applications(self, item: MenuItem) -> str | None:
 				output += f'{tr("Power management")}: {power_management_config.power_management.value}'
 				output += '\n'
 
+			if app_config.firewall_config:
+				firewall_config = app_config.firewall_config
+				output += f'{tr("Firewall")}: {firewall_config.firewall.value}'
+				output += '\n'
+
```
codefiles: Shouldn't this be in a separate pull request?

archinstall/applications/editor.py
h8d13: Tysm, added `Would you like to set an editor globally?`

archinstall/applications/editor.py
```diff
@@ -0,0 +1,30 @@
...
+from archinstall.lib.output import debug
+
+if TYPE_CHECKING:
+	from archinstall.lib.installer import Installer
+
+
+class EditorApp:
+	def _get_editor_binary(self, editor: Editor) -> str:
+		# special handling only name that differs
+		return 'nvim' if editor == Editor.NEOVIM else editor.value
```
codefiles: The paradigm used by this repo is to add functionality like this to the enum. Consider following that convention and moving this method to `Editor` instead and removing the underscore(`_`) that designates it as intending to be private.

Here are two instances of this paradigm being used in the repo (there are more):

- `FilesystemType.installation_pkg`
  https://github.com/archlinux/archinstall/blob/5fbe1deea4816120580a9a017a79ffa5d23d1f84/archinstall/lib/models/device.py#L814-L824
- `GfxDriver.gfx_packages`
  https://github.com/archlinux/archinstall/blob/5fbe1deea4816120580a9a017a79ffa5d23d1f84/archinstall/lib/hardware.py#L83

codefiles: I guess this suggestion can be ignored because you are following the convention set by `archinstall/lib/models/application.py` and other files in `archinstall/applications`.

h8d13: See latest commit ;)

svartkanin:
I'm quite worried at this point to go down a rabbit hole with this new application configuration. If we decide to give this specific configuration option in the menu there would be an argument to simply provide dozens of options for users to configure. 

This is user software to be installed not a system configuration to get Arch up and running. It is not necessary for a working OS at all. The reason an editor was included in the profiles is to provide at least one out of the box so it can be used to edit files, ideally that should be the default a profile ships with otherwise.

PR #4103, opened 2026-01-09
feat: monitor

h8d13:
Same exact train of thought as #4102 , instead of hard-coding in `desktop` offer user a choice and make this extendable to other monitoring tools in the future

Last thing i'm wondering about is `smartmontools` but I'm guessing it's there for a good reason ? Or we might want to remove it too.


codefiles (commented):
These suggestions apply here too: https://github.com/archlinux/archinstall/pull/4102#discussion_r2676153576, https://github.com/archlinux/archinstall/pull/4102#discussion_r2676194701, https://github.com/archlinux/archinstall/pull/4102#discussion_r2676203845

svartkanin:
I'm very against this as also pointed out in https://github.com/archlinux/archinstall/pull/4102#issuecomment-3730891708. 

We have `Additional packages` for this where users can pick the packages they want to install. This will open a box of infinite options for configurations. This is user software that can easily be a post installation step and should not be handled as part of archinstall

PR #4104, opened 2026-01-09
feat: management

h8d13:
Idea credits to @Mattyan89 in discussions

Simply adds a `Management` option with standard tools often used in Arch Linux installations where you can select none or all of them.

### DIFF
--- a/archinstall/lib/applications/application_menu.py
+++ b/archinstall/lib/applications/application_menu.py
@@ -260,11 +260,14 @@ def select_firewall(preset: FirewallConfiguration | None = None) -> FirewallConf
 def select_management(preset: ManagementConfiguration | None = None) -> ManagementConfiguration | None:
 	group = MenuItemGroup.from_enum(Management)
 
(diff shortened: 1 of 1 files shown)


archinstall/lib/models/application.py
```diff
@@ -126,6 +137,22 @@ def parse_arg(arg: dict[str, Any]) -> 'FirewallConfiguration':
...
+class ManagementConfiguration:
+	tools: list[Management]
+
+	def json(self) -> ManagementConfigSerialization:
+		return {
+			'tools': [t.value for t in self.tools],
+		}
+
+	@staticmethod
+	def parse_arg(arg: dict[str, Any]) -> 'ManagementConfiguration':
```
correctmost: Can you use `ManagementConfigSerialization` instead of `dict[str, Any]` here?

That would help catch mismatching dictionary keys and reduce the usage of `Any` in the codebase.

svartkanin:
Same https://github.com/archlinux/archinstall/pull/4103#issuecomment-3730906962

PR #4105, opened 2026-01-09
chore: pre-commit hook pylint

h8d13:
This makes it so that it can find local `archinstall` module properly, was giving me more errors today and end-up using `uninstall` temporarily


svartkanin:
this seems like a bit of a hack, what exactly is the problem you have?

PR #4114, opened 2026-01-11
Change classmethods to staticmethods

codefiles:


archinstall/lib/models/network.py
```diff
@@ -266,8 +266,8 @@ def from_wpa_cli_output(cls, list_networks: str) -> list[WifiConfiguredNetwork]:
 
 		return networks
 
-	@classmethod
-	def _extract_flags(cls, flag_string: str) -> list[str]:
+	@staticmethod
+	def _extract_flags(flag_string: str) -> list[str]:
```
codefiles, line 270: This is unused code. There is a commented out call to this method above.

PR #4118, opened 2026-01-11
Remove unused installation_hooks property

codefiles:


codefiles:
@svartkanin, if you are wondering, you added it in 00b0ae7b and it has never been used. I do not see if it was renamed in that or what it could have been derived from or what its intended use was.

svartkanin:
Thanks, I'll have a look at it

PR #4128, opened 2026-01-15
Remove storage module

codefiles:
`installation_session` is no longer used.


svartkanin:
I reversed the removal of the example from my PR which is using the installation session. Could we remove that example with this PR then

codefiles:
That was just removed with 972e278. That should cover it or am I missing something?

svartkanin:
Ahh missed that removal there

svartkanin (approved):

PR #4132, opened 2026-01-15
feat: libcamera

EliasTheBigMasterofIssues:
- This fix issue: 

## PR Description:

Add a simple menu toggle based on print_service toggle, to add libcamera which can be necessary
to libcamera based applications and can be used to enable webcams

## Tests and Checks
- [YES ] I have tested the code! <br>
- Just when audio server is pipewire

### DIFF
--- a/archinstall/applications/camera.py
+++ b/archinstall/applications/camera.py
@@ -1,0 +1,16 @@
+from typing import TYPE_CHECKING
+
+from archinstall.lib.output import debug
+
+if TYPE_CHECKING:
+	from archinstall.lib.installer import Installer
+
+
+class CameraApp:
+	@property
+	def packages(self) -> list[str]:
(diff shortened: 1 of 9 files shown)


h8d13:
I like this so if i understand correctly we'd need to check the current selection is pipewire ? 

Also any info on which kinds of hardware this is needed ?

EliasTheBigMasterofIssues:
> I like this so if i understand correctly we'd need to check the current selection is pipewire ?
> 
> Also any info on which kinds of hardware this is needed ?

Yes, in this case is IPU3, IPU4 and IPU7 to be used with libcamera-ipa
which provides libcamera softISP, but even uvcvideo cameras can use it
but without simple pipeline, that means no SoftISP IIRC.
But i'm not a programmer, idk how is possible to add support to sysinfo
detect if is necessary libcamera-softisp, i just have copied actual print_service
and adapt it to install libcamera.

svartkanin:
There has been no updates on this for some time now. 
In general I don't think this is something required for a system installation and can easily be a post-installation setup

PR #4137, opened 2026-01-17
Enable quoted-annotation ruff rule and fixes

codefiles:
https://docs.astral.sh/ruff/rules/quoted-annotation/

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -1942,7 +1942,7 @@ class Installer:
 		except SysCallError:
 			return False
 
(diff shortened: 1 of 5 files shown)


archinstall/lib/installer.py
```diff
@@ -1942,7 +1942,7 @@ def chown(self, owner: str, path: str, options: list[str] = []) -> bool:
 		except SysCallError:
 			return False
 
-	def set_vconsole(self, locale_config: 'LocaleConfiguration') -> None:
+	def set_vconsole(self, locale_config: LocaleConfiguration) -> None:
```
codefiles, line 1945: Ruff didn't detect this one for whatever reason.

PR #4139, opened 2026-01-19
Enable GRUB UKI menu entries

codefiles:


svartkanin (approved):

codefiles:
A problem here is if more than one kernel is installed there will be multiple "Arch Linux" entries in the menu with the edit option being the only way to distinguish them.

The menu title comes from the `PRETTY_NAME` in `.osrel`, see [`grub-core/commands/blsuki.c`](https://cgit.git.savannah.gnu.org/cgit/grub.git/tree/grub-core/commands/blsuki.c?id=eaa3b8f0f90605a82c6bfda4c5c4b73c58eb81ac#n1069):

```c
  /*
   * Although .osrel is listed as optional in the UKI specification, the .osrel
   * section is needed to generate the GRUB menu entry title.
   */
  osrel = blsuki_get_val (entry, ".osrel", NULL);
  if (osrel == NULL)
    {
      grub_dprintf ("blsuki", "Skipping file %s with no '.osrel' key.\n", entry->filename);
      goto finish;
    }

  osrel_line = osrel;
  while ((key = uki_read_osrel (&osrel_line, &value)) != NULL)
    {
      if (grub_strcmp ("PRETTY_NAME", key) == 0)
	{
	  title = value;
	  break;
	}
    }
```

This could be alleviated by using [recall previous entry](https://wiki.archlinux.org/title/GRUB/Tips_and_tricks#Recall_previous_entry) with the addition of the following command.

```
grub-editenv /boot/grub/grubenv set blsuki_save_default=true
```

Edit:
This problem should be solved by the following commit (not in a release yet).
https://gitlab.freedesktop.org/gnu-grub/grub/-/commit/9be1b7564daa770fd5ff33e9052edc9893d81f84

svartkanin:
I suppose that's fine for now and we can improve that in the future

codefiles:
Maybe someone will see this and update the [GRUB](https://wiki.archlinux.org/title/GRUB) ArchWiki article.

The `uki` command was introduced in GRUB 2.14rc1:
https://gitlab.freedesktop.org/gnu-grub/grub/-/commit/14c2966c725774aa58a811dada3277c7095cbcb3

With this pull request the `uki` command is used to add menu entries for UKI files. These are the new steps archinstall takes when GRUB is selected along with UKI:

- Create `/etc/grub.d/15_uki` with the following contents:

  ```sh
  #! /bin/sh
  set -e

  cat << EOF
  if [ "\$grub_platform" = "efi" ]; then
    uki
  fi
  EOF
  ```

- Add executable permissions to the file.

  ```
  # chmod +x /etc/grub.d/15_uki
  ```
  
  Now when `grub-mkconfig` is executed it will use this file and generated menu entries for detected UKI files.

- Remove executable permissions from `/etc/grub.d/10_linux` to disable it.

  The file `/etc/grub.d/10_linux` detects kernel files (`vmlinuz-*`) and creates menu entries for them and their corresponding initramfs file (`initramfs-*.img`) when `grub-mkconfig` is executed. The UKI files will contain an initramfs so there is no need for these `initramfs-*.img` files. If `initramfs-*.img` files are not present the file `/etc/grub.d/10_linux` will generate menu entries but they will fail to boot when selected.

  ```
  # chmod -x /etc/grub.d/10_linux
  ```

PR #4147, opened 2026-01-19
Rework has_default_btrfs_vols

### LINKED ISSUE #4146: Enabling BTRFS snapshots while changing the mount point of the ESP partition.
### describe the request

Currently, it is possible to change the mount point of the ESP partition to something else than `/boot`, like `/efi` by manually editing the partition layout.

But doing so, it becomes impossible to enable BTRFS snapshots.

I believe the problem is that the [`has_default_btrfs_vols` ](https://github.com/archlinux/archinstall/blob/38462db2db5e5ed14c3b509738ac6c737cf924e9/archinstall/lib/models/device.py#L203)method which is called by the  [`_check_dep_btrfs`](https://github.com/archlinux/archinstall/blob/38462db2db5e5ed14c3b509738ac6c737cf924e9/archinstall/lib/disk/disk_menu.py#L118) method is too restrictive. 

I don't know why this check is done like that. Maybe only checking if there is at least one BTRFS partition would be enough ?

codefiles:
Fixes #4146

### DIFF
--- a/archinstall/lib/models/device.py
+++ b/archinstall/lib/models/device.py
@@ -201,13 +201,13 @@ class DiskLayoutConfiguration:
 		return config
 
 	def has_default_btrfs_vols(self) -> bool:
-		if self.config_type == DiskLayoutType.Default:
-			for mod in self.device_modifications:
-				for part in mod.partitions:
-					if part.is_create_or_modify():
-						if part.fs_type == FilesystemType.Btrfs:
-							if len(part.btrfs_subvols) > 0:
-								return True
+		for mod in self.device_modifications:
+			for part in mod.partitions:
+				if not (part.is_create_or_modify() and part.fs_type == FilesystemType.Btrfs):
+					continue
+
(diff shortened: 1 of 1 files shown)


svartkanin (approved):

archinstall/lib/models/device.py
```diff
@@ -201,13 +201,13 @@ def parse_arg(
 		return config
 
 	def has_default_btrfs_vols(self) -> bool:
```
svartkanin, line 203: The function name was intended to indicate that we're looking for the default disk config as well so it doesn't make much sense anymore now no?

codefiles: Maybe not but it is still checking for something that relates to the default disk config (a root Btrfs subvolume with the name `Path('@')`), right? What do you suggest?

svartkanin: Yeah not sure what a better name might be, I suppose if someone uses the default layout in the manual partitioning it would also be valid. Let's just leave it for now

PR #4151, opened 2026-01-20
Move script entry point

### LINKED ISSUE #4148: Relocate script entry point out of `__init__.py`
Script entry point:

https://github.com/archlinux/archinstall/blob/c7dbf0105ea89edca9a81655e3c7cb250de4f40a/pyproject.toml#L44-L45

- The entry point for the command `archinstall` is `run_as_a_module`, located in `archinstall/__init__.py`.
- The file `archinstall/__init__.py` is executed any time the package is imported.
- Importing `archinstall` triggers behavior associated with the script entry point.

Consider the following:

```python
import argparse

from archinstall.lib.models.device import SectorSize, Size, Unit

def main() -> None:
	unit_choices = [u.name for u in Unit]

	parser = argparse.ArgumentParser()
	parser.add_argument('value', type=int)
	parser.add_argument('unit', choices=unit_choices)
	parser.add_argument('--sector-size', type=int, default=512)

	mode = parser.add_mutually_exclusive_group()

	mode.add_argument('--to', choices=unit_choices)
	mode.add_argument('--highest', action='store_true')

	args = parser.parse_args()

	sector = SectorSize(args.sector_size, Unit.B)
	src_unit = Unit[args.unit]
	size = Size(args.value, src_unit, sector)

	if args.to:
		target_unit = Unit[args.to]
		converted = size.convert(target_unit, sector)
		print(converted.format_size(target_unit))
	elif args.highest:
		print(size.format_highest())
	else:
		sectors = size.convert(Unit.sectors, sector)
		print(f'sectors: {sectors.value}')

if __name__ == '__main__':
    main()
```

Expected usage:

```
$ python size_cli.py 1024 MiB --to GiB
1 GiB
$ python size_cli.py 2048 MiB --highest
2 GiB
$ python size_cli.py 1 GiB
sectors: 2097152
```

Instead the output will be like the following:

```
usage: size_cli.py [-h] [-v] [--config [CONFIG]] [--config-url [CONFIG_URL]] [--creds [CREDS]] [--creds-url [CREDS_URL]]
                    [--creds-decryption-key [CREDS_DECRYPTION_KEY]] [--silent] [--dry-run] [--script [SCRIPT]] [--mountpoint [MOUNTPOINT]] [--skip-ntp]
                    [--skip-wkd] [--skip-boot] [--debug] [--offline] [--no-pkg-lookups] [--plugin [PLUGIN]] [--skip-version-check] [--skip-wifi-check]
                    [--advanced] [--verbose]
size_cli.py: error: unrecognized arguments: 1024 MiB --to GiB
```

Since archinstall is meant to be usable as a library this is undesirable.

codefiles:
Resolves #4148


archinstall/main.py
```diff
@@ -0,0 +1,129 @@
...
+		Pacman.run('-Sy')
+	except Exception as e:
+		error('Failed to sync Arch Linux package database.')
+		if 'could not resolve host' in str(e).lower():
+			error('Most likely due to a missing network connection or DNS issue.')
+
+		error('Run archinstall --debug and check /var/log/archinstall/install.log for details.')
+
+		debug(f'Failed to sync Arch Linux package database: {e}')
+		sys.exit(1)
```
h8d13, line 57: shoulnd't these also be returns if they are received by `rc` ?

codefiles: This is a code move, not a refactor. It is best to keep a pull request focused on one thing and as small as possible that way it is easier to review amongst other benefits.

There is plenty that can be refactored here but that is irrelevant to this pull request.

Issue #4154, opened 2026-01-21
Cannot manually edit the parttion layout after a partition layout is set


dzamlo:
### Which ISO version are you using?

latest from git

### The installation log

```json
irrelevant
```

#### Description of the issue

I was testing with the latest git version of archinstall. And found an issue:

1. Open archinstall
2. Partition the disk using the best-effort default partition layout or manually
3. Try to manually edit the partition layout:
    i. Select partitioning
    ii. Select manual partitioning
    iii. Select the device you want to edit and press enter
 4. the interface go back to the previous screen instead of allowing you to edit the partition layout

This was working in the latest released version.

After using git bisect it seems that this issue was introduced by commit 0aca992ac506aca01a734b80008ffd01ca489c55 (PR #3997)

svartkanin:
Thanks I think there's a problem with the return value of the disk selection as the selected devices are the same it will just return https://github.com/svartkanin/archinstall/blob/38462db2db5e5ed14c3b509738ac6c737cf924e9/archinstall/lib/interactions/disk_conf.py?plain=1#L182.

svartkanin:
I think this should fix it https://github.com/archlinux/archinstall/pull/4158 if you wanna give this a try. 

I've also found another problem in the manual partitioning, when selecting `Cancel` it will override whatever the partitioning config was before. I raise a separate issue for that

PR #4155, opened 2026-01-21
fix: partionning menu

h8d13:
#4154

all i could find to bring back the disk menu, absolutely have no idea how it works since new changes


svartkanin:
This doesn't actually solve the problem, have you followed the steps from the issue?

h8d13:
it does ? at least first ime i can see the `btrfs | ext4 | ...` menu 

image

Like described in the issue, for the subsequent steps perhaps you can enlighten us

svartkanin:
Did you follow the steps described from the issue, I tried those on your patch and it still happens

PR #4158, opened 2026-01-22
Fix - disk partitioning menu

svartkanin:
This should fix the disk partitioning menu disk select #4154


dzamlo:
This seems to fix the issue, thanks.

Issue #4149, opened 2026-01-20
Switch from singleton pattern to dependency injection


codefiles:
### Singleton pattern

`config.py`

```python
from pathlib import Path

class Config:
    def __init__(self, path, value):
        self.path = path
        self.value = value

config = Config(Path('/mnt'), 42)
```
`general.py`

```python
from config import config

def use_config():
    content = (config.path / 'file').read_text()
    ...
```

`main.py`

```python
from config import config
from general import use_config

def main():
    use_config()
    print(f'config: {config.path}, {config.value}')

if name == '__main__':
    main()
```

### Dependency injection

`config.py`

```python
class Config:
    def __init__(self, path, value):
        self.path = path
        self.value = value
```

`general.py`

```python
def use_config(config):
    content = (config.path / 'file').read_text()
    ...
```

`main.py`

```python
from pathlib import Path

from config import Config
from general import use_config

def main():
    config = Config(Path('/mnt'), 42)
    use_config(config)
    print(f'config: {config.path}, {config.value}')

if name == '__main__':
    main()
```

svartkanin:
Generally I don't mind the pattern, with the dependency injection it becomes more tedious to pass things around though especially when the layers go deep. Also at some point it helped prevent circular dependencies. 

Were you thinking of changing all singletons or just specific ones?

codefiles:
All if possible, but handled incrementally.

codefiles:
Now with #4148 resolved, the example script I provided there works with current master branch. That is a good start and I created this issue following that one for a reason. While `main.py` is a more suitable location for the script entry point than `__init__.py`, it was not the sole issue.

With current master branch let's examine another example script:

```python
import argparse
from pathlib import Path

from archinstall.lib.installer import Installer
from archinstall.lib.models.device import DiskLayoutConfiguration, DiskLayoutType

def main() -> None:
	parser = argparse.ArgumentParser()
	parser.add_argument('--timezone')
	parser.add_argument('--hostname')
	parser.add_argument('mountpoint', type=Path)
	args = parser.parse_args()

	installer = Installer(
		args.mountpoint,
		DiskLayoutConfiguration(DiskLayoutType.Pre_mount),
	)

	if args.timezone:
		installer.set_timezone(args.timezone)

	if args.hostname:
		installer.set_hostname(args.hostname)

if __name__ == '__main__':
	main()
```

Expected usage:

```
# python pre_mount_cli.py --timezone UTC --hostname archlinux /mnt
[...]

PR #4161, opened 2026-01-22
Refactor wifi_handler to use dependency injection

codefiles:
Part of #4149

### DIFF
--- a/archinstall/main.py
+++ b/archinstall/main.py
@@ -10,7 +10,7 @@ import traceback
 from archinstall.lib.args import arch_config_handler
 from archinstall.lib.disk.utils import disk_layouts
 from archinstall.lib.general import running_from_host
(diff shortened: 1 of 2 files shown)


archinstall/main.py
```diff
@@ -79,7 +79,12 @@ def run() -> int:
 	_log_sys_info()
 
 	if not arch_config_handler.args.offline:
-		if not _check_online():
+		if not arch_config_handler.args.skip_wifi_check:
+			wifi_handler = WifiHandler()
+		else:
+			wifi_handler = None
+
+		if not _check_online(wifi_handler):
```
svartkanin, line 87: Is there a benefit in making this an optional parameter instead of creating the instance inside `_check_online` and avoiding the overhead?

codefiles: Yes, testability.

Issue #2684, opened 2024-09-14
EFISTUB not installed if EFI partition not mounted at /boot


Rua:
### Which ISO version are you using?

2024-09-01

### The installation log

```json
https://0x0.st/Xx5T.log
```

#### Description of the issue

I installed in a VM with a pre-mounted partition layout, in which the EFI system partition is mounted at /efi, rather than at /boot. Using EFISTUB as boot option, not using a UKI.

No error appears in the log, but after the installation completes "without any errors", I reboot and no EFI boot menu option is available. If the installation ISO is left enabled, it boots back into that, otherwise I end up in an EFI shell. If I try the same installation with the EFI partition mounted at /boot, the boot menu option is installed correctly.

While I'm not well versed in how `archinstall` works internally, what I found (archinstall/lib/installer.py:1324) seemed like it assumes that the EFI partition is mounted at /boot, so that the kernel is automatically placed in the EFI partition root. But if it's not in /boot, the kernel would need to be copied there first?

#### Virtual machine config:

```xml
<domain type="kvm">
  <name>archlinux-2</name>
  <uuid>1bc157a6-68c2-456c-92f3-6063f3df8456</uuid>
  <metadata>
    <libosinfo:libosinfo xmlns:libosinfo="http://libosinfo.org/xmlns/libvirt/domain/1.0">
      <libosinfo:os id="http://archlinux.org/archlinux/rolling"/>
    </libosinfo:libosinfo>
  </metadata>
  <memory unit="KiB">8388608</memory>
  <currentMemory unit="KiB">8388608</currentMemory>
  <vcpu placement="static">8</vcpu>
  <os firmware="efi">
    <type arch="x86_64" machine="pc-q35-8.2">hvm</type>
    <firmware>
      <feature enabled="no" name="enrolled-keys"/>
      <feature enabled="no" name="secure-boot"/>
    </firmware>
    <loader readonly="yes" type="pflash">/usr/share/OVMF/OVMF_CODE_4M.fd</loader>
    <nvram template="/usr/share/OVMF/OVMF_VARS_4M.fd">/var/lib/libvirt/qemu/nvram/archlinux-2_VARS.fd</nvram>
  </os>
  <features>
    <acpi/>
    <apic/>
    <vmport state="off"/>
  </features>
  <cpu mode="host-passthrough" check="none" migratable="on"/>
  <clock offset="utc">
    <timer name="rtc" tickpolicy="catchup"/>
[...]

Issue #1431, opened 2022-08-24
Is efistub intended to work as is?


miusaky:
I've been attempting to run the installer with the efistub boot option on two different systems, both booted in uefi, both have failed. 

Is this option intended to work without manual intervention? Do I need to take additional steps before running the installer? Couldn't find documentation or previous issues explaining it. Logs below.

```

Hardware model detected: ASUS System Product Name; UEFI mode: True
Processor model detected: AMD Ryzen 5 5600X 6-Core Processor
Memory statistics: 15653476 available out of 16285500 total installed
Could not detect virtual system: ['/usr/bin/systemd-detect-virt'] exited with abnormal exit code [256]: b'none\r\n'
System is not running in a VM: ['/usr/bin/systemd-detect-virt'] exited with abnormal exit code [256]: b'none\r\n'
Virtualization detected: None; is VM: None
Graphics devices detected: dict_keys(['NVIDIA Corporation GP104 [GeForce GTX 1080] (rev a1)'])
[...]

PR #4231, opened 2026-02-15
fix efistub bootloader to use backslashes

h8d13:
As seen in wiki example: 
https://wiki.archlinux.org/title/EFI_boot_stub#efibootmgr

```shell 
efibootmgr --create --disk /dev/sdX --part Y --label "Arch Linux" --loader /vmlinuz-linux \
--unicode 'root=block_device_identifier rw initrd=\initramfs-linux.img'
```

Was just wondering why `EFIstub` never worked to boot and found this :)

#1431 #2684

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -1620,7 +1620,7 @@ class Installer:
 			loader = '/vmlinuz-{kernel}'
 			# EFI standards stipulate backslashes
 			entries = (
-				'initrd=\\initramfs-{kernel}.img',
+				r'initrd=\initramfs-{kernel}.img',
 				*self._get_kernel_params(root),
 			)
 
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -1618,8 +1618,7 @@ class Installer:
 
 		if not uki_enabled:
 			loader = '/vmlinuz-{kernel}'
-			# python escaping for a single backslash
-			# according to wiki EFISTUB when passed to --unicode
+			# EFI standards stipulate backslashes
 			entries = (
 				'initrd=\\initramfs-{kernel}.img',
 				*self._get_kernel_params(root),
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -1618,7 +1618,7 @@ class Installer:
 
 		if not uki_enabled:
 			loader = '/vmlinuz-{kernel}'
-			# python escaping for a single forward slash
+			# python escaping for a single backslash
 			# according to wiki EFISTUB when passed to --unicode
 			entries = (
 				'initrd=\\initramfs-{kernel}.img',
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -1618,9 +1618,10 @@ class Installer:
 
 		if not uki_enabled:
 			loader = '/vmlinuz-{kernel}'
-
+			# python escaping for a single forward slash
+			# according to wiki EFISTUB when passed to --unicode
 			entries = (
-				'initrd=/initramfs-{kernel}.img',
+				'initrd=\\initramfs-{kernel}.img',
 				*self._get_kernel_params(root),
 			)
 


codefiles:
<strike>This does not fix anything, see the note at the bottom of [Unified Extensible Firmware Interface#efibootmgr](https://wiki.archlinux.org/title/Unified_Extensible_Firmware_Interface#efibootmgr).</strike>

Edit: The note is not applicable to kernel parameters.

h8d13:
> This does not fix anything, see the note at the bottom of [Unified Extensible Firmware Interface#efibootmgr](https://wiki.archlinux.org/title/Unified_Extensible_Firmware_Interface#efibootmgr).

Right so try without the patch and then with it? I do not think you had the time to do so ? Are you actively testing changes ? Also did you look at linked issues ?

Like we have never seen firmware quirks, and shouldn't respect the actual specification of provided example (because explicit is better than implicit :) ?

The conversion works for `--loader` not `--unicode` that goes straight to the firmware as is (hence the `'content here'`. Have a good rest of your sunday, Hade

codefiles:
You are correct, the note I mentioned does not apply to the kernel parameters like `initrd`.

Relevant sections of the ArchWiki that support this change:

- https://wiki.archlinux.org/title/Kernel_parameters#Parameter_list
- https://wiki.archlinux.org/title/EFI_boot_stub#Booting_an_EFI_boot_stub

I have caught this before in the past but never submitted changes because I misunderstood the note. I also tested this same thing in the past but could not get it to work in my test environment.

h8d13:
You can also also see it directly here: https://github.com/rhboot/efibootmgr/blob/main/src/efi.c#L330-L361
Which converts `UTF-8` → `UCS-2` directly https://uefi.org/specs/UEFI/2.10/13_Protocols_Media_Access.html#file-names

archinstall/lib/installer.py
```diff
@@ -1618,9 +1618,10 @@ def _add_efistub_bootloader(
 
 		if not uki_enabled:
 			loader = '/vmlinuz-{kernel}'
-
+			# python escaping for a single backslash
+			# according to wiki EFISTUB when passed to --unicode
```
codefiles:
```suggestion
			# EFI standards stipulate backslashes
```

This might be clearer. It is not on account of the ArchWiki but the EFI standard the wiki is relaying information from.

archinstall/lib/installer.py
```diff
@@ -1618,9 +1618,10 @@ def _add_efistub_bootloader(
...
 			entries = (
-				'initrd=/initramfs-{kernel}.img',
+				'initrd=\\initramfs-{kernel}.img',
```
codefiles:
```suggestion
				r'initrd=\initramfs-{kernel}.img',
```

I would use a raw string but it does not matter.

codefiles (approved):
Left some comments that are extremely minor. Good to be merged as is.

codefiles:
I erroneously changed this with bc3b3a3. Here is the relevant line before that change:

https://github.com/archlinux/archinstall/blob/332ec0d6236ca863cb6a2101849555935066549f/archinstall/lib/installer.py#L1096

That commit is from 2023 but #1431 is from 2022. The log from that issue shows version 2.5.0; backslashes were intact:

https://github.com/archlinux/archinstall/blob/c75e6a1da3c243126eaf59ee12f0b2464449e5e2/archinstall/lib/installer.py#L979

In #2684 it was reported that it works when the ESP is `/boot` but not when it is `/efi`. In either of those cases this same line would get used in the kernel parameters, when not using UKI as they reported having done. The ISO given is 2024-09-01 so the code with the erroneous forward slashes would have been in use, eliminating that as a possible cause to the issue.

> Also did you look at linked issues ?

I have seen them before but never determined the source of the issue.

Though I agree with this change, it can be ruled out as fixing the linked issues. If the EFI boot stub option was broken I would expect there would be more than these two old issues with no other users commenting on it.

h8d13:
I can think of dozens of times I've told people to use Grub/Limine or systemd-boot because this one was flaky (when there are 2 reports here, means more on other forums regardless, bbs/reddit/etc, and even from me testing it several times, aside from people not reporting or who do not use efistub) 

I like it (technically one less pkg lol), is why I was trying to fix it :D 

Then I tried it on alpine and it worked flawlessly, so I was was wondering and just thought what if it's something to do with python formatting (and how its passed to the tool), and no worries, shit happens

Also nice UKI patch lore :dancers:

codefiles:
Yes, I am very aware that other places exist that issues with archinstall get discussed but as I demonstrated the two linked issues would not be related to the changes here. That would mean 0 reports of this causing an issue here then not 2.

The standard for `initrd` is indisputably to use backslashes but using forward slashes causing breakage has not been demonstrated. I think it is likely most UEFI implementations accept forward slashes.

From one of your earlier comments:

> Right so try without the patch and then with it?

I just tested and can boot with no problems when forward slashes are used in the `initrd` path. After booting `/proc/cmdline` contained the kerenel parameter `initrd=/initramfs-linux.img` with the starting forward slash. I tested the patch, and after booting `/proc/cmdline` contained the kernel parameter `initrd=\initramfs-linux.img` with the starting backslash.

Are you able to reproduce a failed boot when using forward slashes? It does not look like you have ever stated so directly.

h8d13:
You are testing in VMs or weird hardware ? 

Im not sure how being more correct can be worse than hoping the firmware implementation is well done for others' hardware. 

A lot of these tools are known for quirks with firmware equivalent implementations or straight up not following specs.  Like msi boards not booting without removable bootloader

> I think it is likely most UEFI implementations accept forward slashes.

Emphasis on "most"

codefiles:
> You are testing in VMs or weird hardware ?

QEMU/KVM

> Im not sure how being more correct can be worse than hoping the firmware implementation is well done for others' hardware.

It can't. I am in favor of merging the patch friend. If I had the ability to merge it, I would have already. I desire knowledge and that makes me want to see the evidence of a system that fails to boot when the `initrd` contains forward slashes. This does not mean I in anyway object to the patch in the absence of that evidence. The patch enforces the standard and that is enough reason to merge it.

> A lot of these tools are known for quirks with firmware equivalent implementations or straight up not following specs. Like msi boards not booting without removable bootloader

Good point.

svartkanin (approved):

PR #4232, opened 2026-02-16
Move mount() to disk.utils

codefiles:
Part of #4149

This is meant to work towards decoupling `device_handler` from `installer.py` so that importing it will not cause devices to be probed. None of the `device_handler` methods used in `installer.py` operate on properties of `device_handler` so they do not need a device probe to function.

### DIFF
--- a/archinstall/lib/disk/utils.py
+++ b/archinstall/lib/disk/utils.py
@@ -5,7 +5,7 @@ from pydantic import BaseModel
 from archinstall.lib.command import SysCommand
 from archinstall.lib.exceptions import DiskError, SysCallError
 from archinstall.lib.models.device import LsblkInfo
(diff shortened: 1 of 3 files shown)


archinstall/lib/disk/utils.py
```diff
@@ -67,12 +67,12 @@ def get_lsblk_output() -> LsblkOutput:
 
 def find_lsblk_info(
 	dev_path: Path | str,
-	info: list[LsblkInfo],
+	info_list: list[LsblkInfo],
```
codefiles, line 70: This would seem like an unrelated change but with `info` being added to the imports just above this, pylint will give the following:
https://pylint.pycqa.org/en/latest/user_guide/messages/warning/redefined-outer-name.html

Issue #4223, opened 2026-02-12
Unable to disable service iwd.service - Unit iwd.service does not exist


palanthis:
### Which ISO version are you using?

Feb 1st

### The installation log

```json
[2026-02-12 04:32:46] - DEBUG - Hardware model detected: LENOVO 20JWS0C600; UEFI mode: True
[2026-02-12 04:32:46] - DEBUG - Processor model detected: Intel(R) Core(TM) i7-6600U CPU @ 2.60GHz
[2026-02-12 04:32:46] - DEBUG - Memory statistics: 29569472 available out of 32572924 total installed
[2026-02-12 04:32:46] - DEBUG - Could not detect virtual system: ['/usr/bin/systemd-detect-virt'] exited with abnormal exit code [1]: none

[2026-02-12 04:32:46] - DEBUG - System is not running in a VM: ['/usr/bin/systemd-detect-virt'] exited with abnormal exit code [1]: none

[2026-02-12 04:32:46] - DEBUG - Virtualization detected: None; is VM: False
[2026-02-12 04:32:46] - DEBUG - Graphics devices detected: dict_keys(['Intel Corporation Skylake-U GT2 [HD Graphics 520] (rev 07)'])
[2026-02-12 04:32:46] - DEBUG - Disk states before installing:
{
    "blockdevices": [
        {
            "name": "loop0",
            "path": "/dev/loop0",
            "pkname": null,
            "log_sec": 512,
            "size": "1837 MiB",
            "pttype": null,
            "ptuuid": null,
            "rota": false,
            "tran": null,
            "partn": null,
            "partuuid": null,
            "parttype": null,
            "uuid": null,
            "fstype": "squashfs",
            "fsver": "4.0",
            "fsavail": 0,
            "fsuse_percentage": "100%",
            "type": "loop",
            "mountpoint": "/run/archiso/airootfs",
            "mountpoints": [
                "/run/archiso/airootfs"
            ],
            "fsroots": [
                "/"
            ],
            "children": []
        },
        {
            "name": "sda",
            "path": "/dev/sda",
            "pkname": null,
            "log_sec": 512,
            "size": "244198 MiB",
            "pttype": "gpt",
            "ptuuid": "9373df54-a15c-4636-8157-79fb32d61c8e",
            "rota": false,
            "tran": "sata",
            "partn": null,
            "partuuid": null,
            "parttype": null,
            "uuid": null,
            "fstype": null,
            "fsver": null,
            "fsavail": null,
[...]

PR #4239, opened 2026-02-16
Fix: IWD switch case

h8d13:
also less hidden documentation ;)  #4223

### DIFF
--- a/archinstall/lib/network/network_handler.py
+++ b/archinstall/lib/network/network_handler.py
@@ -16,25 +16,24 @@ class NetworkHandler:
 					enable_services=True,  # Sources the ISO network configuration to the install medium.
 				)
 			case NicType.NM | NicType.NM_IWD:
-				# Install NetworkManager package for both cases
 				packages = ['networkmanager']
-				# Default back-end only for non-iwd
-				if network_config.type == NicType.NM:
-					packages.append('wpa_supplicant')
+				if network_config.type == NicType.NM:  # default
+					packages.extend(['wpa_supplicant', 'wireless_tools'])
+				else:  # iwd better support for some wifi cards
+					packages.append('iwd')
 
 				installation.add_additional_packages(packages)
 
-				# Desktop profile -> Always add applet
-				if profile_config and profile_config.profile:
-					if profile_config.profile.is_desktop_profile():
-						installation.add_additional_packages('network-manager-applet')
+				# in any case if desktop
+				if profile_config and profile_config.profile and profile_config.profile.is_desktop_profile():
+					installation.add_additional_packages('network-manager-applet')
 
 				installation.enable_service('NetworkManager.service')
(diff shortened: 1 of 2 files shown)


archinstall/lib/network/network_handler.py
```diff
@@ -16,25 +16,24 @@ def install_network_config(
 					enable_services=True,  # Sources the ISO network configuration to the install medium.
 				)
 			case NicType.NM | NicType.NM_IWD:
-				# Install NetworkManager package for both cases
 				packages = ['networkmanager']
-				# Default back-end only for non-iwd
-				if network_config.type == NicType.NM:
-					packages.append('wpa_supplicant')
+				if network_config.type == NicType.NM:  # default
+					packages.extend(['wpa_supplicant', 'wireless_tools'])
```
svartkanin, line 21: according to the wiki `wireless_tools` is deprecated https://wiki.archlinux.org/title/Network_configuration/Wireless#Utilities so I don't think we should be installing that any longer

archinstall/lib/network/network_handler.py
```diff
@@ -16,25 +16,24 @@ def install_network_config(
...
+				# in any case if desktop
+				if profile_config and profile_config.profile and profile_config.profile.is_desktop_profile():
+					installation.add_additional_packages('network-manager-applet')
 
 				installation.enable_service('NetworkManager.service')
+
+				# special handling for NM iwd service + conf
 				if network_config.type == NicType.NM_IWD:
-					# NM_IWD special handling
 					installation.configure_nm_iwd()
```
svartkanin, line 35: I know this was here before this PR, but this method should not be part of the installer but rather this handler

archinstall/lib/network/network_handler.py
h8d13: Right can you not edit these directly ? I see codefiles doing that and since I'm unsure you might change your mind 3 times

archinstall/lib/network/network_handler.py
h8d13: Right so it would be for another PR this fixes the immediate problem, you also refactored this file some days ago :)

archinstall/lib/network/network_handler.py
svartkanin: Why would I edit your PRs. I'm leaving feedback for you to think about

archinstall/lib/network/network_handler.py
svartkanin: I have moved the entire handler into its own file but the original code was written by you https://github.com/h8d13/archinstoo/commit/c1eae10e93a94edbd0f9b94da4ad3d1c73ae993c#diff-283d51110d004467e254215ad93e4dfcf0188beb22d9f878bb7088c12076dd2eR771

PR #4245, opened 2026-02-18
Move LVM helpers to dedicated module

codefiles:
Part of #4149, completes #4232

### DIFF
--- a/archinstall/lib/disk/device_handler.py
+++ b/archinstall/lib/disk/device_handler.py
@@ -1,10 +1,7 @@
-import json
 import logging
 import os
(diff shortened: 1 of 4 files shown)


codefiles:
I tried to keep the changes as minimal as possible, that is why I did not move more LVM helpers. I also originally had these in `disk/utils.py` but figured using a dedicated module `disk/lvm.py` was cleaner. Open to any suggestions but ultimately this is intended to finish with removing `device_handler` from `installer.py` more than anything else.

svartkanin (approved):

PR #4263, opened 2026-02-27
Fix a reportUnnecessaryIsInstance Pyright warning

correctmost:
## PR Description:

This PR fixes the following warning:

```
archinstall/archinstall/lib/models/device.py:321:10 - error: Unnecessary isinstance call; "SectorSize" is always an instance of "SectorSize" (reportUnnecessaryIsInstance)
```

### DIFF
--- a/archinstall/lib/models/device.py
+++ b/archinstall/lib/models/device.py
@@ -317,10 +317,6 @@ class Size:
 	unit: Unit
 	sector_size: SectorSize
 
-	def __post_init__(self) -> None:
-		if not isinstance(self.sector_size, SectorSize):
-			raise ValueError('sector size must be of type SectorSize')
-
 	def json(self) -> _SizeSerialization:
 		return {
 			'value': self.value,


svartkanin (approved):

codefiles:
I've thought about removing the following `__post_init__` in the past.

https://github.com/archlinux/archinstall/blob/390b4959bdb91b9a86ba9d5a9a789569053e4dd1/archinstall/lib/luks.py#L43-L45

Here is the `luks_dev_path` type annotation:

https://github.com/archlinux/archinstall/blob/390b4959bdb91b9a86ba9d5a9a789569053e4dd1/archinstall/lib/luks.py#L16-L22

It is checking for `None` instead of using `isinstance` so it does not trigger the reportUnnecessaryIsInstance Pyright warning. This is pretty much the same design defect though, an unnecessary type check in a post-init, right?

correctmost:
> This is pretty much the same design defect though, an unnecessary type check in a post-init, right?

Yeah :).  I would like to enable `mypy --strict-equality-for-none` to catch these extraneous checks, but there are a bunch of places that need to be cleaned up first:

```
$ mypy --strict-equality-for-none
[...snip...]
archinstall/lib/luks.py:44: error: Non-overlapping identity check (left operand type: "Path", right operand type: "None")  [comparison-overlap]
[...snip...]
```

codefiles:
Thanks!

PR #4279, opened 2026-03-02
Remove support for NTFS root file system

### LINKED ISSUE #4276: Remove support for NTFS root file system
This feature was added in 0c96ae0 following the introduction of the NTFS3 driver in Linux Kernel version 5.15.

Reasons for proposal of removal:

- It is not a native Linux file system and not intended to be used as a Linux root file system.
  The contributor stated, "not really practical" in the pull request comment.
- Unnecessary maintenance burden.
- Not maintained for a significant duration.
  After 00b0ae7 mounting was broken; `get_mount_fs_type()` was moved to `FilesystemType.fs_type_mount()` and this is no longer being used in `mount()`. See https://wiki.archlinux.org/title/NTFS#Explicit_file_system_type_required_to_mount

codefiles:
Closes #4276

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -1194,7 +1194,7 @@ class Installer:
 
 			kernel_parameters.append('rw')
 
-		kernel_parameters.append(f'rootfstype={root.safe_fs_type}')
+		kernel_parameters.append(f'rootfstype={root.safe_fs_type.value}')
 		kernel_parameters.extend(self._kernel_params)
 
 		debug(f'kernel parameters: {" ".join(kernel_parameters)}')


archinstall/lib/models/device.py
codefiles, line 791: Existing partitions with NTFS file systems would not display as `ntfs` in disk menus if the line `Ntfs = 'ntfs'` above was removed, right?

https://github.com/archlinux/archinstall/blob/b809c84ad5ed9d383ae091d16f4c261308137b8c/archinstall/lib/disk/device_handler.py#L142-L158

archinstall/lib/installer.py
```diff
@@ -1199,7 +1194,7 @@ def _get_kernel_params(
 
 			kernel_parameters.append('rw')
 
-		kernel_parameters.append(f'rootfstype={root.safe_fs_type.fs_type_mount}')
+		kernel_parameters.append(f'rootfstype={root.safe_fs_type}')
```
svartkanin: Wouldn't this format it as 
```
rootfstype=FilesystemType.Ext4
```

The `fs_type_mount` function was also handling `Fat32` type respectively with `vfat`

codefiles: > Wouldn't this format it as
> 
> ```
> rootfstype=FilesystemType.Ext4
> ```

Yes, that is a mistake. I'm glad you caught that.

> The `fs_type_mount` function was also handling `Fat32` type respectively with `vfat`

The `fs_type_mount` function was only really being used for the `rootfstype` kernel parameter. FAT32 should not be used as a root file system so removing this should be fine.

PR #4280, opened 2026-03-03
Refactor arch_config_handler to use DI

codefiles:
Part of #4149

### DIFF
--- a/archinstall/lib/args.py
+++ b/archinstall/lib/args.py
@@ -549,6 +549,3 @@ class ArchConfigHandler:
 				clean_args[key] = val
 
 		return clean_args
-
-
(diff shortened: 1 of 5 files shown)


svartkanin (approved):

svartkanin:
@codefiles thanks for all the updates, 
Would it make sense to transform the scripts into classes with an abstract something like 

```
class Script(ABC):
     def main(self):
        pass
        
class Guided(Script):
     def __init__(arch_config_handler: ArchConfigHandler | None = None) -> None:
          ...
     
     @override
     def main(self) -> None:
         ....
```

any thoughts?

codefiles:
I'm not sure but I would recommend waiting till `device_handler` is also switched to dependency injection before making a change like that.

PR #4283, opened 2026-03-04
Move all LVM helpers to dedicated module

codefiles:
Follow-up to #4245

### DIFF
--- a/archinstall/lib/disk/lvm.py
+++ b/archinstall/lib/disk/lvm.py
@@ -1,9 +1,11 @@
 import json
 import time
+from collections.abc import Iterable
 from pathlib import Path
 from typing import Literal, overload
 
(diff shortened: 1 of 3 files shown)


archinstall/lib/disk/lvm.py
```diff
@@ -116,6 +118,13 @@ def lvm_vol_change(vol: LvmVolume, activate: bool) -> None:
 	SysCommand(cmd)
 
 
+def lvm_export_vg(vg: LvmVolumeGroup) -> None:
+	cmd = f'vgexport {vg.name}'
+
+	debug(f'vgexport: {cmd}')
+	SysCommand(cmd)
```
codefiles, line 125: This is unused.

PR #4290, opened 2026-03-06
user: add required birth date field to non-root user creation

dylanmtaylor:
Add a required birth date prompt (YYYY-MM-DD) to the user creation flow, stored as a systemd userdb JSON drop-in at `/etc/userdb/<user>.user` on the target system.

## Motivation

Recent age verification laws in California (AB-1043), Colorado (SB26-051), Brazil (Lei 15.211/2025), etc. require platforms to verify user age. Collecting birth date at install time ensures Arch Linux is compliant with these regulations.

## Changes

- `users.py` — add `birth_date` field to `User` dataclass, serialization, and config parsing
- `user_menu.py` — add required birth date prompt with validation after the password step
- `installer.py` — write userdb JSON drop-in with `userName` and `birthDate` after user creation

## Integration

This integrates with [systemd PR #40954](https://github.com/systemd/systemd/pull/40954) which adds `birthDate` support to JSON user records. The drop-in is read by systemd's `io.systemd.DropIn` service and displayed by `userdbctl`.

### DIFF
--- a/archinstall/lib/user/user_menu.py
+++ b/archinstall/lib/user/user_menu.py
@@ -1,4 +1,5 @@
 import re
+from datetime import UTC, date, datetime
 from typing import override
 
 from archinstall.lib.menu.helpers import Confirmation, Input
@@ -91,6 +92,12 @@ class UserList(ListManager[User]):
 			return None
 
 		header += f'{tr("Password")}: {password.hidden()}\n'
+
+		birth_date = self._ask_birth_date(header)
+
+		if birth_date:
+			header += f'{tr("Birth date")}: {birth_date}\n'
+
 		prompt = f'{header}\n' + tr('Should "{}" be a superuser (sudo)?\n').format(username)
 
 		result = Confirmation(
@@ -105,7 +112,35 @@ class UserList(ListManager[User]):
 			case _:
 				raise ValueError('Unhandled result type')
 
-		return User(username, password, sudo)
+		return User(username, password, sudo, birth_date=birth_date)
+
+	def _validate_birth_date(self, value: str | None) -> str | None:
+		if not value:
+			return tr('Birth date is required')
+		try:
+			dt = date.fromisoformat(value)
+		except ValueError:
+			return tr('Invalid date format. Use YYYY-MM-DD')
+		if dt > datetime.now(tz=UTC).date():
+			return tr('Birth date cannot be in the future')
+		if dt.year < 1900:
+			return tr('Birth date year must be 1900 or later')
+		return None
+
+	def _ask_birth_date(self, header: str) -> str:
+		prompt = f'{header}\n' + tr('Enter birth date (YYYY-MM-DD, e.g. 2000-01-01)')
+
+		result = Input(
+			prompt,
+			allow_skip=False,
+			validator_callback=self._validate_birth_date,
+		).show()
+
+		match result.type_:
+			case ResultType.Selection:
+				return result.get_value() or ''
+			case _:
+				raise ValueError('Unhandled result type')
 
 
 def select_users(prompt: str = '', preset: list[User] = []) -> list[User]:
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -1,4 +1,5 @@
 import glob
+import json
 import os
 import platform
 import re
@@ -1942,6 +1943,20 @@ class Installer:
 
 		if user.sudo:
 			self.enable_sudo(user)
+
+		if user.birth_date:
+			self._write_userdb_dropin(user)
+
+	def _write_userdb_dropin(self, user: User) -> None:
+		userdb_dir = self.target / 'etc' / 'userdb'
+		userdb_dir.mkdir(parents=True, exist_ok=True)
+
+		dropin = userdb_dir / f'{user.username}.user'
+		record = {'userName': user.username, 'birthDate': user.birth_date}
+		dropin.write_text(json.dumps(record))
+		dropin.chmod(0o644)
+
+		info(f'Wrote userdb drop-in for {user.username} with birth date')
 
 	def set_user_password(self, user: User) -> bool:
 		info(f'Setting password for {user.username}')
--- a/archinstall/lib/models/users.py
+++ b/archinstall/lib/models/users.py
@@ -107,6 +107,7 @@ UserSerialization = TypedDict(
 		'sudo': bool,
 		'groups': list[str],
 		'enc_password': str | None,
+		'birth_date': NotRequired[str],
 	},
 )
 
@@ -158,6 +159,7 @@ class User:
 	password: Password
 	sudo: bool
 	groups: list[str] = field(default_factory=list)
+	birth_date: str = ''
 
 	@override
 	def __str__(self) -> str:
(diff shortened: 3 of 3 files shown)


AdrianVovk:
Note that this isn't how userdb drop-ins work. As I mentioned [elsewhere](https://github.com/canonical/ubuntu-desktop-provision/pull/1338), the userdb drop-in plugin will race with the NSS plugin for who gets to report information about the user.

Note that the [portal](https://github.com/flatpak/xdg-desktop-portal/pull/1922) backend would have to prompt for age if unknown anyways, for user accounts that were set up prior to the implementation of all of this. Thus for a non-graphical tool like archinstall, it's probably best to leave this information un-collected. The first time an app requests the information, the GUI (whatever it may end up being) will prompt the user to fill it in

dylanmtaylor:
> Note that this isn't how userdb drop-ins work. As I mentioned [elsewhere](https://github.com/canonical/ubuntu-desktop-provision/pull/1338), the userdb drop-in plugin will race with the NSS plugin for who gets to report information about the user.
> 
> Note that the [portal](https://github.com/flatpak/xdg-desktop-portal/pull/1922) backend would have to prompt for age if unknown anyways, for user accounts that were set up prior to the implementation of all of this. Thus for a non-graphical tool like archinstall, it's probably best to leave this information un-collected. The first time an app requests the information, the GUI (whatever it may end up being) will prompt the user to fill it in

Prompting after installation when an app requests an age doesn't comply with the regulation. The key one is California's  AB-1043, which requires operating system providers to collect age information during account/device setup and expose an API that categorizes users into age brackets. Per [Goodwin Law's analysis](https://www.goodwinlaw.com/en/insights/publications/2025/12/alerts-practices-dpc-californias-year-end-privacy-wave), the law "requires operating system providers to require users to provide age information as they set up their accounts."

codefiles:
Why would you waste time with this? It's not April 1st yet.

AdrianVovk:
There is an exception for accounts set up before Jan 1, 2027:

> With respect to a device for which account setup was completed before January 1, 2027, an operating system provider shall, before July 1, 2027, provide an accessible interface that allows an account holder to indicate the birth date, age, or both, of the user of that device for the purpose of providing a signal regarding the user’s age bracket to applications available in a covered application store

Basically: if the account was set up before Jan 1 2027, the OS will need to prompt the user to fill in the information. This is what the portal will have implemented.

But yes, I suppose after Jan 1 2027 you're right. Archinstall is where account setup happens, and so archinstall would have to fill in this data. Which would suck because different desktops might choose to store this data in different places. Perhaps archinstall needs to stop setting up the user account, and the chosen desktop's initial-setup tool will need to run and handle account creation

dylanmtaylor:
> Why would you waste time with this? It's not April 1st yet.

Unfortunately, the new laws go into effect January 1, 2027. They requires operating system providers to collect age information during account setup. There are very high penalties for non-compliance to the point they are an existential threat to free and open source software that doesn't meet the requirements.

dylanmtaylor:
> There is an exception for accounts set up before Jan 1, 2027:
> 
> > With respect to a device for which account setup was completed before January 1, 2027, an operating system provider shall, before July 1, 2027, provide an accessible interface that allows an account holder to indicate the birth date, age, or both, of the user of that device for the purpose of providing a signal regarding the user’s age bracket to applications available in a covered application store
> 
> Basically: if the account was set up before Jan 1 2027, the OS will need to prompt the user to fill in the information. This is what the portal will have implemented.
> 
> But yes, I suppose after Jan 1 2027 you're right. Archinstall is where account setup happens, and so archinstall would have to fill in this data. Which would suck because different desktops might choose to store this data in different places. Perhaps archinstall needs to stop setting up the user account, and the chosen desktop's initial-setup tool will need to run and handle account creation

This isn't an option; not all desktops have initial-setup tools. Some people want a minimal install with no desktop.

dylanmtaylor:
> But yes, I suppose after Jan 1 2027 you're right. Archinstall is where account setup happens, and so archinstall would have to fill in this data

Right. I'd like to have a fix in place before then.

codefiles:
Let's get ahead of the curve and add personal identification with face scans too, right?

AdrianVovk:
> Some people want a minimal install with no desktop.

It is not clear that the law applies at all in this case

dylanmtaylor:
> Let's get ahead of the curve and add personal identification with face scans too, right?

As the law is currently written, a birth date just needs to be provided. No validation is required. This is pretty non-intrusive, all things considered.

codefiles:
Let's wait till this is addressed by Arch Linux leadership.

Also, archinstall should conform to the ArchWiki and so it should follow the [users and groups](https://wiki.archlinux.org/title/Users_and_groups) article.

archinstall/lib/user/user_menu.py
```diff
@@ -91,6 +92,12 @@ def _add_user(self) -> User | None:
 			return None
 
 		header += f'{tr("Password")}: {password.hidden()}\n'
+
+		birth_date = self._ask_birth_date(header)
```
svartkanin, line 96: Creating users is not mandatory, one can also only set a root password

svartkanin:
`archinstall` is a simple wrapper to provide convenience for installing Arch Linux. No user is required to use it and can simply follow the wiki to install the system to their liking. 
How is Arch Linux going to address that? How would one even enforce that, prevent `pacstrap` or `pacman` from executing unless a birth date is set?

I understand that this law is upcoming but solving this in `archinstall` alone doesn't make any sense as it's just a convenience wrapper. Unless this is actually enforced on the installation level either `pacstrap` or `pacman` then I don't see how adding this would be compliant and at that point, yes, `archinstall` will need to provide the parameter input to allow the installation.

archinstall/lib/user/user_menu.py
dylanmtaylor: I know this, you know this, but the laws that just got adopted are written in a stupid way that possibly doesn't exempt TUI installers. I don't know if will be "safe" for the project to not ask for a birth date when adding a user when these laws take effect in 2027.

dylanmtaylor:
> How is Arch Linux going to address that? How would one even enforce that, prevent `pacstrap` or `pacman` from executing unless a birth date is set?

Truthfully, they won't. This will be completely ineffective at preventing anyone from lying about their age or using tooling on the live media before making a user. 

> at that point, yes, archinstall will need to provide the parameter input to allow the installation.

After reading the bill text, this is the conclusion I came to - arch install is an OS installer, the law asks for users to provide birth date when installing an OS. Is that going to be hilariously pointless and ineffective? Yes.

CelestifyX:
Bro, I'm fucking amazed by you.

Torxed:
I appreciate the work ahead of time, and the law is the law.
@svartkanin raised this PR internally within staff channels, and the feedback is that we'll wait until there's an overall stance from Arch Linux on this before merging this, and preferably involve legal representatives on this matter on what the best way forward is for us.

But from a personal reflection it's clear that there's a disconnect between law makers intent and how things like this will be implemented in reality, and once a law is in place - we might have to implement inconvenient things..

So I'll leave this open for now, but I'll also lock the conversation because experience from the mailing lists on this topic has told us this thread will get out of hand quickly.

@dylanmtaylor: this stance does not mean that we won't merge this. And despite locking this thread, I think you, me and other contributors and maintainers can still comment (which is fine, and good).

Torxed:
Linking a possible alternative if we decide we can't/won't merge this specific PR: https://github.com/archlinux/archinstall/issues/4330

PR #4349, opened 2026-03-31
Don't write encryption keyfiles to an unencrypted root partition

### LINKED ISSUE #1856: Encrypting /home but not / leads to keyfile being written unencrypted to disk
If you set an encryption passphrase, this passphrase is only used to encrypt the / partition while all other partitions are encrypted via keyfile, which is stored somewhere under /etc. This way you don't have to enter your passphrase multiple times for each partition beeing unlocked.

But if / is chosen to not be encrypted, archinstall still uses keyfiles to encrypt the other partitions, and doesn't use the passphrase at all, which is not was most users would want.

dylanmtaylor:
Fixes #1856

When encrypting partitions like /home but not /, the keyfiles were written to the unencrypted root filesystem in plaintext, defeating the purpose of the encryption.

Now keyfiles are only generated when root is also encrypted. Otherwise, systemd-cryptsetup will prompt for the passphrase at boot instead.

### DIFF
--- a/archinstall/lib/disk/luks.py
+++ b/archinstall/lib/disk/luks.py
@@ -204,6 +204,18 @@ class Luks2:
 		self._add_key(key_file)
 		self._crypttab(crypttab_path, kf_path, options=['luks', 'key-slot=1'])
 
+	def create_crypttab_entry(self, target_path: Path) -> None:
+		"""
+		Add a crypttab entry without a keyfile so systemd prompts
+		for the passphrase at boot.
+		"""
+		if self.mapper_name is None:
+			raise ValueError('Mapper name must be provided')
+
+		crypttab_path = target_path / 'etc/crypttab'
+		crypttab_path.parent.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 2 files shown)


dylanmtaylor:
This might be as straightforward as it seems, needs tested still, so in draft. I suspect we will run into an issue because we still need a crypttab entry but with none as the keyfile so systemd prompts for the passphrase.

Edit: This is the case, I updated the PR to address ^^

dylanmtaylor:
Testing:

Manual partitioning:
Screenshot From 2026-03-31 20-58-48
Encryption enabled on just /home
Screenshot From 2026-03-31 21-01-03
/etc/crypttab contains home partition with "none" set
image
/etc/cryptsetup-keys.d does NOT have the key for home:
image
home password prompt on boot;
image

dylanmtaylor:
This is tested and confirmed to work correctly. ^^ 

After entering the /home password, I was able to log into the system normally.

svartkanin (approved):

Issue #3735, opened 2025-08-13
Non-Latin languages display issue


killertofus:
### Which ISO version are you using?

2025.08.01

### The installation log

```json
n/a
```

### describe the problem

<img width="425" height="357" alt="Image" src="https://github.com/user-attachments/assets/5a5eacb8-70c6-44da-8a2c-447fc7e8369d" />

nobhkybsh:
Install any Japanese font and type `setfont YOUR_FONT`

Exorcism0666:
The setfont command in a classic Linux TTY console doesn't support Japanese characters, as these fonts generally require multibyte encoding (UTF-8 with CJK), whereas setfont only loads 8×16 or similar bitmap fonts for the console.

So I would like to know if there is a solution to this problem because I see no point in putting the Japanese translation and other languages have non-printable characters outside this range.

svartkanin:
That's a good question, I haven't tried out CJK languages and how to get them displayed. From the looks you're correct that they can't be displayed in the Arch ISO by default or installing fonts. 

Maybe https://code.google.com/archive/p/fbterm/ this might be an option but I haven't tried it. 

However, I would like to note that some users will run archinstall from non ISO environments such as an existing Arch installation and/or VM which means that a proper terminal with available fonts should be able to make use of these languages. I can't speak for how many that would be though. 

From my terminal with Noto they are displayed fine 

<img width="485" height="512" alt="Image" src="https://github.com/user-attachments/assets/a8654436-0bba-4dfa-8eea-1a3db9ed477f" />

ykla:
# Non-Latin languages display issue

This issue is not only related to Japanese, I suggest modifying the title. Thank you.


Non-Latin character languages, including CJK languages and Russian, cannot be displayed properly. I understand this is due to the Linux kernel not including CJK patches and the lack of pre-installed fonts.

Given this situation, is there a simpler way to address the display issue? see also <https://github.com/bigshans/cjktty-patches>、<https://archlinux.org/packages/extra/any/wqy-microhei/>

Alternatively, I would suggest considering the removal of translations for these languages, as they currently do not function properly or require complex steps to enable support. Since the current situation tends to confuse newcomers about what's going on and how to resolve it, experienced experts often don't need to switch languages and may not even need this tool.

Thank you for considering this suggestion.

<img width="506" height="387" alt="Image" src="https://github.com/user-attachments/assets/b8626c10-b392-44f9-83e8-4b9ba598c1bb" />

[...]

PR #4356, opened 2026-04-02
Set console font automatically when selecting language

Softer:
- This fix issue: #3735

## PR Description:

  - Add console_font field to languages.json and Language dataclass. When a language with a mapped font is selected (e.g. Ukrainian → UniCyr_8x16), the console font is set automatically via setfont.
  - Save and restore the original console font on startup/exit using setfont -O (full font + unicode map) and -om (console map) with temp files.
  - Support FONT environment variable to override language font mapping. If FONT is set but invalid, fall back to language mapping and inform the user.
  - Font changes happen inside the Textual app lifecycle (on_mount / _on_exit_app) to prevent color artifacts caused by 256/512 glyph count transitions when Textual's alternate screen is active.
  - Activate translation when loading language from config files (set_font=False to defer font change until TUI starts).
  - Update language selection note to reflect automatic font handling.

## Tests and Checks

  - Run archinstall, select Ukrainian → font changes to UniCyr_8x16, exit → original font restored
  - Run FONT=UniCyr_8x16 archinstall → font applied on TUI start, restored on exit
  - Run FONT=nonexistent archinstall → info message shown, language mapping used as fallback
  - Run archinstall --config with archinstall-language: Ukrainian → font applied inside TUI, not before
  - Verify no color artifacts when switching between 256/512 glyph fonts
  - Run via SSH → no font changes attempted (setfont fails gracefully)


svartkanin:
I need to test this thoroughly, we had quite some issues in the past with this. 
For example rhe font will be set when exiting archinstall which is not desired

Softer:
Perhaps we can simply reset the font to the "default8x16" when exiting?

h8d13:
On a side note @softer : I believe picking any terminus font will cause an error because target would need an extra package `terminus-fonts` when set through:

https://github.com/archlinux/archinstall/blob/6c9f66265a5e940902e911feae1da04c5d4e501c/archinstall/lib/installer.py#L1978-L1986

This is the case because `releng` profile has this added in https://gitlab.archlinux.org/archlinux/archiso/-/raw/master/configs/releng/packages.x86_64 around line 110~ which means its on the ISO but not in our target

I know a lot of folks would love to be able to set larger font through installer directly (aside from language specific fonts) and would persist on target !

Ideally we can have an option for it which would edit the value above and if something `ter` we add the correct dependency ^^

Softer:
@h8d13 
I think we should at least start... :)

h8d13:
Right but, I'm pointing out related more important stuff aha

Also I believe any arabic, greek, asian languages will not work with these mappings if I understand correctly (lots of issues related to this)

Softer:
You are right, but as far as I understand, these languages are not working correctly at the moment. So nothing will change because of that.

I am thinking about giving the user the ability to set a different font for the installer. We cannot get the current font name, for example, the one set by setfont, so what if we use an environment variable? Let it be named FONT - yes, like in vconsole.conf.

If the user sets a font name there, the installer will use that font and will not change it based on my mappings.

The logic would be as follows: if the variable is set, use only the font from it. If it is not set, use my mapping logic.

What do you think?

h8d13:
Well I think what is more useful is to try to make a PR for locale to actually store the font and edit this https://github.com/archlinux/archinstall/pull/4356#issuecomment-4176135457

And get the available ones similar to how we get https://github.com/archlinux/archinstall/blob/6c9f66265a5e940902e911feae1da04c5d4e501c/archinstall/lib/locale/utils.py#L18

Then you'd have it working inside your install too + add `terminus-font` when anything `ter` is selected 

image

About the initial PR: an ENV var seems reasonable, it would be better to hook it up to startup functions before any TUI is being displayed

Softer:
What do you mean? You wanna save selected font (via setfont or my mapping) for target system?

h8d13:
Yes currently it's hardcoded as shown in snippet above. So this would be a nice target ! 

And only need to hook it up said: `font_vconsole = your_selection`

Then it would just work at first boot for target system too :)

Softer:
I'll think what can I do... :)

Softer:
Support FONT environment variable is now done.

When FONT env var is set, use it as the console font instead of the language-specific font mapping from languages.json. The font is applied at startup and preserved across language switches.

On exit (success or failure), restore the console font to default8x16.

I will work on saving the font selection for the target system in a separate PR, as it appears to be a separate task. :)

Softer:
I'm waiting for this PR, it blocks me with the next task :)

codefiles:
From just a cursory glance the code quality could use some work.

This still needs to be reviewed and tested.

Would you like me to review this or will you work on it some more?

Softer:
I totally agree - I'm not a pro with python :) 
It would be amazing if you show me my mistakes.

archinstall/main.py
```diff
@@ -79,6 +79,9 @@ def run() -> int:
 	OR straight as a module: python -m archinstall
 	In any case we will be attempting to load the provided script to be run from the scripts/ folder
 	"""
+	if _ENV_FONT:
+		_set_console_font(_ENV_FONT)
```
codefiles:
```suggestion
	_set_console_font(_ENV_FONT)
```

Since the call with `None` argument should be removed from `main()`.

archinstall/lib/translationhandler.py
```diff
@@ -134,6 +160,8 @@ def activate(self, language: Language) -> None:
 		# The install() call has the side effect of assigning GNUTranslations.gettext to builtins._
 		language.translation.install()
 
+		_set_console_font(_ENV_FONT if _ENV_FONT else language.console_font)
```
codefiles:
```suggestion
		_set_console_font(_ENV_FONT or language.console_font)
```

archinstall/lib/translationhandler.py
```diff
@@ -31,6 +33,29 @@ def json(self) -> str:
...
+	try:
+		result = subprocess.run(['setfont', target], capture_output=True, check=False, timeout=10)
+		if result.returncode != 0 and target != _DEFAULT_FONT:
+			subprocess.run(['setfont', _DEFAULT_FONT], capture_output=True, check=False, timeout=10)
+	except (FileNotFoundError, subprocess.TimeoutExpired, OSError):
+		if target != _DEFAULT_FONT:
+			try:
+				subprocess.run(['setfont', _DEFAULT_FONT], capture_output=True, check=False, timeout=10)
+			except (FileNotFoundError, subprocess.TimeoutExpired, OSError):
+				pass
```
codefiles:
```suggestion
	try:
		SysCommand(['setfont', font_name])
	except SysCallError:
```

Also add a debug log message under the exception.

codefiles (changes requested):
Rather than use a default font, use the `tempfile` python module and on startup store the current font with <code>setfont -O _tmp_</code>. Restore the stored font from the tempfile and remove the tempfile on exit.

If you need me to further elaborate on how to accomplish this, have any questions, or concerns about this suggestion don't hesitate, I'm happy to help.

Softer:
SysCommand - yes, you are right.

Saving and restoring the font is a good idea. I tried to implement it, and it works almost fine, except for one edge case. There may be more, but so far I have only caught this one. 
If we set the font to ter-v16n (or any 512-glyph font), run the installer with another non-512-glyph font (FONT or a language mapping), and then exit, we always get color glitches. The background and foreground colors of all the text above the current prompt change.
As far as I know, there is no way to fix it.

I am not sure if that is acceptable for us. Personally, I think it is okay, because... how often do you change the font before running archinstall? We needed it for proper localization, but now it can work correctly without setting the font before running archinstall (yes, I remember the issue with Arabic and Asian languages :) ).

So, if it is acceptable, I will push it to the PR :)

codefiles:
Let's see if we can figure out a solution. I need some time to look into this.

Softer:
Ok, I pushed the changes for saving and restoring the font. :)

Softer:
@codefiles Any updates?

codefiles:
Following the steps you provided, it is reproducible. I tried a screen redraw using `reset` following the font restore but the color glitch is still visible just before the reset happens. I'm not really sure. I have not given up though and would like to figure out why this happens.

Softer:
I tried it too. On exit, we have some useful information (for example, a stacktrace), so it is not a good idea to clear the screen...

codefiles:
Yes, I agree.

codefiles:
Could the TUI be the source? It does not seem to be reproducible when the console font code is used in isolation.

Partial code example:

```python
def main() -> None:
	_save_console_font()
	print('before font change')
	_set_console_font('UniCyr_8x16')
	print('after font change')
	_restore_console_font()
	print('after font restore')


if __name__ == '__main__':
	main()
```

I can see a slight jitter from the font changing but this does not appear to produce the color glitch.

codefiles:
I created a minimal Textual app that just has a toggle to change the font to `UniCyr_8x16` and uses the font save before running the app and font restore after app exit. It produces the same color glitch and is devoid of any archinstall code other than the console font handling from this pull request.

I do not have the motivation to look into why this happens with Textual. I guess that is it for me, sorry.

svartkanin:
Could you share that minimal example you created?

codefiles:
Steps to reproduce:

- Run `setfont ter-v16n` beforehand.
- Run script.
- Use `d` to toggle.
- `Ctrl` + `q` to exit.

```python
from pathlib import Path
from subprocess import run

from textual.app import App, ComposeResult
from textual.widgets import Footer, Header

_FONT_BACKUP = Path('/tmp/archinstall_font_backup')
_CMAP_BACKUP = Path('/tmp/archinstall_cmap_backup')


class ConsoleFontApp(App):
	BINDINGS = [("d", "toggle_font", "Toggle font")]

	def compose(self) -> ComposeResult:
		yield Header()
		yield Footer()

	def action_toggle_font(self) -> None:
		run(['setfont', 'UniCyr_8x16'])


def _save_console_font() -> None:
	run(['setfont', '-O', str(_FONT_BACKUP), '-om', str(_CMAP_BACKUP)])


def _restore_console_font() -> None:
	if _FONT_BACKUP.exists():
		run(['setfont', str(_FONT_BACKUP), '-m', str(_CMAP_BACKUP)])

	_FONT_BACKUP.unlink(missing_ok=True)
	_CMAP_BACKUP.unlink(missing_ok=True)


if __name__ == '__main__':
	_save_console_font()
	app = ConsoleFontApp()
	app.run()
	_restore_console_font()
```

Softer:
@codefiles 
I have found a very dirty hack, but it looks like it works.

```
import array
import fcntl
from pathlib import Path
from subprocess import run

from textual.app import App, ComposeResult
from textual.widgets import Footer, Header

_FONT_BACKUP = Path('/tmp/archinstall_font_backup')
_CMAP_BACKUP = Path('/tmp/archinstall_cmap_backup')

VT_GETHIFONTMASK = 0x560D


class ConsoleFontApp(App):
	BINDINGS = [("d", "toggle_font", "Toggle font")]

	def compose(self) -> ComposeResult:
		yield Header()
		yield Footer()

	def action_toggle_font(self) -> None:
		run(['setfont', 'UniCyr_8x16'])


def _get_hi_font_mask() -> int:
	"""Get the hi-font mask via ioctl. Returns 0 for 256-glyph, non-zero for 512-glyph."""
	try:
		buf = array.array('H', [0])
		with open('/dev/tty', 'r') as tty:
			fcntl.ioctl(tty, VT_GETHIFONTMASK, buf)
		return buf[0]
	except OSError:
		return 0


def _reset_console_attrs() -> None:
	"""Reset color attributes of all cells on current screen, preserving text and hi-font bit."""
	try:
		hi_mask = _get_hi_font_mask()
		is_512 = hi_mask != 0

		vcsa = Path('/dev/vcsa')
		data = bytearray(vcsa.read_bytes())
		rows, cols = data[0], data[1]
		expected = 4 + rows * cols * 2
		if len(data) != expected:
			print(f'vcsa size mismatch: {len(data)} != {expected}, aborting')
			return

		count = 0
		for i in range(5, len(data), 2):
			old = data[i]
			if is_512:
				data[i] = (old & 0x01) | 0x0E
			else:
				data[i] = 0x07
			if data[i] != old:
				count += 1

		vcsa.write_bytes(bytes(data))
		print(f'hi_font={is_512}, reset {count}/{rows * cols} attrs')
	except OSError as e:
		print(f'Failed to reset console attrs: {e}')


def _save_console_font() -> None:
	run(['setfont', '-O', str(_FONT_BACKUP), '-om', str(_CMAP_BACKUP)])


def _restore_console_font() -> None:
	if _FONT_BACKUP.exists():
		run(['setfont', str(_FONT_BACKUP), '-m', str(_CMAP_BACKUP)])

	_FONT_BACKUP.unlink(missing_ok=True)
	_CMAP_BACKUP.unlink(missing_ok=True)


if __name__ == '__main__':
	_save_console_font()
	app = ConsoleFontApp()
	app.run()
	_restore_console_font()
	_reset_console_attrs()
	print('done')

```

codefiles:
That does not work for me, it produces more color and visual glitches.

I can trigger the bug by using the `curses` module and an _alternate screen_ along with the font changing. This is without any use of Textual. That eliminates the possibility the bug is caused by Textual. It appears the font restore should happen before exiting the alternate screen.

Here is an updated example that should not produce the bug:

```python
from pathlib import Path
from subprocess import run
from tempfile import NamedTemporaryFile

from textual.app import App, ComposeResult
from textual.widgets import Footer, Header


class ConsoleFontApp(App):
	BINDINGS = [('d', 'toggle_font', 'Toggle font')]

	_FONT_BACKUP = Path(NamedTemporaryFile(delete=False).name)
	_CMAP_BACKUP = Path(NamedTemporaryFile(delete=False).name)

	def __init__(self) -> None:
		super().__init__()

		run(
			[
				'setfont',
				f'--output-fullfont={self._FONT_BACKUP}',
				f'--output-consolemap={self._CMAP_BACKUP}',
			]
		)

	def compose(self) -> ComposeResult:
		yield Header()
		yield Footer()

	def action_toggle_font(self) -> None:
		run(['setfont', 'UniCyr_8x16'])

	def exit(self) -> None:
		if self._FONT_BACKUP.exists() and self._CMAP_BACKUP.exists():
			run(
				[
					'setfont',
					str(self._FONT_BACKUP),
					f'--consolemap={self._CMAP_BACKUP}',
				]
			)

			self._FONT_BACKUP.unlink()
			self._CMAP_BACKUP.unlink()

		super().exit()


if __name__ == '__main__':
	app = ConsoleFontApp()
	app.run()
```

In the update to the pull request, you didn't use the `tempfile` module like I asked, so I put that in there. Is there a reason that you didn't?

codefiles:
Below is the `curses` version I mentioned in my last comment. Follow the same steps as before to reproduce the bug. If the script is run with the argument `--fix` the bug should not be produced. The only difference in using the `--fix` argument versus not using it is the location of the font restore; before the alternate screen exit and it works without issue, after the alternate screen exit and the bug is produced. This script is intentionally minimal and is meant to just demonstrate the bug and show how I isolated it.

```python
import curses
import sys
from curses import window, wrapper
from pathlib import Path
from subprocess import run
from tempfile import NamedTemporaryFile


def curses_app(stdscr: window) -> None:
	curses.curs_set(0)

	stdscr.addstr(1, 2, "Press 'd' to toggle font")
	stdscr.addstr(2, 2, "Press 'q' to quit")
	stdscr.refresh()

	while True:
		c = stdscr.getch()
		if c == ord('d'):
			run(['setfont', 'UniCyr_8x16'])
		elif c == ord('q'):
			break


def main() -> None:
	# Enter alternate screen
	sys.stdout.write('\033[?1049h')
	sys.stdout.flush()

	font_backup = Path(NamedTemporaryFile(delete=False).name)
	cmap_backup = Path(NamedTemporaryFile(delete=False).name)

	run(
		[
			'setfont',
			f'--output-fullfont={font_backup}',
			f'--output-consolemap={cmap_backup}',
		]
	)

	try:
		wrapper(curses_app)
	finally:
		fix = '--fix' in sys.argv

		if fix and font_backup.exists() and cmap_backup.exists():
			run(['setfont', str(font_backup), f'--consolemap={cmap_backup}'])

		# Exit alternate screen
		sys.stdout.write('\033[?1049l')
		sys.stdout.flush()

		if not fix and font_backup.exists() and cmap_backup.exists():
			run(['setfont', str(font_backup), f'--consolemap={cmap_backup}'])

		font_backup.unlink(missing_ok=True)
		cmap_backup.unlink(missing_ok=True)


if __name__ == '__main__':
	main()
```

Softer:
Here is the next version. It seems to be a working solution for me. @codefiles Could you check it please?

archinstall/lib/translationhandler.py
```diff
@@ -31,10 +37,75 @@ def json(self) -> str:
...
+
+
+class _FontState:
+	def __init__(self) -> None:
+		self.font_backup: Path | None = None
+		self.cmap_backup: Path | None = None
+		self.using_env_font: bool = False
+
+
+_font_state = _FontState()
```
codefiles: Can you figure out a way to do this without using a module level singleton?

codefiles (changes requested):

archinstall/lib/translationhandler.py
Softer: Yes, I can. Done.

archinstall/lib/translationhandler.py
```diff
@@ -31,10 +37,68 @@ def json(self) -> str:
...
+		translation_handler._font_backup = Path(font_path)
+		translation_handler._cmap_backup = Path(cmap_path)
+		SysCommand(f'setfont -O {translation_handler._font_backup} -om {translation_handler._cmap_backup}')
+	except SysCallError as err:
+		debug(f'Failed to save console font: {err}')
+		translation_handler._font_backup = None
+		translation_handler._cmap_backup = None
+
+
+def _restore_console_font() -> None:
```
svartkanin: As this is exported it should not be private same with `_save_console_font`

archinstall/tui/ui/components.py
```diff
@@ -1245,6 +1252,17 @@ def action_trigger_help(self) -> None:
 			_ = self.screen.mount(HelpPanel())
 
 	def on_mount(self) -> None:
+		import archinstall.lib.translationhandler as th
+
+		if th._ENV_FONT:
```
svartkanin: This logic has nothing to do with the tui and should all be moved to the translation handler module exposed via a single function

archinstall/main.py
```diff
@@ -159,6 +161,8 @@ def main() -> int:
 			_error_message(exc)
 			rc = 1
 
+		_restore_console_font()
```
svartkanin: are we sure that this will be execute on any exception otherwise it will not be reset?

archinstall/tui/ui/components.py
```diff
@@ -1275,9 +1293,14 @@ class TApp:
 	app: _AppInstance[Any] | None = None
 
 	def run(self, main: InstanceRunnable[ValueT] | Callable[[], Awaitable[ValueT]]) -> ValueT:
+		import archinstall.lib.translationhandler as th
+
 		TApp.app = _AppInstance(main)
 		result: ValueT | Exception | None = TApp.app.run()
 
+		if th._ENV_FONT and not th.translation_handler._using_env_font:
```
svartkanin: such infos shouldn't be logged conditionally but rather the state should be part of the log message

archinstall/main.py
Softer: Yes, it's covered at two levels:

* **_on_exit_app** - called by Textual on normal exit and on exceptions within the app, before stop_application_mode()
* **finally in main.py** - catches anything that bypasses Textual's cleanup. Second call is safe since backup files are already removed by the first.

As far as I understand - only uncovered case is SIGKILL, which can't be handled at all.

Am I missing any other scenario?

archinstall/lib/translationhandler.py
```diff
@@ -31,10 +37,68 @@ def json(self) -> str:
 		return self.name_en
 
 
+_DEFAULT_FONT = 'default8x16'
+_ENV_FONT = os.environ.get('FONT')
+
+
+def _set_console_font(font_name: str | None) -> bool:
```
svartkanin: The return value is never used, we can probably remove it

archinstall/lib/translationhandler.py
```diff
@@ -127,12 +198,36 @@ def get_language_by_abbr(self, abbr: str) -> Language:
...
+		if _ENV_FONT:
+			if _set_console_font(_ENV_FONT):
+				self._using_env_font = True
+				debug(f'Console font set from FONT env var: {_ENV_FONT}')
+			else:
+				debug(f'FONT={_ENV_FONT} could not be set, falling back to language font mapping')
+				if self.active_font:
+					_set_console_font(self.active_font)
+					debug(f'Console font set from language mapping: {self.active_font}')
+		elif self.active_font:
```
svartkanin, line 235: I would not want this to be set on my actual host but only on the ISO, so can we check with https://github.com/archlinux/archinstall/blob/deb80803224ff03a696ddef27be895a0d93e0920/archinstall/lib/utils/util.py?plain=1#L8

archinstall/lib/translationhandler.py
```diff
@@ -31,10 +37,68 @@ def json(self) -> str:
...
+
+	try:
+		SysCommand(f'setfont {target}')
+		return True
+	except SysCallError as err:
+		debug(f'Failed to set console font {target}: {err}')
+		return False
+
+
+def save_console_font() -> None:
```
svartkanin: these are module functions but they seem to be tightly coupled with the actual translation handler and only operate in that context. Shall we make them member functions of the translation handler instead

archinstall/lib/args.py
```diff
@@ -142,6 +142,7 @@ def from_config(cls, args_config: dict[str, Any], args: Arguments) -> Self:
 
 		if archinstall_lang := args_config.get('archinstall-language', None):
 			arch_config.archinstall_language = translation_handler.get_language_by_name(archinstall_lang)
+			translation_handler.activate(arch_config.archinstall_language, set_font=False)
```
svartkanin, line 145: why is `set_font` false here?

archinstall/lib/translationhandler.py
```diff
@@ -43,6 +57,63 @@ def __init__(self) -> None:
...
 
+	@property
+	def active_font(self) -> str | None:
+		if self._active_language is not None:
+			return self._active_language.console_font
+		return None
+
+	def _set_font(self, font_name: str | None) -> bool:
+		"""Set the console font via setfont. Only runs on ISO. Returns True on success."""
+		from archinstall.lib.utils.util import running_from_iso
```
svartkanin: does this cause a circular dep otherwise can we move it to the top?

archinstall/lib/translationhandler.py
```diff
@@ -43,6 +57,63 @@ def __init__(self) -> None:
...
+			self._cmap_backup = Path(cmap_path)
+			SysCommand(f'setfont -O {self._font_backup} -om {self._cmap_backup}')
+		except SysCallError as err:
+			debug(f'Failed to save console font: {err}')
+			self._font_backup = None
+			self._cmap_backup = None
+
+	def restore_console_font(self) -> None:
+		"""Restore console font (with unicode map) and console map from backup."""
+		if self._font_backup is None or not self._font_backup.exists():
```
svartkanin, line 103: Can we add an explicit iso check here as well to not run anything if we're not on the ISO?

archinstall/lib/translationhandler.py
svartkanin: @Softer can we address this?

archinstall/lib/args.py
Softer: from_config runs before the TUI starts. Calling setfont here would trigger color artifacts due to 256/512 glyph count transitions once Textual activates the alternate screen. Font application is deferred to on_mount so the transition happens inside the TUI lifecycle.

archinstall/lib/translationhandler.py
Softer: now used in apply_console_font:227 for FONT fallback

Softer:
This PR is one of the most commented ones :)

archinstall/lib/args.py
svartkanin: This may be good to have as a comment on that call

svartkanin (approved):

PR #4358, opened 2026-04-02
Patch plasma profile

h8d13:
Continue conversation: #4336

Instead of reverting and pulling in the whole bag again, I think we can just add bare minimum.

### DIFF
--- a/archinstall/default_profiles/desktop.py
+++ b/archinstall/default_profiles/desktop.py
@@ -27,7 +27,7 @@ class DesktopProfile(Profile):
 	def packages(self) -> list[str]:
 		return [
 			'nano',
-			'vim', ## should be ex-vi-compat
+			'vim',
 			'openssh',
 			'htop',
 			'wget',
--- a/archinstall/default_profiles/desktop.py
+++ b/archinstall/default_profiles/desktop.py
@@ -27,7 +27,7 @@ class DesktopProfile(Profile):
 	def packages(self) -> list[str]:
 		return [
 			'nano',
-			'vim',
+			'vim', ## should be ex-vi-compat
 			'openssh',
 			'htop',
 			'wget',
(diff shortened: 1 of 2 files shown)


h8d13:
Tested on NVIDIA hardware, this makes it so that the sound options and display settings of KDE appear on first boot, sorry for any inconveniences caused by the previous PR

ghost:
This is needed. I tried using archinstall this morning (first noticing there was no graphics options and rage baited because it said null in the config.json, but it wasn't it) and then realized that plasma-desktop was too minimal, basically, I got greeted with a mouse cursor after plasma-login-manager. Soooooo, after manually reinstalling plasma-meta / dolphin / konsole, it was all working again, and not gonna lie, it's better to have flatpak installed by default. I know Arch must be KISS compliant, but if users use Arch, it's not because they want to customize everything but more liking having a good WORKING system, not being greeted with something like what I had. Also, KDE Plasma right now is being considered one of the best for beginners, so if a beginner wants to try Arch (for some reason) and they choose it, they'll have a pain to figure out why they only have a cursor. It is easily fixed after installing 3 packages (including a meta-package) but still frustrating for a beginner. I hope everything is fine if this goes through next month with a new Arch ISO but it's bad since now everyone has to suffer an archinstall script not working for a month (at least). I don't know if pacman -S archinstall can upgrade it since they might only package it with a new version right before an Arch ISO is pushed but voila. You saw someone ragebaiting a simple tool trying to get too simple for something that should be careful and user friendly.

svartkanin (approved):

ghost:
Actually.... My bad. I didn't know (and yelled at people) , apparently you can select multiple DEs, and the new UI doesn't auto select the one under the cursor when pressing enter when there is no other options selected. Meaning I was pressing enter for KDE Plasma and it selected nothing -_- 

So now that I know I would like to say sorry to like everyone I yelled at today. And reinstall Arch but this time by smashing my space bar like for every options JUST TO MAKE SURE.

So yeah sry @h8d13 and @svartkanin

PR #4359, opened 2026-04-02
Check user's jurisdiction and prevent installation in certain regions

yangwenbo99:
Stop providing installation to certain jurisdictions

## Motivation

Recent age verification laws in California (AB-1043), Colorado (SB26-051), Brazil (Lei 15.211/2025), etc. require OS providers to verify user age. I hope that, by stopping providing Arch Linux installation to these regions, we can be fully compliant to the requirement.  

### Changes

Adding a field for users to declare whether they are located in regions that require an OS to provide or facilitate age, name, sex, or any other identifying information.  If they are, simply do not allow installation.  

I believe this is a less less intrusive alternative to  PR #4290. 

## Tests and Checks

### DIFF
--- a/archinstall/lib/args.py
+++ b/archinstall/lib/args.py
@@ -72,6 +72,7 @@ class ArchConfig:
 	hostname: str = 'archlinux'
 	kernels: list[str] = field(default_factory=lambda: ['linux'])
 	ntp: bool = True
+	jurisdiction_restriction: bool = False
 	packages: list[str] = field(default_factory=list)
 	parallel_downloads: int = 0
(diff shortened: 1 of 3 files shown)


christian-heusel:
@yangwenbo99 we're currently still in the process of evaluating the exact legal situation here, so this PR will be put on hold similar to #4290.

See https://gitlab.archlinux.org/archlinux/project-management/-/work_items/61 for our tracking item.

svartkanin:
I will close this for now as we will need to wait for https://gitlab.archlinux.org/archlinux/project-management/-/work_items/61

PR #4363, opened 2026-04-02
Add translation support for TUI help groups and binding descriptions

Softer:
Help group names shown via F1 (General, Navigation, Selection, Search) and key binding descriptions in the Textual footer (Down, Up, Cancel, Confirm, etc.) were hardcoded in English and never went through the translation system.

## PR Description:

This adds tr() wrapping for help group titles and introduces a mechanism to translate Textual binding descriptions at runtime. The tricky part is that Textual evaluates bindings at class definition time (before any language is activated), and stores them as frozen dataclasses. To work around this, _translate_bindings() rebuilds binding objects on every change of screen, so now we always show the current language. The app-level bindings are also re-translated when the user switches language mid-session.

New translatable strings are added to base.pot and translated for Ukrainian.


Torxed (approved):
Haven't tested this, but the code looks good.
Only thing that worry me is that we're calling private functions on the widgets which might be subject to change in the future.
But I also recognize that textual might not have thought of this behavior during translations so, probably best option.

archinstall/tui/ui/components.py
```diff
@@ -27,6 +27,13 @@
 ValueT = TypeVar('ValueT')
 
 
+def _translate_bindings(widget: Any) -> None:
```
svartkanin: Instead of passing a `Any` widget it'd be better to pass the `BindingMap` as that will avoid a private attribute access 

```
def _translate_bindings(binding_map: BindingsMap) -> None:
	for key, bindings in binding_map.key_to_bindings.items():
		binding_map.key_to_bindings[key] = [replace(b, description=tr(b.description)) if b.description else b for b in bindings]
```

I'm not quite following with this logic why is it using the `_merged_bindings` and then writing the translations to `_bindings`? Shouldn't it update teh same map?

archinstall/tui/help.py
```diff
@@ -97,7 +97,7 @@ def get_help_text() -> str:
 		max_key_width = max([help.get_key_width() for help in help_texts])
 
 		for help_group in help_texts:
-			help_output += f'{help_group.group_id.value}\n'
+			help_output += f'{tr(help_group.group_id.value)}\n'
```
svartkanin: This file is part of the deprecated `curses` GUI and there's no need to update it

archinstall/lib/global_menu.py
```diff
@@ -269,12 +269,17 @@ def _update_lang_text(self) -> None:
...
+		from archinstall.tui.ui.components import TApp, _translate_bindings
+
 		new_options = self._get_menu_options()
 
 		for o in new_options:
 			if o.key is not None:
 				self._item_group.find_by_key(o.key).text = o.text
 
+		if TApp.app is not None:
+			_translate_bindings(TApp.app)
```
svartkanin: we shouldn't export private functions from other modules, instead lets add helper functions to trigger the translation updates 

```
class TApp:
...
    def update_translation(self) -> None:
        app.update_translations()
```

archinstall/tui/ui/components.py
Softer: If I understand right - when switching languages, _translate_bindings is called again. If descriptions are read from _bindings, they are already translated (e.g. "Вниз" for Ukrainian). tr("Вниз") won't find a match in another language's .po file since keys are in English. The string would remain in the old language.

Correct me if I'm wrong...

archinstall/lib/global_menu.py
```diff
@@ -269,12 +269,16 @@ def _update_lang_text(self) -> None:
 		The options for the global menu are generated with a static text;
 		each entry of the menu needs to be updated with the new translation
 		"""
+		from archinstall.tui.ui.components import TApp
```
svartkanin: Can we move the import to the top, or does it cause circular dep?

archinstall/tui/ui/components.py
```diff
@@ -1291,5 +1322,11 @@ def exit(self, result: Any) -> None:
 		assert TApp.app
 		TApp.app.exit(result)
 
+	@staticmethod
```
svartkanin: why make this a staticmethod not a member one?

archinstall/tui/help.py
svartkanin: @Softer can we reverrt this

archinstall/tui/ui/components.py
svartkanin: Yeah I'm not 100% sure tbh, I tried to find any docs on this from textual but couldn't

archinstall/lib/global_menu.py
Softer: No, you're right.

archinstall/tui/ui/components.py
Softer: I thought it would be better that way, but your question made me reconsider. :)

archinstall/tui/help.py
Softer: Ok. Done.

archinstall/tui/ui/components.py
codefiles: It never uses `self`. 😕

Softer: But we need the first parameter for this method. Am I right?

svartkanin (approved):

svartkanin:
Nice work, thatnks for the changes

PR #4376, opened 2026-04-05
add mirror source switching functionality during slow database sync

xhdndmm:
- This fix issue: 

## PR Description:

add mirror source switching functionality during slow database sync

When the mirror is slow , use the TUI to switch to a new mirror

## Tests and Checks

### DIFF
--- a/archinstall/lib/switch_mirror.py
+++ b/archinstall/lib/switch_mirror.py
@@ -1,0 +1,52 @@
+from pathlib import Path
+
+from archinstall.lib.installer import Installer
+from archinstall.lib.models.device import DiskLayoutConfiguration, DiskLayoutType
+from archinstall.lib.mirror.mirror_handler import MirrorListHandler
+from archinstall.lib.mirror.mirror_menu import MirrorMenu
+from archinstall.lib.network.wifi_handler import WifiHandler
+from archinstall.lib.output import debug, error, info, warn
+from archinstall.lib.translationhandler import tr
+from archinstall.lib.networking import ping
+from archinstall.tui.ui.components import tui
+
+
+def check_online(wifi_handler: WifiHandler | None = None) -> bool:
+    try:
+        ping('1.1.1.1')
+    except OSError as ex:
+        if 'Network is unreachable' in str(ex):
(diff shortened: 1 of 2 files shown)


svartkanin:
Can you add a explanation and description for this PR. 
Also this usrs `input` and not the TUI menu.

xhdndmm:
> Can you add a explanation and description for this PR. Also this usrs `input` and not the TUI menu.

okay

svartkanin:
In addition, this should not be in the main.py

xhdndmm:
> In addition, this should not be in the main.py

So do I need to make some changes?

svartkanin:
The tui should not get included into the main.py, so I would suggest to extract all that functionality out of the main

xhdndmm:
> The tui should not get included into the main.py, so I would suggest to extract all that functionality out of the main

Okay, I changed it.

svartkanin:
This is still using 
```
		return input(prompt).strip().lower() in ('y', 'yes', '')
```

also why was the `check_online` moved? 

If you're just using an LLM, please don't as these PRs are on the very bottom of my priority list.

xhdndmm:
> This is still using
> 
> ```
> 		return input(prompt).strip().lower() in ('y', 'yes', '')
> ```
> 
> also why was the `check_online` moved?
> 
> If you're just using an LLM, please don't as these PRs are on the very bottom of my priority list.

okay, I'll check them, sorry

xhdndmm:
I fixed it, and now can use tui to switch

svartkanin:
Can you explain what problem you're actually trying to solve here? 
I just run the code I was prompted after a 2 second delay if I want to switch mirrors which doesn't make sese. 

The `retry` parameter is always set to `false` so there's no point in that. Also even if `pacman -Sy` succeeds it will still prompt for a mirror selection just because it was slow? Then why do we need this special menu hook instead of the user just doing it right in the global menu as the is shown right after?

xhdndmm:
In some cases, mirror sites are very slow. I want to solve this by allowing users to switch to a different mirror site.

PR #4387, opened 2026-04-07
Missing one more essential #4382

h8d13:


Torxed (commented):
This should ideally only be added if the user selects `NetworkManager`, otherwise it's a redundant package.

I'm inclined to go ahead on this anyway to solve the immediate issue. Not sure anyone would object until we get it into the next update.. as they won't become aware until it's in their system.

h8d13:
Yeah I agree, would have preferred to keep it minimal but I guess not everyone sees it this way. 

Eitherway perhaps profiles should be aware of the certain picks (especially sound / network), not sure where this could go if not here

PR #4389, opened 2026-04-07
Allow granular plasma configuration 

svartkanin:
Attempts to fix #4336 
Allows the user to select a specific flavor of Plasma 

image

I have reduced the package list fully though to the respective package groups so that we are following the wiki https://wiki.archlinux.org/title/KDE and let the package group maintainers be in charge what they think should be installed


archinstall/default_profiles/desktops/plasma.py
```diff
@@ -15,18 +39,46 @@ def __init__(self) -> None:
 	@property
 	@override
 	def packages(self) -> list[str]:
-		return [
-			'plasma-desktop',
-			'kscreen',
```
h8d13, line 20: these should still be kept for the minimal version i think

Torxed:
We'll merge this, and then push a release within the day to get a KDE fix out there so that people have options.

Torxed (approved):

Issue #1121, opened 2022-05-03
Suggestion: enable pacman color and parallel downloads by default


Schweber:
I think this would lead to a faster installation and a more visually pleasing installation.

Personally, i am using Parallel Downloads = 10

Torxed:
Possible duplicate of #220 and

Issue #930, opened 2022-02-01
Parallel downloads 


RedCommand:
Why not activate parallel download by default or maby an option durring the setup ?

Torxed:
We rely on the defaults used on the ISO. We could add an option for it and present it, sure.
But we are already getting one too many questions during install I think, so this would have to be up for discussion before we commit to this change.

Do people want one more menu entry? Or should a petition be started upstream to `archiso` and namely the `releng` configuration?

l4zy0n3:
I think the petition way would be better, but this is just my humble opinion, as there would be no end to adding optional features as menu entries. Instead, for now, we could leverage plugins, custom commands to add these optional features to try to keep the installer modular and minimal. Installer stability and repeatable builds could be prioritised over adding non defaults. Again, just my 2 cents 🙂

Torxed:
> I think the petition way would be better, but this is just my humble opinion, as there would be no end to adding optional features as menu entries. Instead, for now, we could leverage plugins, custom commands to add these optional features to try to keep the installer modular and minimal. Installer stability and repeatable builds could be prioritised over adding non defaults. Again, just my 2 cents slightly_smiling_face

I fully agree. I could perhaps wrap up a plugin example, consider it the same value `AUR` has to the normal packaging.
It's optional things people can hook into their systems. And a plugin in this case wouldn't be a bad idea :)

For the uninitiated, as a concept example: `archinstall --plugin https://github.com/Torxed/archinstall-paralell-downloads.py`

edl2:
whichever of the three options it is i would say this is really important to the user experience and should not be ignored

Torxed:
I'll ask around and see why it's not enabled by default in upstream archiso releng configuration.
I suspect there might be a reason. I too like the faster downloads using parallel downloads :)

matteoturini:
forgot to enable parallel downloads, speeds suffering

Torxed:
I've gotten some info on the topic. And one of the main reasons is for the mirrors to not suffer as much.
If everyone had X amount of parallel downloads enabled by default that would hog a lot of connection resources for mirrors.

One could probably argue that they would survive, but at some cost.
But that's the reason and we'll honor it. It's relatively easy to modify `/etc/pacman.conf` before running any installation, so users who know what they're doing can easily do so without additional menu entries.

I'm not ruling it out for the future, but I'll close this one for now as the recommendation is to not enable it for a wider audience :)

godalming123:
[...]

PR #4404, opened 2026-04-10
Add Pacman settings submenu with Color and ParallelDownloads

Softer:
- Related to: #1121, #930 

## PR Description:

Replace the standalone "Parallel Downloads" menu item with a "Pacman" submenu (hidden behind the --advanced flag, same as before) containing two options: 
* ParallelDownloads (default 5) 
* Color (default on). 

ParallelDownloads is applied to the live system immediately for faster package installation. Both settings are written to the target system's /etc/pacman.conf by uncommenting and setting the corresponding lines.

Old config files using "parallel_downloads": N are still parsed correctly via a fallback in from_config().

## Tests and Checks
  - Run with --advanced, open Pacman submenu, change both values
  - Verify target /etc/pacman.conf has correct ParallelDownloads and Color after install
  - Run without --advanced — Pacman menu hidden, defaults (5, Color on) applied
  - Load old config with --config containing "parallel_downloads": 3 — parsed correctly


svartkanin:
I like the change but I still need to test and review it properly. 
One thing though, lets not put the entire pacman config menu entry behind the `--advanced` but only the parallel downloads as we may want to have actual future settings in that sub-menu that are fine without advanced.

Softer:
> One thing though, lets not put the entire pacman config menu entry behind the `--advanced` but only the parallel downloads as we may want to have actual future settings in that sub-menu that are fine without advanced.

I agree, but right now we have only two options in this menu, and both of them are advanced in my opinion. :)

archinstall/lib/models/pacman.py
```diff
@@ -0,0 +1,37 @@
...
+@dataclass
+class PacmanConfiguration:
+	parallel_downloads: int = 5
+	color: bool = True
+
+	@classmethod
+	def default(cls) -> Self:
+		return cls()
+
+	def json(self) -> dict[str, Any]:
```
svartkanin: we shouldn't be using `Any` but rather a dedicated serialization type, see https://github.com/svartkanin/archinstall/blob/a48ecf044bcfc59da01c804ff683ebe0cd7a20ae/archinstall/lib/models/authentication.py?plain=1#L14

svartkanin:
Yes I saw that, my question then is why is the color selection behind the advanced flag? 
The reason the parallele downloads is is because if it's visible by default a lot of users may increase that value and hit the Arch server too hard.

Softer:
> Yes I saw that, my question then is why is the color selection behind the advanced flag? The reason the parallele downloads is is because if it's visible by default a lot of users may increase that value and hit the Arch server too hard.

I thought about this but wasn't sure what's better - I followed the same approach as the existing ParallelDownloads. What do you think?

svartkanin:
Keeping the parallel downloads behind the advanced is fine but i think the color can be enabled by default, therefore I'd suggest to make the pacman config menu entry enabled by default, show the color option and put the parallel downloads behind the advanced\

Softer:
Done. I'll test it and then push it.

svartkanin (approved):

PR #4408, opened 2026-04-11
Warn when no network configuration is selected

Softer:
## PR Description:                                                                                                                                                                                               
                                                          
Users frequently forget to configure network during installation, resulting in no internet after reboot. The original version of this PR made the network menu mandatory (same pattern as Disk configuration),   but discussion suggested a softer approach.
                                                                                                                                                                                                                 
This PR instead:                                                                                                                                                                                                 
 
- Adds a `NicType.NONE` option ("No network configuration") for users who intentionally want no network on the target system. Selecting it is treated as a conscious choice and produces no warning.             
- Shows a yellow warning at the configuration confirmation step when the network menu was skipped entirely (config still `None`). The warning explains that network will need to be set up manually on the
installed system.                                                                                                                                                                                                
- Introduces a small install-warning infrastructure in `ConfigurationOutput` (`get_install_warnings()` / `_render_install_warnings()`) so future "soft" warnings can be added the same way.
                                                                                                                                                                                                                 
The network options list follows the enum order instead of alphabetical sorting, with "No network configuration" at the bottom.
                                                                                                                                                                                                                 
## Tests and Checks                                                                                                                                                                                              
- Skip network menu entirely -> yellow warning shown at confirm step, install still proceeds
- Select "No network configuration" explicitly -> no warning, no network packages installed                                                                                                                      
- Select any real network option (ISO / NetworkManager / iwd / manual) -> no warning, install proceeds                                                                                                           
- Old config with `network_config` via `--config` -> works as before                                                                                                                                             
- Old config without `network_config` -> warning shown in interactive mode


svartkanin:
I'm not sure this is a good argument, users should not be forced into selections that are not required for an installation.
Right now only the required settings for a working system are required and this deviates from that pattern.

Softer:
I agree with the principle of minimalism, but I think that in 2026, the network, in most cases, meets the minimum requirements.

svartkanin:
Considering there is a fair amount of people installing it on VMs and such it may still be somewhat impactfull

Softer:
In any case, we are not forcing the user but helping them not to forget to configure (or not configure) the network. They either consciously choose “No network” or do not forget to set it up and will not end up with a system without a network after installation.

svartkanin:
@Torxed what's your thoughts on it?

Torxed:
I don't wish this to be forced. Main reason being the tool should be versatile. Especially the main menu items are not to be enforced IMHO. Some desktop stuff, maybe? But even then, I've prepared laptops that are intended to be used without networking..

So I believe the argument for this is weaker than the intention. It's also perfectly possibly to configure systemd-networkd post-install if you forgot networking. It's 1-2 text files with 5-10 lines and you're up and running :)

Softer:
We're not forcing the user to configure the network or making it mandatory. Especially if they don't need it - that's exactly what the "No network" option is for :) Same pattern as Disk configuration - the user just needs to make a conscious choice.

What about a non-blocking warning instead? Show a note in the Install preview when network is not configured, but don't block the installation.

Torxed:
> We're not forcing the user to configure the network

I must have misunderstood something, I'll re-read when I get back to a computer.

h8d13:
I think explicit None options are nice (but still doesn't need to be mandatory to select)

Just like for greeter and to go through everything even if selecting "nothing"

svartkanin:
The installer doesn't have "None" as a selection in most cases so that doesn't align with other options. The very few that do (Audio for example) should probably be moved away from that.

Softer:
Could you elaborate on why "None" is considered problematic here? Is it about UX consistency across menus, or is there a deeper reason (e.g. config schema, serialization)?

The motivation for adding an explicit "No network" entry was precisely to avoid the "user skipped the menu" ambiguity - an unselected network item currently means "offline install" by omission, which is easy to do accidentally. A required selection with an explicit opt-out felt safer than silent fallthrough. 

If the preferred direction is to remove "None" from the few places that still have it (Audio etc.), I'd be happy to align with that - but I'd like to understand the target pattern first. How should a user signal "I intentionally want nothing here" vs. "I haven't decided yet"?

Torxed:
I don't think the users should be forced to make "No I don't want this".
If you skip it, you choose to not have it as part of the installation process — aka, you don't want us to do anything and we should reliably assume one of two things:
1) You need a custom setup prior/after we're done
2) You made a mistake 

I don't think we need to make life complicated for 99.9% of the other users because a few make a mistake which is easily solved with 3min manual work post-install or, just re-do the installation again.

I might sound harsh, but for some reason I feel very strongly about not forcing users to have to make a "No" choice :/

Softer:
Fair enough - if the core idea doesn't land, I'd rather not drag it out. Should I just close this PR, or is there still room to explore?

One thing I wanted to check before closing it: the non-blocking warning idea I floated earlier (just flagging in the Install preview when no network is configured, no forced choice) - is that worth a shot, or does it fall under the same "don't nag the user" line? Happy to drop the whole thing if it doesn't fit the philosophy, just don't want to close prematurely if there's a softer path that works.

Torxed:
I just wanna clarify that this is just my opinions, and I'm not a dictator so if others feel I'm wrong I'll bow down, but as there hasn't been any additional info I think we can say the "No" aspect can be removed.

However, the warning is sound I think. And serves as a friendly reminder as most people will probably want networking, and those that don't won't mind the warning.

Perhaps in the "Do you want to continue?" prompt we could add it? Unless we want to keep that clean, in which case I'm open to suggestions on where to put it.

h8d13:
> The installer doesn't have "None" as a selection in most cases so that doesn't align with other options. The very few that do (Audio for example) should probably be moved away from that.

Again I think "None" option is the "skip-knowingly" option.

 Instead of having to clear a field for a miss-selection you can just pick "None", if desired. Seems like the clean solution in a lot of cases (and yes is missing in some places).

svartkanin:
With "None" I meant the UX oart as most other settings rely on
* Esc for skipping 
* C-c for resetting 

This pattern is pretty consistent throughout the menu. So introducing a new version means inconsistency. If it is a generally desired addition because users are confusing about the above mentioned key bindings - which I haven't seen so far - then this is a discussion about changing that pattern globally.

I like the idea of having a warning around no network setup, either on the install preview, or the confirmation screen after or even both.

h8d13:
I meant in each menu having an actual "None" IS simpler and more explicit to the user then whatever warnings or key-binds... Then everything is "clear" on user facing side, he can explicitly select " I don't want this" and that satisfies the mandatory needs-selection mechanism still. Without needing to make new warnings: and yes this already done for `Audio` and should probably be added to `Greeter` too IMO

Softer:
On the warning: tried adding it both on the confirm screen and in the Install preview - really liked how it turned out :) That said, while working on the Install preview part I ended up with a broader take on it that I'd rather split out into a separate PR.

On NONE: agreeing with @h8d13 - an explicit "None" option feels more transparent to me than the current "skip = not wanted" behavior.

svartkanin:
>  meant in each menu having an actual "None" IS simpler and more explicit to the user then whatever warnings or key-binds...

Not sure the overhead of that is more valuable, looking at some of the existing settings, like bool, multi-select, tables etc. what value would "None" add there? And if they shouldn't have it there then inconsistencies start popping up again. 

I'm happy to have a discussion around it, maybe an explicit issue may better to track it properly as this PR is now for something else

Softer:
So, going with the warning on the confirm screen? Then later in a separate PR (if it works out :)) I'll show an implementation in the install preview - if you like it, we can move the warning there.

Or should we drop it? :(

svartkanin:
Sounds good, I think the warning is a good enhancement.

Softer:
I've updated the PR and its description.

svartkanin:
The PR still contains the introduction of `None` as a new network config type which I think we should remove from this change and leave it to the outcome of discussion https://github.com/archlinux/archinstall/issues/4464

svartkanin (approved):

PR #4414, opened 2026-04-12
Remove plugin system

### LINKED ISSUE #3021: Problem with plugin
There are some bugs in file "archinstall/lib/plugins.py" and "archinstall/__init__.py". 

# First problem #

If I add `--plugin https://raw.githubusercontent.com/phisch/archinstall-aur/master/archinstall-aur.py` in archinstall command, this error appears:

```
Loading plugin from url ParseResult(scheme='https', netloc='raw.githubusercontent.com', path='/phisch/archinstall-aur/refs/heads/master/archinstall-aur.py', params='', query='', fragment='')
  File "/home/archi/MyArchIso/install.py", line 3, in <module>
    from my_global_menu import Menu
  File "/home/archi/MyArchIso/my_global_menu.py", line 1, in <module>
    from archinstall.lib.global_menu import GlobalMenu
  File "/usr/lib/python3.12/site-packages/archinstall/__init__.py", line 282, in <module>
    post_process_arguments(arguments)
  File "/usr/lib/python3.12/site-packages/archinstall/__init__.py", line 275, in post_process_arguments
    load_plugin(path)
  File "/usr/lib/python3.12/site-packages/archinstall/lib/plugins.py", line 95, in load_plugin
    localized = _localize_path(path)
                ^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/archinstall/lib/plugins.py", line 38, in _localize_path
    converted_path = Path(f'/tmp/{path.stem}_{hashlib.md5(os.urandom(12)).hexdigest()}.py')
                                  ^^^^^^^^^
AttributeError: 'str' object has no attribute 'stem'
```

Seems that the problem is in function `post_process_arguments` in file `archinstall/__init__.py`. This function passes a string to the function `load_plugin` instead of a 'Path' object.
I changed line 276 to `path = Path(args['plugin'])` and the problem was solved.

#Second Problem #
At this point new problem came out: 

```
Loading plugin from url ParseResult(scheme='https', netloc='', path='/raw.githubusercontent.com/phisch/archinstall-aur/refs/heads/master/archinstall-aur.py', params='', query='', fragment='')
  File "/home/archi/MyArchIso/install.py", line 3, in <module>
    from my_global_menu import Menu
  File "/home/archi/MyArchIso/my_global_menu.py", line 1, in <module>
    from archinstall.lib.global_menu import GlobalMenu
  File "/usr/lib/python3.12/site-packages/archinstall/__init__.py", line 282, in <module>
    post_process_arguments(arguments)
  File "/usr/lib/python3.12/site-packages/archinstall/__init__.py", line 275, in post_process_arguments
    load_plugin(path)
  File "/usr/lib/python3.12/site-packages/archinstall/lib/plugins.py", line 95, in load_plugin
    localized = _localize_path(path)
                ^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/archinstall/lib/plugins.py", line 41, in _localize_path
    temp_file.write(urllib.request.urlopen(url.geturl()).read().decode('utf-8'))
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/urllib/request.py", line 215, in urlopen
    return opener.open(url, data, timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/urllib/request.py", line 512, in open
    req = meth(req)
          ^^^^^^^^^
  File "/usr/lib/python3.12/urllib/request.py", line 1268, in do_request_
    raise URLError('no host given')
urllib.error.URLError: <urlopen error no host given>
```

The problem seems localized in funcion `_localize_path` in file `archinstall/lib/plugins.py`,  where funcion url.geturl() return "https:///raw.githubusercontent.com/phisch/archinstall-aur/master/archinstall-aur.py" (there are 3 backslash after "https:").

codefiles:
Closes #3021


svartkanin:
Fine by me, but I'll await @Torxed opinion

Torxed:
Need to consider this a bit. It breaks a few known users use cases, mine included.

Is it just in favor of removing code that otherwise looks unused? Or just to clean up complexity?

h8d13:
think yes it's a lot of complexity in main installer file

We could also document to make modification into `guided.py` at the stage they need with some things that can help them to do this properly like `installation.something` and `profile.provision` or `post_install` probably a few more that are easier to play with properly

codefiles:
Maybe removing it is not the way to go.

svartkanin:
If it is broken but still serves a purpose then there is the option to remove it and introduce a reworked version that is more aligned with the current state of archinstall (the plugin system was introduced a long time ago where the installer was not as flexible and library driven). The other option is the keep it and fix it, if it fits into the current structure and model of the code. 

To me the word plugin sounds a bit more like "apps" and not what it actually does which are hooks during the installation. So maybe re-branding it to installation hooks may be better?  

If it is broken right now and does not provide any value then removing it is probably the right way forward.

codefiles:
@Torxed, could we at minimum remove the following?

https://github.com/archlinux/archinstall/blob/ba7dbeadfc69bebb493b0531909224be136c265a/archinstall/__init__.py#L1-L3

In commit 78969c5, I left this in `__init__.py` because I did not want to further break the plugin system. This has the effect of causing any import of `archinstall` to execute the module-level code in `plugins.py`. I do not think that is ideal with archinstall also intended as a library.

Issue #4405, opened 2026-04-10
Standardize Enum member naming


codefiles:
Python documentation examples for the `enum` module use `UPPER_CASE` for member names; archinstall uses `UPPER_CASE` and `PascalCase`.

- https://docs.python.org/3/library/enum.html
- https://docs.python.org/3/howto/enum.html

PR #4417, opened 2026-04-12
Refactor FilesystemType

codefiles:
Part of #4405

### DIFF
--- a/archinstall/lib/disk/disk_menu.py
+++ b/archinstall/lib/disk/disk_menu.py
@@ -508,17 +508,17 @@ def _boot_partition(sector_size: SectorSize, using_gpt: bool) -> PartitionModifi
 		start=start,
 		length=size,
 		mountpoint=Path('/boot'),
-		fs_type=FilesystemType.Fat32,
+		fs_type=FilesystemType.FAT32,
 		flags=flags,
 	)
 
 
 async def select_main_filesystem_format() -> FilesystemType:
 	items = [
-		MenuItem('btrfs', value=FilesystemType.Btrfs),
-		MenuItem('ext4', value=FilesystemType.Ext4),
-		MenuItem('xfs', value=FilesystemType.Xfs),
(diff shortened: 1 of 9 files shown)


archinstall/lib/models/device.py
```diff
@@ -131,9 +131,14 @@ def parse_arg(
...
 
+				if fs_type := partition.get('fs_type'):
+					fs_type = FilesystemType(fs_type)
+				else:
+					fs_type = None
+
 				device_partition = PartitionModification(
 					status=ModificationStatus(partition['status']),
-					fs_type=FilesystemType(partition['fs_type']) if partition.get('fs_type') else None,
+					fs_type=fs_type,
```
codefiles, line 141: Without this change mypy produces this error:

```
archinstall/lib/models/device.py:136: error: Argument 1 to "FilesystemType" has incompatible type "str | None"; expected "str"  [arg-type]
```

archinstall/lib/disk/disk_menu.py
```diff
@@ -508,17 +508,17 @@ def _boot_partition(sector_size: SectorSize, using_gpt: bool) -> PartitionModifi
...
 	)
 
 
 async def select_main_filesystem_format() -> FilesystemType:
 	items = [
-		MenuItem('btrfs', value=FilesystemType.Btrfs),
-		MenuItem('ext4', value=FilesystemType.Ext4),
-		MenuItem('xfs', value=FilesystemType.Xfs),
-		MenuItem('f2fs', value=FilesystemType.F2fs),
+		MenuItem(FilesystemType.BTRFS.value, value=FilesystemType.BTRFS),
```
svartkanin, line 518: FYI there's a helper function https://github.com/archlinux/archinstall/blob/bf9b9cb7c17c87595676a52210803a2ebb02af48/archinstall/tui/ui/menu_item.py?plain=1#L147
it was introduced after this was implemented so one day we can move it :)

svartkanin (approved):

PR #4418, opened 2026-04-12
Fix DataclassInstance annotation and libcrypt loading

AmmaarBakshi:
## PR Description

This PR fixes two startup issues:

1. `archinstall/lib/output.py`
   - Changed `DataclassInstance` to a forward reference (`'DataclassInstance'`) so the annotation is only evaluated during type checking and does not crash at runtime.

2. `archinstall/lib/crypt.py`
   - Made `libcrypt` loading POSIX-only using `find_library('crypt')`.
   - Added a runtime guard so import does not fail immediately on Windows when `libcrypt` is unavailable.

## Tests and Checks
- [x] `python -m py_compile archinstall/lib/output.py archinstall/lib/crypt.py`
- [ ] Additional Linux runtime testing pending

### DIFF
--- a/archinstall/lib/crypt.py
+++ b/archinstall/lib/crypt.py
@@ -1,6 +1,7 @@
 import base64
 import ctypes
 import os
+from ctypes.util import find_library
 from pathlib import Path
 
 from cryptography.fernet import Fernet, InvalidToken
@@ -8,7 +9,13 @@ from cryptography.hazmat.primitives.kdf.argon2 import Argon2id
 
 from archinstall.lib.output import debug
 
-libcrypt = ctypes.CDLL('libcrypt.so')
+libcrypt_name = find_library('crypt') or 'libcrypt.so'
+try:
+	libcrypt = ctypes.CDLL(libcrypt_name)
+except OSError as exc:
+	raise ImportError(
+		'Could not load system libcrypt library; make sure libcrypt is installed and accessible.'
+	) from exc
 
 libcrypt.crypt.argtypes = [ctypes.c_char_p, ctypes.c_char_p]
 libcrypt.crypt.restype = ctypes.c_char_p


codefiles:
> - Changed `DataclassInstance` to a forward reference (`'DataclassInstance'`) so the annotation is only evaluated during type checking and does not crash at runtime.

A few things you missed:

- `ruff check`

  ```
  UP037 [*] Remove quotes from type annotation
    --> archinstall/lib/output.py:20:6
     |
  18 |     @staticmethod
  19 |     def _get_values(
  20 |         o: 'DataclassInstance',
     |            ^^^^^^^^^^^^^^^^^^^
  21 |         class_formatter: str | Callable | None = None,  # type: ignore[type-arg]
  22 |         filter_list: list[str] = [],
     |
  help: Remove quotes
  ```
  
  https://docs.astral.sh/ruff/rules/quoted-annotation/

- https://github.com/archlinux/archinstall/blob/bf9b9cb7c17c87595676a52210803a2ebb02af48/archinstall/lib/output.py#L13-L14

- d57709c

> - Made `libcrypt` loading POSIX-only using `find_library('crypt')`.

Why?

> - Added a runtime guard so import does not fail immediately on Windows when `libcrypt` is unavailable.

archinstall is not intended to run on Windows.

LLM generated?

AmmaarBakshi:
> > * Changed `DataclassInstance` to a forward reference (`'DataclassInstance'`) so the annotation is only evaluated during type checking and does not crash at runtime.
> 
> A few things you missed:
> 
> * `ruff check`
>   ```
>   UP037 [*] Remove quotes from type annotation
>     --> archinstall/lib/output.py:20:6
>      |
>   18 |     @staticmethod
>   19 |     def _get_values(
>   20 |         o: 'DataclassInstance',
>      |            ^^^^^^^^^^^^^^^^^^^
>   21 |         class_formatter: str | Callable | None = None,  # type: ignore[type-arg]
>   22 |         filter_list: list[str] = [],
>      |
>   help: Remove quotes
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   https://docs.astral.sh/ruff/rules/quoted-annotation/
> * https://github.com/archlinux/archinstall/blob/bf9b9cb7c17c87595676a52210803a2ebb02af48/archinstall/lib/output.py#L13-L14
> * [d57709c](https://github.com/archlinux/archinstall/commit/d57709cb2f23aad2de5e546780ced0a8889a77d2)
> 
> > * Made `libcrypt` loading POSIX-only using `find_library('crypt')`.
> 
> Why?
> 
> > * Added a runtime guard so import does not fail immediately on Windows when `libcrypt` is unavailable.
> 
> archinstall is not intended to run on Windows.
> 
> LLM generated?

good catch on the ruff flag, added from __future__ import annotations to the module so the quoted annotation isn't needed anymore.

for crypt.py, yeah fair enough, dropped the os.name guard entirely. kept find_library('crypt') though since hardcoding libcrypt.so skips the linker search path which could break on distros that put it somewhere non-standard. the try/except gives a clear error if it's missing instead of a raw OSError.

and no lol, just overthought the windows compat thing.

force pushed with the fixes.

archinstall/lib/output.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
codefiles: a92cd50 :unamused:

AmmaarBakshi: should i remove "from __future__ import annotations" 😭 ?

AmmaarBakshi: i have done some changes, tell me about them.....

codefiles:
> for crypt.py, yeah fair enough, dropped the os.name guard entirely. kept find_library('crypt') though since hardcoding libcrypt.so skips the linker search path which could break on distros that put it somewhere non-standard. the try/except gives a clear error if it's missing instead of a raw OSError.

The `archinstall` package has `libcrypt.so` as a dependency.

https://gitlab.archlinux.org/archlinux/packaging/packages/archinstall/-/blob/main/PKGBUILD?ref_type=heads#L23

I do not think the change is warranted without demonstrating it actually solves a problem.

AmmaarBakshi:
> > for crypt.py, yeah fair enough, dropped the os.name guard entirely. kept find_library('crypt') though since hardcoding libcrypt.so skips the linker search path which could break on distros that put it somewhere non-standard. the try/except gives a clear error if it's missing instead of a raw OSError.
> 
> https://gitlab.archlinux.org/archlinux/packaging/packages/archinstall/-/blob/main/PKGBUILD?ref_type=heads#L23
> 
> The `archinstall` package has `libcrypt.so` as a dependency. I do not think the change is warranted without demonstrating it actually solves a problem.

Ah, that makes total sense. I didn't realize the PKGBUILD strictly enforced the .so dependency. Since the output.py forward reference issue is also natively resolved by Python 3.14 now, there isn't really anything left of value in this PR. I'll go ahead and close it! Thanks for the review.

PR #4420, opened 2026-04-13
Add optional "Additional fonts" selection to Applications menu

### LINKED ISSUE #4412: Add optional Fonts selection to Applications menu
### describe the request

Currently there's no easy way to install emoji and CJK font packages during installation. Most desktop users end up with missing emoji (squares in browsers) and no CJK support, then have to figure out which packages to install manually.

What if we add a "Additional fonts" option to the Applications menu - same place as Audio, Bluetooth, Firewall? It would be a simple multi-select with two (or more) choices:

  - noto-fonts-emoji (color emoji for browsers and apps)
  - noto-fonts-cjk (Chinese, Japanese, Korean characters)

Nothing is pre-selected - the user picks what they need, or skips entirely. No extra packages installed unless explicitly chosen.

This doesn't change the minimalist approach - just gives users an easy way to grab common font packages without having to remember package names in "Additional packages".

I have a working implementation if there's interest. Here's how it looks:

Image

Image

Softer:
Closes #4412

## PR Description:
Adds a multi-select "Additional fonts" option to the Applications menu, allowing users to install common font packages during installation:
- noto-fonts (Unicode font coverage for most languages)
- noto-fonts-emoji (color emoji for browsers and apps)
- noto-fonts-cjk (Chinese, Japanese, Korean characters)

Nothing is pre-selected - users pick what they need or skip entirely. Follows the same pattern as Audio, Bluetooth, and Firewall applications.

## Tests and Checks
- Open Applications -> Additional fonts -> select one or more -> install completes with packages
- Skip fonts entirely -> no font packages installed
- Save/load config with fonts_config via --config

### DIFF
--- a/archinstall/lib/applications/application_menu.py
+++ b/archinstall/lib/applications/application_menu.py
@@ -80,7 +80,7 @@ class ApplicationMenu(AbstractSubMenu[ApplicationConfiguration]):
 				key='firewall_config',
 			),
 			MenuItem(
-				text=tr('Fonts'),
+				text=tr('Additional fonts'),
 				action=select_fonts,
 				value=self._app_config.fonts_config,
 				preview_action=self._prev_fonts,
@@ -128,7 +128,7 @@ class ApplicationMenu(AbstractSubMenu[ApplicationConfiguration]):
 		if item.value is not None:
 			config: FontsConfiguration = item.value
 			packages = ', '.join(f.value for f in config.fonts)
-			return f'{tr("Fonts")}: {packages}'
+			return f'{tr("Additional fonts")}: {packages}'
 		return None
 
 
--- a/archinstall/lib/applications/application_menu.py
+++ b/archinstall/lib/applications/application_menu.py
@@ -237,6 +237,7 @@ async def select_firewall(preset: FirewallConfiguration | None = None) -> Firewa
 
 async def select_fonts(preset: FontsConfiguration | None = None) -> FontsConfiguration | None:
 	descriptions = {
+		FontPackage.NOTO: tr('Unicode font coverage for most languages'),
 		FontPackage.EMOJI: tr('color emoji for browsers and apps'),
 		FontPackage.CJK: tr('Chinese, Japanese, Korean characters'),
 	}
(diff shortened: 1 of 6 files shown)


exilonium (commented):
lgtm

svartkanin:
Thnks for the effort, I'll need some time to test it properly

archinstall/lib/applications/application_menu.py
```diff
@@ -217,3 +233,36 @@ async def select_firewall(preset: FirewallConfiguration | None = None) -> Firewa
 			return FirewallConfiguration(firewall=result.get_value())
 		case ResultType.Reset:
 			return None
+
+
+async def select_fonts(preset: FontsConfiguration | None = None) -> FontsConfiguration | None:
+	descriptions = {
```
svartkanin: Can we move this into the `FontPackage` itself I think that will make this more readable with a 

```
class FontPackage(StrEnum):
...
	def description(self) -> str:
		match self:
			case FontPackage.NOTO:
				...
			case FontPackage.EMOJI:
				...
			case FontPackage.CJK:
				...
```

and then 

```
	items = [MenuItem(f.description(), value=f) for f in FontPackage]
	group = MenuItemGroup(items)
```

there is also a helper function 
```
MenuItemGroup.from_enum(FontPackage)
```

but that won't handle the custom description part, so we could enhance that with 

```
	@classmethod
	def from_enum(
		cls,
		enum_cls: type[Enum],
		sort_items: bool = False,
		preset: Enum | None = None,
	) -> Self:
		text = elem.description() if hasattr(enum_cls, 'description') else elem.value
		items = [MenuItem(text, value=elem) for elem in enum_cls]
```

or something along those lines

svartkanin (approved):

svartkanin:
A test is failing

Softer:
> A test is failing

Do you mean pytest?

svartkanin:
Yes https://github.com/archlinux/archinstall/actions/runs/24398747290/job/71263299254#step:7:1

Softer:
Is it better now? :)

Softer:
Added ttf-liberation and ttf-dejavu to the font selection. Liberation provides metric-compatible Arial/Times/Courier replacements with full Cyrillic support - essential for Steam client and Valve games (Dota 2, Source engine). DejaVu offers wide Unicode coverage and works as a reliable fallback font.

Both of them are optiononal - user can install if he needs.

PR #4427, opened 2026-04-14
Revert pydantic update due to upstream bug

codefiles:
reverts commit: 88366ba
upstream issue: pydantic/pydantic#13074

PR #4434, opened 2026-04-16
ci: Update renovate to track Arch Linux repos using repology

dylanmtaylor:
Pin runtime deps to current Arch package versions and configure Renovate to use the Repology datasource (Arch repos) instead of PyPI for version bumps. This prevents mismatches like #4427 where a PyPI release landed before Arch shipped it. Dev deps stay on PyPI since they only run in CI.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,11 +18,11 @@ classifiers = [
     "Operating System :: POSIX :: Linux",
 ]
 dependencies = [
-    "pyparted>=3.13.0",
+    "pyparted==3.13.0",
     "pydantic==2.12.5",
-    "cryptography>=45.0.7",
-    "textual>=5.3.0",
-    "markdown-it-py[linkify]>=4.0.0",
+    "cryptography==46.0.7",
+    "textual==8.2.3",
+    "markdown-it-py[linkify]==4.0.0",
 ]
 
 [project.urls]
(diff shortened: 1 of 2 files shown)


dylanmtaylor:
One consideration, repology scrapes the versions from Arch packages, but might slightly lag - in practice this means the PR would be opened a bit later. Probably irrelevant but worth pointing out.

codefiles:
Did you consider any of the other datasources?

https://docs.renovatebot.com/modules/datasource/git-tags/
https://docs.renovatebot.com/modules/datasource/gitlab-tags/
https://gitlab.archlinux.org/archlinux/packaging/packages/python-pydantic/-/tags

dylanmtaylor:
> Did you consider any of the other datasources?
> 
> https://docs.renovatebot.com/modules/datasource/git-tags/
> https://docs.renovatebot.com/modules/datasource/gitlab-tags/
> https://gitlab.archlinux.org/archlinux/packaging/packages/python-pydantic/-/tags

I picked repology because I'm familiar with it from my work on some NixOS packages. I suspect that repology will work well for us.

codefiles:
I will submit a pull request to query the version information from Arch Linux channels rather than Repology then. Is that ok with you?

dylanmtaylor:
That's fine

svartkanin (approved):

PR #4440, opened 2026-04-17
Enable power management services after package installation

### LINKED ISSUE #2557: power-profiles-daemon package should be installed by default

Softer:
Fixes #2557 

## PR Description:
The power management handler installs the selected packages but does not enable their systemd services. After reboot, neither power-profiles-daemon nor tuned starts automatically, making the power management selection ineffective.

This adds enable_service() calls for power-profiles-daemon.service and tuned.service, matching the pattern already used by Bluetooth, Firewall, and Print Service handlers.

## Tests and Checks
- Select "Power management" in Applications menu on a system with battery
- Choose power-profiles-daemon, complete installation, reboot
- Verify systemctl is-enabled power-profiles-daemon.service returns enabled
- Repeat with tuned, verify systemctl is-enabled tuned.service returns enabled

### DIFF
--- a/archinstall/applications/power_management.py
+++ b/archinstall/applications/power_management.py
@@ -21,6 +21,18 @@ class PowerManagementApp:
 			'tuned-ppd',
 		]
 
+	@property
+	def ppd_services(self) -> list[str]:
+		return [
+			'power-profiles-daemon.service',
+		]
+
+	@property
+	def tuned_services(self) -> list[str]:
+		return [
+			'tuned.service',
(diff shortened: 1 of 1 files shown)


h8d13:
#4015 same conversation on the original pull request, probably should be done explicitly!

svartkanin (approved):

PR #4445, opened 2026-04-19
Allow console font selection during install

h8d13:
Since `'default8x16'` was hard-coded to satisfy previous issue in PR #3928

This allows to scan locally available fonts: then populate the list and write the correct config into `/etc/vconsole.conf`

One edge case that is important to note is that `terminus-font` is available in ISO but not in the target hence the naive `ter-` check.

Continuation of conversation on #4356 by @Softer but instead setting it in live, we just set it up for the target.


Softer:
I'm already working on the same task - fonts and locales for the target system, just wait for merging #4356... :)

h8d13:
> I'm already working on the same task - fonts and locales for the target system, just wait for merging #4356... :)

Great I'll close this one, was just making sure :) Think its more important then setting it during the menu tbh

PR #4460, opened 2026-04-21
Update KDE Plasma flavor descriptions and add core applications to plasma and plasma-meta

lapointek:
## PR Description:

- Update descriptions for KDE Plasma flavors to better clarify their differences.
- Add core Plasma applications (Konsole, Kate, Dolphin, Ark) to the plasma and plasma-meta flavors to provide a more complete default desktop environment.

## Tests and Checks
I have tested changes.

### DIFF
--- a/archinstall/default_profiles/desktops/plasma.py
+++ b/archinstall/default_profiles/desktops/plasma.py
@@ -29,7 +29,7 @@ class PlasmaFlavor(StrEnum):
 		match self:
 			case PlasmaFlavor.Meta:
 				ty = tr('Package')
-				desc = tr('A meta package that depends on a curated set of components. Installing it pulls in the full desktop \nenvironment including extras. Recommended for most users who want a complete desktop.')
+				desc = tr('A meta package that depends on a curated set of packages. Installing it pulls in the full desktop \nenvironment, including extras. Recommended for most users who want a complete desktop.')
 				info = available_package(self.value)
 
 				if info is not None:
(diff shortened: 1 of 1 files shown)


archinstall/default_profiles/desktops/plasma.py
```diff
@@ -57,9 +57,9 @@ def package_details(self) -> str:
 	def packages(self) -> list[str]:
 		match self:
 			case PlasmaFlavor.Meta:
-				return ['plasma-meta']
+				return ['plasma-meta', 'konsole', 'dolphin', 'kate', 'ark']
```
svartkanin, line 60: The reason for adding this granular setting was to avoid any additional packagtes. Why do we need these?

lapointek: Using the TTY after installing a desktop environment just to install a terminal doesn’t provide a great first-user experience, in my opinion. I had assumed that was the purpose of the minimal install, rather than the plasma-meta/full plasma package. I didn’t realize there was a deliberate decision not to include applications like Dolphin or Konsole. In that case, disregard this PR.

Torxed: tbh, there's not any clear consensus what to do here but we've gotten some pretty taxing feedback from both packagers of KDE and users whenever we try to fiddle with the KDE profile.

So until there's consensus from both sides, we should consider the KDE profile and on-off switch.
If you want a working thing, listen to the KDE packagers and install **everything**, and if you want choices, use the non working options and build a .json conf with additional packages and re-use that.

Apologies that we can't do anything with the KDE profile, but that's where we are at trying our best in the middle.

PR #4469, opened 2026-04-24
Add console font selection to Locales menu

Softer:
## PR Description:
Adds a new `Console font` item (4th) to the Locales submenu, allowing the user to pick a font for the target system's TTY (written to `/etc/vconsole.conf`).
 
Font list is built from `/usr/share/kbd/consolefonts` on the running ISO.
 
If the user picks a `ter-*` terminus font, the `terminus-font` package is automatically added to the target install. 

## Why 

Currently the target TTY always gets `default8x16` regardless of the language the user picked. That breaks for users with languages that need a wider Unicode coverage (partialy addresses #3735). Giving a menu entry is the simplest fix that doesn't force heuristics on anyone.
 
## Behavior / backward compatibility 
 
- Default remains `default8x16`, so existing user configs and default installs produce the exact same `vconsole.conf` as before.
- `LocaleConfiguration` gains `console_font: str = 'default8x16'`. Old configuration JSONs without that field keep working via the default value. 
- The `json()` / `_load_config()` / `preview()` sides are updated symmetrically. 
 
## Scope 
One thing only - menu selection + vconsole write. A follow-up PR will offer to propagate the installer's chosen language (sys_lang / sys_enc / console_font) to the target locale configuration.
 
## Test

- Default install: `/etc/vconsole.conf` contains `FONT=default8x16` as before. 
- Pick `ter-v32b` in the menu: `FONT=ter-v32b`, `terminus-font` installed.
- Load an old `user_configuration.json` with no `console_font`: default kicks in, no crash.

### DIFF
--- a/archinstall/lib/locale/utils.py
+++ b/archinstall/lib/locale/utils.py
@@ -1,3 +1,5 @@
+import os
+
 from archinstall.lib.command import SysCommand
 from archinstall.lib.exceptions import ServiceException, SysCallError
 from archinstall.lib.output import error
@@ -24,6 +26,17 @@ def list_locales() -> list[str]:
 				locales.append(line.rstrip())
 
 	return locales
+
+
+def list_console_fonts() -> list[str]:
+	fonts: set[str] = set()
+
+	for entry in os.listdir('/usr/share/kbd/consolefonts'):
+		if entry.endswith('.gz'):
+			name = entry.removesuffix('.gz').removesuffix('.psfu').removesuffix('.psf').removesuffix('.cp').removesuffix('.fnt')
+			fonts.add(name)
+
+	return sorted(fonts)
 
 
 def list_x11_keyboard_languages() -> list[str]:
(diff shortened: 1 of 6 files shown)


archinstall/lib/locale/utils.py
```diff
@@ -26,6 +28,17 @@ def list_locales() -> list[str]:
 	return locales
 
 
+def list_console_fonts() -> list[str]:
+	fonts: set[str] = set()
+
+	for entry in os.listdir('/usr/share/kbd/consolefonts'):
```
svartkanin: Can we change this to use `pathlib.Path` instead and make the suffix removal less targetted 

```
def list_console_fonts():
    directory = Path('/usr/share/kbd/consolefonts')
    
    fonts = {
        path.name.split('.')[0] 
        for path in directory.glob('*.gz')
    }

    return sorted(fonts)
```

archinstall/lib/locale/utils.py
```diff
@@ -26,6 +28,17 @@ def list_locales() -> list[str]:
 	return locales
 
 
+def list_console_fonts() -> list[str]:
```
svartkanin, line 33: Add `@lru_cache` to prevent repeated loading

svartkanin (approved):

PR #4470, opened 2026-04-24
Add developer VM script for faster testing iteration

Softer:
## PR Description
 
Sharing a small helper I've been using locally to iterate on archinstall changes inside a VM. It saved me enough time that I figured other contributors might want it too. 

### The pain 

When I change something in archinstall and want to test it end-to-end (run the installer in a VM, walk the menus, let it actually partition and install), the current options all share the same shape.
`build_iso.sh` rebuilds the full ISO on every source change, carrying gcc/pip/uv/wheel build each time. The README "Testing" section and the wiki "Building and Testing" describe the same loop manually: boot
vanilla Arch ISO, `pacman -S git python-pip gcc pkgconf`, clone repo, run `python -m archinstall`. Both work, but both rebuild the world on every change.

What I wanted was: edit a `.py` on the host, relaunch the VM, run `archinstall`, see the change. No wheel build. No ISO rebuild. Just a hot-reload dev loop. 
 
### What this PR adds

`dev_vm.sh` at the repo root, sibling of `build_iso.sh`. Two independent pieces: 

1. A minimal dev ISO built with `archiso` that carries only archinstall's runtime deps - no gcc, no pip, no wheel build. The dep list is derived at build time from `pyproject.toml`'s `[project.dependencies]`
(PEP 503 normalized, mapped to the Arch `python-<name>` convention, which holds for every current dep), so adding or removing a runtime dep upstream does not require touching the script. The ISO is tiny and
only needs rebuilding when the runtime dep list actually changes.
2. A QEMU launcher that shares the project source read-only to the guest over 9p. Host edits appear live inside the VM. On guest login, the 9p share auto-mounts at `/root/archinstall-dev`, the shell `cd`-s
into it, and `archinstall` is aliased to `python -m archinstall` from the shared source. A 9p mount failure prints a clear error, so a broken virtfs setup does not silently turn into "No module named
archinstall".
 
So the inner loop becomes: edit code on host, `./dev_vm.sh k`, type `archinstall` inside the guest. No rebuild, no reinstall.

### Modes

- `./dev_vm.sh` - build ISO if missing, fresh disk, boot 
- `./dev_vm.sh rebuild | r` - force ISO rebuild, fresh disk, boot
- `./dev_vm.sh keep | k` - reuse disk, boot ISO
- `./dev_vm.sh boot | b` - boot from installed disk (no ISO)
- `./dev_vm.sh clean | c` - remove disk, NVRAM, ISO
- `./dev_vm.sh -h` - help
 
### Host artifacts (all git-ignored) 

- `.dev-iso/` - generated dev ISO
- `.dev-disk.qcow2` - VM disk
- `.dev-ovmf-vars.fd` - persistent UEFI NVRAM
- `.dev-configs/` - optional, user-created, shared rw to guest at `/root/cfg` (handy for keeping test `user_configuration.json` / `user_credentials.json` between runs)
 
### Host requirements
 
Arch-based host (Arch, Manjaro, EndeavourOS, ...) because `mkarchiso` is Arch-only. The script reads `/etc/os-release` and fails fast with a clear message on non-Arch hosts, before ever touching pacman. 
Runtime deps: `qemu-base`, `edk2-ovmf`, `sudo`, `python` (for parsing `pyproject.toml`). OVMF paths are probed across the common layouts (Arch `edk2-ovmf`, Debian/Ubuntu `ovmf`, Fedora `edk2-ovmf`) so the same
 script works across Arch-based distros with different OVMF packaging. 9p `mapped-xattr` mode needs xattr support on the filesystem holding the repo (ext4/btrfs work out of the box; ZFS needs `xattr=sa`). 

### Changes in this PR 

- `dev_vm.sh` - new file 
- `.gitignore` - three lines: `!dev_vm.sh` as an exception under the existing `**/**.sh` blanket, plus `/.dev-iso/` and `/.dev-configs/` for the artifact folders
 
Open to feedback on placement (repo root vs a `scripts/` folder), naming, and scope. I kept artifact names dot-prefixed to hide them from `ls` and group them under one visual namespace.
 
Note: the CI ISO artifact attached to this PR is the regular `build_iso.sh` output, unaffected by `dev_vm.sh` (which is a separate dev-only tool and does not touch the CI build path).


test_tooling/dev_vm/dev_vm.sh
```diff
@@ -0,0 +1,272 @@
+#!/bin/bash
```
svartkanin, line 1: lets move this under a `scripts/` folder rather than having this on the root

dev_vm.sh
```diff
@@ -0,0 +1,272 @@
...
+    fi
+}
+
+# Parse runtime deps from pyproject.toml and map to Arch package names.
+# Assumes the `python-<name>` convention, which holds for every archinstall
+# dependency today. If a future dep breaks the convention, fix here.
+derive_packages() {
+    command -v python3 >/dev/null 2>&1 || err "python3 not found on host (needed to parse pyproject.toml)"
+    local result
+    result=$(python3 - "$PROJECT_DIR/pyproject.toml" <<'PY'
```
svartkanin: To make this less confusing this should probably be extracted into it's own python file.
At which point there is the question would it make more sense to write this whole script in python rather?

svartkanin:
@Torxed I haven't tested this, any thoughts on including this?

dev_vm.sh
Softer: The bulk of dev_vm.sh is QEMU invocation, 9p mount setup, kernel/initrd extraction and shell-script-y environment glue, where bash is fine; the only piece that benefited from a real language was the pyproject.toml parser, which is now its own Python module.

Softer:
I added one small QoL change: /var/log/archinstall is now a third 9p share (backed by .dev-logs/ on the host), and the archinstall wrapper wipes the log directory contents before each invocation. So you can tail -f .dev-logs/install.log on the host across iterations without rebooting the VM. BTW, we can run tests without rebooting the VM in most cases :)

Softer:
Moved to test_tooling/dev_vm/ instead - after #4539 landed with test_tooling/mkosi/ and test_tooling/qemu/, this felt like a better home than a top-level scripts/ folder.

Torxed:
> @Torxed I haven't tested this, any thoughts on including this?

Same, and maintaining bash isn't really something that tickle an itch :)

But I'll look through it today, and see what it does

Softer:
> > @Torxed I haven't tested this, any thoughts on including this?
> 
> Same, and maintaining bash isn't really something that tickle an itch :)
> 
> But I'll look through it today, and see what it does

Any updates here? :)

PR #4472, opened 2026-04-24
Fix OVMF paths in README QEMU examples

Softer:
All three `qemu-system-x86_64` examples in README pointed both `-drive if=pflash` entries at the same file. OVMF needs `OVMF_CODE` first and `OVMF_VARS` second; as written, EFI NVRAM is not initialized correctly. 

Path `/usr/share/ovmf/x64/` is also stale - Arch replaced `ovmf` with `edk2-ovmf`, which installs under `/usr/share/edk2/x64/`.

Fixed in three places (testimage loop, base Boot ISO, espeakup variant). Paths verified via `pacman -Qo` against `edk2-ovmf 202508-1`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ This can be done by installing `pacman -S arch-install-scripts util-linux` local
     # losetup --partscan --show ./testimage.img
     # pip install --upgrade archinstall
     # python -m archinstall --script guided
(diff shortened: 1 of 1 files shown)


Torxed (approved):

Torxed:
Thank you for the cleanup, was a while since they changed this in packaging hehe.

PR #4473, opened 2026-04-24
Fix argv injection in _create_user and gpasswd loop

Softer:
Same class of bug as #4443, two sibling call sites that were missed back then. `_create_user` and the `gpasswd` loop interpolate `user.username` and `group` into f-strings that reach `SysCommand(f'arch-chroot -S {target} {cmd}')`, where `shlex.split` parses the payload as argv. A `username` like `--uid 0` or `-G wheel,root` is accepted as `useradd`/`gpasswd` flags and gives trivial escalation. Vector is the same as #4443: custom `install.py`, plugins, tampered `--config` JSON.

## Changes
`_create_user` now builds an argv list and calls `run()`. Failures are logged via `debug()`, and the `SystemError` re-raise is kept so guided install still aborts.

The `gpasswd` loop gets the same switch. Also wrapped it in `try/except` with `debug()` - previously a failing `gpasswd` was silently ignored.

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -1922,16 +1922,17 @@ class Installer:
 		if not handled_by_plugin:
 			info(f'Creating user {user.username}')
 
-			cmd = 'useradd -m'
+			cmd = ['arch-chroot', '-S', str(self.target), 'useradd', '-m']
 
 			if user.sudo:
-				cmd += ' -G wheel'
+				cmd += ['-G', 'wheel']
 
-			cmd += f' {user.username}'
+			cmd.append(user.username)
 
 			try:
-				self.arch_chroot(cmd)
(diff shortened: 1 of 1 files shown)


archinstall/lib/installer.py
```diff
@@ -1922,16 +1922,17 @@ def _create_user(self, user: User) -> None:
 		if not handled_by_plugin:
 			info(f'Creating user {user.username}')
 
-			cmd = 'useradd -m'
+			cmd = ['arch-chroot', '-S', str(self.target), 'useradd', '-m']
```
svartkanin: There are quite a few of these commands now wouldn't it make sense to create a new helper function that takes the command to be run and adds the `arch-chroot -S <target>`?

svartkanin (approved):

PR #4474, opened 2026-04-24
Extend validate_bootloader_layout with UEFI-dependent checks

Softer:
Follow-up to #4442. The validator introduced there covered two Limine-specific layout cases; everything else that makes a config unbootable still lived in GlobalMenu._validate_bootloader, which meant the silent / --config path through guided.py and the safety-net call inside Installer._add_limine_bootloader never saw those checks. 
 
This PR grows validate_bootloader_layout with a is_uefi parameter and three checks that were missing or menu-only: 
 
- Systemd-boot / Efistub / rEFInd on a BIOS system - none of them boot without UEFI. Previously only the rEFInd case was caught, and only in the TUI.
- Efistub with a non-FAT boot partition - the UEFI firmware reads the kernel directly from that partition, so it must be FAT.
- The rEFInd UEFI-only guard moves from GlobalMenu._validate_bootloader into the util, so all three call sites share it. 
 
GlobalMenu, guided.py and Installer._add_limine_bootloader now pass SysInfo.has_uefi() as the third argument. No changes to what currently valid configs look like.
 
Tested in QEMU against 7 scenarios (BIOS and UEFI, happy paths and every new failure branch) - all behave as expected.

### DIFF
--- a/archinstall/lib/bootloader/utils.py
+++ b/archinstall/lib/bootloader/utils.py
@@ -2,6 +2,7 @@ from dataclasses import dataclass
 from enum import Enum, auto
 from pathlib import Path
 
+from archinstall.lib.hardware import SysInfo
 from archinstall.lib.models.bootloader import Bootloader, BootloaderConfiguration
 from archinstall.lib.models.device import DiskLayoutConfiguration
 
@@ -22,7 +23,6 @@ class BootloaderValidationFailure:
 def validate_bootloader_layout(
 	bootloader_config: BootloaderConfiguration | None,
 	disk_config: DiskLayoutConfiguration | None,
-	is_uefi: bool,
 ) -> BootloaderValidationFailure | None:
 	"""Validate bootloader configuration against disk layout.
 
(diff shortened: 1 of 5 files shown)


archinstall/lib/bootloader/utils.py
```diff
@@ -17,9 +19,13 @@ class BootloaderValidationFailure:
 	description: str
 
 
+_UEFI_ONLY_BOOTLOADERS = (Bootloader.Systemd, Bootloader.Efistub, Bootloader.Refind)
```
svartkanin: Again, can we keep these things inside the actual type definitions? Why are extracting that into global variables? Similar to https://github.com/archlinux/archinstall/pull/4442#pullrequestreview-4151352765

archinstall/scripts/guided.py
```diff
@@ -217,6 +218,7 @@ def main(arch_config_handler: ArchConfigHandler | None = None) -> None:
 	if failure := validate_bootloader_layout(
 		arch_config_handler.config.bootloader_config,
 		arch_config_handler.config.disk_config,
+		SysInfo.has_uefi(),
```
svartkanin: this is a global system setting that is always constant, we shouldn't have to pass that into the function but simply retrieve that in the function itself

Softer:
Thanks, agreed.

svartkanin (approved):

Issue #4466, opened 2026-04-23
Request: Add firmware selection


h8d13:
### describe the request

Currently in `__base_packages__` we always add the full meta [`linux-firmware`](https://archlinux.org/packages/core/any/linux-firmware/) and pulls in ~600mb

For VM, minimal server setups this is massive overkill we could extend to allow explicit selection over https://archlinux.org/packages/?sort=&q=linux-firmware using: `Full | Minimal | Vendor` and vendor being a multi-select

Torxed:
I do like the idea, not sure what minimal would be (other than, not installing any firmware)? And my only wish is that full is default unless users choose otherwise.

h8d13:
Yes same pattern as profiles minimal = nothing :D And still would run fine on VM for example !

And yes agree on full default. Going to test this afternoon on fork(c258f01e1b1484900c479726b2b7eaf775bab9e9)

dylanmtaylor:
I feel like firmware selection belongs in an advanced option to not confuse people. I don't know if --advanced is still a thing or not.

Softer:
Supportive of this. The ~600MB `linux-firmware` meta is painful for VMs, containers, and minimal server setups, and Arch already split the package specifically to enable this granularity.
 
One safety angle worth considering: `Minimal` (and arguably `Vendor` with a narrow selection) is a footgun on bare metal - user picks Minimal, reboots, no WiFi or GPU firmware, has no idea why. Two complementary safeguards come to mind: 
 
1. Gate `Minimal` / `Vendor` behind `--advanced`, as @dylanmtaylor suggested.
2. Emit an install-time warning on the confirm screen when `firmware_type != FULL`, so the user sees a yellow "device firmware may fail to load" line before committing.
 
For (2), there's infrastructure already sitting in PR #4408 - `ConfigurationOutput.get_install_warnings()` / `_render_install_warnings()` driven by a list of short yellow strings shown in the confirmation header. It was added for the "no network configuration" case but is generic. I'd like to iterate on that system (more scenarios, richer messages), so plumbing a Minimal/Vendor-firmware warning through it would be a natural fit and wouldn't need a new mechanism. 

[...]

PR #4475, opened 2026-04-24
Show install summary when configuration is valid

Softer:
## About

Adds a "Ready to install" confirmation plus a two-column summary of the current configuration to the Install menu preview. Previously the preview was empty when everything was valid, giving the user no
explicit "ready" signal. 
 
## Rows 
 
The summary shows one row per meaningful section, skipping any whose underlying config is not set: disks with FS and optional LUKS flag, bootloader, kernel, profile, greeter, additional package count, network,
 locale, timezone. Column width adapts to the longest translated label so i18n does not break the alignment. 

## How does it look?
image

 ## Test plan

- [x] Hover Install with an empty config: missing configurations list appears (unchanged). 
- [x] Hover Install with an invalid bootloader config: existing error text appears (unchanged).
- [x] Hover Install with a valid config on UEFI with LUKS: "Ready to install" followed by the summary, disks row shows (fs + LUKS).
- [x] Non-desktop profile (Server, Minimal): Profile row shows profile name, Greeter row is skipped. 
- [x] Desktop profile with greeter selected: Profile row shows DE names, Greeter row shows greeter value.
- [x] Locales other than English: labels translate and column alignment still holds.

### DIFF
--- a/archinstall/lib/global_menu.py
+++ b/archinstall/lib/global_menu.py
@@ -6,7 +6,7 @@ from archinstall.lib.args import ArchConfig
 from archinstall.lib.authentication.authentication_menu import AuthenticationMenu
 from archinstall.lib.bootloader.bootloader_menu import BootloaderMenu
 from archinstall.lib.bootloader.utils import validate_bootloader_layout
-from archinstall.lib.configuration import save_config
+from archinstall.lib.configuration import ConfigurationOutput, save_config
 from archinstall.lib.disk.disk_menu import DiskLayoutConfigurationMenu
 from archinstall.lib.general.general_menu import select_hostname, select_ntp, select_timezone
 from archinstall.lib.general.system_menu import select_kernel, select_swap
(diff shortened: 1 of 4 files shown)


archinstall/lib/global_menu.py
```diff
@@ -498,6 +498,70 @@ def _validate_bootloader(self) -> str | None:
 
 		return None
 
+	def _install_summary(self) -> str:
```
svartkanin: Can we move this into the `configuration.py` instead as it aligns with the other configuration outputs of JSON.

archinstall/lib/global_menu.py
```diff
@@ -508,7 +572,10 @@ def _prev_install_invalid_config(self, item: MenuItem) -> str | None:
 		if error := self._validate_bootloader():
 			return tr(f'Invalid configuration: {error}')
 
-		return None
+		summary = self._install_summary()
+		if summary:
+			return f'{tr("Ready to install")}\n\n{summary}'
```
svartkanin, line 514: Would it make more sense to just display the `Ready to install` and then display the detailed summary on the confirmation screen? 
I was planning to rework that JSON output there to make it prettier as it's not very user friendly right now, but that can be done as this change

Softer:
First comment: done. ConfigurationOutput.as_summary() lives next to the JSON methods now, preview delegates to it. 

On the second one I'd argue the opposite. The Install-button preview is where the user decides; the confirm screen is just a safety net. Putting the summary on the preview is exactly what lets the decision be informed - showing it again on confirm doesn't help anyone who didn't already read the preview. So I'd rather keep the summary on preview and strip confirm_config down to a plain yes/no (drop the JSON dump entirely). Then your "rework the JSON to be prettier" task disappears - that screen stops being an info surface.
 
Where the preview can grow is colour: red for blockers (missing config, invalid bootloader), yellow for non-blocking warnings (no network from #4408, future firmware case from #4466), green for ready-to-install. One go / wait / fix signal at the moment of decision.
 
Happy to do colour + drop the JSON from confirm in a follow-up PR after this merges. :)

svartkanin (approved):

svartkanin:
@Softer thanks makes sense. 
I'm happy to get this merged for now as is and we can work on the confirmation screen later on. I don't want to remove the summary/full details from the confirmation screen as it's a nice visualization of what has been chosen and what the user is actually confirming. 

I know I suggested to move this into the configuration.py so I think this is good to merge for now. But having had another look at it and the https://github.com/archlinux/archinstall/blob/bf9b9cb7c17c87595676a52210803a2ebb02af48/archinstall/lib/args.py?plain=1#L99 the config gets assembled in a JSON in there whereas now it's distributed across multiple files.

PR #4479, opened 2026-04-25
Include core packages in all Plasma settings

georgemoodier:
## PR Description:

- adds additional core packages, seen as commonalities in a plasma installation (à la #4460)
- the `plasma` setting includes the full range of core packages previously seen in archinstall's Plasma profile, while `plasma-meta` only includes `konsole`and `dolphin` and `plasma-desktop` only includes `konsole`. This is to keep up the overall profile's granularity (since having a GUI terminal is the bottom line when looking at other desktop profiles)

## Tests and Checks

### DIFF
--- a/archinstall/default_profiles/desktops/plasma.py
+++ b/archinstall/default_profiles/desktops/plasma.py
@@ -57,11 +57,11 @@ class PlasmaFlavor(StrEnum):
 	def packages(self) -> list[str]:
 		match self:
 			case PlasmaFlavor.Meta:
-				return ['plasma-meta']
+				return ['plasma-meta', 'konsole', 'dolphin', 'kate', 'ark']
 			case PlasmaFlavor.Plasma:
-				return ['plasma']
+				return ['plasma', 'konsole', 'dolphin', 'kate', 'ark']
 			case PlasmaFlavor.Desktop:
-				return ['plasma-desktop']
+				return ['plasma-desktop', 'konsole', 'dolphin']
 
 
 class PlasmaProfile(Profile):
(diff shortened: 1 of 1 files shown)


Softer:
In my humble opinion, plasma-desktop should stay minimal - that is its whole point as a flavor. The differentiation is already encoded in package_details(): Meta is "Curated", Plasma is "Extensive", Desktop is "Minimal". Adding konsole and dolphin to Desktop conflicts with that label - if a user picks the minimal option, they explicitly opt out of GUI extras, and a file manager is an extra. Anyone who wants more can use the "Additional packages" step.
 
A more graduated approach that respects the existing semantics:
- plasma-desktop (Minimal): keep as-is, no extras
- plasma (Extensive): add konsole + dolphin - terminal and file manager as the bare working baseline 
- plasma-meta (Curated): add all four - konsole, dolphin, kate, ark 
 
This way each flavor becomes a meaningful step up rather than three slight variations on the same set.

georgemoodier:
I agree that including `dolphin` for `plasma-desktop` may be pushing it, but i think it should at least include `konsole`. As I've stated in the description, every other desktop profile (including ones for standalone WMs) has an included terminal emulator, so that users aren't solely reliant on a TTY. I don't see any reason as to why `plasma-desktop` should be an exception in regards to this.
As for the other two, I believe flipping the package list would be better, as the group (by default) installs more packages compared to `plasma-meta` (hence the "Extensive" label, i suppose), whereas those packages are deemed optional by the latter.

Torxed:
See my comment [here](https://github.com/archlinux/archinstall/issues/4480#issuecomment-4320555349) why I'm closing this. It's nothing personal.

PR #4485, opened 2026-04-27
Fix sway+nvidia confirmation dialog (#4481)

### LINKED ISSUE #4481: Incorrect menu behavior sway x nvidia

April 1st 2026

### The installation log

```json
Nothing of note, everything is smooth
Archinstall 4.3 latest
```

### describe the problem

When trying to install sway + nvidia proprietary drivers (open version since fully proprietary is no longer available) you get a menu saying "sway does not support nvidia proprietary drivers are you ok with that ? "
Option defaults to "no" but if you want that choice you are supposed to answer "yes". 
The thing is that it is the "no" option that actually allows you to select the proprietary drivers and not the "yes".

You also get the warning if you select the other nvidia driver "nouveau" which is fully open source which is not intended behavior as it is actually recommended by the sway team

Softer:
Closes #4481.
 
The Sway+Nvidia confirmation dialog had two bugs:
 
1. Inverted boolean. When the user confirmed "yes, I'm okay with issues", the code reverted the driver to the previous selection instead of keeping the chosen Nvidia driver. The user got the opposite of what they confirmed.
 
2. False-positive warning for nouveau. The warning triggered for any Nvidia driver, including NvidiaOpenSource (nouveau), which is the open-source Mesa driver officially supported by Sway. Users selecting nouveau got a misleading warning that they were about to break their Sway setup.
 
This PR adds two methods on GfxDriver: 
- is_nvidia_proprietary() — true only for NvidiaOpenKernel (nvidia-open-dkms, proprietary userspace).
- is_nvidia_nouveau() — true only for NvidiaOpenSource (nouveau, fully open). 

The Sway warning now uses is_nvidia_proprietary() instead of is_nvidia(), so nouveau users no longer see it. The boolean check is also flipped so confirming "yes" actually keeps the driver.
 
is_nvidia_nouveau() is added for symmetry and future use (e.g. when other Wayland compositors get similar checks).

### DIFF
--- a/archinstall/lib/profile/profile_menu.py
+++ b/archinstall/lib/profile/profile_menu.py
@@ -105,8 +105,7 @@ class ProfileMenu(AbstractSubMenu[ProfileConfiguration]):
 						preset=False,
 					).show()
 
-					if not result.get_value():
-						return preset
+					return driver if result.get_value() else preset
 
 		return driver
 
--- a/archinstall/lib/profile/profile_menu.py
+++ b/archinstall/lib/profile/profile_menu.py
@@ -95,7 +95,7 @@ class ProfileMenu(AbstractSubMenu[ProfileConfiguration]):
 				driver = await select_driver(preset=preset)
 
 			if driver and 'Sway' in profile.current_selection_names():
-				if driver.is_nvidia():
+				if driver.is_nvidia_proprietary():
 					header = tr('The proprietary Nvidia driver is not supported by Sway.') + '\n'
 					header += tr('It is likely that you will run into issues, are you okay with that?') + '\n'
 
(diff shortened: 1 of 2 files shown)


svartkanin:
But `is_nvidia_nouveau` is never used right now? Shall we remove that until we actually need it

archinstall/lib/profile/profile_menu.py
```diff
@@ -105,7 +105,7 @@ async def _select_gfx_driver(self, preset: GfxDriver | None = None) -> GfxDriver
 						preset=False,
 					).show()
 
-					if result.get_value():
+					if not result.get_value():
```
svartkanin: `allow_skip` is set to `false` so the user has to make a choice. You can just 
```
return result.get_value()
```

Softer: Collapsed the Sway+Nvidia confirmation handling into return driver if result.get_value() else preset since allow_skip=False guarantees a boolean. If you had a different shape in mind for the result.get_value() simplification, happy to iterate.

svartkanin (approved):

PR #4489, opened 2026-04-28
Add kernel enum

codefiles:
- Add `DEFAULT_KERNEL` constant so it is in one location.
- Small refinement to `preset` parameter of `MenuItemGroup.from_enum()`.

### DIFF
--- a/archinstall/tui/ui/menu_item.py
+++ b/archinstall/tui/ui/menu_item.py
@@ -1,4 +1,4 @@
-from collections.abc import Awaitable, Callable
+from collections.abc import Awaitable, Callable, Iterable
 from dataclasses import dataclass, field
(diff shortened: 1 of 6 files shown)


archinstall/tui/ui/menu_item.py
```diff
@@ -148,7 +148,7 @@ def from_enum(
 		cls,
 		enum_cls: type[Enum],
 		sort_items: bool = False,
-		preset: Enum | None = None,
+		preset: Iterable[Enum] | Enum | None = None,
```
codefiles, line 151: `preset` can be an iterable like it is in this case with `list[Kernel]`.

svartkanin (approved):

PR #4492, opened 2026-04-28
Replace time.time with time.monotonic to avoid clock drift issues

correctmost:
## PR Description:

Using time.time() can be inaccurate if the system clock gets updated in between calls.

## Tests and Checks
- I tested the ping and pacman-lock paths and didn't see any issues


Torxed (approved):
This one really tickles my mind for some reason, nice improvement!

PR #4494, opened 2026-04-28
Add basic Pyrefly configuration to pyproject.toml

correctmost:
## PR Description:

This makes it easier to use the Pyrefly type checker with various IDEs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -172,6 +172,23 @@ score = false
 [tool.pylint.variables]
 init-import = true
 
+[tool.pyrefly]
+python-version = "3.14"
+
+replace-imports-with-any = [
+    "parted",  # pyparted doesn't have type hints
+]
+
+[tool.pyrefly.errors]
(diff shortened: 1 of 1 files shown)


svartkanin:
I'm not against different type checkers, but I'd be good to settle on one. Currently we're using `mypy` heavily, but there  is `ty` as well (not sure how mature it is at this point). Would it make sense to evaluate them properly on the code base and stick to one?

correctmost:
I think it's best to stick with mypy for now:
- mypy - stable and actively maintained
- Pyright - stable, but less actively maintained
- Pyrefly, ty, and zuban - still in beta

The newer type checkers have language server features, so I would expect more developers to start using them while working on archinstall.  I use them with VS Code and thought it would be helpful to add configs that reduce warning noise.

I think it's okay to commit basic configs for tools like Pyright and Pyrefly to support local development, even if the tools don't end up running on CI.

svartkanin (approved):

PR #4495, opened 2026-04-28
Offer to apply archinstall language to target system locale and timezone

Softer:
## Summary
 
After picking an archinstall language that maps to a known system locale, prompt the user to apply that locale (and matching console font / timezone, where unambiguous) to the target system. Confirms with a single Yes/No dialog listing every field that will change. 

Single confirmation dialog. Lists only the fields that would actually change; if all three already match the target, the prompt is skipped silently.

Builds on #4469 (console font in Locales menu): that PR added the `console_font` mapping and a manual selector; this PR wires the same data into the language picker so a user who picks Ukrainian once gets the locale, font, and timezone offered together rather than having to find each menu separately. 

## What changes
 
- **Locale prompt.** When the chosen language has a `sys_lang` mapping (e.g. `uk` -> `uk_UA.UTF-8`), offer to set `sys_lang` / `sys_enc` on `LocaleConfiguration`.
- **Timezone suggestion.** New `default_timezone` field on `Language`, populated for ~30 unambiguous single-zone country mappings (`uk_UA` -> `Europe/Kyiv`, `de_DE` -> `Europe/Berlin`, `ja_JP` -> `Asia/Tokyo`, ...). Multi-zone countries (`en_US`, `es_ES`, `pt_BR`, ...) are intentionally left without a default and fall through unchanged. 
- **Reset on re-pick.** Each language-derived field (`sys_lang`, `console_font`, `timezone`) is offered whenever its target value differs from the current setting, so picking English after Ukrainian resets `console_font` to `default8x16` and `timezone` to `UTC` instead of leaving stale Ukrainian values behind.
- **Validation.** Candidate timezones are validated against `list_timezones()` - the same source the timezone menu reads - so we never offer a value the user could not have selected manually.
- **Yes/No label refresh.** `MenuItem.yes()` / `.no()` cache their translated text on the class. With the new prompt firing right after a language switch, the cached labels stayed in the previous language. Added `MenuItem.invalidate_yes_no()` and a call from `TranslationHandler.activate()` to drop the singletons on language change.
 
## Test plan 
 
- [x] Pick Ukrainian -> prompt offers `uk_UA.UTF-8`, `Lat2-Terminus16`, `Europe/Kyiv`. Accept -> all three applied.
- [x] Pick Ukrainian, accept; then re-open language menu and pick English -> prompt offers reset to `en_US.UTF-8`, `default8x16`, `UTC`. 
- [x] Pick a multi-zone language (Spanish) -> prompt offers locale only, timezone row is absent.
- [x] Pick a language with no `sys_lang` mapping -> prompt does not appear.
- [x] Decline the prompt -> no fields change.
- [x] Yes/No buttons re-translate after language switch (verified the cache invalidation visually).


Softer:
Up front: I understand this is a big PR, and Python isn't my strongest language yet. Any feedback is very welcome :)

svartkanin:
This adds a lot of complexity and maintainance for some convenience for a user, which I'm not sure is worth it. Having to set these manually in the menu is not that big a deal and with this the static JSON list needs to be maintained. 

In addition, this can be set once, then the config can be safed and loaded on repeadet installations. I'd assume a user will use the same settings again so that is already done via the config save flow.

Softer:
The JSON already exists and is already maintained for sys_lang and console_font (#4356, #4469). This PR adds default_timezone only for locales whose country has a single mainland IANA zone (ar_SA, ku_TR, ur_PK, etc.). Locales with regional ambiguity (ca_ES, gl_ES) intentionally have no timezone, so they fall through to UTC. The mapping is essentially write-once.

On save/load covering the same ground: save/load has its own friction - save before reboot, pull the file out of live, push it back into the next live ISO. Worth the setup if you reinstall weekly; not really if you reinstall once a year. And it only helps from the second install onward. PR #4356 (auto console font from language) is the same idea: sane defaults from the language choice, instead of asking the user to repeat it.

On user choice: the apply step is opt-in via a confirmation dialog that shows the full diff (locale, encoding, font, timezone). Default is no-op; nothing is overridden silently.

svartkanin:
I need to dive a bit deeper into this and see what the implications are as well as testing it properly

Softer:
Rebased on master to resolve conflicts

Softer:
> I need to dive a bit deeper into this and see what the implications are as well as testing it properly

Haven't watched it yet?...

PR #4497, opened 2026-04-28
Remove the old curses-based TUI code

correctmost:
## PR Description:

The new Textual-based TUI shipped in the April .iso.

## Tests and Checks
- [x] I have tested the code!


svartkanin:
As a follow up we should also bring the new menu one level up

svartkanin (approved):

PR #4510, opened 2026-05-01
Fix truncated package metadata in additional packages preview (#3580)

### LINKED ISSUE #3580: Additional packages metadata is truncated
## Steps to reproduce
- Select "Additional packages"
- Select the sqlite-doc package
  - Notice that the Description field is truncated: "most of the static HTML files that comprise this website,"

## Analysis

The additional-packages code runs `pacman -S --info` using pty.  On my system, pty seems to default to 80 columns:

```
(Pdb) SysCommand('tput cols')
80
```

This causes the description field to be truncated because archinstall's parsing code is line-based and because pacman [formats its output differently](https://gitlab.archlinux.org/pacman/pacman/-/blob/f13d7d480c7c4284a7394c111a6aef6aaae5438a/src/pacman/util.c#L1001-1010) depending on the number of columns.

We can see the truncation in action by running with different `COLUMNS` values:

```
(Pdb) SysCommand('pacman -S --info sqlite-doc')
Repository      : core-testing
Name            : sqlite-doc
Version         : 3.50.1-1
Description     : most of the static HTML files that comprise this website,
                  including all of the SQL Syntax and the C/C++ interface specs
                  and other miscellaneous documentation
[...snip...]

(Pdb) SysCommand('pacman -S --info sqlite-doc', environment_vars={'COLUMNS': '120'})
Repository      : core-testing
Name            : sqlite-doc
Version         : 3.50.1-1
Description     : most of the static HTML files that comprise this website, including all of the SQL Syntax and the
                  C/C++ interface specs and other miscellaneous documentation
[...snip...]

(Pdb) SysCommand('pacman -S --info sqlite-doc', environment_vars={'COLUMNS': '500'})
Repository      : core-testing
Name            : sqlite-doc
Version         : 3.50.1-1
Description     : most of the static HTML files that comprise this website, including all of the SQL Syntax and the C/C++ interface specs and other miscellaneous documentation
[...snip...]
```

This particular bug seems cosmetic, but I wanted to at least document it because it could affect other areas that rely on the parsing of `SysCommand` output.

In the long run, [using libalpm](https://github.com/archlinux/archinstall/pull/2787) would likely increase the correctness and performance of pacman interactions.

## Version info
d0f1fdf4b40e

Softer:
Fixes #3580.

## Summary

`pacman -S --info` wraps long fields like `Description` across multiple lines when the pty width is small (default 80 cols). archinstall's package info parser pre-stripped each output line before joining continuations, which discarded the leading whitespace it relied on to detect continuation lines, so the wrapped portions were silently dropped.

Example before fix, for `sqlite-doc`:
```
Description : most of the static HTML files that comprise this website,
```

After fix:
```
Description : most of the static HTML files that comprise this website,
including all of the SQL Syntax and the C/C++ interface specs
and other miscellaneous documentation
```

## What changed
- `Pacman.run()` gained an optional `environment_vars` parameter that is forwarded to `SysCommand`.                                                                                                              
- The two `pacman -S --info` call sites in `archinstall/lib/packages/packages.py` now pass `{'COLUMNS': '500'}`, forcing pacman to emit single-line field values.
- `AvailablePackage.info()` now word-wraps each value to roughly half the terminal width using `textwrap.fill`, so the preview pane does not get a horizontal scrollbar on long fields. The continuation lines are aligned to the value column to match pacman's own visual style.

Scope is limited to `-S --info` calls; sync/install paths are unaffected.


h8d13:
I don't think you need the env var plumbing if you fix parser call sites

svartkanin:
I don't think truncated is the right description for this as it's not truncated but horizontally scrollable. 

This can also be solved much simpler with CSS to do text wrap, if you add this 

```
	#preview_content {
		width: 100%;
		height: auto;
	}
```

to the `SelectListScreen` class it will wrap it nicely. 
I think this nees to be added conditionally though as we don't want this to be applied all the time, for example displayed tables in the preview would potentially be malformed with this.

Softer:
It makes sense... Let's try :)

svartkanin:
The snipplet I provided was an example and as mentioned it needs to be conditionally applied as we don't want to have that for every preview as that will malform tables. Have a look at this example code https://github.com/Softer/archinstall/blob/5136b3190d9d8c6d137fc060dbc774e0263a0839/archinstall/tui/ui/components.py#L881

Softer:
Addressed: CSS wrap is now conditional via wrap_preview parameter. Only the packages menu enables it - other previews with tables are unaffected.

archinstall/tui/components.py
```diff
@@ -280,7 +282,10 @@ def compose(self) -> ComposeResult:
 				with Container():
 					yield option_list
 					yield Rule(orientation=rule_orientation)
-					yield ScrollableContainer(Label('', id='preview_content', markup=False))
+					preview_label = Label('', id='preview_content', markup=False)
```
svartkanin, line 290: Did you test this? 
You have only defined the `wrap-preview` CSS for `SelectListScreen`

Softer: Sure. It works. But I think I understand what you mean...

svartkanin (approved):

Issue #833, opened 2022-01-02
Plugins module not imported


ilario:
Testing the master code I just got an error message due to `plugins` not being available here (I suppose it needs to be imported inside the very function?) as implemented in #821:
https://github.com/archlinux/archinstall/blob/240f688ccefa7bddb56efe5bc41a457c96792e42/archinstall/__init__.py#L178

Torxed:
Yepp, some changes have been made to `__init__.py` over the last day or so.
This line should be moved outside of the locale scope function: https://github.com/archlinux/archinstall/blob/240f688ccefa7bddb56efe5bc41a457c96792e42/archinstall/__init__.py#L160

Torxed:
Moved the import, and fixed in b533bd0 :) Thank you for reporting this! Apologies that it took some time. Work has been hectic.

ilario:
Thank you for your great work!!!

PR #4511, opened 2026-05-01
Add share-log subcommand to upload install.log to paste.rs

Softer:
Replaces the broken curl ... https://0x0.st hint with an integrated archinstall share-log command that uploads /var/log/archinstall/install.log to paste.rs and prints a shareable URL. Also updates the same hint in the README, the bug report docs, and the GitHub bug issue template, so all four user-facing references now point to the working command.

Refs #4396 (0x0.st has disabled uploads). Supersedes #4507, which removes the hint without offering a replacement. 

Why paste.rs

- Text-oriented pastebin with extension-based syntax highlighting, so install.log renders as a readable page rather than a raw download
- 10 MiB upload limit (well above realistic install.log sizes)
- Curl-friendly, no account, no server-side state to maintain
 
The endpoint is referenced exactly once - a constant in archinstall/lib/output.py. If a self-hosted paste service is set up later, or 0x0.st returns, swapping it is a one-line change.
 
Behavior

$ archinstall share-log
About to upload /var/log/archinstall/install.log (12345 bytes) to https://paste.rs
The log may contain hostname, mirror URLs, package list and partition layout.
The uploaded paste is public. 
Continue?
https://paste.rs/abc.def
 
- Runs as a subcommand (archinstall share-log), not a flag - it is a standalone command, not an installer option
- Runs before root check so non-root users can share logs too
- TUI confirmation via ConfirmationScreen (consistent with the rest of archinstall)
- Diagnostics go through info(); only the final URL goes to raw print() so it can be piped (archinstall share-log | xclip -sel clip)
- Logs exceeding 10 MiB are truncated from the bottom up (latest entries kept) instead of refusing to upload
- File-not-found, empty file, network failure and unexpected response are each handled with a clear message and non-zero exit
 
Test plan

- [x] archinstall share-log on a live ISO - URL is printed and the paste is readable 
- [x] archinstall share-log answering No - exits cleanly without uploading
- [x] archinstall share-log with install.log missing - reports the error and exits non-zero
- [x] archinstall share-log with no network - reports the error and exits non-zero 
- [x] Trigger an install error and confirm the hint says archinstall share-log


Torxed:
Just note that I started a conversation about an arch official channel for posting these logs. I created a ticket to track the progress on [gitlab@archlinux #833](https://gitlab.archlinux.org/archlinux/infrastructure/-/work_items/833)

Softer:
We can change --share-log behavior transparently for users any time we need without changing user's expectation. :)

archinstall/lib/args.py
```diff
@@ -431,6 +432,12 @@ def _define_arguments(self) -> ArgumentParser:
 			default=False,
 			help='Enabled verbose options',
 		)
+		parser.add_argument(
```
svartkanin: this sholdn't be an argument as an argument implies that you're running `archinstall` with that flag, but what you're doing here is running a command associated with `archinstall`. So it would make more sense to have it as 

```
archinstall share-log
```

instead

archinstall/lib/share_log.py
```diff
@@ -0,0 +1,71 @@
+import sys
```
svartkanin: I think we can move this into the `output.py` file as all log logic is consolidated there

archinstall/lib/share_log.py
```diff
@@ -0,0 +1,71 @@
...
+	if size == 0:
+		print(f'Log file is empty: {log_path}', file=sys.stderr)
+		return 1
+
+	if size > _PASTE_MAX_SIZE:
+		print(
+			f'Log file is too large to share: {size} bytes (limit: {_PASTE_MAX_SIZE} bytes). Trim it or upload manually.',
+			file=sys.stderr,
+		)
+		return 1
```
svartkanin: Instead of returning we should truncate it from bottom up so we get the latest installation attempts. I don't envision many users running into this but it's better to get some logs than none at all

archinstall/lib/share_log.py
```diff
@@ -0,0 +1,71 @@
...
+
+	print(f'About to upload {log_path} ({size} bytes) to {_PASTE_URL}', file=sys.stderr)
+	print(
+		'The log may contain hostname, mirror URLs, package list and partition layout. The uploaded paste is public.',
+		file=sys.stderr,
+	)
+
+	try:
+		answer = input('Continue? [y/N]: ').strip().lower()
```
svartkanin: We should be using the TUI here not the raw input stuff, have a look at the `WifiHandler` as that is also a standalone menu triggered from `main.py`

archinstall/lib/share_log.py
```diff
@@ -0,0 +1,71 @@
...
+	mirror URLs, package list, partition layout and other system details which
+	become public on upload.
+
+	All diagnostic output goes to stderr instead of the standard log helpers,
+	so the file we are about to upload is not modified by this command.
+	"""
+	log_path = logger.path
+
+	if not log_path.exists():
+		print(f'Log file not found: {log_path}', file=sys.stderr)
```
svartkanin: Use `info` instead as that will also appear in the logs and is testable (I promise one day we'll have tests)

Softer:
Changes per review:

- archinstall share-log subcommand instead of --share-log flag
- Code moved into output.py, share_log.py deleted
- Large logs truncated from bottom up instead of refusing upload
- TUI ConfirmationScreen instead of input()
- info() for diagnostics; final URL stays as print() for pipe-friendliness

archinstall/lib/output.py
```diff
@@ -333,3 +333,93 @@ def log(
...
 	if level != logging.DEBUG:
 		print(text)
+
+
+_PASTE_URL = 'https://paste.rs'
+_PASTE_MAX_SIZE = 10 * 1024 * 1024
+
+
+def share_install_log() -> int:
+	from archinstall.lib.command import SysCommand
```
svartkanin: Why are they not global?

archinstall/lib/output.py
```diff
@@ -333,3 +333,93 @@ def log(
...
+	else:
+		content = log_path.read_bytes()
+
+	header = f'About to upload {log_path} ({len(content)} bytes) to {_PASTE_URL}\n\n'
+	header += 'The log may contain hostname, mirror URLs, package list and partition layout.\n'
+	header += 'The uploaded paste is public.\n\n'
+	header += 'Continue?'
+
+	try:
+		from archinstall.tui.ui.components import tui
```
svartkanin: Why is this not global?

archinstall/lib/output.py
```diff
@@ -333,3 +333,93 @@ def log(
...
+
+		confirmed: bool = tui.run(lambda: _confirm_share(header))
+	except Exception:
```
svartkanin: why do we need a try here?

archinstall/lib/output.py
```diff
@@ -333,3 +333,93 @@ def log(
...
+		confirmed = False
+
+	if not confirmed:
+		info('Cancelled.')
+		return 1
+
+	import tempfile
```
svartkanin: Why not global?

archinstall/lib/output.py
```diff
@@ -333,3 +333,93 @@ def log(
...
+
+	if size > _PASTE_MAX_SIZE:
+		fd, tmp_path_str = tempfile.mkstemp(suffix='.log')
```
svartkanin: we can use pathlib here as it's the main pattern in the code base

```
       if size > _PASTE_MAX_SIZE:
            with tempfile.TemporaryDirectory() as tmp_dir:
                upload_path = Path(tmp_dir) / 'archinstall_log_upload.log'
                upload_path.write_bytes(content)
                ...
        else:
            ...
```

In addition lets not call `curl` from syscommand but actually use Python build ins 

```
    try:
        req = urllib.request.Request(_PASTE_URL, data=content)
        
        with urllib.request.urlopen(req) as response:
            url = response.read().decode().strip()
            
    except urllib.error.URLError as e:
        info(f'Upload failed: {e}')
        return 1

```

archinstall/lib/output.py
```diff
@@ -333,3 +333,93 @@ def log(
...
+	if not url.startswith('http'):
+		info(f'Unexpected response from {_PASTE_URL}: {url[:200]!r}')
+		return 1
+
+	# raw print so the URL is pipe-friendly (no ANSI colors, no log prefix)
+	print(url)
+	return 0
+
+
+async def _confirm_share(header: str) -> bool:
```
svartkanin: The TUI code should not be part of this file as `archinstall` can be used as a lib it should only expose the share function to consumers, but the consumers should be able to call it and use it independently from the TUI logic. 

You can move the confirmation out to the main.py where the function is called, and after confirming it will just call the share function which will then run without interruption

Softer: I see the point about lib purity, but share_install_log is specifically a CLI subcommand helper - it's invoked from main.py as archinstall share-log and there's no realistic scenario where someone calls it programmatically as a library function. Moving the confirmation out would split a 20-line function into two halves across two files for a separation that nobody will consume. Happy to do it if you feel strongly.

archinstall/lib/output.py
Softer: Circular import: command.py imports debug, error, logger from output.py, so a top-level from archinstall.lib.command import SysCommand here creates a cycle. Removed in the latest commit anyway - switched to urllib.request so these imports are no longer needed.

archinstall/lib/output.py
Softer: Circular import: tui.ui.components imports debug from output.py, so a top-level import here creates a cycle. Has to stay local.

archinstall/lib/output.py
svartkanin: It's about dependency management and footprint, this makes output.py dependent on the tui module which is not necessary. 
you create a new public function for sharing the log file, so anyone using `archinstall` as a lib or otherwise would be able to use it, regardless if they need to or not. If anyone wants to create their own ISOs and use `archinstall` they are dependent on this hardcoded implementation of using the tui. 
Also it makes isolated functionality testing very difficult as now a test needs to mock a full blown tui instead of being able to test the actual sharing logic in an isolated way. 

In addtion, these 

```
_PASTE_URL = 'https://paste.rs'
_PASTE_MAX_SIZE = 10 * 1024 * 1024
```

should be parameters into the function as well so the function can be tested easily by providing a simple mock url.

Softer: You know better. Decoupled - share_install_log now takes paste_url, max_size, and confirm as parameters, zero TUI imports in output.py. The TUI confirmation lives in main.py as _tui_confirm and is passed via callback.

Softer:
Since share_install_log is now a standalone function in output.py with no TUI dependencies, I figured it would be a good candidate for unit tests. :-)

Added tests/test_share_log.py covering all code paths: 
* file not found
* empty file
* user cancellation
* successful upload
* large file truncation
* network error
* unexpected server response

README.md
```diff
@@ -101,9 +101,9 @@ If you come across any issues, kindly submit your issue here on GitHub or post y
 When submitting an issue, please:
 * Provide the stacktrace of the output if applicable
 * Attach the `/var/log/archinstall/install.log` to the issue ticket. This helps us help you!
-  * To extract the log from the ISO image, one way is to use<br>
+  * To upload the log from the ISO image and get a shareable URL, run<br>
     ```shell
-    curl -F'file=@/var/log/archinstall/install.log' https://0x0.st
+    archinstall --share-log
```
svartkanin: update to 
```
archinstall share-log
```

tests/test_share_log.py
```diff
@@ -0,0 +1,94 @@
+# pylint: disable=redefined-outer-name
```
svartkanin, line 1: Thank you for adding tests!

docs/help/report_bug.rst
```diff
@@ -15,7 +15,7 @@ When submitting a help ticket, please include the :code:`/var/log/archinstall/in
 It can be found both on the live ISO but also in the installed filesystem if the base packages were strapped in.
 
 .. tip::
-   | An easy way to submit logs is ``curl -F 'file=@/var/log/archinstall/install.log' https://0x0.st``.
+   | An easy way to submit logs is ``archinstall --share-log``, which uploads ``install.log`` to paste.rs and prints a shareable URL.
```
svartkanin: update to 
```
archinstall share-log
```

.github/ISSUE_TEMPLATE/01_bug.yml
```diff
@@ -41,8 +41,8 @@ body:
     attributes:
       value: >
         **Note**: Assuming you have network connectivity,
-        you can easily post the installation log using the following command:
-        `curl -F'file=@/var/log/archinstall/install.log' https://0x0.st`
+        you can easily upload the installation log and get a shareable URL by running:
+        `archinstall --share-log`
```
svartkanin: same here

svartkanin (approved):

PR #4512, opened 2026-05-01
Color-code install preview: red for errors, yellow for warnings, green for ready

Softer:
### Summary
* Add preview_markup opt-in field to MenuItem - when True, the preview panel renders Rich markup; when False (default), text is auto-escaped so existing previews are unaffected 
* Move network warning from the confirmation dialog to the install preview, where the user sees it earlier - while still configuring, not after pressing Install 
* All other preview panels (profile, disk, locale, etc.) are unchanged - preview_markup=False by default escapes any [...] sequences in their output 

### Color codes
- **Red**: missing mandatory configs, bootloader validation errors (blocking)
- **Yellow**: warnings like "no network configuration selected" (non-blocking) 
- **Green**: "Ready to install"
 
### Examples
image
image
image
image

### Test plan

- [x] Open archinstall, leave disk config empty - Install preview should show red "Missing configurations" list
- [x] Configure disk but skip network - Install preview should show green "Ready to install" + yellow "Warnings" with network message
- [x] Leave both disk and network unconfigured - should show red missing configs + yellow network warning together 
- [x] Configure everything including network - green "Ready to install" + summary, no yellow 
- [x] Check that other menu items (profile, locale, mirrors) still show plain-text previews without markup artifacts 
- [x] Press Install, confirm dialog should no longer contain the network warning line


Softer:
Converted into draft because looks like I found an error. Will check a bit later...

Softer:
Done, now it works correctly :-)

archinstall/tui/ui/menu_item.py
```diff
@@ -19,6 +19,7 @@ class MenuItem:
 	dependencies_not: list[str] = field(default_factory=list)
 	display_action: Callable[[Any], str] | None = None
 	preview_action: Callable[[Self], str | None] | None = None
+	preview_markup: bool = False
```
svartkanin: This should be set on the actual menu not on the item alongside 
```
preview_location: Literal['bottom'] | None = None,
preview_markup: bool = False,
```

archinstall/lib/global_menu.py
```diff
@@ -495,20 +498,41 @@ def _validate_bootloader(self) -> str | None:
...
-			return text[:-1]  # remove last new line
+		self.sync_all_to_config()
+		config_output = ConfigurationOutput(self._arch_config)
 
-		if error := self._validate_bootloader():
-			return tr(f'Invalid configuration: {error}')
+		warnings = config_output.get_install_warnings()
+		warnings_text = ''
+		if warnings:
+			warnings_text = f'\n\n[yellow]{_escape_markup(tr("Warnings:"))}\n'
```
svartkanin: Lets reuse the pattern we have with the Input validation 

```
class InputInfoType(Enum):
	MsgInfo = auto()
	MsgWarning = auto()
	MsgError = auto()


@dataclass
class InputInfo:
	message: str
	info_type: InputInfoType
```

which means the `preview_action` return value should be changed from a `str` to a dataclass holding different attributes, for now only the type. We can probably rename the `InputInfoType` to `MsgLevelType` and reuse that. 

Then the `MenuItem` becomes something like 

```
@dataclass
class PreviewActionResult:
	message: str
	msg_level_type: MsgLevelType


preview_action: Callable[[Self], PreviewActionResult | None] | None = None
```

Softer:
Renamed InputInfoType to MsgLevelType, added PreviewResult dataclass, removed preview_markup from MenuItem. The widget now maps level to color instead of business logic producing Rich markup. preview_action accepts str | PreviewResult | list[PreviewResult] | None so sections with different levels can be mixed in one preview.

archinstall/lib/configuration.py
svartkanin: This should be moved into the global_menu.py now as there's no purpose for this here anymore

archinstall/lib/global_menu.py
```diff
@@ -494,21 +494,40 @@ def _validate_bootloader(self) -> str | None:
...
 		if missing := self._missing_configs():
-			text = tr('Missing configurations:\n')
-			for m in missing:
-				text += f'- {m}\n'
-			return text[:-1]  # remove last new line
+			errors += f'{tr("Missing configurations:")}\n'
+			errors += '\n'.join(f'- {m}' for m in missing)
+
+		disk_item = self._item_group.find_by_key('disk_config')
+		if disk_item.has_value():
```
svartkanin, line 518: This is not the same logic as the existing code, if there is a bug it'd be good to make these changes incremental to understand why they have changed

archinstall/tui/menu_item.py
```diff
@@ -18,7 +33,7 @@ class MenuItem:
 	dependencies: list[str | Callable[[], bool]] = field(default_factory=list)
 	dependencies_not: list[str] = field(default_factory=list)
 	display_action: Callable[[Any], str] | None = None
-	preview_action: Callable[[Self], str | None] | None = None
+	preview_action: Callable[[Self], str | PreviewResult | list[PreviewResult] | None] | None = None
```
svartkanin: This should be changed to only support the new wrapper type via 
```
preview_action: Callable[[Self], PreviewResult | None] | None = None
```

The `PreviewResult` should then handle the different chunks or whatever a preview may return. This will have to be updated for all previews, so I recommend to do this in an incremental way with multiple PRs. 

To make life easier we can have something like 

```
@dataclass
class PreviewResult:
	messages: list[tuple[str, MsgLevelType]]
```

archinstall/tui/components.py
```diff
@@ -21,12 +21,39 @@
...
 from archinstall.lib.output import debug
 from archinstall.lib.translationhandler import tr
-from archinstall.tui.ui.menu_item import MenuItem, MenuItemGroup
+from archinstall.tui.ui.menu_item import MenuItem, MenuItemGroup, MsgLevelType, PreviewResult
 from archinstall.tui.ui.result import Result, ResultType
 
 ValueT = TypeVar('ValueT')
 
 
+_LEVEL_STYLE: dict[MsgLevelType, str] = {
```
svartkanin: Please don't create these global variables. This is a component of the type 

```
class MsgLevelType(Enum):
	MsgNone = auto()
	MsgInfo = auto()
	MsgWarning = auto()
	MsgError = auto()

	def style(self) -> str:
		match self:
			case MsgLevelType.MsgInfo:
				return 'green'
            case MsgLevelType.MsgNone:
				return 'white'
			case MsgLevelType.MsgWarning:
				return 'yellow'
			case MsgLevelType.MsgError:
				return 'red'
```

svartkanin:
@Softer when you addressed the comments I will have another review

Softer:
@svartkanin Both comments addressed in c5d5135.

Also merged master to resolve the conflict status.

archinstall/lib/configuration.py
svartkanin: @Softer this is still not addressed

svartkanin:
@Softer my previous comments have not all been addressed

Softer:
> @Softer my previous comments have not all been addressed

@svartkanin Could you clarify which comment you feel wasn't addressed?

svartkanin:
@Softer apologies, I just realized that I never sent the comments and they all remained in the pending state.

Softer:
@svartkanin Addressed all comments:

1. configuration.py - Moved get_install_warnings() to GlobalMenu._get_install_warnings() since it's the only caller (9c40f8e)
2. components.py - Replaced _LEVEL_STYLE dict with MsgLevelType.style() method on the enum (139bd10)
3. menu_item.py - Changed PreviewResult to messages: list[tuple[str, MsgLevelType]] so a single result holds multiple sections. Dropped list[PreviewResult] from preview_action signature. Existing str-returning previews still work, will convert them in a follow-up PR (a44edab)
4. global_menu.py - The logic change in _prev_install_invalid_config is intentional: the original does early return on the first error, so the user only sees one problem at a time (missing configs hides the bootloader error). The new version collects all errors into one red section and always shows warnings alongside. Also added a guard: bootloader validation only runs when disk_config has a value.

archinstall/tui/components.py
```diff
@@ -21,12 +21,28 @@
...
+	if result is None:
+		widget.update('')
+		return
+
+	if isinstance(result, str):
+		widget.update(result)
+	else:
+		text = Text()
+		for i, (message, level) in enumerate(result.messages):
+			if i > 0:
```
svartkanin, line 40: why are we throwing away the first entry?

archinstall/tui/menu_item.py
```diff
@@ -1,12 +1,37 @@
...
+class MsgLevelType(Enum):
+	MsgNone = auto()
+	MsgInfo = auto()
+	MsgWarning = auto()
+	MsgError = auto()
+
+	def style(self) -> str:
+		match self:
+			case MsgLevelType.MsgNone:
+				return 'white'
```
svartkanin: Please make the stules enums so they can be type checked

```
class MsgLevelStyle(Enum):
    Green = 'green'
    ...

archinstall/tui/components.py
Softer: Nothing is dropped. The if i > 0 only skips the separator before the first block - text.append(message, ...) runs on every iteration, including i == 0. The first message has no leading blank line; the rest get one as a separator between blocks.

archinstall/tui/menu_item.py
Softer: Agree. Done.

svartkanin (approved):

PR #4513, opened 2026-05-02
Keep standalone initramfs for grub-btrfs when UKI is enabled

### LINKED ISSUE #4505: btrfs + snapper + grub-btrfs produces unbootable snapshots: UKI-only mkinitcpio preset leaves grub-btrfs without standalone initramfs
# btrfs + snapper + grub-btrfs produces unbootable snapshots: UKI-only mkinitcpio preset leaves grub-btrfs without standalone initramfs

## Summary

When installing Arch via `archinstall` with btrfs + snapper + grub-btrfs, the resulting system shows snapshots in the GRUB menu but kernel-panics when one is selected. Normal boot works fine.

The root cause is that `archinstall` configures `mkinitcpio` to produce **only a Unified Kernel Image (UKI)** and no standalone initramfs file. `grub-btrfs` requires a standalone `initramfs-linux.img` to generate working snapshot entries; without it, it generates entries with a `linux` line but no `initrd` line, which kernel-panics on boot.

The two features (UKI boot for the main system, grub-btrfs for snapshots) are each individually configured correctly, but the combination is not coherent — `archinstall` enables them together without ensuring the glue between them.

## Steps to reproduce

1. Boot the Arch ISO and run `archinstall`.
2. Choose:
   - Disk layout / filesystem: **btrfs**
   - Default subvolume layout: **yes** (produces `@`, `@home`, `@log`, `@pkg`)
   - Optional repos / packages including **snapper** and **grub-btrfs**
   - Bootloader: **GRUB**
3. Complete the install and reboot.
4. Install/enable a snapshot generator if not already (e.g., snap-pac) and trigger a snapshot:
   ```
   sudo snapper -c root create --description test
   ```
5. Run:
   ```
   sudo grub-mkconfig -o /boot/grub/grub.cfg
   ```
6. Reboot and select a snapshot from the GRUB menu.

## Expected behavior

The snapshot boots into a read-only view of the system at that point in time.

## Actual behavior

Kernel panic — the kernel cannot find/mount the root filesystem because no initramfs was loaded.

## Diagnostic evidence

Editing the GRUB snapshot entry (press `e` in the menu) shows:

```
linux "/vmlinuz-linux" root=UUID=<uuid> loglevel=3 quiet \
  rootflags=rw,relatime,compress=zstd:3,ssd,discard=async,space_cache=v2,subvol="@/.snapshots/4/snapshot"
```

There is **no `initrd` line** in the entry.

The reason is visible in `/boot/`:

```
$ ls /boot/
EFI/  grub/  intel-ucode.img  loader/  vmlinuz-linux
```

There is no `initramfs-linux.img`. The UKI exists at `/boot/EFI/Linux/arch-linux.efi` and is used by the (working) normal boot entry via `chainloader`, but `grub-btrfs` cannot use a UKI for snapshot entries — it generates traditional `linux` + `initrd` entries and looks for a standalone initramfs in `/boot/`.

The generated `/etc/mkinitcpio.d/linux.preset` confirms why no standalone initramfs exists:

```
PRESETS=('default')
#default_image="/boot/initramfs-linux.img"
default_uki="/boot/EFI/Linux/arch-linux.efi"
##fallback_image="/boot/initramfs-linux-fallback.img"
fallback_uki="/boot/EFI/Linux/arch-linux-fallback.efi"
```

`default_image` is commented out and `fallback` is not in `PRESETS`, so `mkinitcpio -P` only ever produces the UKI.

## Workaround

Edit `/etc/mkinitcpio.d/linux.preset` so that the standalone initramfs files are also generated:

```
PRESETS=('default' 'fallback')

default_image="/boot/initramfs-linux.img"
default_uki="/boot/EFI/Linux/arch-linux.efi"
default_options="--splash /usr/share/systemd/bootctl/splash-arch.bmp"

fallback_image="/boot/initramfs-linux-fallback.img"
fallback_uki="/boot/EFI/Linux/arch-linux-fallback.efi"
fallback_options="-S autodetect"
```

Then:

```
sudo mkinitcpio -P
sudo grub-mkconfig -o /boot/grub/grub.cfg
```

After this, `/boot/grub/grub-btrfs.cfg` contains snapshot entries with proper `initrd` lines (verified by `grep -A 6 "Loading Snapshot" /boot/grub/grub-btrfs.cfg`) and snapshots boot successfully.

## Suggested fix

When the user selects a configuration that includes `grub-btrfs`, `archinstall` should ensure `mkinitcpio` produces a standalone initramfs. Concretely, one of:

1. Generate `/etc/mkinitcpio.d/linux.preset` with `PRESETS=('default' 'fallback')` and the `*_image` lines uncommented (in addition to `*_uki` lines if UKI is desired). This is the smallest change and keeps both UKI and traditional booting available.
2. Use a traditional `linux` + `initrd` setup (no UKI) when `grub-btrfs` is selected, matching the configuration shipped by CachyOS and other Arch-derived distros that have snapshot booting working out of the box.
3. At minimum, surface a warning in the installer summary that snapshot booting requires additional manual configuration with the current defaults.

Option 1 is the smallest, safest change; the cost is ~50–250 MB of additional space in `/boot` for the standalone initramfs files.

## Reference: working comparable system

CachyOS (Arch-based, ships snapshot booting working out of the box) uses the systemd-style `mkinitcpio` HOOKS *and* generates standalone `initramfs-linux.img` alongside any UKI. This issue is not present there.

## Environment

- archinstall version: 4.3.1
- Arch ISO date: 2026.04
- mkinitcpio: 40-4
- grub-btrfs: 4.14-1
- snapper: 0.13.1-2
- Filesystem layout: default archinstall btrfs subvolumes (`@`, `@home`, `@log`, `@pkg`, with `.snapshots` nested inside `@`)
- Hardware: `<brief; CPU vendor matters insofar as intel-ucode/amd-ucode is in the entry>`

Softer:
Fixes #4505

When UKI and Grub are used together with btrfs snapshots (Snapper or Timeshift), _config_uki() no longer comments out *_image= lines in mkinitcpio presets and no longer deletes initramfs files. This lets mkinitcpio -P generate both UKI and standalone initramfs, so grub-btrfs can find initramfs-linux.img for snapshot boot entries.

Without this fix, _config_uki() unconditionally disables standalone initramfs generation, leaving grub-btrfs without a bootable kernel image for snapshots.

### DIFF
--- a/archinstall/scripts/guided.py
+++ b/archinstall/scripts/guided.py
@@ -115,21 +115,10 @@ def perform_installation(
 			installation.setup_swap(algo=config.swap.algorithm)
 
 		if config.bootloader_config and config.bootloader_config.bootloader != Bootloader.NO_BOOTLOADER:
-			keep_initramfs = False
-			if (
-				config.bootloader_config.uki
-				and config.bootloader_config.bootloader == Bootloader.Grub
-				and disk_config.has_default_btrfs_vols()
-				and disk_config.btrfs_options
-				and disk_config.btrfs_options.snapshot_config
-			):
-				keep_initramfs = True
-
 			installation.add_bootloader(
 				config.bootloader_config.bootloader,
 				config.bootloader_config.uki,
 				config.bootloader_config.removable,
-				keep_initramfs,
 			)
 
 		if config.network_config:
--- a/archinstall/scripts/guided.py
+++ b/archinstall/scripts/guided.py
@@ -115,7 +115,22 @@ def perform_installation(
 			installation.setup_swap(algo=config.swap.algorithm)
 
 		if config.bootloader_config and config.bootloader_config.bootloader != Bootloader.NO_BOOTLOADER:
-			installation.add_bootloader(config.bootloader_config.bootloader, config.bootloader_config.uki, config.bootloader_config.removable)
+			keep_initramfs = False
+			if (
+				config.bootloader_config.uki
+				and config.bootloader_config.bootloader == Bootloader.Grub
+				and disk_config.has_default_btrfs_vols()
+				and disk_config.btrfs_options
+				and disk_config.btrfs_options.snapshot_config
+			):
+				keep_initramfs = True
+
+			installation.add_bootloader(
+				config.bootloader_config.bootloader,
+				config.bootloader_config.uki,
+				config.bootloader_config.removable,
+				keep_initramfs,
+			)
 
 		if config.network_config:
 			install_network_config(
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -1806,7 +1806,6 @@ class Installer:
 		bootloader: Bootloader,
 		uki_enabled: bool = False,
 		bootloader_removable: bool = False,
-		keep_initramfs: bool = False,
 	) -> None:
 		"""
 		Adds a bootloader to the installation instance.
@@ -1856,6 +1855,12 @@ class Installer:
 				bootloader_removable = False
 
 		if uki_enabled:
+			keep_initramfs = (
+				bootloader == Bootloader.Grub
+				and self._disk_config.has_default_btrfs_vols()
+				and self._disk_config.btrfs_options is not None
+				and self._disk_config.btrfs_options.snapshot_config is not None
+			)
 			self._config_uki(root, efi_partition, keep_initramfs)
 
 		match bootloader:
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -1754,6 +1754,7 @@ class Installer:
 		self,
 		root: PartitionModification | LvmVolume,
 		efi_partition: PartitionModification | None,
+		keep_initramfs: bool = False,
 	) -> None:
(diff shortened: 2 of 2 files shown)


archinstall/scripts/guided.py
```diff
@@ -115,7 +115,22 @@ def perform_installation(
 			installation.setup_swap(algo=config.swap.algorithm)
 
 		if config.bootloader_config and config.bootloader_config.bootloader != Bootloader.NO_BOOTLOADER:
-			installation.add_bootloader(config.bootloader_config.bootloader, config.bootloader_config.uki, config.bootloader_config.removable)
+			keep_initramfs = False
+			if (
+				config.bootloader_config.uki
```
svartkanin: Given that this isn't really a user setup or config but installation logic we should move this inside the actual `add_bootloader`. If anyone writes their own script they shouldn't have to worry about writing this logic

Softer: Good point - moved the logic inside add_bootloader() where self._disk_config is already available.

svartkanin (approved):

PR #4518, opened 2026-05-03
Fix bspwm black screen: add provision() delegation and default configs

Softer:
bspwm profile installs packages but does not provision configuration files. After login via lightdm the user sees a black screen with no keybindings because ~/.config/bspwm/bspwmrc and ~/.config/sxhkd/sxhkdrc do not exist.
 
Two fixes: 

- DesktopProfile did not override provision() to delegate to sub-profiles (unlike ServerProfile), so BspwmProfile.provision() was never called. Added the missing delegation.
- BspwmProfile.provision() now copies example configs from /usr/share/doc/bspwm/examples/ to each user's ~/.config/ and makes bspwmrc executable. This matches what the Arch Wiki recommends for a minimal working setup.

### DIFF
--- a/archinstall/default_profiles/desktops/bspwm.py
+++ b/archinstall/default_profiles/desktops/bspwm.py
@@ -1,6 +1,10 @@
-from typing import override
+from typing import TYPE_CHECKING, override
 
 from archinstall.default_profiles.profile import DisplayServerType, GreeterType, Profile, ProfileType
+
+if TYPE_CHECKING:
+	from archinstall.lib.installer import Installer
+	from archinstall.lib.models.users import User
 
 
 class BspwmProfile(Profile):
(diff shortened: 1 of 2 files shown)


archinstall/default_profiles/desktops/bspwm.py
```diff
@@ -1,7 +1,11 @@
-from typing import override
+from typing import TYPE_CHECKING, override
 
 from archinstall.default_profiles.profile import DisplayServerType, GreeterType, Profile, ProfileType
 
+if TYPE_CHECKING:
```
svartkanin: Why are they behind a type check?

Softer: As far as I understand - python 3.14 evaluates annotations lazily (PEP 649), so Installer and User are never resolved at runtime - they're only used by type checkers. TYPE_CHECKING avoids importing heavy modules (Installer pulls in a lot) and sidesteps potential circular imports.

svartkanin: Are there actual circular deps? If not we shouls define these as regular imports

Softer: Yes, there are no cyclic dependencies right now. I didn't want to have them in the future. But you're right, it's probably overengineering.

svartkanin (approved):

archinstall/default_profiles/desktops/bspwm.py
svartkanin: Sometimes it's good to encounter them as that means the code not well structured and probably needs refactoring

PR #4483, opened 2026-04-26
Remove disabled translation-check workflow

Softer:
Fully commented out since #2119 (2023-09-28). With no `on:` triggers, GitHub records a `failure` run on every push - 1961 so far, all with zero jobs. PR check-runs are unaffected; the Actions tab is just permanently noisy. Examples: https://github.com/archlinux/archinstall/actions/runs/24960320003, https://github.com/archlinux/archinstall/actions/runs/24953074773. Full list: https://github.com/archlinux/archinstall/actions/workflows/translation-check.yaml.

PR #4519, opened 2026-05-04
Add translation CI validation

Softer:
Fix broken localization in `global_menu.py:505`: `tr(f'Invalid configuration: {error}')` evaluates the f-string before `tr()` runs, so the translation catalog never matches. Switched to `tr('...{}').format(...)` and updated the msgid in `base.pot`. 

Add `scripts/pot_tools.py` developer utility: `stats`, `list`, `add_missing [--dry-run]` for managing `base.pot`.  

Add `translation-check` CI workflow with two jobs:
 - **validate-po** (on .po changes): `msgfmt --check` on changed `.po` files, `.mo` sync warning. On .py changes: grep for `tr(f"...")` anti-pattern.
- **validate-pot** (on .py changes): verify all `tr()` strings exist in `base.pot`.
- Workflow triggers only on `.py`/`.po`/`.pot` file changes — no unnecessary runs. 

 
### Context
The previous translation-check workflow (added 2023, removed in #4483) failed because it validated all 35 languages on every PR, causing merge conflicts between translators. This new workflow checks only changed files and separates `.po` validation from `.pot` freshness checks.


Softer:
I summon @svartkanin for discussion :-)

svartkanin:
I'll have a look at it when I get some time

.github/scripts/check_pot_freshness.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env python3
```
svartkanin: Lets not roll our own parser, there is `msgcmp` to do this

scripts/pot_tools.py
```diff
@@ -0,0 +1,151 @@
...
+			print(f'=== MISSING ({len(missing)}): in code but not in base.pot ===')
+			for s in missing:
+				print(f'  + {s}')
+		else:
+			print('No missing strings')
+	finally:
+		fresh_pot.unlink(missing_ok=True)
+
+
+def cmd_add_missing(dry_run: bool = False) -> None:
```
svartkanin: Don't use manual appending `msgcat` and `msgmerge`

scripts/pot_tools.py
```diff
@@ -0,0 +1,151 @@
...
+import sys
+import tempfile
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+ARCHINSTALL_DIR = REPO_ROOT / 'archinstall'
+BASE_POT = ARCHINSTALL_DIR / 'locales' / 'base.pot'
+
+
+def extract_msgids(path: Path) -> set[str]:
```
svartkanin: why is this implemented twice? 
Also please do not do this. You are hand rolling your own parser which firstly is not trivital and secondly doesn't work as it doesn't handle multi line messages.

scripts/pot_tools.py
```diff
@@ -0,0 +1,151 @@
...
+				ids.add(current)
+			current = None
+
+	if current:
+		ids.add(current)
+
+	return ids
+
+
+def generate_fresh_pot() -> Path:
```
svartkanin: This is already part of https://github.com/archlinux/archinstall/blob/b5501d9507d824851318df36ed361097c18336d1/archinstall/locales/locales_generator.sh?plain=1#L43

We shouldn't create multiple scripts for doing the same thing. Either we enhance the existing one or we replace it.

svartkanin:
There are a lot of things happening here. As pointed out there is already a helper script for minimal things https://github.com/archlinux/archinstall/blob/b5501d9507d824851318df36ed361097c18336d1/archinstall/locales/locales_generator.sh?plain=1#L43. I'd prefer if we either enhance that or replace it. 
But given that there are a lot of moving pieces can we please do this iteratively with multiple smaller PRs. 

In addition, please do not roll your own parser. It is not trivial and prone to edge cases that are difficult to handle. There are already existing tools `msg*` available that allow you to do everything you want.

Softer:
Is it better now?

.github/workflows/translation-check.yaml
```diff
@@ -0,0 +1,138 @@
...
+        paths:
+            - 'archinstall/**/*.py'
+            - 'archinstall/locales/**'
+            - '.github/workflows/translation-check.yaml'
+    pull_request:
+        paths:
+            - 'archinstall/**/*.py'
+            - 'archinstall/locales/**'
+            - '.github/workflows/translation-check.yaml'
+jobs:
```
svartkanin, line 13: We can simplify this even more, for example 
```
jobs:
    translations:
        name: Validate translations
        runs-on: ubuntu-latest
        steps:
            - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6
            - name: Install gettext
              run: sudo apt-get update && sudo apt-get install -y gettext

            - name: Validate .po syntax
              run: |
                  set -e
                  find archinstall/locales -name '*.po' -print0 \
                      | xargs -0 -n1 msgfmt --check --output-file=/dev/null

            - name: Reject tr(f-string) anti-pattern
              run: |
                  if grep -rnE "tr\(\s*f['\"]" archinstall --include='*.py'; then
                      echo "::error::Use tr('...{}').format(...) instead of tr(f'...')"
                      exit 1
                  fi

            - name: Verify base.pot is up to date
              run: |
                  cp archinstall/locales/base.pot /tmp/committed.pot
                  ( cd archinstall && find . -type f -iname '*.py' \
                      | xargs xgettext --no-location --omit-header --keyword=tr \
                            -d base -o locales/base.pot )
                  if ! diff -q /tmp/committed.pot archinstall/locales/base.pot; then
                      echo "::error::base.pot is out of date - run archinstall/locales/locales_generator.sh all"
                      diff -u /tmp/committed.pot archinstall/locales/base.pot || true
                      exit 1
                  fi
```

All these checks are only in the GH action which makes them difficult to test, so I suggest to put these into the already existing `locale_generator.sh` file and call them from here. That means during dev they can actually be called and tested and there is a single source of truth. 
At that point this whole action can be reduced to 

```
jobs:
    translations:
        name: Validate translations
        runs-on: ubuntu-latest
        steps:
            - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6
            - name: Install gettext
              run: sudo apt-get update && sudo apt-get install -y gettext
            - name: Run translation checks
              run: bash archinstall/locales/locales_generator.sh check
```

and the checks implemented in the bash file instead.  As an example you can then have this in the script 

```
...
cmd_check_no_tr_fstring() {
	echo "Checking for tr(f-string) anti-pattern..."
	if grep -rnE "tr\(\s*f['\"]" . --include='*.py'; then
		echo "ERROR: use tr('...{}').format(...) instead of tr(f'...')" >&2
		return 1
	fi
}

cmd_check() {
	cmd_check_no_tr_fstring
     ...
	echo "All translation checks passed."
}

case "${1}" in
	check)    shift; cmd_check "$@" ;;
	generate) shift; cmd_generate "$@" ;;
	-h|--help) usage ;;
	*)        cmd_generate "$@" ;;  # back-compat: bare `<lang>` means generate
esac
```

svartkanin (approved):

svartkanin:
Thanks for the changes and addressing the feedback

PR #4526, opened 2026-05-07
Enhance log sharing capability

svartkanin:
Enhancement for PR https://github.com/archlinux/archinstall/pull/4511

* Tests use `hypothesis` to allow for arbitrary test values and wider test range
* `share-log` definition created for argument parser so it appears in the help output
* Added logfile preview to the confirmation screen so the user knows what they are uploading


Softer:
Preview... How did I not think of that...

Torxed (approved):

PR #4528, opened 2026-05-07
Add IWD standalone option to network configuration + fix NM_IWD

h8d13:
This PR enables `iwd` to be picked standalone (**WITHOUT** the dep chain `networkmanager → wpa_supplicant → pcsclite → polkit`) which enables minimal `seatd` setups better. 

`iwd` handles DHCP on the wireless link, `systemd-resolved` handles DNS ( Services: `iwd.service` + `systemd-networkd.service` + `systemd-resolved.service`) using stub symlink. 

> Would benefit from a small helper `Installer.link_resolved_stub()` since this is very similar to `Copy from ISO` option, but I limited scope there. (Perhaps for a follow-up).

_Misc: was missing parse case for `NM_IWD` my mistake in PR #4025 saved configs with this type were silently falling through._  [L:131](https://github.com/evoquus/archinstall-patches/blob/49c4206c8691f12e486d4e6745de5eb1f0cbd971/archinstall/lib/models/network.py#L131)

**Tests:** Both cases several times over, check existence of config files in target, ping.

### DIFF
--- a/archinstall/lib/models/network.py
+++ b/archinstall/lib/models/network.py
@@ -23,7 +23,7 @@ class NicType(Enum):
 			case NicType.NM_IWD:
 				return tr('Use Network Manager (iwd backend)')
 			case NicType.IWD:
-				return tr('Use iwd standalone (no Network Manager)')
+				return tr('Use standalone iwd')
 			case NicType.MANUAL:
 				return tr('Manual configuration')
 
--- a/archinstall/lib/models/network.py
+++ b/archinstall/lib/models/network.py
@@ -11,6 +11,7 @@ class NicType(Enum):
 	ISO = 'iso'
 	NM = 'nm'
 	NM_IWD = 'nm_iwd'
+	IWD = 'iwd'
 	MANUAL = 'manual'
 
 	def display_msg(self) -> str:
@@ -21,6 +22,8 @@ class NicType(Enum):
 				return tr('Use Network Manager (default backend)')
 			case NicType.NM_IWD:
 				return tr('Use Network Manager (iwd backend)')
+			case NicType.IWD:
+				return tr('Use iwd standalone (no Network Manager)')
 			case NicType.MANUAL:
 				return tr('Manual configuration')
 
@@ -125,6 +128,10 @@ class NetworkConfiguration:
 				return cls(NicType.ISO)
 			case NicType.NM:
 				return cls(NicType.NM)
+			case NicType.NM_IWD:
+				return cls(NicType.NM_IWD)
+			case NicType.IWD:
+				return cls(NicType.IWD)
 			case NicType.MANUAL:
 				nics_arg = config.get('nics', [])
 				if nics_arg:
--- a/archinstall/lib/network/network_handler.py
+++ b/archinstall/lib/network/network_handler.py
@@ -32,6 +32,12 @@ def install_network_config(
 				_configure_nm_iwd(installation)
 				installation.disable_service('iwd.service')
 
+		case NicType.IWD:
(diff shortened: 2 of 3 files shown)


archinstall/lib/models/network.py
```diff
@@ -21,6 +22,8 @@ def display_msg(self) -> str:
 				return tr('Use Network Manager (default backend)')
 			case NicType.NM_IWD:
 				return tr('Use Network Manager (iwd backend)')
+			case NicType.IWD:
+				return tr('Use iwd standalone (no Network Manager)')
```
svartkanin: Proposed change: `Use standalone iwd` 
There is no need to mention network manager

archinstall/lib/models/network.py
```diff
@@ -125,6 +128,10 @@ def parse_arg(cls, config: _NetworkConfigurationSerialization) -> Self | None:
 				return cls(NicType.ISO)
 			case NicType.NM:
 				return cls(NicType.NM)
+			case NicType.NM_IWD:
```
svartkanin, line 131: Interesting, this should have been caught by `mypy`

archinstall/lib/network/network_handler.py
```diff
@@ -45,3 +52,21 @@ def _configure_nm_iwd(installation: Installer) -> None:
...
 	_ = iwd_backend_conf.write_text('[device]\nwifi.backend=iwd\n')
+
+
+def _configure_iwd_standalone(installation: Installer) -> None:
+	# iwd manages wireless only; systemd-networkd handles wired DHCP.
+	iwd_conf_dir = installation.target / 'etc/iwd'
+	iwd_conf_dir.mkdir(parents=True, exist_ok=True)
+
+	main_conf = iwd_conf_dir / 'main.conf'
+	_ = main_conf.write_text('[General]\nEnableNetworkConfiguration=true\n\n[Network]\nNameResolvingService=systemd\n')
```
svartkanin: Make this more readable 

```
	content = textwrap.dedent("""\
		[General]
		EnableNetworkConfiguration=true

		[Network]
		NameResolvingService=systemd
	""")

	_ = main_conf.write_text(content)
```

archinstall/lib/network/network_handler.py
```diff
@@ -45,3 +52,21 @@ def _configure_nm_iwd(installation: Installer) -> None:
...
+
+	networkd_dir = installation.target / 'etc/systemd/network'
+	networkd_dir.mkdir(parents=True, exist_ok=True)
+	wired_conf = networkd_dir / '20-wired.network'
+	_ = wired_conf.write_text('[Match]\nType=ether\nKind=!*\n\n[Network]\nDHCP=yes\n')
```
svartkanin: Same here

archinstall/lib/models/network.py
correctmost: The optional `exhaustive-match` rule is not yet enabled in `pyproject.toml`: https://mypy.readthedocs.io/en/stable/error_code_list2.html#check-that-match-statements-match-exhaustively-exhaustive-match

Enabling the rule produces about 10 warnings right now.

svartkanin: Right, thanks for clarifying

h8d13:
Should be good :) untested now but do not see reason why should have changed expected outcome

Torxed:
> Should be good :) untested now but do not see reason why should have changed expected outcome

I commend your optimism ^^ I'm testing this right now, and if all goes well I think we can merge this.

svartkanin (approved):

PR #4496, opened 2026-04-28
Add installer teardown for LUKS and LVM layouts

kill74:
- This fix issue: Fixes #4491

<html>
<body>
<html><head></head><body><p><strong>Fixes #4491</strong></p>
<h2>PR Description</h2>
<p>This PR adds a guarded teardown path to <code>Installer</code> so archinstall cleans up the target storage stack after an installation attempt exits.</p>
<p>The teardown runs from <code>Installer.__exit__()</code> and is triggered only after <code>mount_ordered_layout()</code> has started. It handles:</p>
<ul>
<li>Recursive unmounting of the installation target</li>
<li>Swap deactivation</li>
<li>LUKS mapper closure</li>
<li>LVM logical volume deactivation</li>
<li>LVM volume group deactivation</li>
<li>Final udev sync</li>
</ul>
<p>The cleanup order follows the reverse of the setup topology:</p>

Configuration | Cleanup Order
-- | --
Plain LUKS | Unmount → close LUKS mappings
Plain LVM | Unmount → deactivate LVM
LVM on LUKS | Unmount → deactivate LVM → close LUKS
LUKS on LVM | Unmount → close LUKS → deactivate LVM

<p>Cleanup is best-effort so teardown does not mask the original installation error.</p>
<h2>Tests and Checks</h2>
<ul>
<li>[x] I have tested the code!</li>
</ul>
<p>Ran successfully:</p>
<pre><code class="language-bash">python3 -m compileall archinstall/lib/installer.py archinstall/lib/disk/lvm.py
</code></pre>
<p>Attempted:</p>
<pre><code class="language-bash">python -m pytest tests
</code></pre>
<p>The pytest suite did not complete in the local WSL environment because test collection fails before reaching this change with an unrelated error from <code>archinstall/lib/output.py</code>:</p>
<pre><code>NameError: name 'DataclassInstance' is not defined
</code></pre>
<p>That file is outside the scope of this PR.</p>
<p><strong>Manual validation still needed on Arch/live ISO for:</strong></p>
<ul>
<li>[ ] Plain LUKS</li>
<li>[ ] Plain LVM</li>
<li>[ ] LVM on LUKS</li>
<li>[ ] LUKS on LVM</li>
<li>[ ] No leftover <code>/mnt</code> mounts after installer exit</li>
<li>[ ] No leftover install-created LUKS mappings after installer exit</li>
<li>[ ] LVM volumes/groups deactivated after installer exit</li>
</ul></body></html>
</body>
</html>

Issue #4491, opened 2026-04-28
Request: Add teardown function


h8d13:
### describe the request

Currently when LVM or LUKS or a combination of them are selected. `/mnt` is left hanging on the host

Cleaning this up properly would allow for faster iteration (especially when testing new modifications to archinstall). I believe there is just a small dance depending on if its `LUKS_ON_LVM` `LVM_ON_LUKS` or just `vg` or just `LUKS`. 

But since we create this, we should be able to close it properly.

This also relates to #4035 in the way that you'd want to be able to retry installs with new changes quickly without rebooting the ISO/host (git pull changes, retry patch kinda deal).

PR #4536, opened 2026-05-12
Teardown after installation #4491

h8d13:
Moves the large def out of installer into `archinstall/lib/disk/utils.py`

Also preserves @kill74 's commit in #4496 which did the heavy lifting

### DIFF
--- a/archinstall/lib/disk/lvm.py
+++ b/archinstall/lib/disk/lvm.py
@@ -118,6 +118,14 @@ def lvm_vol_change(vol: LvmVolume, activate: bool) -> None:
 	SysCommand(cmd)
 
 
+def lvm_vg_change(vg: LvmVolumeGroup, activate: bool) -> None:
+	active_flag = 'y' if activate else 'n'
+	cmd = ['vgchange', '-a', active_flag, vg.name]
+
+	debug(f'vgchange volume group: {" ".join(cmd)}')
+	SysCommand(cmd)
+
+
 def lvm_export_vg(vg: LvmVolumeGroup) -> None:
 	cmd = f'vgexport {vg.name}'
 
--- a/archinstall/lib/disk/utils.py
+++ b/archinstall/lib/disk/utils.py
@@ -1,11 +1,17 @@
+import os
+from contextlib import suppress
 from pathlib import Path
+from typing import TYPE_CHECKING
 
 from pydantic import BaseModel
 
 from archinstall.lib.command import SysCommand
 from archinstall.lib.exceptions import DiskError, SysCallError
-from archinstall.lib.models.device import LsblkInfo
+from archinstall.lib.models.device import EncryptionType, FilesystemType, LsblkInfo
 from archinstall.lib.output import debug, info, warn
+
+if TYPE_CHECKING:
+	from archinstall.lib.installer import Installer
 
 
 class LsblkOutput(BaseModel):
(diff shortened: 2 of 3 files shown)


h8d13:
Reverted this a draft:

Test matrix is very large. Missing no enc generic umount + lazy fallback for when procs might still be active

archinstall/lib/disk/lvm.py
```diff
@@ -118,6 +118,14 @@ def lvm_vol_change(vol: LvmVolume, activate: bool) -> None:
 	SysCommand(cmd)
 
 
+def lvm_vg_change(vg: LvmVolumeGroup, activate: bool) -> None:
+	active_flag = 'y' if activate else 'n'
+	cmd = ['vgchange', '-a', active_flag, vg.name]
+
+	debug(f'vgchange volume group: {" ".join(cmd)}')
+	SysCommand(cmd)
```
svartkanin, line 126: I feel like an error handling might be useful here for debugging purposes

archinstall/lib/disk/utils.py
```diff
@@ -1,12 +1,18 @@
...
 from pydantic import BaseModel
 
 from archinstall.lib.command import SysCommand
 from archinstall.lib.exceptions import DiskError, SysCallError
-from archinstall.lib.models.device import LsblkInfo
+from archinstall.lib.models.device import EncryptionType, FilesystemType, LsblkInfo
 from archinstall.lib.output import debug, info, warn
 
+if TYPE_CHECKING:
+	from archinstall.lib.installer import Installer
```
svartkanin, line 14: Is this causeing circular dependencies? If not can we move this to the top

archinstall/lib/disk/utils.py
```diff
@@ -196,3 +202,119 @@ def swapon(path: Path) -> None:
...
+
+
+def teardown(installer: Installer) -> None:
+	if not installer._layout_teardown_required:
+		debug('No mounted layout registered for teardown')
+		return
+
+	info('Tearing down installation target')
+
+	with suppress(Exception):
```
svartkanin, line 214: We shouldn't supress things, either handle it or log it

archinstall/lib/disk/utils.py
```diff
@@ -196,3 +202,119 @@ def swapon(path: Path) -> None:
...
+	for vol in installer._disk_config.lvm_config.get_all_volumes():
+		if vol.fs_type == FilesystemType.LINUX_SWAP:
+			_swapoff_path(vol.dev_path)
+
+			if vol.mapper_name:
+				_swapoff_path(Path('/dev/mapper') / vol.mapper_name)
+
+
+def _teardown_lvm(installer: Installer) -> None:
+	from archinstall.lib.disk.lvm import lvm_vg_change, lvm_vol_change
```
svartkanin, line 272: Move to the top

archinstall/lib/disk/utils.py
```diff
@@ -196,3 +202,119 @@ def swapon(path: Path) -> None:
 		SysCommand(['swapon', str(path)])
 	except SysCallError as err:
 		raise DiskError(f'Could not enable swap {path}:\n{err.message}')
+
+
+def teardown(installer: Installer) -> None:
```
svartkanin, line 207: We should be passing the configuration here not the installer

archinstall/lib/installer.py
```diff
@@ -131,45 +132,51 @@ def __init__(
...
 		self.pacman = Pacman(self.target, silent)
 
 	def __enter__(self) -> Self:
 		return self
 
 	def __exit__(self, exc_type: type[BaseException] | None, exc_value: BaseException | None, traceback: TracebackType | None) -> bool | None:
-		if exc_type is not None:
-			error(str(exc_value))
+		try:
+			if exc_type is not None:
```
svartkanin, line 143: Can we move this whle thing out of the exit as that will only be run when Installer is used in the `with` context. So it would be better to have this as a function 
```
def cleanup(...)
```

that can be called 

```
installer = Installer(...)
# install steps
installer.cleanup()

```

Issue #4527, opened 2026-05-07
Request: Move new `pacman` menu and `parralel downloads` to Mirrors and repositories + Rename


h8d13:
### describe the request

Instead of creating 3 menus that interact with the same stuff, we could rename the menu item to `Pacman configuration` and nest all related items under it.

PR #4537, opened 2026-05-12
Pacman menu logic #4527

h8d13:
This PR moves `Color` and `Parallel Downloads` into `Pacman Configuration` (renamed from `Mirrors and repositories`). 

This aligns better 3 menus that essentially **interact with the same exact subject.** Also removes the `--advanced` gate for `Parallel Downloads` since its already a sane default if unchanged. 

A bit of a large blast radius (search and replace mostly, less code tho net -99 lines) but is much cleaner/logical in my opinion for long term options that might get added. 

image
image


svartkanin:
According to @Torxed https://github.com/archlinux/archinstall/discussions/3897#discussioncomment-14860284 the parallel option should not be shown by default. 

I'm not sure renaming the mirror main entry is user friendly. It is very clear what the mirror/region config refers to but putting that behind a `Pacamn configuration` namespace isn't as clear. 

I think it makes sense to potentially move parallel downloads option (behind the --advanced) into the mirrrors config as it is related to the downloads. 

The pacman config color option however doesn't make much sense in there but I agree that it's a bit odd to have that as a global namespace with a single entry for color. So we can find a new home for that. 

I haven't looked at the code at all yet, but be aware that any of these changes will require backwards compability with config parsing. 

Also it'd be great to have this as iterative, smaller PRs to make reviewing easier and faster.

h8d13:
Right sorry about the large blast to be honest don't expect it to be merged, was just showing that these things SHOULD be grouped.  

A full entry for Color = True is wild, idk how that ended up in master. 

And yes Generally using `Something Configuration` > All items related to that something would make more sense. Here all these are related to `pacman.conf` was my point :)

Torxed:
> According to @Torxed [#3897 (reply in thread)](https://github.com/archlinux/archinstall/discussions/3897#discussioncomment-14860284) the parallel option should not be shown by default.
> 
> I'm not sure renaming the mirror main entry is user friendly. It is very clear what the mirror/region config refers to but putting that behind a `Pacamn configuration` namespace isn't as clear.
> 
> I think it makes sense to potentially move parallel downloads option (behind the --advanced) into the mirrrors config as it is related to the downloads.
> 
> The pacman config color option however doesn't make much sense in there but I agree that it's a bit odd to have that as a global namespace with a single entry for color. So we can find a new home for that.
> 
> I haven't looked at the code at all yet, but be aware that any of these changes will require backwards compability with config parsing.
> 
> Also it'd be great to have this as iterative, smaller PRs to make reviewing easier and faster.

I fully agree with this, the parallel thing can move as suggested:)

PR #4547, opened 2026-05-17
Add AUR helper installation support

pentago:
## Acknowledgement before review

I am aware of the project's current stance, codified in the README under [FAQ → AUR](https://github.com/archlinux/archinstall/blob/master/README.md#aur) (added in #4468), pointing at the [arch-dev-public mailing-list consensus](https://lists.archlinux.org/archives/list/arch-dev-public@lists.archlinux.org/thread/VYOULH2GOJLFM2BXOFLWH3D754YXFPSL/). I have read that thread, #4447, and #4468 in full.

I am not opening this PR to relitigate the consensus. I am opening it because:

1. The previous attempt (#4447) was closed before its implementation was iterated on. The clean redesign sketched in [#4447 (comment)](https://github.com/archlinux/archinstall/pull/4447#issuecomment-4307149816) (chroot-time build, sudoers fenced, user-gated, no first-boot service) was never built. This PR is that implementation, plus the VM-discovered fixes that proposal did not anticipate (libalpm soname mismatch ruling out `-bin`; `TMPDIR` redirect required because chroot `/tmp` has no tmpfs).
2. If the consensus is ever revisited, having a reviewed, tested implementation on file lowers the cost of "yes." If it is not revisited, closing this with a link to the README costs you one comment.

I am explicitly not asking for special treatment, a poll re-run, or a maintainer-vs-consensus override. Close at will; the implementation will remain on my fork either way.

## What this changes

- New **"Enable AUR"** entry in the global menu (placed after "Applications") that opens a submenu offering **paru** or **yay**.
- A new `AURConfiguration` model that serializes alongside the rest of the config to `aur_config.helper_config.helper`, fully round-tripping through `--config` / `--silent` installs.
- A new `AURHandler` that runs after `install_applications` and before `install_profile_config`, installing the helper as the first sudo user (or first user) from source.

## How it works

- Helpers are built from source rather than the `-bin` variants, so the resulting binary links against the chroot's actual libalpm rather than a soname snapshot. This avoids `libalpm.so.X: cannot open shared object file` after first boot.
- A temporary `/etc/sudoers.d/99_aur_build` grants the build user `NOPASSWD: /usr/bin/pacman` for the duration of `makepkg -si`. The file is removed in a `finally` block, so a failed build never leaves a passwordless-pacman entry on the installed system.
- Builds happen in `$HOME/.cache/aur-build/<helper>` and `TMPDIR` is redirected to `$HOME/.cache/aur-build/tmp`, because `arch-chroot -S` runs the build under a systemd-run unit and the chroot's `/tmp` keeps its on-disk perms (`0755`, root-owned) since no boot-time tmpfs mount has happened. Without the `TMPDIR` redirect, build tools that default to `/tmp` (e.g. Go for `yay`) fail with `permission denied`.
- The menu item is hidden unless a regular user account exists. `--silent` installs without a regular user raise `RequirementError` rather than failing silently.

## Differences vs. #4447

| Concern raised on #4447 | Status here |
| --- | --- |
| Sudoers cleanup not guaranteed on failure | `try` / `finally` around the entire `makepkg -si` invocation; sudoers file is removed on any exit path |
| Unescaped username injection into systemd unit / sudoers | No systemd unit. Username is passed via `arch-chroot -u <user>` and `shlex`-quoted in the sudoers content |
| Dangling `multi-user.target.wants/` symlink | No service is ever created; install happens inside the chroot at install time |
| Only `users[0]` gets the helper silently | Helper installs once for the first sudo user; non-sudo users are not silently skipped — the menu item is hidden when no sudo user exists |
| `--advanced` gating | Not implemented in this PR. Happy to add if that is a merge prerequisite. |
| go/rust toolchain left systemwide | `base-devel`, `go`, and `rust` are installed via `--asdeps`, so they are eligible for `pacman -Rs $(pacman -Qdtq)` post-install. They are not explicitly removed by this PR to keep the diff minimal. |

## Scope

This PR does **not**:

- Touch `pacman.conf`, multilib, or any third-party repository.
- Add a CLI flag (configured via the existing `--config` JSON).
- Pre-install any AUR packages — only the helper itself.

## Testing

Static (all green on the patched files):

- `ruff check` + `ruff format --check`
- `mypy` (strict mode, 8 files)
- `pylint`, `flake8` (per CONTRIBUTING.md)
- `msgfmt -c` on `base.pot`
- `pytest tests/` — 27/27, including a new `test_aur_config_roundtrip` that exercises the full JSON `--config` parse/serialize path
- `pre-commit run -a` clean

End-to-end on a QEMU VM (both helpers, separately):

- Live Arch ISO booted, target disk `/dev/vda`, regular sudo user.
- Ran the full guided install with **paru** selected, then again with **yay** selected.
- Install completed in both cases; rebooted into the installed system.
- `paru --version` (paru 2.1.0) and `yay --version` (yay 12.5.7) run.
- `pacman -Qi paru` / `pacman -Qi yay` confirm installation.
- `/etc/sudoers.d/99_aur_build` is absent post-install (the `finally`-block cleanup ran).

Diff: 12 files (6 new, 6 modified). No pre-existing logic changed.

## Tests and Checks

- [x] I have tested the code!

### DIFF
--- a/archinstall/applications/aur_helper.py
+++ b/archinstall/applications/aur_helper.py
@@ -1,0 +1,69 @@
+import shlex
+from typing import TYPE_CHECKING
+
+from archinstall.lib.exceptions import PackageError, SysCallError
+from archinstall.lib.models.aur import AURHelperConfiguration
+from archinstall.lib.models.users import User
+from archinstall.lib.output import debug, info
+from archinstall.lib.translationhandler import tr
+
+if TYPE_CHECKING:
+	from archinstall.lib.installer import Installer
+
+
+class AURHelperApp:
+	"""Bootstraps an AUR helper (paru / yay) inside the chroot.
+
+	The helper package name comes from a controlled ``AURHelper`` enum value,
+	and the build username is validated upstream by archinstall's user-creation
+	path, so shell interpolation of these values is safe. Dynamic file system
+	paths are still passed through ``shlex.quote``.
+	"""
+
+	_SUDOERS_FILE = 'etc/sudoers.d/99_aur_build'
+
+	def install(
+		self,
+		install_session: Installer,
+		helper_config: AURHelperConfiguration,
+		build_user: User,
+	) -> None:
+		install_session.add_additional_packages(['base-devel', 'git'])
+
+		sudoers_path = install_session.target / self._SUDOERS_FILE
+		sudoers_path.write_text(f'{build_user.username} ALL=(ALL) NOPASSWD: /usr/bin/pacman\n')
+		sudoers_path.chmod(0o440)
+
+		helper_pkg = helper_config.helper.value
+		# Build inside the user's $HOME rather than /tmp: arch-chroot -S spawns a
+		# transient systemd-run unit and the chroot's /tmp is the on-disk
+		# directory (mode 0755, root-owned) since no boot-time tmpfs mount has
+		# happened, so non-root writes to /tmp fail. Build tools like Go also
+		# default TMPDIR to /tmp, so we redirect TMPDIR for the same reason.
+		build_subdir = f'.cache/aur-build/{helper_pkg}'
+		quoted_build = shlex.quote(build_subdir)
+		tmp_env = 'TMPDIR="$HOME/.cache/aur-build/tmp"'
+
+		try:
+			info(tr('Installing AUR helper {}').format(helper_config.helper.value))
+			install_session.arch_chroot(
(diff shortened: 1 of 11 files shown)


svartkanin:
As you pointed out rightfully the consensus was to not have any AUR support. This was decided by Arch as a public stance https://github.com/archlinux/archinstall#aur and there is no revisiting the decision at this point

pentago:
All good, thanks for the time!

Issue #1796, opened 2023-05-06
Add support for Plymouth


svartkanin:
It is not an essential thing to have to get the system up and running, but considering that setting up Plymouth requires changes in the mkinitcpio.conf it may be much more convenient to offer it through the installer

Torxed:
I agree, I would like to introduce a submenu called "optimizations" that contains non-trivial and community driven options that has zero requirements, but where we can shove popular demands that don't meet the core values of the project.

That way we can still offer the ideas - without being in the way and clutter the experience.

Thought?

svartkanin:
That sounds like a great idea. I know we have the `--advanced` flag - which I think gets overlooked by users quite a bit - would it make sense to put such things under `advanced` in the menu? 

I'm aware that the current advanced flag is for special logic triggers as well not only special menu menu options so it might get more confusing that way

Torxed:
Yea the `--advanced` flag are for those that know and respect the breaking changes it can bring using the features we hide behind the flag.

Proposal:

`--advanced` only purpose is to show a new menu option in the main menu called `experimental`.
A new menu option called `optimizations` or `community features` and we throw non-important features in there - but still vet them and put only those that many want

svartkanin:
That sounds great, i like the idea!

gallium-stanza:
Hi! Are there any recent plans regarding implementing the change discussed above?

svartkanin:
Work hasn't started yet (lots of other things were prioritised), any contributions are more than welcome

PR #4555, opened 2026-05-25
Add Plymouth configuration setup

mariovictorrs:
Add Plymouth boot splash support as opt-in in installation with the possibility to choose the theme (Only default plymouth themes available).

- This fix issue: #1796

## Tests and Checks
- [x] I have tested the code!


archinstall/lib/models/plymouth.py
```diff
@@ -0,0 +1,57 @@
...
+from archinstall.lib.models.config import SubConfig
+from archinstall.lib.translationhandler import tr
+
+
+class PlymouthConfigSerialization(TypedDict):
+	plymouth: str
+
+
+class PlymouthTheme(Enum):
+	DISABLED = 'Disabled'
```
svartkanin: This is to disable the config right? In that case that shouldn't be a theme but the entire plymouth config should then no longer be stored.

svartkanin:
Can you remove all the translation file changes from this PR otherwise it's massive. 

Also this should not be a new root entry in the menu, we can probably fit this under `Bootloader`

mariovictorrs:
> Can you remove all the translation file changes from this PR otherwise it's massive.

Done, although the CI for translations will fail, but I can add it after the review no problem.

> Also this should not be a new root entry in the menu, we can probably fit this under `Bootloader`

Sure, seems better to me, I'll make those changes and address the `DISABLED` entry.

svartkanin:
I'll address the translation pipeline, the pipeline was recently added and shouldn't be failing on such PRs

archinstall/lib/models/plymouth.py
ComicShrimp: `DISABLED` "theme" removed. Now when not set, plymouth will be set to `None`

svartkanin:
https://github.com/archlinux/archinstall/pull/4576 should fix the broken check

svartkanin:
@ComicShrimp could you rebase master the build should pass now

mariovictorrs:
@svartkanin done. All checks pass now.

Torxed (approved):

svartkanin (approved):

Issue #4565, opened 2026-05-31
archinstall fails after "Sorting your selected mirrorlist" with "failed to syncronize all databases"


stawze:
### Which ISO version are you using?

2026-05-01

### The installation log

```json
https://pastebin.com/0chTbXLd
```

#### Description of the issue

On my T490 laptop the archinstall fails after configuring the mirrorlist

This appears similar to #4305

The suggestion in the replies from @Abhinav08bhatt does not work, and time is synced, however the suggestion in this forum post does work

https://forum.endeavouros.com/t/error-gpgme-error-no-data-error-failed-to-synchronize-all-databases/56920/4

Running these commands, then archinstall _without_ configuring mirrors, results in a successful installation.

It appears there is an issue with how archinstall configures mirrors

Issue #2213, opened 2023-11-13
Refresh archlinux-keyring for old keyrings


benmordecai:
Since archinstall is included on the default archlinux installation media, it would be common for arch uses to keep an old usb drive with arch installation media on hand for emergency recovery. Also with tools like ventoy it is more popular to keep a usb drive full of isos as a toolkit. 

As these Arch isos age, the keyrings they come with expire leading to failures to install packages in the middle of the archinstall script - after destructive steps like disk partitioning and filesystem creation have been completed. The simplest solution is to reboot into the arch  install media, run `pacman -S archlinux-keyring`, then run archinstall, but new users might not know this and the user might not have access to a secondary device to easily look up commands. 

Archinstall could prevent this problem by refreshing the archlinux-keyring early in the script, or at least to offer the option to refresh the keyring before a point of no return.

Torxed:
There's an automated key-ring-update-service called [archlinux-keyring-wkd-symc.service](https://gitlab.archlinux.org/archlinux/archlinux-keyring/-/blob/643790b09ed27e00bce7d644c63b43b1af50511c/wkd_sync/archlinux-keyring-wkd-sync.service.in) on the Arch ISO as of a couple of months ago.

The issue is the race condition of network not being up during it's first execution, delaying the update but a substantial amount.

The best way forward would be to stop monkey-patching `archinstall` to deal with these things, as we've been battling this for years now. And it's a moving target.

Instead, there's two things that should happen (from a personal perspective):
 1. Arch Linux key ring must be stable over the course of 1+ month (gap between ISO's)
 2. The auto-update service needs to be more robust

benmordecai:
When I initially made this comment I was working with the 2023.06 image and thought that it was basically my fault for using an old image, so it seemed like it would be simple to add a check but as I have learned more (with your comments as well) its clear this is much more complicated and I agree with you as to the approach. I updated to 2023.11.01 and found that I could not get the archinstall script to work at all and I ended up having to redownload the 2023.06 to get it working again and before running archinstall I manually ran `pacman-key --init` and `pacman -Sy archlinux-keyring`. After that I was able to successfully install with the 2023.06 media. 

All this to say I agree that the main solution should probably be making the keyring init and update process bulletproof, which is out of the responsibility of this project.

Torxed:
The one thing I want to convey, and I truly mean this, I understand the frustration! And I agree that this is an issue - and I am pushing for change to get this fixed.

[...]

PR #4577, opened 2026-06-06
Skip custom mirror config when keyring sync fails

Softer:
Related to #4565

I couldn't reproduce the exact crash, but there is a clear gap in the code: _verify_service_stop() waits for archlinux-keyring-wkd-sync.service to reach a terminal state but never checks whether it succeeded or failed. If it failed, set_mirrors() triggers mirror speed sorting (downloading core.db from each mirror), followed by pacman -Syy which can crash with "GPGME error: No data" due to uninitialized keyring.

This change checks the service exit state after waiting. If it failed, custom mirror configuration is skipped and installation continues with default mirrors (from reflector or ISO).

What changed:
- installer.py: track _keyring_synced flag, set to False when wkd-sync service state is failed
- guided.py: check installation.keyring_synced before calling set_mirrors()


h8d13:
- #2213 and numerous other reports of this...

My thought is that perhaps there should be a fall-back path that uses steps in [comment](https://github.com/archlinux/archinstall/issues/2213#issuecomment-2625443477)

And I don't think how you handled it in code will help actual state tracking of the service:

- Rather you'd want to use stderr of the `pacman -Syy` command itself IMO

Softer:
Good point, thanks for linking #2213. You're right that checking the service state alone isn't enough - the keyring can be broken even when wkd-sync reports success (e.g. network wasn't ready when it ran).

Updated the approach: instead of skipping mirrors, we now catch the error where it actually happens - in pacman -Syy. If the sync fails with a GPGME/keyring error, we automatically reinit the keyring (killall gpg-agent, pacman-key --init, pacman-key --populate archlinux) and retry the sync. The wkd-sync state check stays as an early warning in the log.

archinstall/lib/pacman/pacman.py
```diff
@@ -53,15 +53,45 @@ def ask(self, error_message: str, bail_message: str, func: Callable, *args, **kw
...
+				)
+
 		self.synced = True
 
+	@staticmethod
+	def _reinit_keyring() -> None:
+		try:
+			SysCommand('killall gpg-agent')
+		except SysCallError as err:
+			debug(f'killall gpg-agent failed (may not be running): {err}')
```
svartkanin: you should validate if it is actually running before killing it and not hoping the error is related to that state

archinstall/lib/pacman/pacman.py
```diff
@@ -53,15 +53,45 @@ def ask(self, error_message: str, bail_message: str, func: Callable, *args, **kw
...
-		self.ask(
-			'Could not sync a new package database',
-			'Could not sync mirrors',
-			self.run,
-			'-Syy',
-			default_cmd='pacman',
-		)
+
+		try:
+			self.run('-Syy', default_cmd='pacman')
```
svartkanin: `default_cmd` is already `pacman`

archinstall/lib/pacman/pacman.py
```diff
@@ -53,15 +53,45 @@ def ask(self, error_message: str, bail_message: str, func: Callable, *args, **kw
...
+		except SysCallError as err:
```
svartkanin, line 59: can be simplified to 
```
def sync(self) -> None:
		if self.synced:
			return

		try:
			self.run('-Syy')
		except SysCallError as err:
			if b'GPGME' in err.worker_log or b'keyring' in err.worker_log.lower():
				warn('Pacman sync failed with keyring error, attempting keyring reinit')
				self._reinit_keyring()
				msg = 'Could not sync a new package database after keyring reinit'
			else:
				msg = 'Could not sync a new package database'

			self.ask(msg, 'Could not sync mirrors', self.run, '-Syy', default_cmd='pacman')

		self.synced = True
```

archinstall/lib/pacman/pacman.py
Softer: Right, dropped the redundant default_cmd='pacman' here as well as in the retry through ask.

archinstall/lib/pacman/pacman.py
Softer: Done

archinstall/lib/pacman/pacman.py
Softer: Hm... Maybe you're right. Done.

archinstall/lib/pacman/pacman.py
```diff
@@ -53,15 +53,45 @@ def ask(self, error_message: str, bail_message: str, func: Callable, *args, **kw
...
+				msg = 'Could not sync a new package database'
+
+			self.ask(msg, 'Could not sync mirrors', self.run, '-Syy')
+
 		self.synced = True
 
+	@staticmethod
+	def _is_running(process: str) -> bool:
+		try:
+			SysCommand(f'pgrep -x {process}')
```
svartkanin, line 74: Given that `gpg-agent` is a systemd service we can use `systemctl` rather 

```
	def _is_running() -> bool:
	return os.system('systemctl is-active --quiet gpg-agent.service') == 0 
```

Maybe even put that into the utils https://github.com/archlinux/archinstall/blob/af2120c0e94ada1bb7e9eea44283d19323ce2f27/archinstall/lib/utils/util.py

Softer: Good call on systemctl, that's cleaner and I see we already do exactly this for espeakup.

One thing I'm unsure about though: in the live ISO, gpg-agent is usually spawned on-demand by gpg/pacman-key rather than started as a systemd service, so systemctl is-active --quiet gpg-agent.service might report it inactive even while the process is actually running. In that case we'd skip the killall, and that's the one spot where killing it matters (a stale agent holding the old /etc/pacman.d/gnupg open before --init). pgrep catches the process directly regardless of how it was started, which is why I reached for it.

That said, I might be overthinking the ISO behavior here. Have you actually seen gpg-agent show up as an active unit in that context? If so I'm happy to switch to systemctl, otherwise maybe pgrep is the safer bet. What do you think?

svartkanin: Fair point, it seeems to run as a process not as a service 
image

svartkanin (approved):

PR #4578, opened 2026-06-06
Add users to seat group when seatd is selected

### LINKED ISSUE #4471: Sway + seatd greeter lock

archlinux-2026.04.01-x86_64.iso

### The installation log

```json
Nothing of note, installation go as planned without errors
```

### describe the problem

Installing on a bare metal laptop following a clean gpt table overwrite preceding the boot used for archinstall itself.
Using archinstall after pacman-key --init as advised and verifying boot mode.
I use archinstall --advanced for the record but this is unrelated
Archinstall latest 4.3.

In the options I select desktop profile then sway, for hardware access seatd instead of polkit, amd graphic drivers for my iGPU and lightdm slick.

Archinstall process goes smoothly.

The issue is when I manage to boot the installed system afterwards, I get the lightdm screen and enter the correct user/password it fades like it is going to authenticate and then goes back to the lightdm menu.

So i tested with various other greeters since I am more used to sddm, and using sddm/plasma login/ly it does not work as well. Only when using ly it throws a rapidly disapearing error message on the top left regarding something about seatd.

Now about the problem cause, I generally use kde + sddm/plasma login and already did a sucessful install with archinstall 4.3. So issue is not disk config or graphics driver but specifically tied to the DE or greeter.
Since I also managed to install kde with ly without ly throwing errors, I think it is tied to sway WM.

Now i have not been able to test if sway + polkit works, but it should be investigated.
And again archinstall itself works smoothly, it just looks like it installed a non functional configuration for that WM

I'll be happy to help further for that issue, if you have more questions feel free to ask

Softer:
Fixes #4471

When seatd is chosen as seat access for sway/hyprland/niri/labwc, the compositor needs the user to be in the seat group to access input devices. Without this, the compositor fails to start after installation.

Verified on real hardware: installed sway + seatd, confirmed seat group was missing from user groups. After the fix, groups test shows seat.

What changed:
- desktops/utils.py: added provision_seat_access() helper that runs usermod -a -G seat for each user when seatd is selected
- sway.py, hyprland.py, niri.py, labwc.py: added provision() override calling the helper


archinstall/default_profiles/desktops/hyprland.py
```diff
@@ -1,8 +1,12 @@
-from typing import override
+from typing import TYPE_CHECKING, override
 
-from archinstall.default_profiles.desktops.utils import select_seat_access
+from archinstall.default_profiles.desktops.utils import provision_seat_access, select_seat_access
 from archinstall.default_profiles.profile import CustomSetting, DisplayServerType, GreeterType, Profile, ProfileType
 
+if TYPE_CHECKING:
+	from archinstall.lib.installer import Installer
```
svartkanin: why are they in the TYPE_CHECKING scope?

archinstall/default_profiles/desktops/utils.py
```diff
@@ -1,16 +1,31 @@
...
 
 class SeatAccess(Enum):
 	seatd = 'seatd'
 	polkit = 'polkit'
 
 
+def provision_seat_access(
+	install_session: Installer,
+	users: list[User],
+	seat_access: str | None,
```
svartkanin: The function itself is pointless if the seat_access is passed as `None`, the consumers of the function should perform this validation and only call it if it is not None

archinstall/default_profiles/desktops/hyprland.py
```diff
@@ -45,6 +49,10 @@ def services(self) -> list[str]:
 			return [pref]
 		return []
 
+	@override
+	def provision(self, install_session: Installer, users: list[User]) -> None:
```
svartkanin: all these profiles essentially implement the same provision logic, instead we can move this into https://github.com/archlinux/archinstall/blob/13944f3ccad8ad6ec7de26791a65eab9aed872d1/archinstall/default_profiles/desktop.py?plain=1#L92

```
	@override
	def provision(self, install_session: Installer, users: list[User]) -> None:
		for profile in self.current_selection:
			profile.provision(install_session, users)

			if seat_access := profile.custom_settings.get(CustomSetting.SeatAccess):
				provision_seat_access(install_session, users, seat_access)

```

Softer:
Thanks @svartkanin! I reworked it so all three are covered at once. The seat-access setup now lives in the base DesktopProfile.provision() instead of being copy-pasted into each Wayland profile, so Hyprland/Sway/niri/labwc lost their duplicated overrides (and with them the TYPE_CHECKING imports). The None check moved to the caller, so the helper just takes a str now.

archinstall/default_profiles/desktops/utils.py
```diff
@@ -1,16 +1,31 @@
 from enum import Enum
+from typing import TYPE_CHECKING
 
 from archinstall.lib.menu.helpers import Selection
 from archinstall.lib.translationhandler import tr
 from archinstall.tui.menu_item import MenuItem, MenuItemGroup
 from archinstall.tui.result import ResultType
 
+if TYPE_CHECKING:
```
svartkanin: Remove the TYPE_CHECKING here

Softer: Oops, missed that one - leftover from the refactor. Fixed.

svartkanin (approved):

PR #4587, opened 2026-06-10
Extract TUI binding descriptions into base.pot and translate raw widget bindings

### LINKED ISSUE #4584: Some strings aren't being generated by `locales_generator.sh`
Strings from the bottom help bar and the F1-triggered help section aren't being generated by the script.
<details>
<summary>Screenshot</summary>
Image
</details>

The Polish translation shows up as 100% done, but it really isn't. If I manually put:
```po
msgid "Up"
msgstr "W górę"
```
inside the Polish translation file, it works correctly:

Image

However, this edit will be commented out after running `./locales_generator.sh pl`. If I also put the strings in `base.pot`, the script will remove them.

I think this is because the script only detects `tr("string goes here")`, and these remaining strings are translated from a variable:
https://github.com/archlinux/archinstall/blob/c6cb7b319d1878e1c9cc008d33c3aa5c5857d3ff/archinstall/tui/components.py#L46-L55
This was added in #4363. The PR also manually added the strings to `base.pot`, but they got removed by the script in #4557.

Softer:
Fixes #4584.

The help bar and F1 panel strings never made it into base.pot: they are binding descriptions translated at runtime from a variable (tr(b.description)), and xgettext only knew the literal tr("...") form. #4363 worked around that by adding the strings to base.pot manually, but the regeneration in #4557 wiped them again (and with the freshness check gone since #4576, nothing noticed).

What's here:

- locales_generator.sh passes two more keywords to xgettext: Binding:3 extracts the description argument of our own Binding() definitions, tr_noop marks strings that exist only inside the textual package.
- New tui/binding_descriptions.py lists those textual built-ins. The list covers the full MRO of the widgets we use - _translate_bindings() works on the merged bindings map, so inherited descriptions like "Scroll Up" are translated at runtime too. They carry live translations in 18 locales today; leaving them out of the pot would have obsoleted those on the next msgmerge.
- Raw Input and Button widgets never went through _translate_bindings(), so their help-panel descriptions stayed English no matter what was in the .po - subclassed them (_Input/_Button) following the existing _OptionList pattern.
- test_tooling/check_binding_descriptions.py - a small manual checker to re-verify the list after a textual upgrade (not wired into CI).

Note: the pot regen also picks up the three Plymouth strings from #4555 - master's pot simply hadn't been regenerated since that merge.


Softer:
The mypy job died on infra - Docker Hub timed out while pulling the container, mypy itself never ran (it passes locally). Could someone with the button re-run the failed job?

Torxed:
Yeah the infra somewhat often dies and runner results geta screwed. Those are fine, I manually check them usually or re-run from admin panel on GH.

svartkanin:
This looks quite gnarly. I need to figure out if this is really the best way of doing this

Softer:
> This looks quite gnarly. I need to figure out if this is really the best way of doing this

Any updates? :)

svartkanin:
I haven't forgotten about it, but I had some personal things ongoing so I'll get to it hopefully shortly

svartkanin (approved):

svartkanin:
I played around with this today and it seems your approach is the only working way I encountered as well. As textual doesn't have any support for translations we are forced to extract the default bindings out and do the fake translation calls. 

Could you fix the mypy action and add a small entry in the `/locale/README.md` around how to generate the descriptions then it's good to go

Softer:
> Could you fix the mypy action and add a small entry in the `/locale/README.md` around how to generate the descriptions then it's good to go

Actually, I can't... This issue is not related to my code... Look at the @Torxed comment above...

```
Starting job container
  /usr/bin/docker pull archlinux/archlinux:latest
  Error response from daemon: Get "https://registry-1.docker.io/v2/": context deadline exceeded
  Warning: Docker pull failed with exit code 1, back off 6.801 seconds before retry.
  /usr/bin/docker pull archlinux/archlinux:latest
  Error response from daemon: Get "https://registry-1.docker.io/v2/": context deadline exceeded
  Warning: Docker pull failed with exit code 1, back off 3.207 seconds before retry.
  /usr/bin/docker pull archlinux/archlinux:latest
  Error response from daemon: Get "https://registry-1.docker.io/v2/": context deadline exceeded
  Error: Docker pull failed with exit code 1
```

svartkanin:
can you rebase master?

PR #4588, opened 2026-06-10
Select the installer language before wifi menu

acuteenvy:
## PR Description:

The wifi menu was always shown in English, even though translations exist. This is because it's launched before the main menu, and the user has no chance to select the archinstall language.

Now there's a language selection menu before the internet connection check.

## Tests and Checks
- [x] I have tested the code!

### DIFF
--- a/archinstall/lib/general/general_menu.py
+++ b/archinstall/lib/general/general_menu.py
@@ -107,7 +107,8 @@ async def select_archinstall_language(languages: list[Language], preset: Languag
 
 	title = 'NOTE: Console font will be set automatically for supported languages.\n'
 	title += 'For other languages, fonts can be found in "/usr/share/kbd/consolefonts"\n'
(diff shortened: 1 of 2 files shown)


svartkanin:
I am not in favor of this as it will always prompt for the language selection. 
IF we want this the language selection could be triggered with a special key binding instead when in the pre-menu selection such as the wifi setup.

acuteenvy:
> it will always prompt for the language selection

Yeah, that was the point. I thought it's not that big of a deal - English is the default, just press enter and the menu is gone. Many installers ask for the language first.

vscsilva:
+1 for this
I really don´t think choosing the language will bother anyone

Torxed:
I need to think about this one too, it is a step that I think will be redundant for (wild guess obv) 80%+ of the user base, but I'll think and look at this one when I'm not in a rush.

PR #4598, opened 2026-06-20
fix ValueError exception when no disks are detected in the disk menu

brurmonmemment:
so in lib/disk/disk_menu.py, the menu items are created using info from a list called "devices" (device_handler.devices). when no block devices are detected, the list is empty. right now, theres no empty list detection so it goes ahead and creates zero MenuItems, and causes this ValueError from tui:

ValueError: Menu must have at least one item

i added a check to see if the length of the list is shorter than 1 (empty list) before constructing the menu. if the list is empty, the user will be notified that no disks were detected. otherwise, the menu is created and behaves as normal


Torxed (approved):
Looks like a sane thing, I'll test it as I'm pushing a release atm.

Issue #4585, opened 2026-06-10
Plymouth on NVIDIA: splash needs early module loading, LUKS prompt is the worst case


Softer:
After #4555 landed I dug into how the new Plymouth option interacts with NVIDIA setups (a topic I find interesting), and there's a gap: we install nvidia-open but never put the kernel modules into the initramfs.

  Two consequences:

  - The splash starts on SimpleDRM, and when nvidia-drm finally loads from the root fs it kicks SimpleDRM out. Depending on hardware that's anywhere from a quick mode switch to a few seconds of black screen (plymouth upstream issue 264 is about exactly this).
  - The nasty one: with LUKS, current plymouth (26.134.222) deliberately skips SimpleDRM and waits for a real DRM driver. There is no nvidia-drm in the initramfs, so at the password prompt stage it never shows up - the user gets a delayed/text/black prompt. And nvidia + LUKS + Plymouth is a combination archinstall happily produces now.

  The good news: everything else just works in 2026. modeset=1 and fbdev=1 are driver defaults, and the kms hook is harmless next to nvidia (modconf puts the blacklists into the image). Early loading is the only missing piece.

  So the suggestion: when the gfx driver is NVIDIA and Plymouth is selected, add

  `MODULES=(nvidia nvidia_modeset nvidia_uvm nvidia_drm)`

  and probably ship a pacman hook to rebuild the initramfs on driver updates, otherwise the modules go stale (CachyOS early-loads unconditionally via chwd, and stale modules are their most common breakage).

  This overlaps with #1239, but I'd argue that one has outlived itself: it asked to early-load for everyone back when modeset wasn't a default and basic functionality was the concern. In 2026 late loading just works for the general case - Plymouth is the only scenario left where early loading buys something the user actually sees. So it might even make sense to close #1239 in favor of this issue, with its much narrower scope (only nvidia users who explicitly turned Plymouth on).

h8d13:
Plymouth has always been hell, i swear most bricked reports i've seen is something do with playing around with Plymouth. 
All that to modify somethign purely cosmetic (really not needed at all for installing an operating system). I would have left it out of `archinstall` to be honest

Torxed:
Nice catch, I was planning on playing around with this as well. Plymouth might be deserving of `--advanced` or at least a fairly obvious warning as it does indeed cause trouble at times. But it is a nice feature to have if you want a more pretty experience :)

mariovictorrs:
[...]

PR #4604, opened 2026-06-28
Warn before enabling Plymouth in the bootloader menu

Softer:
When enabling Plymouth in the bootloader menu, show a yellow confirmation warning that lists the most common ways it breaks booting, and ask the user to confirm. When Plymouth is already enabled the user is only switching the theme, so the warning is skipped in that case.

The warning text:

> Plymouth adds a cosmetic boot splash but can cause boot problems on some setups:
>
>   - black screen with the NVIDIA driver
>   - hidden password prompt with disk encryption (LUKS)
>
> Would you like to enable it?

## Why

Plymouth is purely cosmetic but is a frequent source of "system does not boot" reports, most notably a black screen with the NVIDIA driver and a hidden LUKS passphrase prompt (you cannot tell the system is waiting for your password). A confirmation step makes the risk visible before the user commits to it, while keeping the feature available for those who want it.

This follows the discussion in #4585, where a warning on selection was preferred over hiding Plymouth behind `--advanced`.

## Notes

- The warning only fires when going from disabled to enabled; theme changes are untouched.
- New strings are added to `base.pot`.

Part of #4585.

### DIFF
--- a/archinstall/lib/bootloader/bootloader_menu.py
+++ b/archinstall/lib/bootloader/bootloader_menu.py
@@ -130,6 +130,34 @@ class BootloaderMenu(AbstractSubMenu[BootloaderConfiguration]):
 		return bootloader
 
 	async def _select_plymouth(self, preset: PlymouthTheme | None) -> PlymouthTheme | None:
+		# Plymouth is purely cosmetic and a frequent source of boot breakage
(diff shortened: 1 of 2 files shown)


Torxed (approved):
Fair enough to add a warning to Plymouth I recon.

PR #4607, opened 2026-06-28
Version bump

Torxed:


Torxed:
Not going to lie, this terrifies me:

image

`./locale_generator.sh all` did something different this time.
But seeing as it's "only" language changes, I'm going for it!

Torxed:
image


image

Ok something is really funky, since I can mainly verify Swedish/English - I looked at it and it seems that it's commented out a bunch of things or made them "obsolete translation entries"?

I will revert the language changes, and bump version from another branch.. Would love an extra set of eyes on this one to see if this is intended behavior or not @svartkanin, in your own time.

svartkanin:
I'll have a look what's going on

Torxed:
> I'll have a look what's going on

Appreciate it! Because I couldn't make sense of it :/

This is what I ran if it helps:

```shell
(cd archinstall/locales/ && ./locales_generator.sh all)
```

Softer:
Looks like it's just a reorder... But not sure.

svartkanin:
@Torxed I did some analysis and I think it's safe. 
With the changes in https://github.com/archlinux/archinstall/pull/4519 the `--join-existing` flag was removed which told `xgettext` (https://www.man7.org/linux//man-pages/man1/xgettext.1.html) to add new strings to base.pot. But that was removed and so what happens now is that obsolete strings - translation strings no longer used in the code base - are marked with "#~" prefix and then the file is sorted. This leads to the massive diff, but in future reruns it shouldn't be this big again. 

Example of unused string prefix change  https://github.com/archlinux/archinstall/pull/4607/changes#diff-bb45424a7f212764b65d5ec1b97f25934f61a70692e14422bd99d3d9315a6691R1248

```
#~ msgid "[!] A log file has been created here: {} {}"
#~ msgstr "[!] مِلَف سِجِل أُنشِأ هُنا: {} {}"
```

PR #4620, opened 2026-07-08
Enable more mypy and Pylint checks for test_tooling/

correctmost:
## PR Description:
I think it's worthwhile to type check test code.  This also makes it easier to run other type checkers on the codebase without configuring exclusions.

## Tests and Checks
- I'm not able to test the changes locally, but I did review the code a few times before submitting the PR

### DIFF
--- a/test_tooling/qemu/qemu.py
+++ b/test_tooling/qemu/qemu.py
@@ -11,6 +11,8 @@ import sys
 import time
 from argparse import ArgumentParser
 from collections.abc import Iterator
+from functools import cache
+from pathlib import Path
 from select import EPOLLHUP, EPOLLIN, epoll
 from shutil import which
 from types import TracebackType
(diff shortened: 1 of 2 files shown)


test_tooling/qemu/qemu.py
```diff
@@ -334,81 +343,82 @@ def decode(self, encoding: str = 'UTF-8') -> str:
 		return self._trace_log.decode(encoding)
 
 
-def ensure_sudo():
-	global sudo_password  # pylint: disable=global-statement
+@cache
+def get_sudo_password() -> str:
```
correctmost, line 347: One flaw with this code and the previous `ensure_sudo` code is that a mistyped password cannot be corrected after the first attempt is tried.

svartkanin (approved):

PR #4630, opened 2026-07-13
Fix Galician locale

cachelous:
## PR Description:

The Galician (gl) translation was in poor shape — a large share of the strings were actually Spanish rather than Galician ("Se ha creado un archivo de registro", "¿Quiere continuar?", etc.), mixed with non-normative Galician, and the file had drifted out of sync with the template.

This PR rewrites the locale from scratch:

- Synced `base.po` with the current template and removed the obsolete (`#~`) entries, which only preserved the old Spanish strings.
- Retranslated all 380 strings into normative Galician, with consistent terminology throughout.
- Every string was reviewed by a native speaker (me).

## Tests and Checks
- [ ] `msgfmt --check` and `./locales_generator.sh check` pass; verified all placeholders and `python-brace-format` flags are intact (380 translated, 0 fuzzy).


svartkanin (approved):

gaelgnz:
"make it exist then make it better", i made this translation with a lot of search and replace though i did some lines manually, i had planned to make it better in the future but i got busy

PR #4633, opened 2026-07-15
fix: Updated WiFi SSID Parsing Logic

CooperWang0912:
## PR Description:

- The current code relies on `split()` for parsing returns from `wpa_cli scan_results`
- Therefore WiFi with SSID that contains spaces cannot be parsed correctly
- The new parsing logic relies on tabs, and uses a `max_split` argument for when tabs are not present

## Tests and Checks

### DIFF
--- a/archinstall/lib/models/network.py
+++ b/archinstall/lib/models/network.py
@@ -178,8 +178,12 @@ class WifiNetwork:
 			if not line:
 				continue
 
-			parts = line.split()
-			if len(parts) != 5:
+			if '\t' in line:
+				parts = line.split('\t')
+			else:
+				parts = line.split(maxsplit=4)
+
+			if len(parts) < 5:
 				continue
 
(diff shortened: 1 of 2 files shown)


nmarley:
Nice catch on the root cause, that's the same place I landed.

One thing to watch: the scan_results header is space-separated (`bssid / frequency / ...`), not tab-separated. With the
`split(maxsplit=4)` fallback that header can parse as 5 parts and show up as a fake network. Also `len(parts) < 5` drops empty-SSID rows (hidden APs).

I opened #4634 with `split(None, 4)`, an explicit header skip, empty-SSID handling, and unit tests covering those cases. Happy to align if you prefer tab-first parsing; the important part is not requiring exactly 5 whitespace-split fields.

CooperWang0912:
@nmarley Thanks for the review, your PR looks great and we can go ahead with that fix. Closing my draft PR now!

Issue #4632, opened 2026-07-14
WiFi SSIDs with spaces missing from TUI list


nmarley:
### Which ISO version are you using?

2026-07-01

### The installation log

```json
I don't have network connectivity to be able to capture this from the device.
```

### describe the problem

When using the guided installer's WiFi setup (wpa_supplicant / `wpa_cli`), networks whose SSIDs contain spaces never appear in the selection menu, even though they show up in `wpa_cli scan_results` / `iwctl`.

### To Reproduce:

1. Boot Arch ISO (reproduced on `archlinux-2026.07.01`, `archinstall 4.4-1`, kernel `7.0.14-arch1-1`).
2. Have a nearby AP whose SSID contains a space (e.g. `My Home Network`).
3. Start `archinstall` and use the WiFi connection prompt.
4. Observe: SSIDs **without** spaces are listed; SSIDs **with** spaces are missing.

### Expected:

All scanned SSIDs from `wpa_cli scan_results` are shown, including names with spaces.

### What Actually Happens:

SSIDs with spaces are not present in the menu.

CooperWang0912:
I believe I have located the root cause, which is `line 181-183` under `lib/models/network`. 

The current code that parses `wpa_cli scan_results` is structured to be:
````
parts = line.split()
if len(parts) != 5:
	continue
````
In this case, when parsing SSID with spaces, the length of the parsed array will exceed 5, and fail silently due to the conditional check.

I have created a draft PR that works around this by using `\t` as the delimiter, and falling back to using `split(maxsplit=4)` when tabs are not present. Will probably create tests later for this to verify that it works.

h8d13:
Should be good to close :)

PR #4634, opened 2026-07-15
Fix WiFi SSIDs with spaces missing from scan list

nmarley:
- This fixes issue: #4632

## PR Description:

`WifiNetwork.from_wpa` used `line.split()` and required exactly 5 fields. SSIDs with spaces produced more tokens and were dropped silently, so they never appeared in the TUI list.

- Parse with `split(None, 4)` so the SSID is the remainder (spaces preserved)
- Skip the header row explicitly
- Allow empty SSID (hidden networks)

Unit tests cover SSIDs with and without spaces, empty SSID, header/blank line skip, and field preservation.

## Tests and Checks
- [x] I have tested the code

### DIFF
--- a/archinstall/lib/models/network.py
+++ b/archinstall/lib/models/network.py
@@ -178,12 +178,24 @@ class WifiNetwork:
 			if not line:
 				continue
 
-			parts = line.split()
-			if len(parts) != 5:
+			if line.lower().startswith('bssid'):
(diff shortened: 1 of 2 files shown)


svartkanin (approved):

svartkanin:
Thank you for adding tests

PR #4635, opened 2026-07-15
Add rtkit to improve sound on Linux

davgar99:
..


svartkanin:
This seems to be more of a nice to have rather than a required package to get pipewire up and running no?

svartkanin:
`archinstall` tends to primarily focuses on installing essential packages for something to work and stays away from nice to haves if not explicitly pointed to by the wiki. There are always use cases for certain packages to be valuable but at the same time they are opinionated. So to keep things simple it shouldn't install things that are not strictly required for pipewire to work out of the box.

PR #4639, opened 2026-07-17
Update translations

castillofrancodamian:
- This fix issue: 

## PR Description:

1. Update translation template
2. Update translation files for all languages
3. Update Spanish translation

## Tests and Checks

### DIFF
--- a/.github/workflows/github-pages.yml
+++ b/.github/workflows/github-pages.yml
@@ -22,7 +22,7 @@ jobs:
       options: --privileged
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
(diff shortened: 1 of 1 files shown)


svartkanin:
Please don't submit a update for all languages but only for the spanish translation if that's what the PR is for

castillofrancodamian:
@svartkanin Why not update all the files? That seems like the best approach, in addition to updating the Spanish translation.

Softer:
@svartkanin I think we need to clean garbage anyway. Why not in this PR? Or create a separate one...

svartkanin:
Because that convolutes the history and it's difficult to follow where what happened. All translation files need to be updated to meet a reformatting which should be done as its own PR to check any problems in the future and make debugging easier

PR #4648, opened 2026-07-20
Fix too-many-statements-in-try-clause violations

codefiles:
`ruff check --preview` gives errors on these.

https://docs.astral.sh/ruff/rules/too-many-statements-in-try-clause/

### DIFF
--- a/archinstall/lib/network/wifi_handler.py
+++ b/archinstall/lib/network/wifi_handler.py
@@ -72,16 +72,17 @@ class WifiHandler(InstanceRunnable[bool]):
 
 		try:
 			SysCommand(f'wpa_supplicant -B -i {wifi_iface} -c {self._wpa_config.config_file}')
-			result = self._wpa_cli('status')  # if it it's running it will blow up
-
-			if result.success:
-				debug('successfully enabled wpa_supplicant')
-				return True
-			else:
-				debug(f'failed to enable wpa_supplicant: {result.error}')
-				return False
 		except SysCallError as err:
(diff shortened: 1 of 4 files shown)


archinstall/lib/network/wifi_handler.py
```diff
@@ -72,18 +72,19 @@ async def _enable_supplicant(self, wifi_iface: str) -> bool:
...
-				debug('successfully enabled wpa_supplicant')
-				return True
-			else:
-				debug(f'failed to enable wpa_supplicant: {result.error}')
-				return False
 		except SysCallError as err:
 			debug(f'failed to enable wpa_supplicant: {err}')
 			return False
 
+		result = self._wpa_cli('status')  # if it it's running it will blow up
```
svartkanin: If the comment is still valid than calling `status` will throw an exception so it should be tried with return false no

codefiles: The comment is probably a duplicate from this line:

https://github.com/archlinux/archinstall/blob/389f4cfc64b8b4dbb174a03fc1c0da6450fe87d3/archinstall/lib/network/wifi_handler.py#L61

`_wpa_cli` uses `SysCommand` with a handled `SysCallError`.

https://github.com/archlinux/archinstall/blob/389f4cfc64b8b4dbb174a03fc1c0da6450fe87d3/archinstall/lib/network/wifi_handler.py#L198-L214

The comment should be removed and this additional try block can be cleaned up too.

svartkanin (approved):

PR #4663, opened 2026-07-24
feat: Menu Status Indicator

### LINKED ISSUE #4658: [Feat] Menu Item Status Indicator
## Motivation
- Currently the menu does not indicate whether a field is configured or not
- Users can only find out if critical fields are missing in the `Install` page
- There are already functions inside `global_menu` that check for missing configuration, and can be easily reused

## Proposed Addition
- With `Textual` already included in the dependencies, colored symbols can be used to indicate required fields that are missing, configured fields, and optional fields
- The addition will hopefully make the UI more user friendly and straight forward

## Sample Menu Render

Image

## Please feel free to leave feedback on the proposal inside this issue!

CooperWang0912:
## PR Description
- Created status indicator for the installation menu
image

## Motivation
- Visually indicate which fields are preventing the user from installing
- Create instant feedback when user configures a field
- Provide better user experience for people unfamiliar with the installation process
- Closes #4658 

## Technical Description
1. Added method `_get_status_prefix` for menu item completion status, with red exclamation marks for required and uncompleted fields, yellow dots for optional and uncompleted fields, and green checks for completed fields.[¹] The checks are achieved using the following logic: 
```python
if item.read_only or item.key in (SpecialMenuKey.SAVE.value, SpecialMenuKey.INSTALL.value, SpecialMenuKey.ABORT.value):
			return ''

		if item.key == 'auth_config':
			auth_config: AuthenticationConfiguration | None = item.value
			is_auth_valid = auth_config is not None and (auth_config.root_enc_password is not None or auth_config.has_superuser())
			if is_auth_valid:
				return '[bold green][✓][/bold green] '
			return '[bold red][!][/bold red] '

		# Standard mandatory or configured item check
		if item.has_value():
			return '[bold green][✓][/bold green] '
		elif item.mandatory:
			return '[bold red][!][/bold red] '
		else:
			return '[bold yellow][•][/bold yellow] '
```

2. Added method `_update_item_labels` that updates the menu render based on `_get_status_prefix`. The code is as follows: 
```python
def _update_item_labels(self) -> None:
		"""
		Re-applies translated titles and status prefixes to all active menu items.
		"""
		raw_options = self._get_menu_options(wrap_actions=False)
		new_options = {o.key: o.text for o in raw_options if o.key is not None}

		for item in self._item_group.items:
			if item.key in new_options:
				base_title = new_options[item.key]
				prefix = self._get_status_prefix(item)
				item.text = f'{prefix}{base_title}'
```

3. In order to avoid stale feedback due to background process not instantly updating values, created a wrapper that calls the `_update_item_labels` when called, and is triggered with `_get_menu_options`.[²] The implementation is as follows:
```python
def _wrap_action(self, item: MenuItem, action: Callable[..., Any]) -> Callable[..., Any]:
		@wraps(action)
		async def wrapper(*args: Any, **kwargs: Any) -> Any:
			try:
				if inspect.iscoroutinefunction(action):
					result = await action(*args, **kwargs)
				else:
					result = action(*args, **kwargs)
					if inspect.isawaitable(result):
						result = await result

				item.value = result

				return result

			finally:
				self._update_item_labels()

		return wrapper
```

## Tests and Checks
- [x] I have tested the code!

## Notes
[¹] The color scheme and symbols are up for discussion, and changes to these would be fairly straight forward
[²] I am using this primarily to contain the blast radius inside `global_menu`, and I would be happy to rework the logic if needed


svartkanin:
I'm not sure this is going in the right direction. as these symbols are not very intuitive. 
What is the exact problem with the install preview pointing to the missing settings?

CooperWang0912:
@svartkanin Thanks for the feedback. I am completely open to swapping out the symbols for something more intuitive, or just ditching the symbols altogether. Some other paths I can think of right now are to color code the fields directly or use colored texts instead of symbols, but I do think that any labelling would be a net positive for user experience, since currently there is quite literally no way to tell field configuration status from the menu bar. 

As for the install preview, I think it provides limited functionality since it only indicates critical missing settings, which means that the users wouldn't know what other non-critical fields are not configured, such as the additional packages option. Having a status menu also means that the user gets an instant feedback when a field is configured, which I think would help with the overall flow of the installation process.

CooperWang0912:
@svartkanin Here is the rendered menu for the other option of using colored tags, and we can also decide what the tags and their colors should be. In addition, I would also be happy to open up a discussion and see what the community prefers in terms of the UI design. 
 
image

KsanDjordje:
I wonder if colored tags are the right approach here. The installation summary already clearly shows which configuration is missing before the installation starts, so adding multiple colored status tags to the main menu might introduce more visual noise than value.

If we want to indicate progress in the menu itself, perhaps a simpler approach would be enough. For example, using a plain + for options that haven't been configured yet and a > (or another simple marker) for configured options, without the surrounding []. That would still provide a quick visual indication without drawing too much attention away from the menu itself. Simplicity and readability are key.

That's just my take.

CooperWang0912:
@KsanDjordje Yeah that makes a lot of sense. I guess the main problem is figuring out simple and clear symbols to indicate configured vs missing fields.  That was actually why I leaned into color coding, which I thought was easier to understand compared to the symbols and added a second level of safety. 

Would love to see some community suggestions for implementation ideas.

CooperWang0912:
Here is another render with removed brackets to avoid too much visual noise and hopefully more readable symbols:
image

svartkanin:
To be honest none of these looks very intuitive to me. I would not know what to associate `+` or `*` with at all.
I would rather only have missing ones marked to very easily tell what's it about. I think maybe the exclamation mark would be suitable for it. 

As a note on the code, the validation to show it or not should probably go to `components.py` and be evaluated in there generically, reason being, your implementation only caters for the global menu at this point, but other sub-menus may also benefit from this and it keeps things consistently then

CooperWang0912:
@svartkanin Thanks again for your input. I've refactored the `_get_status_prefix` function to be inside `components.py`, and changed the logic so that only missing fields are reported using yellow exclamation marks. 

One thing that might be worth noting is that I had to use `item.text` to check for special keys such as `Install`, `Abort`, and `← Back`, since I cannot import `SpecialMenuKey` from `abstract_menu.py` due to circular imports, and the `← Back` button also doesn't have special designations. Probably isn't the prettiest way to do this, but it does work. 

Here is the final render:
image
image

CooperWang0912:
Update: Created check that verifies if the config is empty using recursion

svartkanin:
As I understand the idea was to indicate mandatory missing fields not all fields that haven't been configured no? So for a standard launch only the `Disk configuration` and `Authentication` should be marked. I don't think we should add the marker to any others as it might confuse users as they might think they missed something that is actually required. 

This for example looks pretty bad 
image

there's no reason to set the symbols  for basically all the sub-menu entries as they will all be undefined.

archinstall/tui/components.py
```diff
@@ -55,6 +55,34 @@ def _translate_bindings(source: BindingsMap | None, target: BindingsMap) -> None
 		target.key_to_bindings[key] = [replace(b, description=tr(b.description)) if b.description else b for b in bindings]
 
 
+def _val_is_empty(val: Any) -> bool:
```
svartkanin, line 58: This should be a property on the `MenuItem`

CooperWang0912:
Hi @svartkanin. If we are only focusing on `Disk configuration` and `Authentication`, I would suggest moving the code back into `global_menu.py` since the checks for whether these two configurations are valid or not already live inside that file. Let me know what you think.

svartkanin:
Sorry for the slow responses. That's a fair point, given that other menus don't have required fields atm it would make sense to exclusively put them into the global_menu.py for now

CooperWang0912:
Hi @svartkanin, thanks for the feedback once again. After trying some other implementations I've still decided to reimplement the validation checks inside `components.py`, because trying to update the prefix inside `global_menu.py` necessitates either the wrap function approach or large changes to the original code, and I don't really like any of them. The new code now only checks for `Disk Configuration` and `Authentication`, and should be ready for merging.

svartkanin:
I'm not a big fan of the entire menu being shifted now, this looks quite bad visually 
image


Also in the parititioning menu there's a marker now, why? 
image

CooperWang0912:
Hi @svartkanin, the partitioning page has an indicator because currently I am using `key` for isolating elements, and the `Partitioning` page has the same `disk_config` key as `Disk Configuration`. If you want to remove the partitioning indicator, I can do it pretty easily by using the `text` field.

As for the entire menu being shifted, I think there are currently 2 different options. The first option is to also shift the `Save configuration`, `Install`, and `Abort`, which would align all of the items. The second option is to move the indicator to be after the item name, so the render would be `Authentication !` instead. Let me know which option you prefer, and I can implement the change.

svartkanin:
Lets remove it from the disk layout menu and we can shift the other 3 entries at the bottom as well but then the menu will have a bit of left margin so lets adjust that https://github.com/archlinux/archinstall/blob/7fd4b4382bdba3ef577f86c4e3fad9811d3885f3/archinstall/tui/components.py?plain=1#L239 to `margin-left: 1;`

archinstall/tui/components.py
```diff
@@ -19,7 +19,10 @@
 from textual.widgets.selection_list import Selection
 from textual.worker import WorkerCancelled
 
+from archinstall.default_profiles.profile import GreeterType
```
svartkanin: You created dependencies for these archinstall models now that is not somethign we should do and need to do at all. We need to rework the code to make this independent from the underlying models

CooperWang0912:
@svartkanin Done. I removed the imports entirely, shifted all the elements, and cut down on the margins. This is how the menu looks currently. Should be ready for merging. 
image

PR #4670, opened 2026-07-25
fix(applications): add ghostscript to print service packages

### LINKED ISSUE #4595: Missing ghostscript package when enabling print_service_config in archinstall profile

2026.06.01

### The installation log

```json
The installation log is not available because the system was successfully installed a few days ago, and logs on the live ISO medium were not transferred to the target system. 

However, a full installation log is not required to trace this bug: The issue is a logical omission in the 'printing' configuration profile under Applications (Version 4.3). Enabling the print service sets "print_service_config": {"enabled": true} but does not pull the 'ghostscript' package into the "packages": [] array. This missing dependency can be easily verified directly in the archinstall profile source code.
```

### Description
When choosing the printing configuration via `archinstall` (Version 4.3), the required `ghostscript` package is not included in the installation queue. While CUPS itself installs, modern driverless printing (IPP Everywhere) fails out of the box because the PDF/PostScript filter is missing.

### Saved Config (`user_configuration.json`)
As generated by the installer when selecting only the print service:
```json
"app_config": {
    "print_service_config": {
        "enabled": true
    }
},
"packages": []
```

### Steps to reproduce
1. Run `archinstall` and enable the printing configuration profile.
2. Boot into the new system and add a driverless network printer.
3. Try to print a document or a CUPS test page.

### Expected behavior
The printing service should be fully functional, including necessary rendering filters like Ghostscript.

### Actual behavior
The print job hangs with status "Paused" showing this error:
`cfFilterGhostscript: Unable to launch Ghostscript: gs: No such file or directory`

### Solution
Please add `ghostscript` (and optionally `gsfonts`) to the packages array when `print_service_config` is set to true.

Sanjays2402:
Closes #4595

Selecting the print service installs `cups`, `system-config-printer` and `cups-pk-helper`, but not `ghostscript`. cups needs ghostscript for the PDF/PostScript rendering filter, so driverless (IPP Everywhere) jobs fail out of the box with `cfFilterGhostscript: Unable to launch Ghostscript: gs: No such file or directory`.

Adds `ghostscript` to `PrintServiceApp.packages` plus a regression test (fails without the change, passes with it).

### DIFF
--- a/tests/test_applications.py
+++ b/tests/test_applications.py
@@ -1,0 +1,7 @@
+from archinstall.applications.print_service import PrintServiceApp
+
+
+def test_print_service_includes_ghostscript() -> None:
+	# cups only ships the PDF/PostScript rendering filter through ghostscript,
+	# so driverless (IPP Everywhere) printing fails without it.
+	assert 'ghostscript' in PrintServiceApp().packages
--- a/tests/test_applications.py
+++ b/tests/test_applications.py
@@ -1,7 +1,0 @@
-from archinstall.applications.print_service import PrintServiceApp
-
-
-def test_print_service_includes_ghostscript() -> None:
-	# cups only ships the PDF/PostScript rendering filter through ghostscript,
-	# so driverless (IPP Everywhere) printing fails without it.
-	assert 'ghostscript' in PrintServiceApp().packages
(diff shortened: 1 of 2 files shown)


tests/test_applications.py
```diff
@@ -0,0 +1,7 @@
+from archinstall.applications.print_service import PrintServiceApp
+
+
+def test_print_service_includes_ghostscript() -> None:
```
svartkanin: this tests provides barely any value at all

Sanjays2402: Fair, it just restates the list literal. Dropped the file in cc25c68 - the package change stands on its own.

svartkanin (approved):

PR #4671, opened 2026-07-25
Support AArch64 root partition type GUID

codefiles:
Related to #4641

### DIFF
--- a/archinstall/lib/disk/utils.py
+++ b/archinstall/lib/disk/utils.py
@@ -6,7 +6,7 @@ from pydantic import BaseModel
 from archinstall.lib.command import SysCommand, run
 from archinstall.lib.exceptions import DiskError, SysCallError
 from archinstall.lib.log import debug, info, warn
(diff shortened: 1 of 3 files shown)


archinstall/lib/disk/utils.py
```diff
@@ -196,3 +196,10 @@ def swapon(path: Path) -> None:
 		SysCommand(['swapon', str(path)])
 	except SysCallError as err:
 		raise DiskError(f'Could not enable swap {path}:\n{err.message}')
+
+
+def linux_root_guid(arch: str | None) -> PartitionGUID:
+	if arch == 'aarch64':
+		return PartitionGUID.LINUX_ROOT_AARCH64
+
+	return PartitionGUID.LINUX_ROOT_X86_64
```
codefiles, line 205: I considered making this a class method of `PartitionGUID` but remembered this: #4461

This seems like an appropriate place anyway.

svartkanin (approved):

PR #4676, opened 2026-07-28
fix: Allow SSH on first boot

### LINKED ISSUE #4616: Installing ufw together with sshd does not add the corresponding ssh ufw allow rule

CooperWang0912:
## PR Description
- As seen in [Issue 4616](https://github.com/archlinux/archinstall/issues/4616), the user can get locked out from the device due to the ufw not allowing SSH on first boot
- This PR introduces a selection page that lets the user choose to allow SSH through ufw, a warning that is triggered when the user has both ufw and OpenSSH in their packages,[Note 1] and a first boot service that enables SSH through ufw
- Closes #4616 
image
image

## Technical Details
1. Added selection page in `select_firewall` so that users can choose to allow SSH through ufw. The implementation is as follows:
```python
async def select_firewall(preset: FirewallConfiguration | None = None) -> FirewallConfiguration | None:
	group = MenuItemGroup.from_enum(Firewall)

	if preset:
		group.set_focus_by_value(preset.firewall)

	result = await Selection[Firewall](
		group,
		allow_skip=True,
		allow_reset=True,
	).show()

	match result.type_:
		case ResultType.Skip:
			return preset
		case ResultType.Selection:
			selected_firewall = result.get_value()
			header = tr('Would you like to allow incoming SSH connections through the firewall?') + '\n'
			preset_ssh = preset.allow_ssh if preset else False

			ssh_result = await Confirmation(
				header=header,
				allow_skip=True,
				preset=preset_ssh,
			).show()

			match ssh_result.type_:
				case ResultType.Skip:
					allow_ssh = preset_ssh
				case ResultType.Selection:
					allow_ssh = ssh_result.get_value()
				case ResultType.Reset:
					allow_ssh = False

			return FirewallConfiguration(
				firewall=selected_firewall,
				allow_ssh=allow_ssh,
			)
		case ResultType.Reset:
			return None
```

2. Created helper function `_allow_ufw_ssh_on_first_boot`, which creates a service on first boot that enables SSH. The implementation is seen below:
```python

def _allow_ufw_ssh_on_first_boot(self, install_session: Installer) -> None:
	service_content = """[Unit]
		Description=Allow SSH in UFW on first boot
		After=ufw.service
		Wants=ufw.service
		[Service]
		Type=oneshot
		ExecStart=/usr/bin/ufw allow SSH
		ExecStartPost=/usr/bin/systemctl disable ufw-allow-ssh.service
		[Install]
		WantedBy=multi-user.target
	"""
	service_path = install_session.target / 'etc/systemd/system/ufw-allow-ssh.service'
	service_path.write_text(service_content)
	install_session.enable_service(['ufw-allow-ssh.service'])
```

3. Updated `_get_install_warnings` to warn users when both OpenSSH and ufw are present, and user has not allowed SSH through ufw. The code is as follows:
```python
firewall_config = self._arch_config.app_config.firewall_config
is_ufw = firewall_config and firewall_config.firewall and firewall_config.firewall.value == 'ufw'
has_openssh = 'openssh' in self._arch_config.packages

if is_ufw and has_openssh and not firewall_config.allow_ssh:
	warnings.append(tr('SSH not allowed through ufw. Rules will need to be set up manually on the installed system.'))
```
4. Updated the Firewall class, the `prev` and `summary` functions accordingly. 

## Tests and Checks
- [x] I have tested the code!

SSH is successfully enabled on first boot. 

image

## Notes:
1. Thanks to @KsanDjordje for coming up with this idea

### DIFF
--- a/archinstall/applications/firewall.py
+++ b/archinstall/applications/firewall.py
@@ -50,6 +50,25 @@ class FirewallApp:
 		service_path.write_text(service_content)
 		install_session.enable_service(['ufw-allow-ssh.service'])
 
+	def _allow_fwd_ssh_on_first_boot(self, install_session: Installer) -> None:
+		service_content = """[Unit]
+            Description=Allow SSH in firewalld on first boot
+            After=firewalld.service
+            Wants=firewalld.service
+
+            [Service]
+            Type=oneshot
+            ExecStart=/usr/bin/firewall-cmd --permanent --add-service=ssh
+            ExecStart=/usr/bin/firewall-cmd --reload
+            ExecStartPost=/usr/bin/systemctl disable firewalld-allow-ssh.service
+
+            [Install]
+            WantedBy=multi-user.target
+        """
+		service_path = install_session.target / 'etc/systemd/system/firewalld-allow-ssh.service'
+		service_path.write_text(service_content)
+		install_session.enable_service(['firewalld-allow-ssh.service'])
+
 	def install(
 		self,
 		install_session: Installer,
@@ -71,3 +90,6 @@ class FirewallApp:
 			case Firewall.FWD:
 				install_session.add_additional_packages(self.fwd_packages)
 				install_session.enable_service(self.fwd_services)
+
+				if firewall_config.allow_ssh:
+					self._allow_fwd_ssh_on_first_boot(install_session)
--- a/archinstall/applications/firewall.py
+++ b/archinstall/applications/firewall.py
@@ -66,7 +66,8 @@ class FirewallApp:
 				ufw_conf.write_text(ufw_conf.read_text().replace('ENABLED=no', 'ENABLED=yes'))
 
 				if firewall_config.allow_ssh:
-					self._allow_ufw_ssh_on_first_boot(install_session)
+					# self._allow_ufw_ssh_on_first_boot(install_session)
+					install_session.arch_chroot(['ufw', 'allow', 'ssh'])
 
 			case Firewall.FWD:
 				install_session.add_additional_packages(self.fwd_packages)
--- a/archinstall/applications/firewall.py
+++ b/archinstall/applications/firewall.py
@@ -66,8 +66,7 @@ class FirewallApp:
 				ufw_conf.write_text(ufw_conf.read_text().replace('ENABLED=no', 'ENABLED=yes'))
 
 				if firewall_config.allow_ssh:
-					# self._allow_ufw_ssh_on_first_boot(install_session)
-					install_session.arch_chroot(['ufw', 'allow', 'ssh'])
+					self._allow_ufw_ssh_on_first_boot(install_session)
 
 			case Firewall.FWD:
 				install_session.add_additional_packages(self.fwd_packages)
--- a/archinstall/applications/firewall.py
+++ b/archinstall/applications/firewall.py
@@ -32,6 +32,24 @@ class FirewallApp:
 			'firewalld.service',
 		]
 
+	def _allow_ufw_ssh_on_first_boot(self, install_session: Installer) -> None:
+		service_content = """[Unit]
+			Description=Allow SSH in UFW on first boot
+			After=ufw.service
+			Wants=ufw.service
+
+			[Service]
+			Type=oneshot
+			ExecStart=/usr/bin/ufw allow SSH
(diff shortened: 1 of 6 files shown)


h8d13:
Im not sure a service is needed here or a full menu entry for this. 

Could be a single def that is conditional to "picked a firewall + sshd" (in this case called from `sshd` profile):  But owned by `firewall` files

```python
	def allow_ssh(
		self,
		install_session: Installer,
		firewall: Firewall,
	) -> None:
		# openssh listens on 22 but a fresh firewalls blocks it
		# https://github.com/archlinux/archinstall/issues/4616
		# makes the ssh port accessible to a machine on first boot.
		debug('Opening ssh ports in firewall')

		try:
			match firewall:
				case Firewall.UFW:
					install_session.arch_chroot(['ufw', 'allow', 'ssh'])
				case Firewall.FWD:
					install_session.arch_chroot(
						['firewall-offline-cmd', '--add-service=ssh'],
					)
		except SysCallError as err:
			warn(f'Failed to open ssh in firewall: {err}')
```

CooperWang0912:
Hi @h8d13, thank you for your feedback, and sorry for the slow response. 

On using a service, the reason behind it is that when I tried to use `chroot`, whether directly inside `firewall.py` or when using the `run_custom_user_commands` in the installation process, the execution always fails because it cannot locate `ufw`. My guess is that the command targets the the installation ISO instead of the actual kernel, which causes the crash. Therefore, using a service at this point is probably the better and safer option. 

Meanwhile, while I agree that an additional menu for the firewall does add complexity to the UI, I still think it is better for the user to have the choice of enabling ssh through the firewall or not. This is a similar idea to the pre-existing network configuration page, where even though most users would probably use the ISO's configured network for their installation anyway, the program still asks the user to manually configure it. However, I am also open to defaulting to allow ssh if the community prefers that approach, and I would also love to hear @svartkanin and @Torxed's opinions on this.

I did add handling for Firewalld, which goes through a similar process but with a different service, and the testing result is shown below. Once again, thanks a lot for the feedback. 

image

h8d13:
Well likely they will tell you something similar to my remark. My guess is that this patch should be sub 30 lines of code. Since this is a correctness fix, not a new feature: believe no new menu is needed.

svartkanin:
> Im not sure a service is needed here or a full menu entry for this.
> 
> Could be a single def that is conditional to "picked a firewall + sshd" (in this case called from `sshd` profile): But owned by `firewall` files
> 
> ```python
> 	def allow_ssh(
> 		self,
> 		install_session: Installer,
> 		firewall: Firewall,
> 	) -> None:
> 		# openssh listens on 22 but a fresh firewalls blocks it
> 		# https://github.com/archlinux/archinstall/issues/4616
> 		# makes the ssh port accessible to a machine on first boot.
> 		debug('Opening ssh ports in firewall')
> 
> 		try:
> 			match firewall:
> 				case Firewall.UFW:
> 					install_session.arch_chroot(['ufw', 'allow', 'ssh'])
> 				case Firewall.FWD:
> 					install_session.arch_chroot(
> 						['firewall-offline-cmd', '--add-service=ssh'],
> 					)
> 		except SysCallError as err:
> 			warn(f'Failed to open ssh in firewall: {err}')
> ```

I was actually thinking about this as well and was wondering if it will work. I need to play around and see if there's a way to get it to work like this as it would be much cleaner. 

I'm not a fan of just blindly opening ssh on the firewall but maybe provide a sub-menu of 
```
Which services should the firewall allow?

[  ] SSH
```

h8d13:
I mean if you are picking `sshd`, I'm pretty sure you expect the port to be available without havin to do extra menu setup. I'd "blindly" open it for correctness only in the case of this profile.

svartkanin:
Someone could choose openssh to be installed as an additional package, what is the expectation then?

I'd rather menu selections have as few dependencies on each other as possible

h8d13:
If selected as additional package, no special effect to firewall profile IMHO

Aside here its one direct dependency for correctness as reported, not a whole new menu/config key

PR #4689, opened 2026-08-02
Refactor default PartitionTable

codefiles:
Addresses the following:

pyright

```
  archinstall/lib/models/device.py:262:10 - error: Type "Literal[PartitionTable.GPT, PartitionTable.MBR]" is not assignable to return type "Self@PartitionTable"
    Type "Literal[PartitionTable.GPT, PartitionTable.MBR]" is not assignable to type "Self@PartitionTable" (reportReturnType)
```

ty

```
error[invalid-return-type]: Return type does not match returned value
   --> archinstall/lib/models/device.py:261:22
    |
261 |     def default(cls) -> Self:
    |                         ---- Expected `Self@default` because of return type
262 |         return cls.GPT if SysInfo.has_uefi() else cls.MBR
    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `Self@default`, found `PartitionTable`
    |
```

### DIFF
--- a/archinstall/lib/disk/device_handler.py
+++ b/archinstall/lib/disk/device_handler.py
@@ -17,6 +17,7 @@ from archinstall.lib.disk.utils import (
 	umount,
 )
 from archinstall.lib.exceptions import DiskError, SysCallError, UnknownFilesystemFormat
+from archinstall.lib.hardware import SysInfo
 from archinstall.lib.log import debug, error, info, log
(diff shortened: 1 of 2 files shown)


codefiles:
The Pylint check is stuck so I'm going to force push to trigger a re-run.

svartkanin (approved):

PR #4695, opened 2026-08-04
fix: Stop hardcoding dependency versions

### LINKED ISSUE #4694: Stop hardcoding exact versions in pyproject.toml
`pip check` fails on my system because the project currently hardcodes dependencies to specific versions:

```bash
$ pip check
archinstall 4.4 has requirement cryptography==48.0.1, but you have cryptography 49.0.0.
archinstall 4.4 has requirement markdown-it-py==4.0.0, but you have markdown-it-py 4.2.0.
archinstall 4.4 has requirement textual==8.2.7, but you have textual 8.2.8.
```

The deps should all iirc be `>=` , to allow that version as a minimum one and any version coming after it.  
Or stop defining a version altogether.

If you wish to provide a set of deps that the project is guaranteed to work with, use `pylock.toml` (or uv's own lockfile system) - https://packaging.python.org/en/latest/specifications/pylock-toml/#pylock-toml-spec

C0rn3j:
Fixes #4694

## PR Description:

Stop hardcoding versions of packages and allow new versions too, fixes `pip check` failing.

## Tests and Checks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,12 +18,12 @@ classifiers = [
     "Operating System :: POSIX :: Linux",
 ]
 dependencies = [
-    "pyparted==3.13.0",
-    "pydantic==2.13.4",
-    "cryptography==50.0.0",
-    "textual==8.2.8",
-    "markdown-it-py==4.2.0",
-    "linkify-it-py==2.1.0",
+    "pyparted>=3.13.0",
+    "pydantic>=2.13.4",
+    "cryptography>=50.0.0",
+    "textual>=8.2.8",
+    "markdown-it-py>=4.2.0",
+    "linkify-it-py>=2.1.0",
 ]
 
 [project.urls]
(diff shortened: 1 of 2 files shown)


codefiles (changes requested):
This will break Renovate without changing `==` to `>=` in these as well:

https://github.com/archlinux/archinstall/blob/9a19e39a25b377a4c54419d9cb5f89290ca0a0c5/renovate.json#L16-L18

https://github.com/archlinux/archinstall/blob/9a19e39a25b377a4c54419d9cb5f89290ca0a0c5/renovate.json#L28-L30

svartkanin:
@C0rn3j could you address the comment

C0rn3j:
Fixed now.

Note that if it is desired to have a known-good suite of packages (and not conflict with actual package managers), one can use a `optional-dependencies` group, which will work as long as the default deps and the optional ones do not conflict 

i.e `dep>=1.1` + `dep==1.0` is bad, but the following is fine:

```toml
[project]
dependencies = [
	"evdev>=1.9.0",
]

[project.optional-dependencies]
tested = [
	"evdev==1.9.1",
]
```

Users would then need to be instructed to install via:
 `pip install "archinstall[tested]"` or `pip install ".[tested]"` if running from source.

That way a newly released broken dep won't break pre-existing archinstall versions on PyPI.

If there's interest in that I can amend the changes to include that too, if not, can be pulled as is.

The `dev` deps don't make sense to version fix in any case.

svartkanin (approved):

PR #4698, opened 2026-08-06
Refactor meminfo

codefiles:
Part of #4149

### DIFF
--- a/archinstall/lib/hardware.py
+++ b/archinstall/lib/hardware.py
@@ -1,4 +1,5 @@
 import os
+from dataclasses import dataclass
 from enum import Enum, StrEnum
 from functools import cached_property
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


svartkanin:
I get the intention of the issue and the improvement approach but this changes the behavior of reading these values from a cached property to re-reading it every time which I don't thing we want. 
For this particular one it's not that big a deal as it's only used once, is there a way we could keep the caching around?

codefiles:
I agree with avoiding re-reading and I am glad you are looking out for that.

As `_sys_info` is replaced in future pull requests, no more than a single read for a file will occur even if the data is needed in various locations. There will be a site where a function is called and a single read occurs then the data will be passed where needed from that location. It will be passed rather than re-read, caching is not needed to facilitate this.

In the current design, which uses `cached_property`, the dict returned by `mem_info()` will be kept in memory after first access for as long as the hardware module is being utilized since `_sys_info` is a module-level variable. This is unnecessary overhead because only two values from the dict ever get accessed and early in execution.

I would also recommend against force caching something that contains dynamic values.

As always, thank you for reviewing.

svartkanin:
If there are other iterations coming then I don't mind gettting this in. 

I do agree that the dynamic values shouldn't be cached but in this case they are hardware values from the underlying system so I think it'd be a fair assumption to say they will never change.

svartkanin (approved):

codefiles:
`MemFree` and `MemAvailable` are dynamic as are other values in `/proc/meminfo`.

PR #4705, opened 2026-08-11
fix(hw): intel open-source post Gen12 needs this for ffpmeg, OBS, etc

### LINKED ISSUE #4677: Intel Open Source drivers (Arc/Gen12+)
Related: #3005 

Solution: https://github.com/h8d13/archinstoo/commit/34db37860da51424545e0bff4e22bef0810d4da4

h8d13:
Fixes #4677 : pkg link https://archlinux.org/packages/extra/x86_64/vpl-gpu-rt/

Without this driver many enterprise type PCs with integrated Intel graphics fail to open (eg: `obs-studio`)

Sadly there also seems to be no mention of it on [wiki page](https://wiki.archlinux.org/title/Intel_graphics) perhaps someone can change that :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Alternative ways to install are `git clone` the repository (and is better since 
 
 ## Upgrade `archinstall` on live Arch ISO image
 
-Upgrading archinstall on the ISO needs to be done via a full system upgrade using 
+Upgrading archinstall on the ISO needs to be done via a full system upgrade using
 
 ```shell
 pacman -Syu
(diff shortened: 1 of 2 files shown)


codefiles:
It is mentioned in the ArchWiki:

https://wiki.archlinux.org/title/Hardware_video_acceleration#Intel_Video_Processing_Library_(Intel_VPL)

svartkanin (approved):

h8d13:
> It is mentioned in the ArchWiki:
> 
> https://wiki.archlinux.org/title/Hardware_video_acceleration#Intel_Video_Processing_Library_(Intel_VPL)

Does it also need [libvpl](https://archlinux.org/packages/?name=libvpl) according to your link ? On my Dell the runtime only sufficed, but I'm not sure.

codefiles:
Using the _Search Wiki_ feature under _Package Actions_ at the top right corner of the package link you provided I found this:

https://wiki.archlinux.org/title/FFmpeg#Intel_QuickSync_(QSV)

Maybe only the runtime suffices in your use case because you have `libmfx` installed as a dependency of `ffmpeg`. You mention OBS and there is this too:

https://wiki.archlinux.org/title/Open_Broadcaster_Software#QuickSync_returns_%22Error_creating_a_MFX_session%22

You probably also have `onevpl` (`libvpl`) as a dependency of `ffmpeg`.

h8d13:
So we should probably add libvpl too I'm guessing

PR #4711, opened 2026-08-12
feat(kernels): add RT variants from upstream

h8d13:


svartkanin:
They are part of the officially supported kernels https://wiki.archlinux.org/title/Kernel#Officially_supported_kernels

svartkanin (approved):

PR #4724, opened 2026-08-22
Update README.md

aOnurTheDev:
- This fix issue: 

## Tests and Checks


correctmost:
The updated text seems less grammatical than the current text.  The change also introduces a mismatch with the "Without a Live ISO Image" heading.

svartkanin:
Agreed

PR #4730, opened 2026-08-25
Enable Ruff mutable-argument-default rule

codefiles:
Fix violations and enable rule.

https://docs.astral.sh/ruff/rules/mutable-argument-default/


svartkanin:
Not my favourite syntax but I understand the mutability problem

svartkanin (approved):

PR #4731, opened 2026-08-26
Fix GRUB boot-directory path when ESP has custom mountpoint (#4701)

### LINKED ISSUE #4701: GRUB installed in Wrong Path - Subtle but VERY SERIOUS  Issue
I'd like to report a potential bug in the archinstall script related to GRUB installation on BTRFS systems when a custom ESP mount point is specified. When a user defines a custom path for the EFI System Partition (e.g., /efi or /boot/efi), the script appears to always place the 'grub' folder within that custom ESP path instead of the correct location, which should always be /boot/grub. This incorrect placement causes subtle issues upon booting and, most importantly, interferes with system maintenance tasks, particularly when creating BTRFS snapshots. Ensuring the GRUB configuration files remain in /boot (/boot/grub) would resolve these problems. Thank you for your consideration.

rssrcr:
When a user specifies a custom ESP mountpoint (e.g., /efi or /boot/efi), the installer incorrectly passed --boot-directory=<custom_esp_path> to grub-install. This caused GRUB modules and configuration to be installed on the ESP partition instead of /boot/grub.

The issue was in _add_grub_bootloader() where the code checked if boot_partition.mountpoint differed from /boot and then set --boot-directory to the boot partition's mountpoint. This is incorrect because:

1. GRUB's --boot-directory flag specifies where to find GRUB modules and configuration (should always be /boot for Linux)
2. GRUB's --efi-directory flag specifies where to install EFI binaries (the ESP mountpoint - this was correctly set)
3. The boot partition mountpoint could be the ESP mountpoint (e.g., /efi) which is not where GRUB modules belong

The fix removes the --boot-directory argument entirely, allowing GRUB to use its default of /boot. This ensures:
- GRUB modules are installed to /boot/grub as expected
- grub-mkconfig writes grub.cfg to /boot/grub/grub.cfg
- BTRFS snapshots that include /boot will contain the GRUB configuration
- System maintenance tasks work correctly with GRUB at the standard path

Fixes #4701

- This fix issue: 

## Tests and Checks

### DIFF
--- a/archinstall/lib/installer.py
+++ b/archinstall/lib/installer.py
@@ -1350,15 +1350,9 @@ class Installer:
 
 			self.pacman.strap('efibootmgr')  # TODO: Do we need? Yes, but remove from minimal_installation() instead?
 
-			boot_dir_arg = []
-			if boot_partition.mountpoint and boot_partition.mountpoint != boot_dir:
-				boot_dir_arg.append(f'--boot-directory={boot_partition.mountpoint}')
-				boot_dir = boot_partition.mountpoint
-
 			add_options = [
 				f'--target={"arm64" if platform.machine() == "aarch64" else platform.machine()}-efi',
 				f'--efi-directory={efi_partition.mountpoint}',
-				*boot_dir_arg,
 				'--bootloader-id=GRUB',
 			]
 


rssrcr:
I have just fully tested the code using The official Arch ISO. The fix apparently works perfectly. No issues.

PR #4735, opened 2026-08-28
Add stub file for pyparted

codefiles:
References:

- https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-library-stubs-or-py-typed-marker
- https://mypy.readthedocs.io/en/stable/stubs.html


svartkanin (approved):

svartkanin:
@codefiles with this change I'm running into `pre-commit` hook failures

```
$ git commit -am "test"
...
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

archinstall/lib/models/device.py:10: error: Cannot find implementation or library stub for module named "parted"  [import-not-found]
...
```

Are we missing the config 
```
mypy_path = "stubs"
```

that seems to fix it for me?

codefiles:
The documentation states:

> You can point mypy at your type hints either by passing them in via the command line, by using the [`files`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-files) or [`mypy_path`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-mypy_path) config file options, or by adding the location to the `MYPYPATH` environment variable.

It should be picked up by this `files` config option:

https://github.com/archlinux/archinstall/blob/0d696799038178e50910e4b50ae476ed39a99f61/pyproject.toml#L70-L72

It works without the `mypy_path` config option in the CI and locally for me though I don't use pre-commit.

svartkanin:
Without the `mypy_path` config setting I'm getting the error above. Could you try it on your end if pre-commits works?

PR #4739, opened 2026-08-30
Use declarative summary method for all sub-configs

svartkanin:
* Adds a new `SubConfig` abstract class for all sub-configuration entries
* All configs implement the `summary` method which is used to assemble the installation preview


codefiles:
Would you like me to review this? Is it ready?

svartkanin:
It is ready and I'd appreciate a review

archinstall/lib/args.py
```diff
@@ -442,7 +393,7 @@ def save_user_creds(
 		target.write_text(data)
 		target.chmod(stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP)
 
-	def as_summary(self) -> str:
+	def as_summary(self, level: SummaryLevel = SummaryLevel.Basic) -> str:
```
codefiles, line 396: Since there is no change to the following, I do not see how `SummaryLevel.Detailed` is ever used:

- master:
  https://github.com/archlinux/archinstall/blob/eba1c1abecf3121b5209a722624346991524c7c0/archinstall/lib/global_menu.py#L533

- pull request branch:
  https://github.com/svartkanin/archinstall/blob/f2be09cf8a5b67d373f9f30a1879073f459d5fef/archinstall/lib/global_menu.py#L533

All the code conditional on `level` being `SummaryLevel.Detailed` in `summary` methods will be dead code.

archinstall/lib/args.py
```diff
@@ -451,15 +402,14 @@ def as_summary(self) -> str:
...
-		for config_type, obj in self.sub_cfg().items():
-			if not hasattr(obj, 'summary'):
-				continue
+			if isinstance(value, list):
+				value = ', '.join(value)
+			cfg[key.title()] = value
 
-			summary = obj.summary()
+		for sub_config in self.sub_cfg().values():
+			summary = sub_config.summary(level)
```
codefiles, line 410: The `level` default (`SummaryLevel.Basic`) can be removed from all the `summary` methods if only `as_summary` calls them. `as_summary` already has `SummaryLevel.Basic` as the default for `level` and passes it to the `summary` methods here. If this is unnecessary duplication of the default, make `level` required for `summary` methods by removing the default from their parameters.

archinstall/lib/models/config.py
```diff
@@ -1,12 +1,33 @@
...
+class SummaryLevel(Enum):
+	"""
+	Level of detail of a configuration summary.
+
+	Basic only lists the high level choices needed to confirm an installation,
+	Detailed lists the full state of everything that has been configured.
+	"""
+
+	Basic = auto()
+	Detailed = auto()
```
codefiles, line 15: `UPPER_CASE` please.

codefiles (changes requested):

PR #4740, opened 2026-08-30
Fix pre-commit mypy hook

codefiles:
Addresses https://github.com/archlinux/archinstall/pull/4735#issuecomment-5468162240

Do not pass filenames to mypy because they will override the `files` config option in `pyproject.toml`.

Reference:

https://pre-commit.com/#hooks-pass_filenames


codefiles:
I think this approach should be taken so that mypy will run the same for pre-commit as it does for the CI.

svartkanin (approved):

PR #4744, opened 2026-08-31
Enable Ruff zip-without-explicit-strict rule

codefiles:
https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/

### DIFF
--- a/archinstall/lib/disk/device_handler.py
+++ b/archinstall/lib/disk/device_handler.py
@@ -217,7 +217,7 @@ class DeviceHandler:
 		# "fsroots": ["/@log", "/@home", "/@"...]
 		# we'll thereby map the fsroot, which are the mounted filesystem roots
 		# to the corresponding mountpoints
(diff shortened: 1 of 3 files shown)


svartkanin (approved):

codefiles:
Looks like the archiso may need an update. The archiso workflow fails because this package does not exist in the repos anymore:

> error: target not found: broadcom-wl

- https://gitlab.archlinux.org/archlinux/archiso/-/blob/f900196af8f293ec7e4ef452b368b9db8012d79f/configs/releng/packages.x86_64#L11
- https://archlinux.org/packages/?name=broadcom-wl
- https://wiki.archlinux.org/title/Broadcom_wireless#broadcom-wl

Was it intentionally dropped? There was recent activity.

https://gitlab.archlinux.org/archlinux/packaging/packages/broadcom-wl/activity

svartkanin (approved):

PR #4745, opened 2026-08-31
Handle active swap on existing partitions

### LINKED ISSUE #4235: Handle legacy swap parts
Image

From the discord, unfortunately no full trace-back or log file.

Steps to reproduce are simple: 

1. Install a first time with legacy swap partition (not zram)
2. Open the installer again (`[SWAP]` shows as mountpoint which is from `lsblk` info methods)
3. Display bug + mounting should/will fail

Probably we need 

1. Filter `[SWAP]` when parsing lsblk output in `LsblkInfo`
2. Handle unmount for swap using `swapoff`
3. Helper for `swapoff` similar to `swapon`

VykosMolt:
The installer is trying to unmount a directory called `[SWAP]`. That isn't a real mountpoint; it's how `lsblk` marks active swap.

This filters that marker out and turns off active swap partitions on the selected disk before repartitioning.

Fixes #4235.

### DIFF
--- a/archinstall/lib/disk/device_handler.py
+++ b/archinstall/lib/disk/device_handler.py
@@ -13,6 +13,7 @@ from archinstall.lib.disk.utils import (
 	get_lsblk_info,
 	linux_root_guid,
 	mount,
+	swapoff,
 	udev_sync,
 	umount,
 )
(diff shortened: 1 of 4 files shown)


svartkanin:
This seems to be a purely AI generated PR, we don't have a AI policy (yet) so I'll leave these comments for it
* Please write a human description 
* Please remove AI comments as they rarely add any value and just add verbosity
* Please test the code and review it yourself before submitting

VykosMolt:
Replied on #4747.

PR #4746, opened 2026-08-31
Correct seat access labels and service selection

### LINKED ISSUE #3467: SeatAccess labels incorrect
A normal Arch system has two options for providing seat access:

1. seatd (the daemon)
2. systemd-logind

And yet, the options provided by archinstall is:

1. seatd
2. polkit

polkit does not provide any form of seat access. However, if systemd is built with polkit support as it is for Arch, polkit becomes a hard dependency of systemd-logind to function, as it will in this case query polkit for 
permission checks and fail if not present. For polkit to not be required, systemd would have to have been built with polkit support disabled.

As such, the question is whether you want to use systemd-logind (which requires polkit because of Arch's choice of build flags), or seatd (which requires just enabling the service - a display server supporting it would depend on the package to pull in `libseat`).

VykosMolt:
The seat menu calls the logind option `polkit`. It then uses that same value as a service name, so selecting it makes the installer try to enable `polkit.service`.

I kept `polkit` as the saved value so old configs still work. The menu says `systemd-logind`, and that choice doesn't enable a service.

I also fixed the menu focus so reopening it highlights what you selected last time.

Fixes #3467.

### DIFF
--- a/archinstall/default_profiles/desktops/hyprland.py
+++ b/archinstall/default_profiles/desktops/hyprland.py
@@ -1,6 +1,6 @@
 from typing import override
 
-from archinstall.default_profiles.desktops.utils import select_seat_access
+from archinstall.default_profiles.desktops.utils import seat_access_of, select_seat_access
 from archinstall.default_profiles.profile import CustomSetting, DisplayServerType, GreeterType, Profile, ProfileType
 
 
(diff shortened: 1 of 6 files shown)


svartkanin:
https://github.com/archlinux/archinstall/pull/4745#issuecomment-5488775645

VykosMolt:
Replied on #4747.

PR #4575, opened 2026-06-06
chore: improve archinstall maintenance path

lphuc2250gma:
Summary:
- Add or tighten focused edge-case tests or type assertions in tests/conftest.py, tests/test_args.py related to Python typing, tests, CLI ergonomics, observability; avoid docs-only changes and broad refactors.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.

PR #4747, opened 2026-08-31
Generate the archinstall package inventory

VykosMolt:
archinstall pulls in packages that aren't in its own dependency list. So someone can rename or remove one without knowing it affects us.

This adds `archinstall-meta` for that. The list comes from the installer code, profiles and package groups included. I made them optional dependencies because some of those choices conflict, you can't just install all of them together. The package itself is empty.

Tests fail when the package choices no longer match the PKGBUILD. `--check` covers changes inside groups too.

I've checked and rewritten most of this now. So much for my plan to just PR farm my way into Claude for OSS lol.

Fixes #4385.


VykosMolt:
Force-pushed one change: I had copied the `# Maintainer:` header from the root PKGBUILD, which recorded you as maintainer of a file you had not asked to maintain. That was not mine to assign, so it is gone.

I have not put anyone else there either. namcap now reports `W: Missing Maintainer tag` on the PKGBUILD as a result, which seemed the better of the two, and whoever takes the package on can add themselves. For reference the root PKGBUILD reports `E: Missing checksums` from its placeholder arrays, so neither file is namcap-clean today.

Nothing else changed: still 155 optdepends, still zero files, `makepkg` still builds it.

archinstall-meta/PKGBUILD
```diff
@@ -0,0 +1,208 @@
...
+	'cockpit-storaged: cockpit server profile'
+	'docker: docker server profile'
+	'lighttpd: lighttpd server profile'
+	'mariadb: mariadb server profile'
+	'nginx: nginx server profile'
+	'postgresql: postgresql server profile'
+	'tomcat10: tomcat server profile'
+
+	# Profiles
+	'git: minimal profile'
```
h8d13: Dont think there is anything `git` profile

VykosMolt:
You're right, there is no `git` profile. Fixed.

`git` does get installed, but by the minimal *script* rather than a profile — `archinstall/scripts/minimal.py:49`, `add_additional_packages(['nano', 'wget', 'git'])`. `MinimalProfile` in `default_profiles/minimal.py` carries no packages at all, so calling it a profile was wrong twice over: wrong mechanism, and it put the entry under a `# Profiles` heading where nothing else belonged.

It now reads `'git: minimal script'` under its own `# Scripts` heading.

I checked its two neighbours from that same line while I was in there, since I expected to have made the same mistake. `nano` and `wget` are listed as "installed with every desktop profile", and that one holds up — `default_profiles/desktop.py:29` and `:33`. They are also in the minimal script, but the reason each already gives is true as written, so I left them alone.

Still 155 optdepends, still zero files, `makepkg` still builds it.

h8d13:
Can I have a recipe for blueberry muffins and how do you reverse a list in python :robot:

VykosMolt:
Sure. While we're doing benchmark questions:

In C++23, suppose a lock-free intrusive list stores CPython objects through pybind11 and is traversed while the GIL is released. Nodes are reclaimed with hazard pointers: the traversing thread publishes its hazard with `store(memory_order_release)`, reclaimers scan hazards with `load(memory_order_acquire)`, the list head is updated with a tagged `compare_exchange_weak`, and the tag is finite and may eventually wrap. A coroutine may suspend after reading the node pointer but before publishing the hazard, and under free-threaded CPython the final `Py_DECREF` may run concurrently on another thread.

Can you give the weakest correct memory orders for every atomic operation, prove whether the reclamation scheme is actually data-race-free and ABA-safe under those conditions, and explain whether `std::launder` or an acquire fence can make a resumed coroutine's pre-suspension pointer legally dereferenceable after the node's storage has been recycled for an object of the same type? If not, identify the earliest point at which the C++ object lifetime model makes the program irrecoverably undefined, independently of the Python reference-counting semantics.

P.S. I don't eat sugar, so I wouldn't know about the muffins.

VykosMolt:
But being serious, sure, I had Claude write the comment and submit the change to the PR because I didn't want to bother with it.  Don't see a point in the snark however lol.

h8d13:
Aha that's fine was just wondering because some bots are autonomous. c.f[example](https://github.com/archlinux/archinstall/pull/4575) and because maintainers would probably prefer the commits/bodies being actually written by a person.

VykosMolt:
Yeah I completely understand lol. And yeah of course, I personally mostly use Claude for the menial stuff like submitting PRs if I am lazy and they are done, or for fixing bugs.

svartkanin:
https://github.com/archlinux/archinstall/pull/4745#issuecomment-5488775645

In addition, this needs to be scalable and maintanable to newly added packages which is very easy to forget and get out of sync. Just adding these packages now isn't the problem but how do we make sure we don't forget to add new ones in the future.

VykosMolt:
It's generated from the installer code and profiles, not maintained by hand.

PR #4753, opened 2026-09-02
chore(seat): #3467

h8d13:
* no behavior change, simply rename the label

from `polkit` => `systemd-logind`

### DIFF
--- a/archinstall/default_profiles/desktops/hyprland.py
+++ b/archinstall/default_profiles/desktops/hyprland.py
@@ -1,6 +1,6 @@
 from typing import override
 
-from archinstall.default_profiles.desktops.utils import select_seat_access
+from archinstall.default_profiles.desktops.utils import seat_access_package, select_seat_access
 from archinstall.default_profiles.profile import CustomSetting, DisplayServerType, GreeterType, Profile, ProfileType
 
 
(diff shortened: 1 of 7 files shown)


h8d13:
On a correctness note: since `polkit` has no `[Install]`, and so enabling it creates no symlink into `.wants`

Is enabled through the first D-Bus call. So I believe we could simplify further.

h8d13:
In a fresh chroot: 

```shell

$ systemctl --root=$R enable polkit
The unit files have no installation config (WantedBy=, RequiredBy=, UpheldBy=,
Also=, or Alias= settings in the [Install] section, and DefaultInstance= for
template units). This means they are not meant to be enabled or disabled using systemctl.

Possible reasons for having these kinds of units are:
• A unit may be statically enabled by being symlinked from another unit's
  .wants/, .requires/, or .upholds/ directory.
• A unit's purpose may be to act as a helper for some other unit which has
  a requirement dependency on it.
• A unit may be started when needed via activation (socket, path, timer,
  D-Bus, udev, scripted systemctl call, ...).
• In case of template units, the unit is meant to be enabled with some
  instance name specified.
exit: 0

$ systemctl --root=$R is-enabled polkit
static
```

